Issue #3, opened 2023-07-19
Add additional validation rules


brendt:
Each of these rules should implement the `Rule` interface, and have a test to validate it works as expected. Keep in mind that Tempest relies on PHP's type system for rudimentary validation like required, nullable, and type. These shouldn't be validation rules.

- [x] Alpha - all characters are letters [A-Za-z] (https://github.com/tempestphp/tempest-framework/pull/69)
- [x] AlphaNumeric - all characters are alphanumeric (https://github.com/tempestphp/tempest-framework/pull/69)
- [x] Count — an array with a min/max count
- [x] Email — a valid email address
- [x] Enum — a valid enum value (#68)
- [x] Length — a string with a min/max length
- [x] PhoneNumber — a valid phone number
- [x] Between — an integer between two values, edges included
- [x] Boolean (https://github.com/tempestphp/tempest-framework/pull/76)
- [x] ShouldBeTrue — a value that represents a boolean `true` value. (#64)
- [x] ShouldBeFalse — a value the represents a boolean `false` value (#64)
- [x] Url — a valid URL
- [ ] ~Date — a valid date~ we can use DateTimeFormat?
- [x] Time — a valid time (#67)
- [x] DateTimeFormat — a valid date/date time with a specific format (#53)
- [x] Timestamp — a valid timestamp (integer)
- [x] Timezone - a valid timezone
- [x] Regular Expression — value matches the regular expression passed (#72)
- [x] StartsWith - starts with the specified string (https://github.com/tempestphp/tempest-framework/pull/74)
- [x] EndsWith - ends with the specified string (#82)
- [x] DoesNotStartWith - does not start with the specified string
- [x] DoesNotEndWith - does not end with the specified string
- [x] Uppercase - is UPPERCASE (#82)
- [x] Lowercase - is lowercase (#82)
- [x] UUID - is a valid UUID
- [x] ULID - is a valid ULID
- [x] IP Address - is a valid IP address
- [x] MAC Address - is a valid MAC address
- [x] Password - Allows the specification of password complexity with some sensible defaults
- [ ] ~PasswordNotCompromised - Checks the password against haveibeenpwned~ not right now
- [ ] ~SameAs — this value should be the same as another value, for example for password validation~ skip for now since Rules aren't aware of their larger context.

aidan-casey:
@brendt - it might be helpful to flesh out exactly which rules we'd need to see in place before v1. Gives people something to work toward. 🙂

[...]

PR #40, opened 2024-02-16
Adds validator for email addresses.

aidan-casey:
This adds a dependency (the same used by Symfony and Laravel) and exposes dependency classes as parameters to the attribute. Let me know if you want to make adjustments to any of that.

Related to #3.


brendt:
Looks good 👍

PR #41, opened 2024-02-16
Adds phone number validator.

aidan-casey:
@brendt - This adds a fairly hefty dependency, but I believe it might be worth it to support phone number validation natively.

Related to #3.


brendt:
Nice!

PR #46, opened 2024-02-19
Add url validation rule

alibori:
Hi!

Add url validation rule and test related to [issue #3](https://github.com/tempestphp/tempest-framework/issues/3)

Please, tell me if something has to be improved.

Thanks!


aidan-casey:
Hey, @alibori! Thanks for the pull request!

@brendt, are you aware of any current reasons we shouldn't use `filter_var` for URL validation? Both Laravel and Symfony utilize the same regex pattern to validate URLs, but that might be because of the historical problems with `filter_var`, not any current ones.

aidan-casey:
So... It looks like `filter_var` still has some unresolved problems. For example, [this](https://bugs.php.net/bug.php?id=64173) issue is still a thing which returns true for a string like "https://https://www.github.com".

@alibori - Do you mind updating this rule to utilize [this](https://github.com/symfony/validator/blob/7.0/Constraints/UrlValidator.php) Regex pattern Symfony is using and credit them accordingly?

alibori:
Hi @aidan-casey !

Of course, take a look now 😄 

Thanks!

aidan-casey:
Thank you, @alibori!

PR #53, opened 2024-02-19
Add Date validation rule

alibori:
Hi!

Regarding [Issue #3](https://github.com/tempestphp/tempest-framework/issues/3) this PR adds a Date validation rule and it's tests.

Thanks!


alibori:
Hi @aidan-casey!

Ok, I'll try it tomorrow afternoon after work.

Thanks!

brendt:
About timestamps though: I think this should be a separate rule. I've never ran into a case where one field could contain both a textual date _and_ a timestamp, so I'd keep the string check for this one.

brendt:
> By the way, does PHPStorm have a way to hint a date string as a parameter so as to pop the format reference?

No, only for the built-in functions, as far as I'm aware.

> I was thinking less about one field containing a date and a timestamp and more that the DateTimeFormat rule is simply accepting a valid PHP DateTime format string and validating against that, one of which is a timestamp (U).

The way I approach this is from the user input point of view, which is what we're validating, after all. I can't imagine a field that in some cases contains a timestamp, and in other cases contains a formatted date string. 

If you're dealing with an external API or webhook, and want to use a datetime object while the client side is sending a timestamp, I would validate it as a timestamp, and then cast it to a datetime:


```php
#[Timestamp]
#[CastWith(TimestampToDateTime::class)]
public DateTime $date;
```

Writing it like this though, I think this syntax is a bit confusing 🤔 Mainly because of the naming. It would be clearer if it was this:

```php
#[ValidateAsTimestamp]
#[CastToDateTime]
public DateTime $date;
```

I think this discussion is outside the scope of this PR though. How about we merge this PR in its simplest form, and then reiterate over the design of our validation and casting components?

aidan-casey:
@alibori - If you are able to just update the rule name to `DateTimeFormat` we will get this merged as is for now. Thank you for your work here!

alibori:
Hi @brendt and @aidan-casey !

Perfect, rule renamed. Let's go to next rule/issue 😉 

Thanks!

aidan-casey:
Thanks for all your help, @alibori!

Issue #54, opened 2024-02-20
app folder is included twice as a discovery location


brendt:
```
Folders included in Tempest
- /Users/brentroose/Dev/my-app/tests
- /Users/brentroose/Dev/my-app/vendor/tempest/framework/src
- /Users/brentroose/Dev/my-app/vendor/my/package/src
- /Users/brentroose/Dev/my-app/app
- /Users/brentroose/Dev/my-app/app
```

PR #62, opened 2024-02-20
Fixes all the things Brent broke

aidan-casey:


aidan-casey:
@brendt - The reason the app folder was getting loaded twice (#54) is because it counts as a discovery location by default because of the `composer.json` one level up.

If you think we need to maintain an exception for `App` and `Src`, we should explicitly add them and do a `continue` within the foreach when they are hit.

brendt:
Oh, this actually is better now that we scan for composer. Thanks for fixing it!

PR #64, opened 2024-02-20
Add ShouldBeTrue and ShouldBeFalse validation rules

alibori:
Hi!

Regarding [Issue #3](https://github.com/tempestphp/tempest-framework/issues/3) this PR adds ShouldBeTrue and ShouldBeFalse validation rules and it's tests.

Thanks!

### DIFF
--- a/src/Validation/Rules/ShouldBeFalse.php
+++ b/src/Validation/Rules/ShouldBeFalse.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


aidan-casey:
Thanks @alibori! The only change I would like to make (for now) is to remove null as a falsy value. If @brendt is of the mind that should change we can introduce a second PR to adjust that.

alibori:
Done @aidan-casey 👍 

Thanks!

aidan-casey:
Thanks, @alibori!

Issue #58, opened 2024-02-20
Add Code Coverage Script


aidan-casey:
It would be good to add a code coverage report script that can be run locally and in CI. With things ever changing before v1, it would be helpful to ensuring a) we have necessary tests in place and b) there is no lingering dead code.

aidan-casey:
Uh... done. 😅

PR #66, opened 2024-02-20
Adds coverage composer script

aidan-casey:
This PR adds a coverage script to the composer file. As we have a solution in place (Coveralls, Action, etc.) I will update this to include that as a GitHub action.

Related to #58.


brendt:
Should coverage also run locally? I'm inclined to keep `composer qa` as is, and have coverage only run on CI.

brendt:
I've also approved Coveralls integration FYI

aidan-casey:
@brendt - For the time being, I will move coverage out to a separate command. I treat it as a "qa" thing locally, since I prefer to know what my coverage is before I commit.

PR #67, opened 2024-02-20
Adds time rule

aidan-casey:
This rule uses regex to validate time in standard format (hh:mm xm) or military (24-hour) format (hh:mm). It covers the rule mentioned in #3.


aidan-casey:
@brendt - Can you sanity check this one for me?

brendt:
Sanity has been checked ✅

PR #72, opened 2024-02-20
Adds regex validation rule

aidan-casey:
Adds the regex rule (#3).


tests/bootstrap.php
```diff
@@ -1,7 +1,9 @@
 <?php
 
+declare(strict_types=1);
```
brendt, line 3: What's up with these changes? Should we change anything in our qa script?

PR #76, opened 2024-02-21
Adds boolean validation rule

skrskr:
This PR adds an boolean validation rule.


aidan-casey:
@skrskr thanks for the PR! Can you remove the punctuation from the end of the error message?

brendt:
One consideration to make: `Boolean` might be a confusing name. How about `LikeBoolean` or `Booleanish` or something to indicate that these are _not_ boolean values. I've already merged it for now, but let's consider whether we can improve the name.

PR #77, opened 2024-02-21
Add Ip validation rule

skrskr:
This PR adds an ip validation rule for ipv4 and ipv6.

### DIFF
--- a/src/Validation/Rules/Ip.php
+++ b/src/Validation/Rules/Ip.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Validation\Rules;
+
+use Attribute;
+use Tempest\Interface\Rule;
+
+#[Attribute]
+final readonly class Ip implements Rule
+{
+    public function isValid(mixed $value): bool
+    {
+        return (
+            (filter_var($value, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) ? true : false) ||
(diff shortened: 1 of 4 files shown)


aidan-casey:
@skrskr thanks for the PR! Just a few things...

1. Careful, it looks like you may have carried over the changes from the Boolean PR. Can you just clean that up a little bit?

2. Can we uppercase "IP"?

3. Let's actually split these into three rules (`IP`, `IPv4`, and `IPv6`).

Appreciate your work here!

src/Validation/Rules/Ip.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Attribute;
+use Tempest\Interface\Rule;
+
+#[Attribute]
+final readonly class Ip implements Rule
+{
+    public function isValid(mixed $value): bool
+    {
+        return (
```
brendt, line 15: I think this nested conditional is a bit confusing; maybe rewrite it in several steps so that it's more clear to what it does. I'm ok with multiple `if` statements and early returns.

brendt:
Appreciate the contributions! 💪 Let's first fix the PR though, I suggest checkout out main, updating, and creating a new clean branch, next manually copy over those two files — it's only two files anyway.

```php
git checkout main
git pull origin main
git checkout -b ip-validation-rule-new
<manually copy over the two files from the old branch>
```

Apart from fixing the branch, I left one comment about making the check a bit more readable :) 

I'll close this PR to avoid confusion, definitely send a new one with the updated branch and I'll be happy to merge it :)

PR #78, opened 2024-02-21
refactored the Kernel.php

Mavv3006:


brendt:
Could you list and explain the changes in this PR?

Mavv3006:
Kernel.php:
- In the `initConfig` function I extracted the part of scanning for package config files into a seperate method `scanForPackageConfigFiles`
- From my point of view the function `initDiscovery` is very bloated so I decided to extract also a part of that function into another method with the uninspiring name `executeDiscovery`

The other changes are because of the `composer qa` command and a failing test in the `BooleanTest` class.

brendt:
Gotcha! So, I appreciate the effort, and think there's indeed room for improvement, but I actually think we'll need some more refactoring to have a `discovery:cache` command work. Ie. move discovery to an entirely separate place. I plan on doing work on this myself, possibly during a livestream, so I'll close this PR for now. I might come back to it and merge it into a dev branch as a starting point, but that's something that'll need to wait a couple of days.

PR #84, opened 2024-02-21
feat: add not starts with and not ends with validation rules

vsergiu93:
Not sure if should have just used the StartsWith and EndsWith rules directly and negate their return.
Something like this: 
```php
$startsWithRule = new StartsWith($needle);

return !$startsWithRule->isValid($value);

```

### DIFF
--- a/src/Validation/Rules/DoesNotEndWith.php
+++ b/src/Validation/Rules/DoesNotEndWith.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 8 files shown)


aidan-casey:
Thank you, @vsergiu93! This looks fine to me! I'd rather the logic remains in the rule instead of negating the other rules. 🙂

aidan-casey:
Thanks @vsergiu93! This looks fine to me. I'd rather the logic is self contained.

brendt:
@vsergiu93 good to see you here! 💪

PR #87, opened 2024-02-21
Added UUID rule.

sebhd1:
Added UUID rule as stated in #3.


aidan-casey:
Thank you, @sebhd1! I believe we can get this merged, I just want to do some review before doing so.

aidan-casey:
Hey, @sebhd1! UUIDs are case-insensitive. Can you please update this PR to match the spec? [This](https://uuidtools.com/what-is-uuid) page has some regex information.

sebhd1:
Hey! The regex has the `i` modifier which works for case insensitive matching.

I will add some tests for it!

PR #93, opened 2024-02-21
Validating Timezones

didi955:
First ever open-source contribution :)

Let me know if you have any questions or requests about my changes.

Refers to https://github.com/tempestphp/tempest-framework/issues/3

### DIFF
--- a/src/Validation/Rules/Timezone.php
+++ b/src/Validation/Rules/Timezone.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


aidan-casey:
Congrats on the first PR, it looks great! The only change I would request is can you accept timezoneGroup and countryCode parameters that default to the [timezone_identifiers_list](https://www.php.net/manual/en/function.timezone-identifiers-list.php) defaults?

didi955:
Thank you, should be good now if I have understood correctly

PR #98, opened 2024-02-22
Expands IP rule

aidan-casey:


aidan-casey:
@brendt - Having to validate IP addresses a lot, I know for a fact that I went crazy on this one. Let me know if it's too crazy and I'll refactor. The alternative, in my mind, is two separate rules for IPv4 / IPv6.

brendt:
```
        private bool $ipv4 = false,
        private bool $ipv6 = false,
        private bool $allowPrivateRange = true,
        private bool $allowReservedRange = true
```

This, I don't like 😁 I think one rule is best, but we need another way to configure it.

Just some quick thinking aloud:

```php
#[IP(versions: [IP::v4, IP::v6])]
```

But what about the other flags then… idk…

I do think the default should be both versions on true.

aidan-casey:
I know, I know. 🙈 

Technically it defaults to both versions being true. Without specifying one specifically, it does not add a version flag, which just validates any ol' IP address. The default to false is a UX thing because then you have to manually turn off the versions you don't want instead of simply specifying the one you do.

Honestly, the worst thing about this PR is the error messages. 😛

brendt:
> The default to false is a UX thing because then you have to manually turn off the versions you don't want instead of simply specifying the one you do.

So if both are false, both are actually true? I don't think that's clear 😁 What about this?

```php
    public function __construct(
        private array $versions = [IpVersion::v4, IpVersion::v6]
        private bool $allowPrivateRange = true,
        private bool $allowReservedRange = true
    ) {
```

`IpVersion` (or should it be `IPVersion`?) is an enum. Maybe, for convenience, we can do something like this, although I don't know if it's possible?

```php
class IP
{
    public const IpVersion v4 = IpVersion::v4;
    public const IpVersion v6 = IpVersion::v6;
}
```

So that you can write `IP::v4` while still using the enum?

brendt:
I don't think that works, unfortunately :/

brendt:
Scratch that, it actually works, I was using typed constants in 8.2, which didn't work. But this does:

```php
enum IpVersion: string
{
    case v4 = 'v4';
    case v6 = 'v6';
}

class Ip
{
    public const v4 = IpVersion::v4;
    public const v6 = IpVersion::v6;
}
```

aidan-casey:
Yeah, I had gone down that road early on. Was trying to avoid introducing an enum. 😁 

Let me think on this one a little bit. I'm almost leaning toward splitting them out into different rules, because I _think_ that's what I would do if it were a value object.

Then down the line we can really easily introduce a `#[AnyOf()]` rule once we figure out the messaging part.

brendt:
`#[AnyOf]` was my second thought, but I couldn't come up with a proper name. I like `AnyOf` though, so I'm fine with that approach as well 👍

On the other hand, I think most of the time people would want to validate either v4 or v6. So maybe we combine them?

- `#[Ipv4]`
- `#[Ipv6]`
- `#[Ip]`

PS: according to our naming conventions, `Ipv4` and `Ipv5` should be written camelcase, because it's longer than 3 characters, but `#[IP]` should be written uppercase. I don't like that disambiguity 😅

aidan-casey:
I'm not opposed to structuring it like that. By the way, we need to figure out an approach for multiple error messages (see #122).

I think we are in compliance regarding the acronym. The acronym is IP (Internet Protocol) while vX is a standard format for versions. I can see both sides of it though. 😛

brendt:
I made an issue for the multiple messages: https://github.com/tempestphp/tempest-framework/issues/127

So, what's our decision on naming these classes? `#[IPv4]` and `#[IP]` ?

aidan-casey:
ChatGPT says we're good. 😎 

Screenshot 2024-02-23 at 12 04 35 AM

Issue #73, opened 2024-02-20
Add status to response


aidan-casey:
I am not seeing anywhere that we are truly translating the response status to the client. We should get statuses working correctly.

PR #100, opened 2024-02-22
Fixes response status code

aidan-casey:


aidan-casey:
This PR fixes status codes (#73). There is a lot of refactoring I want to do here, but want to make sure @brendt is good with this general direction before going down that road.

I did add the status code to the `response` function. Because the `GenericResponse` class requires it, it seems like we should allow it to be set.

src/Http/IsResponse.php
```diff
@@ -63,6 +63,48 @@ public function status(Status $status): self
         return $this;
     }
 
+    public function sendHeaders(): self
```
brendt, line 66: I don't think a response should send itself. The response is a data object, sending that response is someone else's responsibility. It doesn't have to be in `HttpApplication`, it can be a separate class, but it should not be within Response classes

brendt (changes requested):
Only one thing I think we need to finetune further :)

src/Http/IsResponse.php
aidan-casey: I thought this might happen. 🙃 

What do you think about a [`ResponseEmitter`](https://github.com/slimphp/Slim/blob/4.x/Slim/ResponseEmitter.php) similar to `Slim`?

brendt: Sounds perfect 👍

src/Application/Kernel.php
```diff
@@ -65,6 +66,7 @@ private function initContainer(): Container
             ->singleton(CommandBus::class, fn () => $container->get(GenericCommandBus::class))
             ->addInitializer(new ServerInitializer())
             ->addInitializer(new RequestInitializer())
+            ->addInitializer(new ResponseSenderInitializer())
```
brendt, line 69: You don't need to manually add it here, you could add `#[InitializedBy]` on the `ResponseSender` interface. The reason some initializers are listed here, is because these initializers don't initialize one specific class. They implement an additional `CanInitialize` interface, which adds more dynamic behaviour. This is needed in some cases like for example when you're initializing the current request.

There might be some more intializers in this list that can be refactored to the `#[InitializedBy]` approach btw, I'll look into that after this one is merged.

src/Http/GenericResponseSender.php
```diff
@@ -0,0 +1,42 @@
...
+        if (headers_sent()) {
+            return;
+        }
+
+        foreach ($response->getHeaders() as $key => $value) {
+            header("{$key}: {$value}");
+        }
+
+        /**
+         * TODO: Is there a reason to manually set the headers
```
brendt, line 34: How do you mean? This http_response_code only sets the response code and no headers? https://www.php.net/manual/en/function.http-response-code.php

src/Http/ResponseSender.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Tempest\Http;
+
+interface ResponseSender
```
brendt, line 7: This is the place where you'd add `#[InnitializedBy(ResponseSenderInitializer::class)]`

src/Http/GenericResponseSender.php
aidan-casey: I wanted to look into whether there was any difference between using this function and assembling the status code header ourselves. Most major routers Symfony / Slim / etc. are building this manually instead of using this function. That may very well be legacy code, but I wanted to confirm.

src/Http/ResponseSender.php
aidan-casey: I will revisit this. I had initially had it set up that way but it was not loading in. Not sure if I was doing something wrong or if something wasn't working right with discovery.

aidan-casey: This appears to be working. 👍🏼

aidan-casey:
@brendt - Are you generally on board with the approach in this PR? If so, I can add some tests, etc. before we merge.

src/Http/GenericResponseSender.php
brendt: Gotcha, I'm fine either way

brendt:
@aidan-casey definitely, feel free to go ahead :)

PR #111, opened 2024-02-22
Add HTTP middleware support

### LINKED ISSUE #4: Route Middleware support
- Route attributes should be able to define middleware
- Middleware should implement a dedicated `HttpMiddleware` interface
- We might need to refactor the existing `Middleware` interface to be named `CommandMiddleware`
- We might need a way to define middleware groups for convenience, they are resolved via an intializer
- The router should send the request through the middleware stack

brendt:
Resolves #4

### DIFF
--- a/src/Http/GenericRouter.php
+++ b/src/Http/GenericRouter.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Http;
 
+use Closure;
 use ReflectionClass;
 use Tempest\AppConfig;
 use function Tempest\attribute;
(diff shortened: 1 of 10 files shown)


src/Http/GenericRouter.php
```diff
@@ -82,6 +77,25 @@ public function dispatch(Request $request): Response
...
+    ): Closure {
+        $callable = fn (Request $request) => $this->container->call(
+            $this->container->get($route->handler->getDeclaringClass()->getName()),
+            $route->handler->getName(),
+            ...$routeParams,
+        );
+
+        $middlewareStack = $route->middleware;
+
+        while ($middlewareClass = array_pop($middlewareStack)) {
```
aidan-casey, line 92: Is there a reason we are using `while` here? My biggest concern is that it looks like we are prioritizing the middleware registered _last_, but we might actually want to be prioritizing middleware registered _first_?

`foreach` is generally faster too, though the speed difference is probably so minimal it is not worth mentioning.

brendt: Well you _need_ to apply the middleware in reverse order, so either your first reverse the array and then loop over it, or you use while + array_pop. I doubt the the reverse + foreach approach would be noticeably faster.

PR #114, opened 2024-02-22
test: improve test coverage for length validation rule 

vsergiu93:
@brendt I just saw that you assigned the issue about improving coverage to you so if you already improved this or in the process of doing so just close this one.

Thanks

### DIFF
--- a/tests/Validation/Rules/LengthTest.php
+++ b/tests/Validation/Rules/LengthTest.php
@@ -9,27 +9,117 @@ use Tempest\Validation\Rules\Length;
 
 class LengthTest extends TestCase
 {
-    /** @test */
-    public function test_length()
+    /**
+     * @dataProvider dataSets
+     */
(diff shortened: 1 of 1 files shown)


aidan-casey:
Hey @vsergiu93! Thank you for the PR here. I think this is an excellent improvement.

Your ideas around architecture testing are good as well. Especially for things like validation rules. It'd be cool if we could check certain things like whether they are an attribute in the `Rules` folder. 🙂 

I'll let @brendt speak to Pest, but I don't believe we are planning on heading that direction anytime soon right now. It is important to note however, that Pest's architecture plugin is simply a wrapper around `ta-tikoma/phpunit-architecture-test`.

brendt:
Yeah I was actually thinking on whether Pest would be a good addition 😅

I'm not sure yet, it'll need some more discussing, but right now there are other topics we need to settle first. 

Thanks for the PR though, appreciate it!

PR #118, opened 2024-02-22
feat: add json validation rule

vsergiu93:
I know is not in the list, but I think it might be useful to have it. 

Also it might look ugly, but wanted to use the new function json_validate, maybe instead of adding this polyfill I could just move the body of the function inside the isValid method.

### DIFF
--- a/src/Validation/Rules/JSON.php
+++ b/src/Validation/Rules/JSON.php
@@ -9,9 +9,9 @@ use Tempest\Validation\Rule;
 
 /**
  * We need to polyfill this, because at the moment the minimum required version of php for tempest is ^8.2,
- *  and json_validate was introduced in 8.3
+ * and json_validate was introduced in 8.3
  * @TODO remove this polyfill once the minimum php version is >= than 8.3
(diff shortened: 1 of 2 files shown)


vsergiu93:
@aidan-casey Do you think it would be useful to add [php 8.3 to phpunit action](https://github.com/tempestphp/tempest-framework/blob/main/.github/workflows/quality-control.yml#L65)?

aidan-casey:
Thanks for the PR, @vsergiu93! I actually meant to add this one to the list.

Let me sit on this implementation for a day. I was just speaking with @brendt about bumping the Tempest PHP version (because we can!) to 8.3 anyway.

As far as adding 8.3 to the matrix, I think that's an awesome idea! I would merge a PR for that.

vsergiu93:
Sure, just let me know if you want to change something. 

Thanks

aidan-casey:
@vsergiu93 - Go ahead and convert this to entirely `json_validate`. We will be bumping the PHP dependency. 🙂

vsergiu93:
@aidan-casey Done :)

aidan-casey:
Love it! Thank you.

PR #122, opened 2024-02-22
Validate password

AdamWills:
Added some basic password validation (addresses #3)

Have selected most rules to be disabled by default. NIST guidelines don't recommend any sort of password enforcement. Instead they recommend using a list of compromised passwords to compare against (which we are handling separately).

I'm not sure if we want to have different messages for different issues - but then we may want message() to allow returning an array of errors. Seems like it's something a little out of scope for this one though.

Source: https://pages.nist.gov/800-63-FAQ/#q-b06

### DIFF
--- a/src/Validation/Rules/Password.php
+++ b/src/Validation/Rules/Password.php
@@ -53,6 +53,31 @@ final readonly class Password implements Rule
 
     public function message(): string
     {
-        return "Value should be a valid password";
+        $messages = ["at least {$this->min} characters"];
+
+        if ($this->mixedCase) {
+            $messages[] = 'at least one uppercase and one lowercase letter';
+        }
+        if ($this->numbers) {
+            $messages[] = 'at least one number';
+        }
+        if ($this->letters) {
+            $messages[] = 'at least one letter';
+        }
+        if ($this->symbols) {
+            $messages[] = 'at least one symbol';
+        }
+
+        return 'Value should contain ' . $this->natural_language_join($messages);
+    }
+
+    private function natural_language_join(array $list)
+    {
+        $last = array_pop($list);
+        if ($list) {
+            return implode(', ', $list) . ' ' . 'and' . ' ' . $last;
+        }
+
+        return $last;
     }
 }
--- a/src/Validation/Rules/Password.php
+++ b/src/Validation/Rules/Password.php
@@ -11,18 +11,15 @@ use Tempest\Validation\Rule;
 final readonly class Password implements Rule
 {
     private int $min;
-    private ?int $max;
 
     public function __construct(
         int $min = 12,
-        ?int $max = null,
         private bool $mixedCase = false,
         private bool $numbers = false,
         private bool $letters = false,
         private bool $symbols = false,
     ) {
         $this->min = max(1, $min);
-        $this->max = $max ? max($this->min, $max) : null;
     }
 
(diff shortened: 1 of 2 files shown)


src/Validation/Rules/Password.php
```diff
@@ -0,0 +1,65 @@
...
+use Tempest\Validation\Rule;
+
+#[Attribute]
+final readonly class Password implements Rule
+{
+    private int $min;
+    private ?int $max;
+
+    public function __construct(
+        int $min = 8,
```
aidan-casey, line 17: Despite NIST recommendations, let's revise this to be `12` as our default.

src/Validation/Rules/Password.php
```diff
@@ -0,0 +1,65 @@
...
+        ?int $max = null,
```
aidan-casey, line 18: Let's remove `$max` as an option.

src/Validation/Rules/Password.php
```diff
@@ -0,0 +1,65 @@
...
+        private bool $mixedCase = false,
+        private bool $numbers = false,
+        private bool $letters = false,
+        private bool $symbols = false,
+    ) {
```
aidan-casey, line 23: Let's add an option along the lines of `$restrictRepetitiveCharacters` (if you think of naming more user friendly, I'm open to it!)

src/Validation/Rules/Password.php
```diff
@@ -0,0 +1,65 @@
...
+        if ($this->symbols && ! preg_match('/\p{Z}|\p{S}|\p{P}/u', $value)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public function message(): string
+    {
+        return "Value should be a valid password";
```
aidan-casey, line 63: Yeah... let's add more specific messages. 🙈

aidan-casey (changes requested):
Thanks for the PR, @AdamWills! This is a big one to tackle.

I'd like to change just a few things if you don't mind reviewing!

src/Validation/Rules/Password.php
AdamWills: Can you define this in a little more detail?

src/Validation/Rules/Password.php
AdamWills: Perhaps for this PR, I'll return a single, more specific message. If we want to make a broader change to return multiple messages, let's separate that out.

src/Validation/Rules/Password.php
aidan-casey: Actually, ignore this for now. We can revisit later if we need to.

Was thinking things like "aaaaaa"

src/Validation/Rules/Password.php
aidan-casey: That's fine. We can strategize a bit for more complex scenarios like this.

src/Validation/Rules/Password.php
AdamWills: Done!

src/Validation/Rules/Password.php
AdamWills: Done!

AdamWills:
Hm. Looks like there's failing tests on main...

aidan-casey:
Thanks for the updates, @AdamWills! As far as failing tests go... I blame @brendt. 😉

Issue #117, opened 2024-02-22
Use composer.lock (or installed.json) for package discovery


jtheuerkauf:
https://github.com/tempestphp/tempest-framework/blob/b983830f37ea6293eb602c5768fb1c251f5195d8/src/Application/Kernel.php#L76

i'm examining the lock JSON from another project, and i see that every entry in the root `composer.json` includes the packages each _dependency_ requires. So why not use that already compiled list?

The expectations are the same as the original, with some key differences:

* Read ONE file with the added expectation that `composer install|update` is locked. Since that's the most common approach, especially in production, it seems like a good primary strategy.
* For a no-lock, you could fall back to reading the `autoload_psr4.php` file and all the packages' `composer.json` files.
* The lock file can be pretty large, so stuffing it all into a simple array is probably not the best. Perhaps look for an alternative data loading method?

Just a rough sketch of what *might* work:

```php
private function initDiscoveryLocations(): void
{
    if (is_file(path($this->root . '/composer.lock'))) {
        // WARN: Is there a more memory-efficient way to load and access the JSON?
        $lockContent = file_get_contents(path($this->root, 'composer.lock'));
        $composer = json_decode($lockContent, true);

        foreach ($composer['packages'] ?? [] as $package) {
            // WARN: Is this universally true? Does the package name ALWAYS match the "vendor/___" path?
            $packagePath = path($this->root, 'vendor', $package['name']);

            // Tempest must be required by the package, and must define PSR-4 namespace map
            if (isset($package['require']['tempest/framework'], $package['autoload']['psr-4'])
                && is_array($package['autoload']['psr-4'])
            ) {
                foreach ($package['autoload']['psr-4'] as $namespace => $pkgNamespacePath) {
                    $this->appConfig->discoveryLocations[] = new DiscoveryLocation(
                        namespace: $namespace,
                        path: path($packagePath, $pkgNamespacePath),
                    );
                }
            }
        }
    } else {
        // Do the existing logic?
    }
}
```

aidan-casey:
Yeah, I was look at this too. You can also use `vendor/composer/installed.json`. I wish that `vendor/composer/installed.php` had the same info! 🙂 

The problem is it will always be slower because you have to use `json_decode` vs. `str_contains`.

I started to mess around last night with using regex to do this, but it's not an easy pattern to match.

jtheuerkauf:
[...]

PR #125, opened 2024-02-22
#117 Use installed.json for package discovery

jtheuerkauf:
Replace namespace source with root `composer.json` `"autoload"` data and `vendor/composer/installed.json` `"packages"` data

Root `composer.json` includes `"autoload-dev"` conditionally on `ENVIRONMENT` value

### DIFF
--- a/src/Application/Kernel.php
+++ b/src/Application/Kernel.php
@@ -4,11 +4,11 @@ declare(strict_types=1);
 
 namespace Tempest\Application;
 
-use JsonException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use ReflectionClass;
 use Tempest\AppConfig;
+use Tempest\Application\Exceptions\KernelException;
 use Tempest\Commands\CommandBus;
 use Tempest\Commands\GenericCommandBus;
 use Tempest\Console\ConsoleFormatter;
@@ -27,7 +27,6 @@ use Tempest\Http\RouteBindingInitializer;
 use Tempest\Http\Router;
 use Tempest\Http\ServerInitializer;
 use Throwable;
-use function Tempest\env;
 use function Tempest\path;
 
 final readonly class Kernel
@@ -153,21 +152,19 @@ final readonly class Kernel
     private function discoverInstalledPackageLocations(): void
     {
         $composerPath = path($this->root, 'vendor/composer');
-        $installed = path($composerPath, 'installed.json');
+        $installedPath = path($composerPath, 'installed.json');
 
-        if (is_file($installed)) {
-            $installed = $this->loadJsonFile($installed);
+        $installedJson = $this->loadJsonFile($installedPath);
+        $packages = $installedJson['packages'] ?? [];
+        foreach ($packages as $package) {
+            $packagePath = path($composerPath, $package['install-path']);
 
-            foreach ($installed['packages'] ?? [] as $package) {
-                $packagePath = realpath(path($composerPath, $package['install-path']));
-
-                if (isset($package['require']['tempest/framework'], $package['autoload']['psr-4'])
-                    && is_array($package['autoload']['psr-4'])
-                ) {
-                    foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
-                        $namespacePath = path($packagePath, $namespacePath);
-                        $this->addDiscoveryLocation($namespace, $namespacePath);
-                    }
+            $requiresTempest = isset($package['require']['tempest/framework']);
+            $hasPsr4Namespaces = isset($package['autoload']['psr-4']);
+            if ($requiresTempest && $hasPsr4Namespaces) {
+                foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
+                    $namespacePath = path($packagePath, $namespacePath);
+                    $this->addDiscoveryLocation($namespace, $namespacePath);
                 }
             }
         }
@@ -177,13 +174,8 @@ final readonly class Kernel
     {
         $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
 
-        $autoloads = [$composer['autoload']['psr-4'] ?? []];
-        if (env('ENVIRONMENT') === 'production') {
-            $autoloads[] = $composer['autoload']['psr-4'] ?? [];
-        }
-        $autoload = array_merge(...$autoloads);
-
-        foreach ($autoload as $namespace => $path) {
+        $namespaceMap = $composer['autoload']['psr-4'] ?? [];
+        foreach ($namespaceMap as $namespace => $path) {
             $path = path($this->root, $path);
             $this->addDiscoveryLocation($namespace, $path);
         }
@@ -199,7 +191,11 @@ final readonly class Kernel
 
     private function loadJsonFile(string $path): array
     {
-        return json_decode(file_get_contents($path), true)
-               ?? throw new JsonException('Cannot read json file: ' . str_replace($this->root, './', $path));
+        if (!is_file($path)) {
+            $relativePath = str_replace($this->root, './', $path);
+            throw new KernelException(sprintf('Could not locate %s, try running "composer install"', $relativePath));
+        }
+
+        return json_decode(file_get_contents($path), true);
     }
 }
--- a/src/Application/Kernel.php
+++ b/src/Application/Kernel.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Application;
 
+use JsonException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use ReflectionClass;
@@ -25,8 +26,9 @@ use Tempest\Http\RequestInitializer;
 use Tempest\Http\RouteBindingInitializer;
 use Tempest\Http\Router;
 use Tempest\Http\ServerInitializer;
-use function Tempest\path;
 use Throwable;
+use function Tempest\env;
+use function Tempest\path;
 
 final readonly class Kernel
 {
@@ -73,22 +75,8 @@ final readonly class Kernel
 
     private function initDiscoveryLocations(): void
     {
-        $namespaces = require path($this->root, '/vendor/composer/autoload_psr4.php');
-
-        foreach ($namespaces as $namespace => $path) {
-            $composer = @file_get_contents(path($path[0], '/../composer.json'));
-
-            if (! $composer) {
-                continue;
-            }
-
-            if (str_contains($composer, '"tempest/framework"')) {
-                $this->appConfig->discoveryLocations[] = new DiscoveryLocation(
-                    namespace: $namespace,
-                    path: $path[0],
-                );
-            }
-        }
+        $this->discoverTempestNamespaces();
+        $this->discoverInstalledPackageLocations();
     }
 
     private function initConfig(Container $container): void
@@ -160,5 +148,58 @@ final readonly class Kernel
 
             $discovery->storeCache();
         }
+    }
+
+    private function discoverInstalledPackageLocations(): void
+    {
+        $composerPath = path($this->root, 'vendor/composer');
+        $installed = path($composerPath, 'installed.json');
+
+        if (is_file($installed)) {
+            $installed = $this->loadJsonFile($installed);
+
+            foreach ($installed['packages'] ?? [] as $package) {
+                $packagePath = realpath(path($composerPath, $package['install-path']));
+
+                if (isset($package['require']['tempest/framework'], $package['autoload']['psr-4'])
+                    && is_array($package['autoload']['psr-4'])
+                ) {
+                    foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
+                        $namespacePath = path($packagePath, $namespacePath);
+                        $this->addDiscoveryLocation($namespace, $namespacePath);
+                    }
+                }
+            }
+        }
+    }
+
+    private function discoverTempestNamespaces(): void
+    {
+        $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
+
+        $autoloads = [$composer['autoload']['psr-4'] ?? []];
+        if (env('ENVIRONMENT') === 'production') {
+            $autoloads[] = $composer['autoload']['psr-4'] ?? [];
+        }
+        $autoload = array_merge(...$autoloads);
+
+        foreach ($autoload as $namespace => $path) {
+            $path = path($this->root, $path);
+            $this->addDiscoveryLocation($namespace, $path);
+        }
+    }
+
+    private function addDiscoveryLocation(string $namespace, string $path): void
+    {
+        $this->appConfig->discoveryLocations[] = new DiscoveryLocation(
+            namespace: $namespace,
+            path     : $path,
+        );
+    }
+
+    private function loadJsonFile(string $path): array
+    {
+        return json_decode(file_get_contents($path), true)
+               ?? throw new JsonException('Cannot read json file: ' . str_replace($this->root, './', $path));
     }
 }
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,4 @@
+# See \Tempest\Application\Environment for supported values
 ENVIRONMENT=production
+
 DISCOVERY_CACHE=false
--- a/src/Application/Exceptions/KernelException.php
+++ b/src/Application/Exceptions/KernelException.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Tempest\Application\Exceptions;
+
+use Exception;
+
+class KernelException extends Exception
+{
+
+}


src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
         }
     }
+
+    private function discoverInstalledPackageLocations(): void
+    {
+        $composerPath = path($this->root, 'vendor/composer');
+        $installed = path($composerPath, 'installed.json');
+
+        if (is_file($installed)) {
+            $installed = $this->loadJsonFile($installed);
```
aidan-casey, line 159: So here, I'd be curious what something to the effect of the following would give you. 🙂 

```php
$installedContents = file_get_contents($installed);

// Splits each package section {} into it's own block
preg_match_all('/(?<=\s{8})\{[\s\S]+?\n\s{8}\}/', $installedContents, $packages);

foreach ($packages as $package) {
    // Decodes for the specific package
    $package = json_decode($package, true);
}
```

jtheuerkauf: i'll give it a shot and see what happens.

jtheuerkauf: ## Results

Also added `laravel/framework` to the Tempest `composer.json` (HOLY CRAP Symfony, are you entirely built on polyfills???)

Using the regex:
```php
if (is_file($installed)) {
    $installed = file_get_contents($installed);
    preg_match_all('~(?<=\s{8})\{[\s\S]+?\n\s{8}}~', $installed, $packages);

    foreach ($packages[0] as $package) {
        $package = json_decode($package, true);
        $packagePath = realpath(path($composerPath, $package['install-path']));
```

Benchmarks for regex and my version with Laravel included...

My version:
```
$ for ((i = 0; i < 10; i++)); do php tempest; done

0.22264194488525
0.22401404380798
0.27412104606628
0.24248886108398
0.22104001045227
0.26659607887268
0.20401000976562
0.2078320980072
0.18999195098877
0.23425197601318

(Avg: 0.228 sec)
```

Regex:
```
$ for ((i = 0; i < 10; i++)); do php tempest; done

0.27564597129822
0.25507497787476
0.26790499687195
0.21789503097534
0.20109605789185
0.19769716262817
0.25411415100098
0.21002793312073
0.22865700721741
0.25373101234436

(Avg: 0.236 sec)
```

Roughly speaking, very little performance change. By the way, that's ~100 total package entries in `installed.json` so i imagine it gives a better than average estimation. 😃

jtheuerkauf: i've seen references to more advanced JSON handlers that might help performance, but haven't worked with any of them previously, so maybe it's something we can try in a later revision. Let me know whether to update this PR to the regex or not.

aidan-casey: Okay, so I am getting very different numbers from you. I started the evening genuinely curious and confused and ended the evening over-optimizing microseconds. 😅 

For some reason I cannot replicate your timings at all. I don't know if it is a PHP version mismatch (I am on 8.3) or if you are running on a slower disk (HDD?). See below for the timing I am currently getting and the "optimized" code I am currently using.


```bash
# json_decode
0.0014290809631348n
0.0011060237884521n
0.0010828971862793n
0.0011448860168457n
0.0010919570922852n
0.0010881423950195n
0.0010750293731689n
0.0010759830474854n
0.0011980533599854n
0.0010321140289307n

# RegEx
0.00080204010009766n
0.000762939453125n
0.00078010559082031n
0.00077986717224121n
0.00078606605529785n
0.00073504447937012n
0.0007779598236084n
0.000762939453125n
0.00078701972961426n
0.00080013275146484n
```

```php
private function discoverInstalledPackageLocations(): void
{
    $startup = microtime(true);

    $composerPath = path($this->root, 'vendor/composer');
    $installed = path($composerPath, 'installed.json');

    if (! is_file($installed)) {
        return;
    }

    $content = file_get_contents($installed);
    $packages = json_decode($content, true)['packages'];

    foreach ($packages as $package) {
        $packagePath = realpath(path($composerPath, $package['install-path']));

        if (
            ! isset($package['require']['tempest/framework']) ||
            empty($package['autoload']['psr-4'])
        ) {
            continue;
        }

        foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
            $namespacePath = path($packagePath, $namespacePath);
            $this->addDiscoveryLocation($namespace, $namespacePath);
        }
    }

    die(microtime(true) - $startup);
}
```

```php
private function discoverInstalledPackageLocations(): void
{
    $startup = microtime(true);

    $composerPath = path($this->root, 'vendor/composer');
    $installed = path($composerPath, 'installed.json');

    if (! is_file($installed)) {
        return;
    }

    $content = file_get_contents($installed);

    preg_match_all('/\s{8}\{.*?tempest\/framework.*?\n\s{8}\}/s', $content, $packages);

    foreach ($packages[0] as $package) {
        $package = json_decode($package, true);
        $packagePath = realpath(path($composerPath, $package['install-path']));

        if (
            ! isset($package['require']['tempest/framework']) ||
            empty($package['autoload']['psr-4'])
        ) {
            continue;
        }

        foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
            $namespacePath = path($packagePath, $namespacePath);
            $this->addDiscoveryLocation($namespace, $namespacePath);
        }
    }

    die(microtime(true) - $startup);
}
```

Full disclosure, I did adjust my RegEx slightly so that now it is _only_ pulling the packages that mention Tempest in. It did not improve speed dramatically, but it is a difference from yours.

brendt: I suspect the difference for regex probably has to do with differences in CPU. I still don't think regex is the way to go, as we'll probably miss edge cases leading to subtle bugs.

jtheuerkauf: Yeah, very weird. i'm running 8.3 through WSL. i disabled all extensions to the minimum that would run the script (mbstring and pdo)...

Your version with `json_decode()`:
```
$ for ((i = 0; i < 10; i++)); do php tempest; done

0.21244215965271
0.19042801856995
0.19336795806885
0.19525384902954
0.179358959198
0.20081305503845
0.20244216918945
0.18885803222656
0.22774887084961
0.21510100364685
```

Your version with regex:
```
$ for ((i = 0; i < 10; i++)); do php tempest; done
0.57280683517456
0.57812809944153
0.58618807792664
0.56667304039001
0.58222198486328
0.5756139755249
0.5791130065918
0.5681848526001
0.56858205795288
0.57621693611145
```
That difference from your numbers is _crazy_ ...

i'll put up results using 8.3 from PowerShell in a separate comment.

aidan-casey: @jtheuerkauf - Are you using WSL 2 and have the files on /mnt/c? If so, you are probably running into the slowness of the 9P protocol.

jtheuerkauf: Files are on a different mount, but otherwise normal. You may be right though. Here's 8.3 from PS:

`json_decode()`:
```
> for ($i = 0; $i -lt 10; $i++) { php .\tempest }

0.004680871963501
0.0046839714050293
0.0047390460968018
0.0050721168518066
0.0051999092102051
0.0046918392181396
0.0047049522399902
0.0047979354858398
0.0047011375427246
0.0047698020935059

Avg: 0.0048041582107544 sec
```


Regex:
```
> for ($i = 0; $i -lt 10; $i++) { php .\tempest }

0.00072097778320312
0.00059008598327637
0.00069689750671387
0.00060105323791504
0.00063681602478027
0.00059008598327637
0.00056695938110352
0.0006558895111084
0.00058507919311523
0.00063514709472656

Avg: 0.00062789916992187 sec
```

---

So, at the scale of thousandths of a second, the RegEx is faster. However, i lean toward @brendt's concern about reliable accuracy. But if the time difference is convincing enough, i'm open to switching to it.

My last thoughts on it: practically speaking, once caching is turned on, the time difference only matters once, right? Hell, discovery and caching the packages could be part of the composer post-install scripts, couldn't it? It never even has to occur for actual application requests/commands.

---

Anyway, how do we feel about this version? Any specific changes to make?

brendt: I say we go with the non-regex approach. Sorry @aidan-casey 🫣

Regarding cache: local development won't have it enabled, so we need it to be performant enough still. I think the current numbers look good enough

brendt: Vendor discovery can be cached locally btw, we're already planning on adding a discovery:cache command for that

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+            }
+        }
+    }
+
+    private function discoverTempestNamespaces(): void
+    {
+        $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
+
+        $autoloads = [$composer['autoload']['psr-4'] ?? []];
+        if (env('ENVIRONMENT') === 'production') {
```
brendt, line 181: We shouldn't use env directly, instead use `AppConfig`

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+                }
+            }
+        }
+    }
+
+    private function discoverTempestNamespaces(): void
+    {
+        $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
+
+        $autoloads = [$composer['autoload']['psr-4'] ?? []];
```
brendt, line 180: I don't understand what this part of the code does? Why the check on environment? What's the difference between these lines?

```
$autoloads = [$composer['autoload']['psr-4'] ?? []];

// …

$autoloads[] = $composer['autoload']['psr-4'] ?? [];
```

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+    {
+        $this->appConfig->discoveryLocations[] = new DiscoveryLocation(
+            namespace: $namespace,
+            path     : $path,
+        );
+    }
+
+    private function loadJsonFile(string $path): array
+    {
+        return json_decode(file_get_contents($path), true)
```
brendt, line 202: Does this need a separate function?

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+               ?? throw new JsonException('Cannot read json file: ' . str_replace($this->root, './', $path));
```
brendt, line 203: I would write a more specific exception message. Probably create a new `KernelException`. 

> Could not locate installed.json in {$path}, try running `composer install`

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+
+    private function discoverTempestNamespaces(): void
+    {
+        $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
+
+        $autoloads = [$composer['autoload']['psr-4'] ?? []];
+        if (env('ENVIRONMENT') === 'production') {
+            $autoloads[] = $composer['autoload']['psr-4'] ?? [];
+        }
+        $autoload = array_merge(...$autoloads);
```
brendt, line 184: What's happening here exactly? Looks a bit weird to me that we first store everything in `$autoloads` and then merging it? Are you trying to flatten the array?

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+    private function discoverInstalledPackageLocations(): void
+    {
+        $composerPath = path($this->root, 'vendor/composer');
+        $installed = path($composerPath, 'installed.json');
+
+        if (is_file($installed)) {
+            $installed = $this->loadJsonFile($installed);
+
+            foreach ($installed['packages'] ?? [] as $package) {
+                $packagePath = realpath(path($composerPath, $package['install-path']));
```
brendt, line 162: Why wrapping in `realpath`? If this is necessary, we might need to add it to the `path` helper function so that it is consistent.

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
             $discovery->storeCache();
         }
     }
+
+    private function discoverInstalledPackageLocations(): void
+    {
+        $composerPath = path($this->root, 'vendor/composer');
+        $installed = path($composerPath, 'installed.json');
```
brendt, line 156: rename to `$installedPath` for clarity

src/Application/Kernel.php
```diff
@@ -161,4 +149,57 @@ private function initDiscovery(Container $container): void
...
+
+        if (is_file($installed)) {
+            $installed = $this->loadJsonFile($installed);
+
+            foreach ($installed['packages'] ?? [] as $package) {
+                $packagePath = realpath(path($composerPath, $package['install-path']));
+
+                if (isset($package['require']['tempest/framework'], $package['autoload']['psr-4'])
```
brendt, line 164: I would refactor this for clarity into two boolean variables:

```php
$requiresTempest = isset($package['require']['tempest/framework']);
$hasRootNamespaces = isset($package['autoload']['psr-4']) && is_array($package['autoload']['psr-4']);

if ($requiresTempest && $hasRootNamespaces) {
```

brendt (changes requested):
I've added some remarks, could you take a look?

Good work btw, really appreciate it 👍

src/Application/Kernel.php
aidan-casey: I don't think we need the check for whether it's an array.

src/Application/Kernel.php
aidan-casey: Hey, I get it. You guys don't care about speed. 😢🙃

src/Application/Kernel.php
brendt: True, composer already validates that

src/Application/Kernel.php
jtheuerkauf: Oops. i left out part of the key name inside the `if`. Should look like:

```php
if ( /* not production */ ) {
    $autoloads[] = $composer['autoload-dev']['psr-4'] ?? [];
}
```

We always want the production `"autoload"` set of PSR-4 namespaces. But the original code was also including the framework's `"Tests\\Tempest\\": "tests/"` namespace, which is in `"autoload-dev"`.

I didn't want to include dev namespaces in a production environment, but to keep continuity with the original code, this adds the `"autoload-dev"` section for non-production.

src/Application/Kernel.php
jtheuerkauf: i wouldn't mind seeing it added to `path()` -- i get nervous with relative path segments, and when you try to read them in output or debugging it's easy to get confused. i prefer to have the resolved actual path whenever possible.

Your call: i can make the change to `path()` or remove `realpath()` here.

src/Application/Kernel.php
jtheuerkauf: Agreed on removing `is_array()` ... i'm too accustomed to code with unreliable data structures in it. i should _hope_ composer is more thoughtful than that. 😉 

As for separating the `isset()` ... i agree the variables describe the condition, and i'll make the change.

For what it's worth, i try to balance concise code with descriptive code. Separate variables, each calling `isset()`, felt like variable pollution. Whereas one call that checks both felt sufficient. Personal preference, i don't feel strongly enough about it to protest. 😃

src/Application/Kernel.php
brendt: I don't think we should take that into account here, we could manually add items to appconfig during test setup

jtheuerkauf: i'll cover this in your new question below.

src/Application/Kernel.php
jtheuerkauf: The original functionality that came from using `autoload_psr4.php` was _all_ the defined namespaces are added to discovery - including those in `"autoload-dev"`.

i wanted to preserve that original logic, but it doesn't make sense (to me) to include development level namespaces in production.

The merge keeps the final array unique, to prevent accidental duplication of a namespace key. Duplication might cause other problems, but this doesn't have to be one of them. 😉 

So, all of that said, i can alter it any way you wish. i can always include or always exclude the dev namespaces, which removes the environment check. i could also flip it so the dev set is added first, then production overwrites any accidental duplicate entry... Whatever you'd prefer.

```php
/*
 * Given this structure:
 * {
 *   "autoload": {
 *     "psr-4": {
 *       "Foo\\": "foo",
 *       "Bar\\": "bar"
 *     },
 *  },
 * "autoload-dev": {
 *   "psr-4": {
 *     "Tests\\": "tests",
 *     "Bar\\": "bar-tests" (hypothetical duplicate namespace!)
 *   }
 * }
 */


// Always include dev (prod overwrites dev)
$autoloads = array_merge($composer['autoload-dev']['psr-4'], $composer['autoload']['psr-4'])
/*
 * Result:
 * {
 *   "Tests\\": "tests",
 *   "Foo\\": "foo",
 *   "Bar\\": "bar"
 * }
 */

// Always include dev (dev overwrites prod)
$autoloads = array_merge($composer['autoload']['psr-4'], $composer['autoload-dev']['psr-4'])
/*
 * Result:
 * {
 *   "Foo\\": "foo",
 *   "Tests\\": "tests",
 *   "Bar\\": "bar-tests"
 * }
 */

// Always exclude dev
$autoloads = $composer['autoload']['psr-4'];
/*
 * Result:
 * {
 *   "Foo\\": "foo",
 *   "Bar\\": "bar"
 * }
 */

// Conditionally include dev (dev overwrites prod - current submission)
$autoloads = [ $composer['autoload']['psr-4'] ];
// if not-prod add dev section and merge the two sections to a single set of namespaces
/*
 * Result (prod):
 * {
 *   "Foo\\": "foo",
 *   "Bar\\": "bar"
 * }
 * Result (not prod):
 * {
 *   "Foo\\": "foo",
 *   "Tests\\": "tests",
 *   "Bar\\": "bar-tests"
 * }
 */

// Conditionally include dev (prod overwrites dev)
if ( /* not-production */ ) {
    $autoloads = [ $composer['autoload-dev']['psr-4'] ];
}
$autoloads[] = $composer['autoload']['psr-4'];
$autoloads = array_merge(...$autoloads);
/*
 * Result (prod):
 * {
 *   "Foo\\": "foo",
 *   "Bar\\": "bar"
 * }
 * Result (not prod):
 * {
 *   "Tests\\": "tests",
 *   "Foo\\": "foo",
 *   "Bar\\": "bar"
 * }
 */
```

Perhaps i'm really over-thinking it, but it seems like an important detail, so i want to be thorough...

src/Application/Kernel.php
jtheuerkauf: i took your suggestion for a `KernelException` class to throw when the target file isn't found, and put it here. It cleans this up a little _and_ the spot where `is_file()` was being checked. One less nesting level above, and more consistent flow here: always check for the file, then return the decoded file contents.

Does that make more sense?

src/Application/Kernel.php
jtheuerkauf: Hm. `realpath()` fails on glob patterns like `.../tempest-framework/src/Config/**.php` (coming from `Kernel::initConfig()`

That's unfortunate.

src/Application/Kernel.php
brendt: I think the best approach is to never look at autoload-dev for discovery. The only time we need it is for testing of the framework itself, which we can easily fix within `Tempest\Tests\TestCase::setUp()`

Don't worry about tests breaking if you omit the autoload-dev part, I'm happy to fix that once the rest is in place :) I need to tinker around with it a bit myself, so I can't give you a ready-solution, but don't worry about it :) 

Are you ok with omitting all autoload-dev detection?

src/Application/Kernel.php
brendt: Oh, that's unfortunate indeed. How about we just use `path` as is for now, and we tackle `realpath` problems in a separate PR or issue after this PR gets merged. I like it when PRs focus on one core problem and not diverge too much from the initial goal :)

brendt:
@aidan-casey it might be a good idea, but first let's focus on get this merged in, then think about possible extra refactorings :)

src/Application/Kernel.php
jtheuerkauf: Roger that. It could be as simple as `return realpath($path) ?: $path;` but i don't feel great about "your input has been changed, maybe" results.

src/Application/Kernel.php
jtheuerkauf: i'm _completely_ OK with leaving the dev info out of it. 😄

jtheuerkauf:
New commit pushed.

i agree there's a LOT of code dedicated to discovery in the Kernel right now, and maybe it belongs in a Bootloader or something (i've been tinkering in Spiral lately, so i see a lot of "register Bootloaders in Kernel" so the boot tasks (e.g. Discovery) can be compartmentalized. Probably not the only framework to do it, just what i've been practicing with lately.

jtheuerkauf:
Thanks @brendt! Glad i could contribute something.

PR #134, opened 2024-02-23
Add test helpers

aidan-casey:
@brendt Not sure if you have any input here. I am somewhat following Laravel's syntax for testing for now (there may be some things we change along the way).

A few things that I am running into:

1. We do not currently have a way to set priority for an initializer, right? What's the best way to override a default implementation?
2. Due to the application self-processing the request, we basically cut it out entirely in the test helper. Not sure if we are just fine with that or if you have other thoughts here.

### DIFF
--- a/src/Testing/Http/TestResponseSenderInitializer.php
+++ b/src/Testing/Http/TestResponseSenderInitializer.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Testing\Http;
+
+use Tempest\Container\CanInitialize;
+use Tempest\Container\Container;
+use Tempest\Http\Response;
+use Tempest\Http\ResponseSender;
+
+final class TestResponseSenderInitializer implements CanInitialize
+{
+    public function canInitialize(string $className): bool
+    {
+        return in_array($className, [Response::class, TestResponse::class]);
+    }
+
+    public function initialize(string $className, Container $container): object
+    {
+        $responseSender = new TestResponseSender();
+
+        $container->singleton(ResponseSender::class, fn () => $responseSender);
+
+        return $responseSender;
+    }
+}
--- a/src/Testing/Http/TestsHttpRequests.php
+++ b/src/Testing/Http/TestsHttpRequests.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Testing\Http;
+
+use Tempest\Http\GenericRequest;
+use Tempest\Http\Method;
+use Tempest\Http\Request;
+use Tempest\Http\ResponseSender;
+use Tempest\Http\Router;
+
+trait TestsHttpRequests
+{
+    protected function get(string $path): TestResponse
+    {
+        return $this->performRequest(
+            new GenericRequest(
+                method: Method::GET,
+                uri: $path,
+                body: []
+            )
+        );
+    }
+
+    protected function post(string $path): TestResponse
+    {
+        return $this->performRequest(
+            new GenericRequest(
+                method: Method::POST,
+                uri: $path,
+                body: []
+            )
+        );
+    }
+
+    private function performRequest(Request $request): TestResponse
+    {
+        // Register our test request in the container.
+        $this->container->singleton(Request::class, fn () => $request);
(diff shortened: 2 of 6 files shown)


brendt:
I think you need to rebase, since this branch contains the ResponseSender changes?

brendt:
> We do not currently have a way to set priority for an initializer, right? What's the best way to override a default implementation?

What I've been doing is manually registering singletons in test cases, those get priority. Apart from that, we currently don't have anything else.

> Due to the application self-processing the request, we basically cut it out entirely in the test helper. Not sure if we are just fine with that or if you have other thoughts here.

Could you elaborate about what we're cutting out exactly?

aidan-casey:
> What I've been doing is manually registering singletons in test cases, those get priority. Apart from that, we currently don't have anything else.

This is probably something worth exploring a bit. It _might_ be enough to put the default initializers at the end of the array, or have them configurable.

> Could you elaborate about what we're cutting out exactly?

Instead of calling `HttpApplication::run` we have to build it up manually like so:

```php
        // Register our test request in the container.
        $this->container->singleton(Request::class, fn () => $request);
        $this->container->singleton(ResponseSender::class, fn () => new TestResponseSender());

        $router = $this->container->get(Router::class);
        $responseSender = $this->container->get(ResponseSender::class);

        return $responseSender->send(
            $router->dispatch($request)
        );
```

This is because we have to override the request and response sender, and then get the returned response. Maybe it's okay, just need to keep it in mind.

brendt:
I'd add a `request` helper method in the bast TestCase that does this setup for you. The only thing you'd need to pass in is the request object itself, and this method would do the rest, and return a response.

aidan-casey:
I'm guessing you haven't looked at my PR. 🙃

src/Testing/Http/TestResponseSenderInitializer.php
```diff
@@ -0,0 +1,27 @@
...
+use Tempest\Http\Response;
+use Tempest\Http\ResponseSender;
+
+final class TestResponseSenderInitializer implements CanInitialize
+{
+    public function canInitialize(string $className): bool
+    {
+        return in_array($className, [Response::class, TestResponse::class]);
+    }
+
```
brendt, line 18: On mobile, but shouldn't you also implement Initializer? Not sure

src/Testing/Http/TestsHttpRequests.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Tempest\Testing\Http;
+
+use Tempest\Http\GenericRequest;
+use Tempest\Http\Method;
+use Tempest\Http\Request;
+use Tempest\Http\ResponseSender;
+use Tempest\Http\Router;
+
+trait TestsHttpRequests
```
brendt, line 13: I like it!

tests/TestingTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Tempest;
+
+class TestingTest extends \Tempest\Testing\TestCase
```
brendt, line 7: Isn't this a bit of a weird name?

aidan-casey: Yes, this goes away in the last commit. This is just my dev playground file. 😛

src/Testing/Http/TestResponseSenderInitializer.php
aidan-casey: Somewhere along the way `CanInitialize` got modified to extend `Initializer` so that now throws errors. I meant to track down when that happened.

brendt:
Overall, I like it! It's good that we're moving stuff into `src/Testing` so that packages and users can make use of it 💯

src/Testing/Http/TestResponseSenderInitializer.php
brendt: I refactored it, you now have to implement both. Please rebase on main :)

brendt:
I'm going to merge this one, since I need these features 😅 we can always improve on it later.

aidan-casey:
I mean, it was just a draft, but okay. 😂

I'll put some work on adding to the helpers this week.

brendt:
pre-1.0, everything's possible 😁

aidan-casey:
I'll have to see what other half-baked features I can come up with. 😉

PR #135, opened 2024-02-24
Expands autowire capabilities

aidan-casey:


aidan-casey:
Full disclosure, I think there is plenty of room to continue improving here, but this PR expands the abilities of autowiring quite a bit.

If we can figure out how to tell the container to resolve certain classes (e.g., configuration classes) as singletons/configs, we can potentially set defaults as default parameter values and allow the developer to override simply by creating a config file in the app.

brendt:
I like this PR 💯 

Feel free to merge when QA stuff is passing

Issue #116, opened 2024-02-22
Add Architecture Testing


aidan-casey:
@vsergiu93 brought up the idea of architecture testing in #114. I think this is a great idea and something work exploring.

Looking at `ta-tikoma/phpunit-architecture-test` (which Pest uses) it does not appear they currently have support for attributes. It seems like that would be a pretty key thing considering discovery and validation rules.

There are some alternatives ([phpat](https://www.phpat.dev/), etc.) that do support attributes and are probably worth digging into further.

Plytas:
What about adding Pest? All main tests can still be written using phpunit syntax, but architecture tests could be written using pest which seems to have [attribute support](https://pestphp.com/docs/arch-testing#content-tohaveattribute).

brendt:
So, let's talk about Pest. In a way it would be fitting to use it as the default test runner for Tempest. Pest emerged as a way to break with "the standard", and took a whole different approach to testing. 

Personally, I'm not for or against it, although it would require some time to get used to it.

On top of that, I see a small "marketing" opportunity for Tempest if we embrace Pest as the default test runner.

And even better, "Tem**pest**", I mean, speaking of a match made in heaven :p

aidan-casey:
Personally, I'm not a Pest guy, for no other reason than I prefer the syntax and structure of PHPunit.

As far as what decision gets made here goes, I'd just caution against making that decision based on functionality like architecture testing, which is essentially a wrapper for another package and primarily way the pros (or cons) of Pest's style.

Plytas:
Since Pest is progressive framework it can run on top of phpunit syntax. There's no need to migrate existing tests or even write new tests using pest syntax. Heck, people could write with whichever syntax they prefer and it would still work, with the downside being that tests now have mixed structure and can be confusing.

Because of that, at work we opted to continue writing tests using phpunit syntax as everyone was familiar with it. We did however add pest for the sake of architecture testing. A single file that contains some basic rules to make sure everyone adheres to them.

I understand the argument that we could use the underlying package and implement it ourselves, but at this point it would feel like re-inventing the wheel when there's already a working solution available.

Also, Pest comes with [parallel testing](https://pestphp.com/docs/optimizing-tests#content-parallel-testing), [profiling slow tests](https://pestphp.com/docs/optimizing-tests#content-parallel-testing), [test coverage](https://pestphp.com/docs/test-coverage) and [type coverage](https://pestphp.com/docs/type-coverage) out of the box.

[...]

PR #141, opened 2024-02-24
Added pest architecture test

Plytas:
As discussed in https://github.com/tempestphp/tempest-framework/issues/116, I've added an example of how architecture tests could be implemented with Pest.

I've purposely left it failing to showcase that it's working. `Lowercase`, `Uppercase` and `DoesNotEndWith` rules are not `final` and `readonly` like others are.

### DIFF
--- a/.github/workflows/quality-assurance.yml
+++ b/.github/workflows/quality-assurance.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Run PHPStan
         run: vendor/bin/phpstan --error-format=github
 
-  phpunit:
+  pest:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
@@ -86,7 +86,7 @@ jobs:
         run: composer show -D
 
       - name: Execute tests
-        run: vendor/bin/phpunit --coverage-clover build/reports/clover.xml
+        run: vendor/bin/pest --coverage-clover build/reports/clover.xml --compact
 
       - name: Upload coverage results to Coveralls
         env:
--- a/.github/workflows/quality-control.yml
+++ b/.github/workflows/quality-control.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Run PHPStan
         run: vendor/bin/phpstan --error-format=github
 
-  phpunit:
+  pest:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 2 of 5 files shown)


vsergiu93:
I think I worked on those rules :D, that's why I would like to have architecture testing, sometimes I just forget about this things. 
But I think the benefits of adding this kind of tests is clear for everyone, the questions if what tool to use, should we go with Pest, which changes a lot in terms o syntax and style or just add phpunit plugins. 

Personally I like Pest syntax, nowadays I'm working more in javascript than php, and like the `it('some test case')` syntax.

But more important than syntax would be to pick a tool which cand be used to enforce most or ideally all the architectural conventions. 

@Plytas thanks for looking in to this.

Plytas:
The point of this PR is to show how architecture tests would look like with Pest. It also demonstrates that it doesn't affect other tests that could still be written with phpunit if the team prefers that.

I believe that @brendt will work on an alternative using phpunit before they make a decision.

I briefly looked into those alternatives and while they do achieve the same thing, I believe that Pest's ease of use and developer experience is unmatched.

aidan-casey:
@Plytas - Do you mind running the drift plugin on this PR? I think if we were to switch to Pest we'd want to make the tests consistent, regardless of Pest's compatibility with PHPunit.

Plytas:
@aidan-casey drift plugin kinda fails in this repo. Mainly because tests include [helper classes in the same file](https://github.com/tempestphp/tempest-framework/blob/30f8ab4f7fa45590801628f872fad5f944e81ab6/tests/CommandBus/CommandBusTest.php#L52-L61). They would have to be extracted manually.

aidan-casey:
@Plytas we should do that anyway since we are PSR-4 autoloading tests. I'll create a separate issue for that and we can rebase this when that's done.

Plytas:
@aidan-casey I've opened a separate PR https://github.com/tempestphp/tempest-framework/pull/144 to showcase Pest syntax.

brendt:
FYI, I wanna look into [PHPAT](https://www.phpat.dev/) as an alternative. I do have some other things on my todo list though, so please allow me a couple of days.

Plytas:
Closing in favor of #151 as discussed on #116

Issue #39, opened 2024-02-16
Response Objects


aidan-casey:
@brendt - ASP.NET has some helper responses that allow for the easy generation of API responses. Any reason we could not add some support for a similar concept in object form? Those could in turn be used for some response helpers such as `Ok`.

_Examples_

```php
final readonly class CreatePetController
{
    #[Post(uri: '/api/v1/pets')]
    public function __invoke(CreatePetRequest $request): Created
    {
        $pet = map($request)->to(Pet::class)->save();

        return new Created($pet);
    }
}

final readonly class CreateOwnerController
{
    #[Post(uri: '/api/v1/owners')]
    public function __invoke(CreateOwnerController $request): Created
    {
        $pet = map($request)->to(Owner::class)->save();

        return new Created(
            value: $owner,
            uri: uri(ShowOwnerController::class, $owner) // Sets the Location header
        );
    }
}
```

_References_

- https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.created?view=aspnetcore-8.0
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.ok?view=aspnetcore-8.0
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.createdataction?view=aspnetcore-8.0

brendt:
Looks good! Are you up for sending a PR? :)

aidan-casey:
Absolutely! Just wanted to float the idea before putting in the work. 🙂

alibori:
Hi both!

Which helper responses are you thinking about? I would like to help at this point (or at least try to 😜 ). My approach would be base on this [article](https://juststeveking.uk/articles/laravel-response-classes/)

What do you think?

Thanks!

aidan-casey:
Hey @alibori! Essentially these objects should be classes that use `BaseResponse` and implement best practices for that response type.

For example, in the case of a created response, it might look something like this:

```php
namespace Tempest\Http\Response;

final class Created
{
    use BaseResponse;

    public function __construct(object $value, string $location)
    {
        parent::__construct(Status::CREATED, json_encode($value), ['Location' => $location]);
    }
}
```

Please note that is a super rough example!

The one thing we might want to explore, but instantly makes this more complex is serializers for the response to match what the header requests.

alibori:
Hey @aidan-casey !

[...]

PR #142, opened 2024-02-25
Response objects

alibori:
Hi!

Regarding [Issue #39](https://github.com/tempestphp/tempest-framework/issues/39) this PR adds:

- Adds preconfigured CreatedResponse and it's test.

Thanks!

### DIFF
--- a/tests/Http/Responses/CreatedResponseTest.php
+++ b/tests/Http/Responses/CreatedResponseTest.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Tempest\Http\Responses;
+
+use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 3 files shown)


tests/Http/Responses/CreatedResponseTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Tempest\Http\Responses\CreatedResponse;
+use Tempest\Http\Status;
+
+class CreatedResponseTest extends TestCase
+{
+    public function test_created_response()
+    {
+        $response = new CreatedResponse(json_encode(['foo' => 'bar']));
```
brendt, line 15: @aidan-casey I think these response objects should convert arrays to json automatically, wdyt?

Nothing we should change in this PR, just a random thought I wanted to check up with you

Issue #143, opened 2024-02-25
Move Test Fixtures to Separate files


aidan-casey:
To be a) PSR-4 compliant and b) allow easier migration to Pest IF we go that route, classes in the same file as tests should be moved to separate files under a `Fixtures` directory within the module namespace.

See #141

aidan-casey:
Closed per #146

brendt:
Wrong button, sorry

PR #144, opened 2024-02-25
Converted tests to Pest syntax

Plytas:
This PR includes changes present in #141.

Showcasing tests converted to Pest syntax.

Things to note:
* ~Test fixtures had to be extracted to separate classes, that's done in https://github.com/tempestphp/tempest-framework/commit/7679c1e4e34469068b368f2f81fdf29ff6499dd1~ Done in #146
* Currently not all tests use `\Tests\Tempest\TestCase` as they don't depend on it.
* `\Tests\Tempest\TestCase` usage is implemented on a test basis. It's possible to do that [globally](https://pestphp.com/docs/configuring-tests), but it's a bit tricky because you need to specify a directory or a matching glob expression. Usually you would have a `Feature` directory and you could apply it to all tests in there. It's possible to specify all current folders, but there's a possibility of forgetting to add them when introducing new directories.
* I had to delete a space in `app/Views/include-parent.php` for tests to pass, because for some reason my PhpStorm was automatically deleting trailing space in [this line](https://github.com/tempestphp/tempest-framework/blob/main/tests/View/ViewTest.php#L143)
* PHPStan will report errors in tests because it doesn't know about `$this`. PhpStorm has no problems and can provide autocompletion. I couldn't find a proper solution for this apart from ignoring it or not running PHPStan on tests.


aidan-casey:
Thanks for all your work on this, @Plytas! Would you be up for putting the separation of the test fixtures to fixture directories under the test module namespaces in another PR per #143? If so, I think we should be able to merge that pretty quickly and then we can continue to discuss the benefits of Pest. 🙂

Plytas:
@aidan-casey Sure thing, opened #146.

aidan-casey:
Thanks!

Plytas:
Updated this PR to be in-line with newest changes.

Plytas:
Closing as discussed in #116

PR #153, opened 2024-02-26
Expands test helpers

aidan-casey:
@brendt - Would appreciate your review here. A couple notable things:

1. I approached this differently than the HTTP helpers and I think I like it enough to move the HTTP helpers to the same approach. Rather than being protected methods directly on the test case, it is a helper class that is initialized on a property. This allows some restriction of which methods are shown, which is helpful to the DX.
2. Console testing is kind of hard... purely because we are allowing developers to pass already formatted text into the console, so we have to decide if we want users to also have to pass in formatted text to the assertion methods if they did that or try and use RegEx to strip those out.

**THIS IS A DRAFT** so go ahead and merge it I guess. 😉


aidan-casey:
@brendt - I would have more coverage, but routing is a bit broken. 😉

aidan-casey:
> I preferred the traits more

The problem I have with the traits is that now you get both http and console helper methods as you build up the test case which is both bad DX and impossible to actually use. 😛

> We need one way of doing this. I propose the second because of the flexibility it offers.

Great. Make it so. 🙃

> I propose we keep track of both the formatted and unformatted output.

The problem I was expressing is that with your second way, it's already formatted by the time it gets to us, so we cannot track the unformatted version.

Don't worry though, I hacked together a regex you'll probably hate to let us test that. 😉

> We could change it to....

There are basically methods for this in the PR already.

> I've made some improvements here, you can rebase on main.

I'll take a look.

brendt:
> The problem I have with the traits is that now you get both http and console helper methods as you build up the test case which is both bad DX and impossible to actually use. 😛

How so? What if you had two separate traits? `TestsHttp` and `TestsConsole`, people can choose whatever they want?

> Great. Make it so. 🙃

I will, so please hang on until this is fixed :) 

> The problem I was expressing is that with your second way, it's already formatted by the time it gets to us, so we cannot track the unformatted version.
> Don't worry though, I hacked together a regex you'll probably hate to let us test that. 😉

I think regex is the perfect solution 👍 It shouldn't even be a regex I think, a simple str_replace will do. I'll take care of it.

> There are basically methods for this in the PR already.

I saw, perfect 💯

aidan-casey:
> What if you had two separate traits? TestsHttp and TestsConsole

If you have these traits used by the base test case by default (which we were doing) then you get both mixed together.

I think the DX of extending a test case AND adding a trait is worse though.

brendt:
> We need one way of doing this. I propose the second because of the flexibility it offers.

So, this has been refactored on main, `ConsoleFormatter` is gone, `TestConsoleOutput` has this method: https://github.com/tempestphp/tempest-framework/blob/main/tests/TestConsoleOutput.php#L32

We'll need to refactor it to your PR though, I only added it in `TestConsoleOutput` as a temporary PoC

brendt:
@aidan-casey I've rebased this branch on main. What changed:

- `ConsoleFormatter` is gone, I've made the necessary changes
- `Server` is gone, I've made the necessary changes
- I've reimplemented `HttpRouterTester::sendRequest` and removed all the `$_SERVER` related stuff, as it's not necessary anymore

src/Testing/Console/TestConsoleHelper.php
```diff
@@ -0,0 +1,70 @@
...
+        Assert::assertStringContainsString(
+            $text,
+            $this->output->getTextWithFormatting(),
+            sprintf('Failed to assert that console output included formatted text: %s.', $text)
+        );
+
+        return $this;
+    }
+
+    public function assertContainsError(string $line): self
```
brendt, line 38: I like these assertions, but what about formatting? Right now it are the formatted error lines that are checked, which doesn't make much sense. Alternatively, we can store the unformatted lines, but what if we want to test on formatted error messages then. Adding `assertContainsFormattedError` and `assertContainsUnformattedError` also seems a bit overkill…

I think we need a way to get whatever output, and then optionally call `unformat` or something on it. IDK what exactly that should look like though…

src/Testing/Http/HttpRouterTester.php
```diff
@@ -0,0 +1,86 @@
...
+
+use Exception;
+use Tempest\Container\Container;
+use Tempest\Http\GenericRequest;
+use Tempest\Http\Method;
+use Tempest\Http\Request;
+use Tempest\Http\Router;
+use function Tempest\map;
+
+final class HttpRouterTester
```
brendt, line 15: I'd call it `HttpTester` I think… Not sure though

src/Testing/Http/TestResponseHelper.php
```diff
@@ -0,0 +1,97 @@
...
+use Tempest\Http\Response;
+use Tempest\Http\Status;
+
+final readonly class TestResponseHelper
+{
+    public function __construct(private Response $response)
+    {
+    }
+
+    public function getResponse(): Response
```
brendt, line 17: FYI, I added these getters as shorthands. It doesn't make sense that users should access the wrapped response for such simple methods.

src/Testing/TestCase.php
```diff
@@ -8,16 +8,19 @@
 use Tempest\AppConfig;
 use Tempest\Application\Kernel;
 use Tempest\Container\Container;
-use Tempest\Testing\Http\TestsHttpRequests;
+use Tempest\Testing\Console\ConsoleCommandTester;
+use Tempest\Testing\Http\HttpRouterTester;
 
 abstract class TestCase extends BaseTestCase
```
brendt, line 14: Do we plan on migrating all our tests to extend from this TestCase?

tests/TestCase.php
```diff
@@ -50,6 +48,8 @@ protected function setUp(): void
             ConsoleOutput::class,
             fn () => new TestConsoleOutput(),
         );
+
+        $this->http = $this->container->get(HttpRouterTester::class);
```
brendt, line 52: I temporarily added this one, so that I could already rewrite some tests to use the new syntax.

src/Testing/Http/HttpRouterTester.php
aidan-casey: How about `HttpRequestTester`?

src/Testing/TestCase.php
aidan-casey: I think we need to split out our tests into `Unit` and `Integration` tests. Probably all `Integration` tests would extend `Tempest\Testing\TestCase`.

What are your thoughts on that?

src/Testing/Console/TestConsoleHelper.php
aidan-casey: It's a good point. What I had been playing around with _prior_ to the console changes was passing two parameters (string and variadic styles) like you had for the formatter. If no styles were passed, it would do an unformatted test.

I think my inclination now is that _all_ methods expect the developer to pass the string as they would have passed it to output (so they are dictating the styling) and we have a single `assertContainsText` method that does an unformatted comparison if we are just looking for a single string.

Thoughts?

src/Testing/TestCase.php
brendt: I'm fine with that, but I'd make it a rule that as soon as the framework boots, it's an integration test. It means that 99% of tests are integration tests right now.

src/Testing/Http/TestResponseHelper.php
brendt: You're right. Ideally we add methods like `assertBodyContains` etc. 

Let's say I added these getters so that my current tests could succeed, now we can continue to improve our test helpers. (But first let's get this merged)

src/Testing/Console/TestConsoleHelper.php
brendt: > I think my inclination now is that all methods expect the developer to pass the string as they would have passed it to output (so they are dictating the styling) and we have a single assertContainsText method that does an unformatted comparison if we are just looking for a single string.

I don't really understand, could you give an example?

brendt:
Overall, I'm totally fine merging this in its current form, and then iterate on it further. 

I can take care of the merge conflict and merge it, let me know if that's ok

PR #154, opened 2024-02-26
Forbids dangerous/debug functions from being used

aidan-casey:


brendt:
Hey but I need those functions 😅

aidan-casey:
I'm honestly not sure why you need exec. Composer shouldn't be stripping the executability of a file, so as long as it is committed as an executable you should be good.

PR #156, opened 2024-02-26
Begins to adds PSR-11 support.

### LINKED ISSUE #148: PSR-11 support

aidan-casey:
Things that should be done...

- [ ] Expand on the exceptions and use a not found vs. generic container exception where appropriate.
- [ ] Add tests.

Closes #148

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
         "ext-dom": "*",
         "ext-mbstring": "*",
         "giggsey/libphonenumber-for-php": "^8.13",
-        "egulias/email-validator": "^4.0"
+        "egulias/email-validator": "^4.0",
+        "psr/container": "^2.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^10.2",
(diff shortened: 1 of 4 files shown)


brendt:
See, here's the problem with implementing PSRs. We're just going to extend the interface and add stuff of our own. That means that everywhere where we rely on `Tempest\Container\Container`, we essentially don't support PSR anymore, because we rely on our own interface with additional functionality. That means that someone cannot simply plug in another PSR-compliant container into Tempest anymore.

What about the other way around? Tempest provides a container that's PSR-compliant, so people could _in theory_ use it as a standalone thing. For me, that's not a goal that Tempest should try to fulfil. If people want a PSR-compliant container, they can use other packages.

So, what's left? Why are we adding PSR-11 support, when both ways don't add any value? So that we can say "we're PSR-compliant?" 

A longer form of this rant is here: https://stitcher.io/blog/re-on-using-psr-abstractions

Right now, I see zero benefit from adding PSR support

aidan-casey:
Closing for now.

PR #157, opened 2024-02-26
Adds basic http client wrapper

aidan-casey:
@brendt - This is just a rough draft to give us something to play with API-wise.

I suggest we engage PSR for the underworkings of the client and build our own API on top. The question becomes whether we want that to be a fluent factory (like Laravel) or more just simple `get`, `post`, etc. methods.

I think it would be pretty simple to use [discovery](https://github.com/php-http/discovery) to automatically register the PSR implementations _OR_ we just intentionally set our defaults.

We also have to decide whether we are adding a wrapper for the response or leaving it as a PSR response.

We may also want to break out http router vs. client in separate modules.


brendt:
- Ok, let's say we'll use PSR.
- I like the `HttpClient` implementation, let's definitely _not_ use the fluent interface approach Laravel has, I really dislike it :p 
- So, which client implementation would we actually use? 
- I think our `HttpClient` should also have a method `send(Request $request)`, which is a Tempest `Request` interface

brendt:
> I think it would be pretty simple to use [discovery](https://github.com/php-http/discovery) to automatically register the PSR implementations OR we just intentionally set our defaults.

This is one of those cases where you'd need a `CanInitialize` class that's registered to the container directly :)

brendt:
Please rebase on main, and we'll then have to move `TestCase::send` to the appropriate place

aidan-casey:
> So, which client implementation would we actually use?

I keep going back and forth on this. One the one hand, we could use something like `symfony/http-client` and plan to offer async support down the road. On the other hand we could go super lightweight and use Tobias Nyholm's packages.

At the end of the day I don't think it matters and we should just pick one. Using PSR it can be swapped as a dependency at any time.

> I think our HttpClient should also have a method send(Request $request), which is a Tempest Request interface

See, this is where Tempest should just get with the program and use PSR-7 requests. 😉 🙃

aidan-casey:
@brendt - I think this is all set. Only question is which client to provide by default. Let me know if you have any hard opinions on that.

brendt:
Could you show how the client can be used by end users? I saw this in the tests:


```php

        $this->factory = new Psr17Factory();

        $this->mock = new Client(
            responseFactory: $this->factory,
            streamFactory: $this->factory
        );

        $this->client = new GenericHttpClient(
            client: $this->mock,
            uriFactory: $this->factory,
            requestFactory: $this->factory,
            streamFactory: $this->factory
        );
```

That's a no go for me. Either we provide an initializer that sets this up via the container, or we so something like `HttpClient::new()` or provide a `send(Request)` function, but this is too much setup.

Also, I would like this client to work with our Request and Response classes, not the PSR interfaces. Just like with the container, we're not aiming to publish Tempest code as standalone packages, so we don't need that interop. 

I'm ok with wrapping PSR internally, but Tempest users should never come in contact with it. There's a reason that both Symfony and Laravel don't implement PSR on their request/response classes:

- https://github.com/symfony/http-foundation/blob/7.0/Request.php
- https://github.com/symfony/http-foundation/blob/7.0/Response.php
- https://github.com/illuminate/http/blob/master/Request.php
- https://github.com/illuminate/http/blob/master/Response.php

The fact is, these PSRs aren't user friendly, overly verbose, and framework users don't care about them. I understand that we want to be able to use a PSR-compliant HTTP client, I'm fine with that; but remember that Tempest should always, _always_, go out of its way to make sure the end-user experience is the best. PSR isn't the best end-user experience.

Sorry for the (b)rant

aidan-casey:
> Could you show how the client can be used by end users?

Simply:

```php
class MakeARequest
{
    public function __construct(private HttpClient $client)
    {}

    public function doSomething(): void
    {
        $response = $this->client->get('https://api.whatever.com');
    }
}
```

Per my comment:

> Only question is which client to provide by default.

We can setup an initializer as soon as the client has been decided.

> There's a reason that both Symfony and Laravel don't implement PSR on their request/response classes

I'm not convinced there's a good one beyond the fact that PSR wasn't established when Symfony was created and the lift would have been too great by the time it was. Laravel simply built itself on top of what Symfony already had. See items like this [issue](https://github.com/symfony/symfony/issues/17392) for context.

> The fact is, these PSRs aren't user friendly, overly verbose, and framework users don't care about them.

That's requests in general though. 😜  A request object is simply intended to be a value object representing the actual request received. Syntax sugar can be thrown on top of that.

aidan-casey:
@brendt - Regarding support for Tempest requests... we can only stream strings. How are you proposing we are to handle things like the `$body` array?

aidan-casey:
@brendt - If it makes you happier, I just pushed discovery, so you could also do:

```php
(new HttpClient)->get('https://example.com');
```

Container method is for sure to be preferred though.

aidan-casey:
I don't love adding Guzzle as a dependency, but for ease I have just done so to keep this moving.

brendt:
> Regarding support for Tempest requests... we can only stream strings. How are you proposing we are to handle things like the $body array?

I believe there's a helper somewhere to convert a string to a stream somewhere in some PSR implementation, but I can't remember it. Once again an example of how the 99% use case is disregarded in the name of "flexibility"

> If it makes you happier, I just pushed discovery, so you could also do:

IDK… I'm ok if injecting is required for now. It's a common pattern within the framework, so I don't think we need to make exceptions. Let's just add an `HttpClientInitializer` ?

aidan-casey:
@brendt - As much as it pains me, perhaps this refactor will make you happy. 😢

src/Http/GenericHttpClient.php
```diff
@@ -110,8 +110,12 @@ public function options(string $uri, array $headers = [], ?string $body = null):
         );
     }
 
-    public function sendRequest(RequestInterface $request): ResponseInterface
+    public function sendRequest(Request|RequestInterface $request): ResponseInterface
     {
+        if ($request instanceof Request) {
```
brendt: 🥳

src/HttpClient/Driver/Psr18Driver.php
```diff
@@ -0,0 +1,78 @@
...
+            ),
+        );
+
+        foreach ($tempestRequest->getHeaders() as $name => $value) {
+            $request = $request->withHeader($name, $value);
+        }
+
+        // TODO: This is crappy and doesn't support stuff we need to.
+        // Eventually the array will be a string.
+        if (count($tempestRequest->getBody()) !== 0) {
```
brendt, line 60: https://github.com/tempestphp/tempest-framework/issues/191

brendt: We'll do this after this is merged

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\HttpClient;
+
+use Tempest\Container\InitializedBy;
+use Tempest\Http\Response;
+
+#[InitializedBy(HttpClientInitializer::class)]
+interface HttpClient
```
brendt, line 11: Shouldn't we add a `send(Request $request)` as well?

src/HttpClient/HttpClientDriver.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Tempest\Container\InitializedBy;
+use Tempest\Http\Request;
+use Tempest\Http\Response;
+
+/**
+ * The Tempest HttpClientDriver takes a Tempest request and
+ * produces a Tempest response.
+ */
+#[InitializedBy(HttpClientInitializer::class)]
```
brendt, line 15: You'll have to refactor this

tests/Integration/Http/HttpClientInitializerTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Tempest\HttpClient\GenericHttpClient;
+use Tempest\HttpClient\HttpClient;
+use Tests\Tempest\Integration\TestCase;
+
+class HttpClientInitializerTest extends TestCase
+{
+    public function test_container_can_initialize_http_client()
+    {
+        $httpClient = $this->container->get(HttpClient::class);
```
brendt, line 15: After rebase on main, you'll have to manually call `$this->container->addInitializer` here

brendt (changes requested):
Looks good, some minor details but I'm fine merging this after those are done

src/HttpClient/HttpClientDriver.php
aidan-casey: Only if we want it to work. 😛 😉

tests/Integration/Http/HttpClientInitializerTest.php
aidan-casey: Huh. It won't automatically discover it?

src/HttpClient/HttpClient.php
aidan-casey: Should we? I'm happy to, just not sure how often that'd get used.

tests/Integration/Http/HttpClientInitializerTest.php
brendt: Oh, it's an integration test, nvm 👍

PS: we should rename `Integration\TestCase` to `IntegrationTest`

src/HttpClient/HttpClient.php
brendt: Difficult one 😅

It might be overkill. OTOH, if we decide to add it later, it'll be a breaking change :/ But we might never want to add it.

brendt: Did some more thinking on it, I actually think it'll be very useful in cases where you're integrating with APIs that don't have proper PHP SDKs. You could make a base request class or interface + trait combo that sets up the right headers etc, and then use that for all subsequent requests.

aidan-casey: > You could make a base request class or interface + trait combo that sets up the right headers etc, and then use that for all subsequent requests.


I'd be incapable of doing this, actually. Some other developer might insult PSR in such a manner, however. 😛 

I'll add it.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\HttpClient;
+
+use Tempest\Http\Request;
+use Tempest\Http\Response;
+
+interface HttpClient
```
aidan-casey, line 10: @brendt - Do we want to keep this interface or require all modifications to be through the driver?

aidan-casey: That is an issue with the initializer rework done #184, not this specific PR.

yassiNebeL: I cannot open an issue at the moment, as I know it's still in development, but nothing magical, I'm using Tempest as a package:

Clone/download the repository
Run composer install
Configure the .env file

brendt: Let's stick with the interface 👍

brendt:
🥳

PR #159, opened 2024-02-26
Reworks request objects.

aidan-casey:
@brendt  - I am probably doing something you don't want me to here, but is there any reason we cannot skip the `Server` concept and skip right to building up a request? This way, if our test system sets the request as a singleton, it should just skip this step entirely.

Further, we should probably make requests PSR compatible.

### DIFF
--- a/src/Application/Kernel.php
+++ b/src/Application/Kernel.php
@@ -26,6 +26,7 @@ use Tempest\Http\RequestInitializer;
 use Tempest\Http\RouteBindingInitializer;
 use Tempest\Http\Router;
 use Tempest\Http\ServerInitializer;
+use Tempest\Mapper\ObjectMapper;
 use function Tempest\path;
 use Throwable;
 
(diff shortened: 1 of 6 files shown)


aidan-casey:
(Yes I know tests are failing. 🙃 )

brendt:
> I am probably doing something you don't want me to here, but is there any reason we cannot skip the Server concept and skip right to building up a request?

The goal is to skip `Server` 99% of the time during testing, let's agree to agree 🤝

> Further, we should probably make requests PSR compatible

Why? Please read https://stitcher.io/blog/re-on-using-psr-abstractions 😁

Finally, I've made a bunch of changes on main to Requests, you should probably close this PR and apply the changes that matter on a fresh branch.

aidan-casey:
@brendt yes, I've read your post before. I believe this is a major area in which PSR is helpful though.

The same request/response objects we use for the router should be able to be used for the HTTP client. If we rely on a PSR-18 client under the hood and pass it a PSR compliant request, we'll get a PSR compliant response which Tempest can handle.

brendt:
> As far as the server goes... all I want to know is: what is its point to begin with?!!

To be able to mock so that we can test a couple of core framework functionalities, see `\Tests\Tempest\Route\RequestTest`. 

> The same request/response objects we use for the router should be able to be used for the HTTP client. If we rely on a PSR-18 client under the hood and pass it a PSR compliant request, we'll get a PSR compliant response which Tempest can handle.

Not sure I understand in full, but let's continue this discussion then in #157 ?

PR #161, opened 2024-02-27
#59 Refactor Kernel to prepare for cache implementation

jtheuerkauf:
## CLOSING IN FAVOR OF #176

* Remove `Tempest::__construct()` `$appConfig` parameter - available from Container object
* Add `Kernel::__construct()` `$bootstraps` promoted property
* Rename `Kernel::initContainer()`
* Loop `$bootstraps` to register them in the Container as singletons
* Execute Bootstraps `boot()` in `Kernel::init()`
* Add Bootstrap classes for Discovery and Config
* Set Bootstraps as third Kernel constructor argument
* Fixed some accessors to AppConfig from the `$kernel` object

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -1,12 +1,9 @@
 <?php
 
-use Tempest\AppConfig;
-use Tempest\Application\Environment;
 use Tempest\Tempest;
-use function Tempest\env;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-Tempest::boot(__DIR__ . '/../')->http()->run();
+Tempest::boot(dirname(__DIR__))->http()->run();
 
 exit;
--- a/src/Application/Kernel.php
+++ b/src/Application/Kernel.php
@@ -8,39 +8,38 @@ use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use ReflectionClass;
 use Tempest\AppConfig;
-use Tempest\Application\Exceptions\KernelException;
 use Tempest\Container\Container;
 use Tempest\Container\GenericContainer;
 use Tempest\Database\PDOInitializer;
-use Tempest\Discovery\DiscoveryLocation;
 use Tempest\Http\RequestInitializer;
 use Tempest\Http\RouteBindingInitializer;
 use Tempest\Http\ServerInitializer;
-use function Tempest\path;
 use Throwable;
 
 final readonly class Kernel
 {
     public function __construct(
-        private string $root,
+        public string $root,
         private AppConfig $appConfig,
+        private array $bootstraps = [],
     ) {
     }
 
     public function init(): Container
     {
-        $container = $this->initContainer();
+        $container = $this->createContainer();
+        $container->config($this->appConfig);
 
-        $this->initDiscoveryLocations();
-
-        $this->initConfig($container);
+        foreach ($this->bootstraps as $bootstrap) {
+            $container->get($bootstrap)->boot();
+        }
 
         $this->initDiscovery($container);
 
         return $container;
     }
 
-    private function initContainer(): Container
+    private function createContainer(): Container
     {
         $container = new GenericContainer();
 
@@ -54,30 +53,12 @@ final readonly class Kernel
             ->addInitializer(new RouteBindingInitializer())
             ->addInitializer(new PDOInitializer());
 
-        return $container;
-    }
-
-    private function initDiscoveryLocations(): void
-    {
-        $this->discoverTempestNamespaces();
-        $this->discoverInstalledPackageLocations();
-    }
-
-    private function initConfig(Container $container): void
-    {
-        // Register AppConfig
-        $container->config($this->appConfig);
-
-        // Scan for package config files
-        foreach ($this->appConfig->discoveryLocations as $package) {
-            $configFiles = glob(path($package->path, 'Config/**.php'));
-
-            foreach ($configFiles as $configFile) {
-                $configFile = require $configFile;
-
-                $container->config($configFile);
-            }
+        /** @var class-string<\Tempest\Bootstraps\Bootstrap> $bootstrap */
+        foreach ($this->bootstraps as $bootstrap) {
+            $container->singleton($bootstrap, fn () => new $bootstrap($container));
         }
+
+        return $container;
     }
 
     private function initDiscovery(Container $container): void
@@ -132,56 +113,5 @@ final readonly class Kernel
 
             $discovery->storeCache();
         }
-    }
-
-    private function discoverInstalledPackageLocations(): void
-    {
-        $composerPath = path($this->root, 'vendor/composer');
-        $installedPath = path($composerPath, 'installed.json');
-
-        $installedJson = $this->loadJsonFile($installedPath);
-        $packages = $installedJson['packages'] ?? [];
-        foreach ($packages as $package) {
-            $packagePath = path($composerPath, $package['install-path']);
-
-            $requiresTempest = isset($package['require']['tempest/framework']);
-            $hasPsr4Namespaces = isset($package['autoload']['psr-4']);
-            if ($requiresTempest && $hasPsr4Namespaces) {
-                foreach ($package['autoload']['psr-4'] as $namespace => $namespacePath) {
-                    $namespacePath = path($packagePath, $namespacePath);
-                    $this->addDiscoveryLocation($namespace, $namespacePath);
-                }
-            }
-        }
-    }
-
-    private function discoverTempestNamespaces(): void
-    {
-        $composer = $this->loadJsonFile(path($this->root, 'composer.json'));
-
-        $namespaceMap = $composer['autoload']['psr-4'] ?? [];
-        foreach ($namespaceMap as $namespace => $path) {
-            $path = path($this->root, $path);
-            $this->addDiscoveryLocation($namespace, $path);
-        }
-    }
-
-    private function addDiscoveryLocation(string $namespace, string $path): void
-    {
-        $this->appConfig->discoveryLocations[] = new DiscoveryLocation(
-            namespace: $namespace,
-            path     : $path,
-        );
-    }
-
-    private function loadJsonFile(string $path): array
-    {
-        if (! is_file($path)) {
-            $relativePath = str_replace($this->root, './', $path);
-
-            throw new KernelException(sprintf('Could not locate %s, try running "composer install"', $relativePath));
-        }
-
-        return json_decode(file_get_contents($path), true);
     }
 }
--- a/src/Bootstraps/ConfigBootstrap.php
+++ b/src/Bootstraps/ConfigBootstrap.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Tempest\Bootstraps;
+
+use Tempest\AppConfig;
+use Tempest\Container\Container;
+use function Tempest\path;
+
+final readonly class ConfigBootstrap implements Bootstrap
+{
+    public function __construct(
+        private Container $container
+    ) {}
+
+    #[\Override]
+    public function boot(): void
+    {
+        // Scan for package config files
+        foreach ($this->container->get(AppConfig::class)->discoveryLocations as $package) {
+            $configFiles = glob(path($package->path, 'Config/**.php'));
+
+            foreach ($configFiles as $configFile) {
+                $configFile = require $configFile;
+
+                $this->container->config($configFile);
+            }
+        }
+    }
+}
--- a/src/Tempest.php
+++ b/src/Tempest.php
@@ -4,57 +4,62 @@ declare(strict_types=1);
 
 namespace Tempest;
 
-use Closure;
 use Dotenv\Dotenv;
 use Tempest\Application\Application;
 use Tempest\Application\ConsoleApplication;
 use Tempest\Application\Environment;
 use Tempest\Application\HttpApplication;
 use Tempest\Application\Kernel;
+use Tempest\Bootstraps\ConfigBootstrap;
+use Tempest\Bootstraps\DiscoveryBootstrap;
 use Tempest\Exceptions\ExceptionHandler;
 
 final readonly class Tempest
 {
     private function __construct(
         private Kernel $kernel,
-        private AppConfig $appConfig,
     ) {
     }
 
-    public static function boot(string $root, ?Closure $createAppConfig = null): self
+    public static function boot(string $root): self
     {
         $dotenv = Dotenv::createUnsafeImmutable($root);
         $dotenv->safeLoad();
 
-        $createAppConfig ??= fn () => new AppConfig(
+        $createAppConfig = static fn (): AppConfig => new AppConfig(
             environment: Environment::from(env('ENVIRONMENT', Environment::LOCAL->value)),
             discoveryCache: env('DISCOVERY_CACHE', false),
             enableExceptionHandling: env('EXCEPTION_HANDLING', false),
         );
 
-        $appConfig = $createAppConfig();
-
-        $kernel = new Kernel($root, $appConfig);
+        $kernel = new Kernel(
+            $root,
+            $createAppConfig(),
+            [
+                DiscoveryBootstrap::class,
+                ConfigBootstrap::class,
+            ],
+        );
 
         return new self(
             kernel: $kernel,
-            appConfig: $appConfig
         );
     }
 
     public function console(): ConsoleApplication
     {
         $container = $this->kernel->init();
+        $appConfig = $container->get(AppConfig::class);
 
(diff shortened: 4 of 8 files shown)


src/Application/Kernel.php
```diff
@@ -8,39 +8,38 @@
...
 use Tempest\Http\RouteBindingInitializer;
 use Tempest\Http\ServerInitializer;
-use function Tempest\path;
 use Throwable;
 
 final readonly class Kernel
 {
     public function __construct(
-        private string $root,
+        public string $root,
```
aidan-casey, line 22: Why are we changing the visibility here? If we really need global access, it should probably be in the config file.

src/Application/Kernel.php
```diff
@@ -8,39 +8,38 @@
...
+            $container->get($bootstrap)->boot();
+        }
 
         $this->initDiscovery($container);
 
         return $container;
     }
 
-    private function initContainer(): Container
+    private function createContainer(): Container
```
aidan-casey, line 42: What's the reason for renaming this?

src/Application/Kernel.php
```diff
@@ -54,30 +53,12 @@ private function initContainer(): Container
...
-            $configFiles = glob(path($package->path, 'Config/**.php'));
-
-            foreach ($configFiles as $configFile) {
-                $configFile = require $configFile;
-
-                $container->config($configFile);
-            }
+        /** @var class-string<\Tempest\Bootstraps\Bootstrap> $bootstrap */
+        foreach ($this->bootstraps as $bootstrap) {
+            $container->singleton($bootstrap, fn () => new $bootstrap($container));
```
aidan-casey, line 58: You're passing the container to every bootstrap class, but that's not defined in the interface.

src/Application/Kernel.php
```diff
@@ -54,30 +53,12 @@ private function initContainer(): Container
...
-        foreach ($this->appConfig->discoveryLocations as $package) {
-            $configFiles = glob(path($package->path, 'Config/**.php'));
-
-            foreach ($configFiles as $configFile) {
-                $configFile = require $configFile;
-
-                $container->config($configFile);
-            }
+        /** @var class-string<\Tempest\Bootstraps\Bootstrap> $bootstrap */
+        foreach ($this->bootstraps as $bootstrap) {
```
aidan-casey, line 57: Why is this necessary? The container should resolve those classes and their dependencies when you call `get` above.

src/Tempest.php
```diff
@@ -4,73 +4,79 @@
...
     }
 
-    public static function boot(string $root, ?Closure $createAppConfig = null): self
+    public static function boot(string $root): self
     {
         $dotenv = Dotenv::createUnsafeImmutable($root);
         $dotenv->safeLoad();
 
-        $createAppConfig ??= fn () => new AppConfig(
+        $createAppConfig = static fn (): AppConfig => new AppConfig(
```
aidan-casey, line 29: What's the point of this refactor? Why can we no longer pass in an app config?

public/index.php
```diff
@@ -1,12 +1,9 @@
...
 
-use Tempest\AppConfig;
-use Tempest\Application\Environment;
 use Tempest\Tempest;
-use function Tempest\env;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-Tempest::boot(__DIR__ . '/../')->http()->run();
+Tempest::boot(dirname(__DIR__))->http()->run();
```
aidan-casey, line 7: These are doing the same exact thing. What's the point in this refactor?

src/Bootstraps/ConfigBootstrap.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function __construct(
+        private Container $container
+    ) {}
+
+    #[\Override]
+    public function boot(): void
+    {
+        // Scan for package config files
+        foreach ($this->container->get(AppConfig::class)->discoveryLocations as $package) {
```
aidan-casey, line 21: Why not just making AppConfig a dependency in the constructor and allow the container to resolve it when getting the class?

src/Application/Kernel.php
jtheuerkauf: It's the only part of this process i couldn't find a clean way to get into the Bootstrap logic, where Discovery needs it. i'm fine moving it somewhere else. Which config file are you referring to?

src/Application/Kernel.php
jtheuerkauf: Part of @brendt's recommendations in the issue.

src/Application/Kernel.php
jtheuerkauf: i didn't think constructors are typically defined in interfaces, but i can add it. Or we can go back to requiring it in the `boot()` signature.

@brendt - any preference?

src/Application/Kernel.php
```diff
@@ -8,39 +8,38 @@
...
-use function Tempest\path;
 use Throwable;
 
 final readonly class Kernel
 {
     public function __construct(
-        private string $root,
+        public string $root,
         private AppConfig $appConfig,
+        private array $bootstraps = [],
```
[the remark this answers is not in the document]
jtheuerkauf: It doesn't really fit semantically as configuration... It's a set of tasks the application requires to get started. They aren't meant to hold data throughout the application.

src/Application/Kernel.php
jtheuerkauf: It was among my comments in the issue. The response pointed out that i wouldn't need to pass `$container` to `boot()`, but didn't speak to removing this loop registering the singletons.

You're right, of course, it works without, so i'll remove this.

src/Bootstraps/ConfigBootstrap.php
jtheuerkauf: i'm not sure i follow. `AppConfig` is in the Container (a singleton via `config()`), so why make it a separate dependency?

aidan-casey: Since you are now using the `get` method on the container to create the bootstrap class, the container will provide any dependencies in the `__construct` method. `AppConfig` could simply move to the construct as a dependency. 🙂

src/Tempest.php
jtheuerkauf: Part of @brendt's response in the issue agreed this isn't necessary as a parameter. My initial suggestion was to replace it with the array of Bootstrap classes that can be added to those we have internally, but he recommended we wait on that for now.

src/Application/Kernel.php
aidan-casey: I think I would be inclined to move this to `AppConfig` over exposing it as a public property.

public/index.php
jtheuerkauf: The function doesn't mess with or assume directory separators, where appending the path segment does. It's a common recommendation i've seen, but i can revert it.

src/Application/Kernel.php
jtheuerkauf: i considered that, but i wasn't sure i like adding another parameter to its constructor. It already feels like anything we want to hold in configuration has to be added as a constructor param. i'm not against it, that's just where i fell when i was examining it.

src/Application/Kernel.php
jtheuerkauf: Yeah, i'm removing this loop, so if i understand correctly, i don't need to update the interface. 😃❓

src/Application/Kernel.php
jtheuerkauf: @brendt / @aidan-casey 

i agree with Aidan that `$root` shouldn't _need_ to be exposed publicly. But there's no easy way to get it from Kernel to Container or from Kernel to Bootstraps (or anywhere else). We can add it to AppConfig as a new constructor parameter, but i'm not loving that either. We could add a function to add arbitrary data to AppConfig, but... You get the idea.

Any suggestions?

public/index.php
aidan-casey: I'm not opposed to the change necessarily, but let's move this to a separate PR so we can retain a single focus on this one.

src/Application/Kernel.php
aidan-casey: I think we need to determine _what_ we are bootstrapping here. I am of the mind that we are not bootstrapping the `container`, but the application. In that case, the bootstrapper should _only_ be receiving an instance of the application and not have full control over the container. This will also guard us against circular dependencies and an inappropriate use of bootstrappers.

I recommend looking through how Symfony and Laravel use these for some inspiration.

jtheuerkauf: i'll take some time to review those and a couple others i've worked with recently. i'll try to come up with options soon. If anyone else has suggestions in the meantime, please post.

src/Bootstraps/ConfigBootstrap.php
jtheuerkauf: Here's what i see happening in Spiral:

The Kernel is constructed by passing an _array_ of named directories:

[Tutorial app example](https://github.com/spiral/app/blob/master/app.php#L20)
[Method declaration](https://github.com/spiral/framework/blob/master/src/Boot/src/AbstractKernel.php#L109)

The array becomes a `Directories` object in the Container:
[Kernel constructor](https://github.com/spiral/framework/blob/master/src/Boot/src/AbstractKernel.php#L85)

It may be worth noting the Kernel builds the Container in `Kernel::create()` and passes it to the constuctor... i guess that permits other approaches to construct the Kernel with a different Container implementation, where Tempest's Kernel wants it done with the Tempest Container.

For future consideration, we may want to refactor to permit other Containers, but not today.

---

If there's no opposition, i'm going to add a similar `Directories` class that the Container will be able to resolve, and take care of this exposure issue.

aidan-casey: So in reviewing, I actually think Spiral is a pretty poor example of a bootstrapper implementation.

Laravel, I think, gives a better perspective into what a bootstrapper _should_ be doing. See [here](https://github.com/laravel/framework/blob/698c74b8465daa518cf3954eea133e4e6e3f47d7/src/Illuminate/Foundation/Http/Kernel.php#L42) for example.

A bootstrapper isn't really getting hands on with the application. It is getting and providing the data necessary to help setup the kernel so that the actual application can load. Much like a computers boot process.

To me, there are a few things that are concerning to me about this PR:

1. It seems to have taken a turn away from kernel bootstrapping and toward setting up dependencies in the container. We basically already have this in initializers, so I'm not entirely sure what we are gaining.

2. The scope of the PR keeps changing/getting added to.

I don't say that to discourage you from going down this path, but only to encourage you to target one specific problem to tackle here so that we can merge this specific PR quickly and look to tackle other issues in future PRs. It would be helpful to have a clear argument for what the issue is that we are trying to solve, and how bootloaders/bootstrappers are going to do that for us.

jtheuerkauf: Oops, i meant that Directories comment for the discussion on `__construct()` ...

i apologize for responding to the wrong thread here... i'll copy the comments up to that thread, so hopefully that will get it back on track.

---

Focusing on _this_ thread originally being about passing AppConfig and Container separately to the ConfigBootstrap: the Container has AppConfig already, and we're letting DI pass the Container to the Bootstrap, so all the resources ConfigBootstrap needs are available in one parameter. i'm still not clear what advantage there is to separating them.

src/Application/Kernel.php
jtheuerkauf: (Originally posted in the wrong thread....)

Here's what i see happening in Spiral:

The Kernel is constructed by passing an array of named directories:

[Tutorial app example](https://github.com/spiral/app/blob/master/app.php#L20)
[Method declaration](https://github.com/spiral/framework/blob/master/src/Boot/src/AbstractKernel.php#L109)

The array becomes a Directories object in the Container:
[Kernel constructor](https://github.com/spiral/framework/blob/master/src/Boot/src/AbstractKernel.php#L85)

It may be worth noting the Kernel builds the Container in Kernel::create() and passes it to the constuctor... i guess that permits other approaches to construct the Kernel with a different Container implementation, where Tempest's Kernel wants it done with the Tempest Container.

For future consideration, we may want to refactor to permit other Containers, but not today.

If there's no opposition, i'm going to add a similar Directories class that the Container will be able to resolve, and take care of this exposure issue. i was already headed somewhat in that direction earlier, but it felt like overkill. But after more thought it makes sense - other application types may need to keep track of various file system resources.

jtheuerkauf: @aidan-casey's response:

---

So in reviewing, I actually think Spiral is a pretty poor example of a bootstrapper implementation.

Laravel, I think, gives a better perspective into what a bootstrapper should be doing. See [here](https://github.com/laravel/framework/blob/698c74b8465daa518cf3954eea133e4e6e3f47d7/src/Illuminate/Foundation/Http/Kernel.php#L42) for example.

A bootstrapper isn't really getting hands on with the application. It is getting and providing the data necessary to help setup the kernel so that the actual application can load. Much like a computers boot process.

To me, there are a few things that are concerning to me about this PR:

1. It seems to have taken a turn away from kernel bootstrapping and toward setting up dependencies in the container. We basically already have this in initializers, so I'm not entirely sure what we are gaining.
2. The scope of the PR keeps changing/getting added to.

I don't say that to discourage you from going down this path, but only to encourage you to target one specific problem to tackle here so that we can merge this specific PR quickly and look to tackle other issues in future PRs. It would be helpful to have a clear argument for what the issue is that we are trying to solve, and how bootloaders/bootstrappers are going to do that for us.

jtheuerkauf: Put in its simplest form, this PR intends to move Discovery and Config out of the Kernel itself as a precursor to implementing discovery cache.


The changes in this PR are targeting just that one objective. There were some unrelated changes that you rightfully pointed out should be put back to limit scope, so those reversions are upcoming. But the basic change set remains:

1, Move Discovery-related methods out of the Kernel
2. Move `initConfig()` out of the Kernel
3. Provide the new "Bootstrap" classes made from 1 & 2 to the Kernel so it can construct them and execute the original logic

---

The outline of changes i offered seemed reasonable for doing that. The extractions don't really qualify as "Initializers" as they're not establishing application resources - they're discovering dependencies that will establish those resources...

i looked over the link to Illuminate's Http Kernel, but i honestly don't see it behaving any differently from Spiral. And to be clear, i'm not a Spiral evangelist by any means. i've just been working with it lately.

As to the concerns about the PR:

> It seems to have taken a turn away from kernel bootstrapping and toward setting up dependencies in the container. We basically already have this in initializers, so I'm not entirely sure what we are gaining.

i don't see it this way. The fundamental changes here just pull the Discovery/Config functions out to their own "tasks" - Bootstraps. They still only do what they were doing before, and no new functionality was added to the Kernel itself.

>The scope of the PR keeps changing/getting added to.

The scope changes have been attempting to address concerns raised. If we want to limit that scope creep, we can always set up issues to attack them later. My offer to add the Directories class was an attempt to do this somewhat elegantly, but it doesn't "fix" anything other than moving a public property into an object in the Container.

The `$root` visibility can be solved in a few different ways (maybe more):
1. `public` - minimal change to satisfy what Discovery/Config need
    * Leaves a bit of code smell, although it _is_ still within an object in the Container
2. Add to `AppConfig`
    * Expands PR scope via one of the following:
        a. Requires expanding AppConfig::__construct() signature for another parameter,
        b. OR adding logic to maintain arbitrary data in the AppConfig object
5. Add to Container
    * Requires making it resolvable (Directories class or similar that gets added to Container)
    * Expands PR scope

---

In my opinion, the visibility change is the minimal change that achieves the goal without expanding scope more than necessary. The extraction is functional, and we should be able to move forward with the cache implementation.

We can easily revisit what to do with it in a future adjustment.

Or, maybe what we're calling Bootstrapping for purposes of this change should just remain internal in the Kernel, and as you say, let "Initializers" be the mechanism that does anything else.

@brendt and @aidan-casey
i'll catch up this branch, and let you decide what happens to the PR. If there are specific ways you want it done differently, please let me know.

src/Application/Kernel.php
```diff
@@ -8,39 +8,38 @@
...
-        $this->initDiscoveryLocations();
+        $container = $this->createContainer();
+        $container->config($this->appConfig);
 
-        $this->initConfig($container);
+        foreach ($this->bootstraps as $bootstrap) {
+            $container->get($bootstrap)->boot();
+        }
 
         $this->initDiscovery($container);
```
jtheuerkauf, line 37: @brendt Given the extraction of the Discovery/Config code, this straggler `initDisovery()` feels out of place, but what should i do with it?

Honestly, i'm starting to question whether i was on the right track with this proposal at all. Seems to raise concerns and some clumsiness. Any suggestions are welcome, but i'm also okay with abandoning this for some other approach.

jtheuerkauf:
See #176

PR #155, opened 2024-02-26
Expands on dependency autowiring and allows for default parameter values to be passed

aidan-casey:

PR #163, opened 2024-02-27
Merges expanded dependency autowiring and default parameters with Brent's changes

aidan-casey:


aidan-casey:
@brendt - This is very much so draft mode, FYI and we should probably squash this one when (if) we merge.

Currently, passing the container to something marks it as a circular dependency, which kind of makes sense but also doesn't in this scenario.

src/Container/ContainerLogItem.php
```diff
@@ -9,33 +9,43 @@
...
+use Reflector;
 
 final readonly class ContainerLogItem
 {
     public function __construct(
         public string $id,
-        public ?ReflectionClass $class = null,
-        public ?ReflectionMethod $method = null,
-        public ?ReflectionParameter $parameter = null,
+        public ?Reflector $subject = null
```
brendt, line 18: Oh, this makes much more sense 😅

src/Container/ContainerLogItem.php
```diff
@@ -9,33 +9,43 @@
...
     ) {
     }
 
     public function __toString(): string
     {
-        return match (true) {
-            $this->parameter !== null => $this->parameterAsString(),
-            $this->method !== null => "{$this->method->getDeclaringClass()->getName()}::{$this->method->getName()}()",
-            $this->class !== null => "{$this->class->getName()}",
+        $subjectType = $this->subject ? $this->subject::class : null;
```
brendt, line 24: This makes much more sense as well… what was I thinking?? 😅

src/Container/ContainerLogItem.php
```diff
@@ -9,33 +9,43 @@
...
     }
 
-    private function parameterAsString(): string
+    private function formatParameter(): string
     {
-        $type = $this->shortTypeName($this->parameter->getType());
-        $class = $this->shortTypeName($this->parameter->getDeclaringClass());
+        $type = $this->shortTypeName($this->subject->getType());
+
+        return " [unresolved parameter: {$type} \${$this->subject->getName()}]";
```
brendt, line 38: We do need the location as well though, because otherwise it kind of becomes useless. I hard coded `__construct` because apparently `Parameter` has no reference to its method. We could improve that, but we can't keep it like this.

Also, de the `[ ]` brackets add value?

src/Container/ContainerLogStep.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Container;
+
+final class ContainerLogStep
```
brendt, line 7: Could you explain what this class is used for? Preferably as a comment in the code, so that we don't forget 5 months from now :)

tests/Container/Exceptions/CannotAutowireExceptionTest.php
```diff
@@ -18,10 +19,11 @@ public function test_autowire_without_exception()
...
 
             $container->get(ContainerObjectRequiringC::class);
-        } catch (CannotAutowireException $exception) {
-            $this->assertStringContainsString('string $prop in ContainerObjectC::__construct()', $exception->getMessage());
-            $this->assertStringContainsString('ContainerObjectC $c in ContainerObjectRequiringC::__construct()', $exception->getMessage());
-            $this->assertStringContainsString('Tests\Tempest\Container\Fixtures\ContainerObjectRequiringC', $exception->getMessage());
+            // TODO: Update the exception type.
+        } catch (Throwable $exception) {
+            $this->assertStringContainsString('[unresolved parameter: string $prop]', $exception->getMessage());
+            $this->assertStringContainsString('Tests\Tempest\Container\Fixtures\ContainerObjectC::__construct()', $exception->getMessage());
```
brendt, line 25: Oh maybe you removed the reference to the method because it's listed as the second line?

brendt (changes requested):

tests/Container/Exceptions/CannotAutowireExceptionTest.php
aidan-casey: Yeah, basically it pulls the method name and combines it with the class name to show what failed. Then the attempted types show below that in a tree. Subject to change a little bit because I am still playing around with this tonight.

src/Container/ContainerLogStep.php
aidan-casey: I will make this all more clear in a bit. Still working on the implementation a tad.

src/Container/ContainerLogItem.php
aidan-casey: The "step" keeps the context of the class / method / parameter so that we can output this information. Brackets just add visual appeal. 🙃

src/Container/ContainerLogItem.php
aidan-casey: I'm not sure. 🙃

aidan-casey:
@brendt - I entirely ripped out every logging per #165. This is essentially my original PR #155.

If you are good with these changes, I think we squash and merge.

PR #164, opened 2024-02-28
Remove Server

### LINKED ISSUE #162: Remove `Server`
This is old code that can be removed and simplified, probably.

brendt:
Closes #162 

It took some tinkering, but I think I landed on a clean solution for testing. Instead of writing this:

```php
$router = $this->container->get(Router::class);

$this->server(
    method: Method::POST,
    uri: uri([BookController::class, 'store']),
    body: $body,
);

$response = $router->dispatch(request($uri)->post($body));
```

We can now write this:

```php
$response = $this->send(new CreateBookRequest(
    method: Method::POST,
    uri: uri([BookController::class, 'store']),
    body:  [
        'title' => 'a',
    ],
));
```


brendt:
When this is merged, we'll need to refactor the `send` method to the http test helper in #153

brendt:
Ideally, this should work as well:

```php
$response = $this->send(new GenericRequest(
    method: Method::POST,
    uri: uri([BookController::class, 'store']),
    body:  [
        'title' => 'a',
    ],
));
```

Because that means we could add shorthand helpers like so:

```php
$response = $this->post(
    uri: uri([BookController::class, 'store']),
    body:  [
        'title' => 'a',
    ],
);
```

Right now, the mapping from generic request to the concrete request object is done in the request initializer; which is skipped because our `send` method overwrites that initializer with a concrete `Request` singleton. The benefit of doing that is that we don't have to worry about dealing with `$_SERVER` stuff in our tests. The downside is that we can't just send a generic request object as shown above, it needs to be the correct request class so that's used by the controller:

```
final readonly class BookController
{
    #[Post('/books')]
    public function store(CreateBookRequest $request): Response
    {
 ```

Maybe our router can be smart enough to convert between different instances of request classes? Just one idea

**Comment by brendt** [comment]
> I think it's pretty straight forward

It was! And it has a pretty cool advantage. Say you forget some data in when sending a request, Tempest will tell you:

```php
$this->http->post('/create-post'); // missing data

// Tempest\ORM\Exceptions\MissingValuesException: The following required properties are missing in App\Modules\Posts\PostRequest: title, text
```

HOW COOL IS THAT!!!

PR #151, opened 2024-02-26
Added PHPAT

Plytas:
Couldn't resist and added [PHPAT](https://www.phpat.dev/) as an alternative to Pest for architecture tests. This should be compared to #141

PR #169, opened 2024-02-28
Refactor tests to `Unit` namespace

Plytas:
In preparation for #151


aidan-casey:
@Plytas - do you have an easy way to optimize imports here? Would rather import the classes than use the FQDN.

Plytas:
I just used PhpStorm to move files. I see now where it failed to import. Will fix that.

Plytas:
@aidan-casey should be imported now.

@brendt can pass feedback to JetBrains :D
When moving classes it didn't properly import (or even provide an option to import via context menu) these cases. It seems to be related to using classes from the same namespace as they are not imported by default.
image

aidan-casey:
That's alright it gives me a good reason to add a CS-Fixer rule. 😉

aidan-casey:
Thank you for all your work here, @Plytas!

Plytas:
My pleasure! 🫡

PR #173, opened 2024-02-28
Restructures integration tests

### LINKED ISSUE #171: Create distinction between integration and unit tests
@brendt - I am thinking we should create an distinction between integration and unit tests for the framework. Unit tests would only ever extend `PHPUnit\Framework\TestCase` while integration tests might extend `Tempest\Testing\TestCase`. We could even add architecture tests for this to ensure each directory is being used appropriately.

Any objections to this?

aidan-casey:
This PR, if accepted, would close #171.

I started to play around with the ability to move the app directory, but that is going to require some change beyond the scope of this PR.


brendt:
Looks good! Could you take a look at the merge conflicts? Feel free to merge afterwards 👍

src/Testing/IntegrationTest.php
```diff
@@ -4,14 +4,14 @@
...
-use PHPUnit\Framework\TestCase as BaseTestCase;
+use PHPUnit\Framework\TestCase;
 use Tempest\AppConfig;
 use Tempest\Application\Kernel;
 use Tempest\Container\Container;
 use Tempest\Testing\Console\ConsoleCommandTester;
 use Tempest\Testing\Http\HttpRouterTester;
 
-abstract class TestCase extends BaseTestCase
+abstract class IntegrationTest extends TestCase
```
brendt, line 14: I renamed this to `IntegrationTest`. I don't like it when people write `TestCase` and then have to find to right one based on the namespace.

brendt (approved):
Made a couple of minor changes before merging

PR #175, opened 2024-02-28
Optimize route matching

### LINKED ISSUE #60: Optimize Route matching
Matching the request URL to a route is currently done via a looped Regex.

[Nikita wrote](https://www.npopov.com/2014/02/18/Fast-request-routing-using-regular-expressions.html) about how using one Regex to match all routes can speed up Route matching, which is also used in Symfonys Router.

Is this a change that would be of interest to the framework?

N-Silbernagel:
So I have given the route matching optimization I suggested in #60 a go. It is not quite done yet but I would appreciate some feedback.

## The Changes
Taking big inspiration from https://www.npopov.com/2014/02/18/Fast-request-routing-using-regular-expressions.html and https://nicolas-grekas.medium.com/making-symfonys-router-77-7x-faster-1-2-958e3754f0e1, I have refactored the route matching to utilize one big Regex and use that to match all available routes instead of matching all routes on their own.

The first step here is to do a hash-map lookup for static routes, therefore I changed the routes in the RouteConfig to an assoc array.
The next step is to match dynamic routes via a big Regex. For this, each Route now already has a cached matching regex ready. These routes are then "joined" using an "or" ( | ). To identified which route matched, I am using Marks, so I basically mark each route their respective numeric-index in the array. That looks like "(*Mark:1)".

The Regex will look something like this:
`#^(?|/route-1 (*MARK:0) |/route-2 (*MARK:1) |/99 (*MARK:2))$#`.

Short explanaition:
(?| => [branch reset group](https://www.npopov.com/2014/02/18/Fast-request-routing-using-regular-expressions.html#group-number-reset)
/route-1 => the regex of the route
(*Mark:0) => we can later identify a match by this mark

One problem this implementation still has is that we also match static routes in the dynamic part, but changing that would mean splitting routes in RouteConfig into "staticRoutes" and "dynamicRoutes", which I don't like too much.

## Performance
To test the performance, I created 5 Controllers. Those contain 531 GET Routes and 204 POST routes.
The GET routes consist of 112 static and 419 dynamic routes (1-2 params per dynamic route).

I then checked 4 different routes: The first static route, the last static route, the first dynamic route and the last dynamic route.

### First static route
before:
image
after:
image

1.802μs -> 2.089μs => 15,92% slower

### Last static route
before:
image
after:
image

247,153μs -> 1,822μs => 99,26% faster

### First dynamic route
before:
image
after:
image

3,803μs -> 44,208μs => 1.062,45% slower

### Last dynamic route
before:
image
after:
image

271,565μs -> 44,511μs => 83,60% faster

I used [PhpBench](https://github.com/phpbench/phpbench) and the following Bench class:
```php
<?php

declare(strict_types=1);

namespace Tests\Tempest\Benchmark;

use PhpBench\Attributes\Revs;
use Tempest\AppConfig;
use Tempest\Container\GenericContainer;
use Tempest\Discovery\RouteDiscovery;
use Tempest\Http\GenericRequest;
use Tempest\Http\GenericRouter;
use Tempest\Http\Method;
use Tempest\Http\RouteConfig;

class GenericRouterBench
{
    private GenericRouter $genericRouter;
    private GenericRequest $genericRequest;

    public function __construct()
    {
        $genericContainer = new GenericContainer();
        $appConfig = new AppConfig();


        $routeConfig = new RouteConfig([]);
        $routeDiscovery = new RouteDiscovery($routeConfig);
        $routeDiscovery->restoreCache($genericContainer);

        $this->genericRouter = new GenericRouter(
            $genericContainer,
            $appConfig,
            $routeConfig
        );

        // first static route
        $this->genericRequest = new GenericRequest(Method::GET, '/', []);
        // last static route
        //$this->genericRequest = new GenericRequest(Method::GET, '/99', []);
        // first dynamic route
        //$this->genericRequest = new GenericRequest(Method::GET, '/books/xyz', []);
        // last dynamic route
        //$this->genericRequest = new GenericRequest(Method::GET, '/test/a/b', []);
    }

    #[Revs(10_000)]
    public function benchDispatch(): void
    {
        $this->genericRouter->dispatch($this->genericRequest);
    }
}
```
So I skip loading the routes from the cache etc. and then dispatch the prepared request 10.000 times.

This is on an M2 pro.
I am not a bit benchmark guy, but the results seem reasonable. 
In conclusion, this solution should provide a pretty big performance improvement (speaking about relative numbers) on average.

## Should this be merged
Honestly I am not sure. It does bring some Performance improvements, -99,26% on last static route sounds amazing.
But I don't think it matters too much, we're talking about 200μs to 2μs.

On the other hand, it does add to the complexity of the code and hides a lot of logic behind a Regex.

Maybe it'd be a good compromise do match the static routes via a hash key lookup and leave the dynamic matching as is?

Let me know what you think.

Closes #60


aidan-casey:
First off, nice write up and well done! While @brendt will have the final call, I love my RegEx and this stuff is pretty cool. I think it's certainly a nice value add.

A couple of initial thoughts just from the technical side...

1. We could potentially optimize this more by grouping routes by prefixes, right? See [here](https://nicolas-grekas.medium.com/making-symfony-router-lightning-fast-2-2-19281dcd245b) for the part 2 on the Symfony article. It's an understandable improvement from a RegEx perspective.
2. There are some (very) minor improvements that we could make to the RegEx patterns.
    - You are replacing the route param string with `([\w\d\s]+)`. `\w` will account for `\d` so that is not necessary and adds another thing for RegEx to evaluate. 
    - We could implement some support for more specific matching (e.g., `{id:int}`) that equates to a more specific pattern (e.g., `\d`) which gives RegEx less to evaluate and will improve speed (again, we're dealing in the minors here).
    - We could also add direct support for RegEx matching and put that power back in the developers hands for those who are comfortable with it. Something to think about.
4. There is no reason that we couldn't refactor this to cache the built up RegEx pattern and only build it up once. This could be beneficial in production where these should not change and we could skip `foreach` loops, etc. Do you see any downsides to this?

> But I don't think it matters too much, we're talking about 200μs to 2μs.

Yeah, it is "minor" but it's also nice to be able to say the router is fast and that we likely won't have that limitation anytime soon. 😅 

> On the other hand, it does add to the complexity of the code and hides a lot of logic behind a Regex.

I'd be less concerned about this, personally. Not many people are as quick to jump into the router components like you are here. I think the key is well documenting whatever we opt for.

---

Do you have your benchmarks uploaded anywhere? I'd love to mess around with them.

N-Silbernagel:
Thanks for the very quick and elaborate feedback!

> We could potentially optimize this more by grouping routes by prefixes, right? See [here](https://nicolas-grekas.medium.com/making-symfony-router-lightning-fast-2-2-19281dcd245b) for the part 2 on the Symfony article. It's an understandable improvement from a RegEx perspective.

Yes this could be another optimization, I did not want to go into that for the first step as I think it makes it more complex. Then again we are now here to optimize it, so might as well optimize it to the fullest 😁

> You are replacing the route param string with ([\w\d\s]+). \w will account for \d so that is not necessary and adds another thing for RegEx to evaluate.

True! FastRoute and Symfony are using [^/], which is probably as simple as it gets.

> We could implement some support for more specific matching (e.g., {id:int}) that equates to a more specific pattern (e.g., \d) which gives RegEx less to evaluate and will improve speed (again, we're dealing in the minors here).
We could also add direct support for RegEx matching and put that power back in the developers hands for those who are comfortable with it. Something to think about.

I think this is actually more of a cool "validation" feature than a performance improvement, right? FastRoute does that using the {id:\d+} Syntax, Symfony uses {id<\d+>}. Giving users some "pre-built" patterns (such as the "int" you mentioned) sounds like a DX win.

When thinking about all these improvement and also having in the back of my head that we will probably need some form of setting priorities on routes at some point, it might be wise to jump back to @brendt's initial thought and use FastRoute or Symfony Router internally?

> There is no reason that we couldn't refactor this to cache the built up RegEx pattern and only build it up once. This could be beneficial in production where these should not change and we could skip foreach loops, etc. Do you see any downsides to this?

I did think about that but I am not sure if a cache read is faster than the loop here. I think with the current, serialized cache it would be slower. Maybe if we'd cache to PHP files to leverage opcache it would be faster.

> Do you have your benchmarks uploaded anywhere? I'd love to mess around with them.

I do not currently, I'll try to make a repo probably this evening.

aidan-casey:
> I did not want to go into that for the first step as I think it makes it more complex.

That makes total sense! Let's get some feedback from @brendt on whether he's willing to merge and pursue if so. I'm excited about the possibilities here. 🙂 

> `[^/]`

Oh, that makes total sense.

> I think this is actually more of a cool "validation" feature than a performance improvement, right?

Pretty much. The speed performance should have a negligible difference though, as it is matching a fewer amount of characters. 🙂 

It may be worth _only_ support RegEx for the validation and thinking about adding our own DX later on.

> it might be wise to jump back to @brendt's initial thought and use FastRoute or Symfony Router internally?

Where's the fun in that?! 😉 🙃 

> I did think about that but I am not sure if a cache read is faster than the loop here. I think with the current, serialized cache it would be slower.

You may be right, not sure.

> I do not currently, I'll try to make a repo probably this evening.

Cool, thanks!

N-Silbernagel:
> It may be worth only support RegEx for the validation and thinking about adding our own DX later on.

I also think it would be nice to support RegEx in params. But we'd have to think about what kind of RegEx should be allowed. I think Symfony disallows capturing group syntax for example (although this probably already a concern right now as we just use the route as our regex (minus params)).

> [^/]

I'll change this.

> Where's the fun in that?! 😉 🙃

That's the spirit 😄

> You may be right, not sure.

As always in performance: it's not going to be how we guess. I am going to benchmark this.

> Do you have your benchmarks uploaded anywhere? I'd love to mess around with them.

Just published a little benchmark repo https://github.com/N-Silbernagel/tempest-benchmark. You do need to mange the composer.json to make use of your local tempest repo. And I had to mess with the discoveryLocations which might be a concern for another issue... The discovery did not "see" the framework source because it's not a namespace of the project.

> I see that some tests are failing. Is this because of the changes here or because of the testsuite on that version?

My bad, removed the extended flag on the RegEx. Green now.

I did notice something else... for a route "/users/{userId}" we could match the route when a request goes to "/users/{userId}", because it'll be matched through the hashmap lookup. I'll also look into that.

N-Silbernagel:
> [^/]

Added.

> I did notice something else... for a route "/users/{userId}" we could match the route when a request goes to "/users/{userId}", because it'll be matched through the hashmap lookup. I'll also look into that.

Fixed.

> One problem this implementation still has is that we also match static routes in the dynamic part

Also fixed.

I cache if the route is dynamic now.

brendt:
Ok, first of all, @N-Silbernagel I'm seriously impressed by the effort your put into this, thank you! 

Second, there's a repo called [tempest-benchmark](https://github.com/tempestphp/tempest-benchmark) which contains a setup script for a couple thousand of controllers. Just run `php setup/setup.php` and it generates those controllers. You'll have to change the composer.json though to point to your fork of `tempest/framework`, or symlink it manually. 

Third, I actually think the implementation is pretty readable. I like how each route holds its own part of the regex. I might refactor the router part a bit when I'm not on my laptop, maybe add a separate function to encapsulate building the regex or something, and adding some comments explaining what is happening.

I actually think we'll merge this. 

And again, impressed!

N-Silbernagel:
I'm glad you like it an are willing to merge!

> Second, there's a repo called [tempest-benchmark](https://github.com/tempestphp/tempest-benchmark)

That is really cool, should have looked for that 😄

> I might refactor the router part a bit

Cool, excited to see what you make of it.

Should we keep the other possible changes like adding RegEx patterns to the params and grouping by prefixes for followup PRs?

brendt:
> Should we keep the other possible changes like adding RegEx patterns to the params and grouping by prefixes for followup PRs?

I think so, yes. Let me do a quick review here, and then I'll merge it 👍

src/Http/GenericRouter.php
```diff
@@ -172,4 +160,68 @@ private function createResponse(Response|View $input): Response
...
+    private function matchStaticRoute(Request $request): ?MatchedRoute
+    {
+        $staticRoute = $this->getRoutes()[$request->method->value][$request->getPath()] ?? null;
+
+        if ($staticRoute === null) {
+            return null;
+        }
+
+        // TODO: why do we need to check this?
+        if ($staticRoute->isDynamic) {
```
brendt, line 173: @N-Silbernagel not a blocker, but I suppose we can get rid of this check if we split dynamic and static routes in the config?

src/Http/GenericRouter.php
```diff
@@ -172,4 +160,68 @@ private function createResponse(Response|View $input): Response
...
+
+        if (! $matchResult || ! array_key_exists(self::MARK_TOKEN, $matches)) {
+            return null;
+        }
+
+        $route = $routesForMethod[$matches[self::MARK_TOKEN]];
+
+        // TODO: we could probably optimize resolveParams now,
+        // because we already know for sure there's a match
+        $routeParams = $this->resolveParams($route, $request->getPath());
```
brendt, line 217: @N-Silbernagel my assumption is that we can further optimise resolveParams, wdyt?

src/Http/GenericRouter.php
```diff
@@ -172,4 +160,68 @@ private function createResponse(Response|View $input): Response
...
+        // Next, we'll build one big regex to match the correct route
+        // See https://github.com/tempestphp/tempest-framework/pull/175 for the details
+
+        /** @var \Tempest\Http\Route[] $routesForMethod */
+        $routesForMethod = array_values($routesForMethod);
+        $combinedMatchingRegex = "#^(?|";
+
+        foreach ($routesForMethod as $routeIndex => $route) {
+            if (! $route->isDynamic) {
+                continue; // TODO: why this check?
```
brendt, line 197: Also the same question here: we can probably get rid of this check when we split dynamic and static routes?

src/Http/GenericRouter.php
```diff
@@ -172,4 +160,68 @@ private function createResponse(Response|View $input): Response
...
+
+        $route = $routesForMethod[$matches[self::MARK_TOKEN]];
+
+        // TODO: we could probably optimize resolveParams now,
+        // because we already know for sure there's a match
+        $routeParams = $this->resolveParams($route, $request->getPath());
+
+        // This check should _in theory_ not be needed,
+        // since we're certain there's a match
+        if ($routeParams === null) {
```
brendt, line 221: This check might be unnecessary if we refactor `resolveParams`

src/Http/GenericRouter.php
```diff
@@ -172,4 +160,68 @@ private function createResponse(Response|View $input): Response
...
+
+        // TODO: why do we need to check this?
+        if ($staticRoute->isDynamic) {
+            return null;
+        }
+
+        return new MatchedRoute($staticRoute, []);
+    }
+
+    private function matchDynamicRoute(Request $request): ?MatchedRoute
```
brendt, line 180: I think we're at a point where a dedicated `RouteMatcher` class might be useful.

brendt (approved):

brendt:
Left some comments, but I'll also create a new issue to keep track of further possible improvements.

src/Http/GenericRouter.php
N-Silbernagel: Yes, that would work, what I did not like about that, is that RouteConfig would not have a route param but maybe "staticRoutes" and "dynamicRoutes", which might not be ideal for reuse of the RouteConfig class.
On the other side, RouteConfig is probably only useful for the Router.

src/Http/GenericRouter.php
N-Silbernagel: True! The only case where it would return null right now is if the route Regex does not match which we have ensured before.

N-Silbernagel:
> There is no reason that we couldn't refactor this to cache the built up RegEx pattern and only build it up once. This could be beneficial in production where these should not change and we could skip foreach loops, etc. Do you see any downsides to this?

@aidan-casey 
I've put some time into testing this and the results are quite surprising to me.

For the baseline, I took my GenericRouterBench and pulled restoring the cache into the benchmark, so that caching is included in the timings.

I've got 4 different setups:

### Current main, with opcache
This version is using the loop to create the RegEx "on demand".

image

The difference between static and dynamic matching is quite high, as seen before.

### Caching with file_get_contents, with opcache
Simply writes the RegEx to a file and gets it via "file_get_contents".

image

Static matching is 9μs slower than on main, Dynamic matching 27μs faster. So my initial assumption was wrong.

### Caching with require, without opcache
Write the cache as a php file `file_put_contents(self::CACHE_PATH_REGEX, "<?php return " . var_export($this->routeConfig->regex, true) . ";")`.
And gets it via `require self::CACHE_PATH_REGEX;`.

image

### Caching with require, with opcache
Same as the previous but opcache is enabled.

image

Basically as fast as main for static route but 45μs faster for dynamic routing! Not a lot of difference between static and dynamic routing.

The code is here https://github.com/N-Silbernagel/tempest-framework/tree/cache. Keep in mind it is not what I would want to merge, just a quick implementation for testing.

src/Http/GenericRouter.php
brendt: That's a good point. Let's actually keep it as it is for now, we can always improve on it later if needed.

brendt:
I still think these are promising results. Could you add one more benchmark? Try to store it as a class, eg. `RouteCache` with a method "getRegex" or something, then use this class, relying on composer's autoloader. The reason I'm asking is because I think opcache will be able to cache a class, but not a hardcoded plain PHP file — _I think_.

Also, even with the simple `require` call, 45μs is a 10% performance gain, which I'd say isn't bad at all.

src/Http/GenericRouter.php
N-Silbernagel: Yes I was also thinking about that... the tricky part is finding which params are which matching group in the RegEx. I was thinking of either storing the results of the first "preg_match_all" in resolveParams in the route class to have all the tokens cached or using named capturing groups in the "big RegEx".

N-Silbernagel:
I've now tested with the class and am getting similar results.

From main:
image

With plain php file:
image

With class:
image

I'd say there is not too much of a difference here. What I think is a downside with using autoloading is that using optimized autoloader would probably not work.

brendt:
> Also just for fun I've tried to cache the routes via php files as well. Which results in these timings:

I suppose you mean instead of serializing them? That's actually an impressive result. Would you be able to PR that part as a standalone PR?

For the other changes: I'm totally up for merging another PR with the improvements you've been tinkering with. Probably the plain PHP file is the way to go.

PR #176, opened 2024-02-29
Refactor Kernel to extract Bootstraps

jtheuerkauf:
* Remove `Tempest::__construct()` `$appConfig` parameter - available from Container object
* Add `Kernel::__construct()` `$bootstraps` promoted property
* Rename `Kernel::initContainer()`
* Add Bootstrap classes for Discovery and Config
* Loop `$bootstraps` to register them in the Container as singletons
* Execute Bootstraps `boot()` in `Kernel::init()`
* Set Bootstraps as third Kernel constructor argument
* Fixed some accessors to AppConfig from the `$kernel` object

### DIFF
--- a/src/Application/Kernel.php
+++ b/src/Application/Kernel.php
@@ -9,6 +9,8 @@ use RecursiveIteratorIterator;
 use ReflectionClass;
 use SplFileInfo;
 use Tempest\AppConfig;
+use Tempest\Bootstraps\ConfigBootstrap;
+use Tempest\Bootstraps\DiscoveryBootstrap;
 use Tempest\Container\Container;
 use Tempest\Container\GenericContainer;
 use Tempest\Database\PDOInitializer;
@@ -22,17 +24,19 @@ final readonly class Kernel
     public function __construct(
         public string $root,
         private AppConfig $appConfig,
-        private array $bootstraps = [],
     ) {
     }
 
     public function init(): Container
     {
         $container = $this->createContainer();
-        $container->config($this->appConfig);
 
-        /** @var class-string<\Tempest\Bootstraps\Bootstrap> $bootstrap */
-        foreach ($this->bootstraps as $bootstrap) {
+        $bootstraps = [
+            DiscoveryBootstrap::class,
+            ConfigBootstrap::class,
+        ];
+
+        foreach ($bootstraps as $bootstrap) {
             $container->get($bootstrap)->boot();
         }
 
@@ -48,7 +52,8 @@ final readonly class Kernel
         GenericContainer::setInstance($container);
 
         $container
-            ->singleton(__CLASS__, fn () => $this)
+            ->config($this->appConfig)
+            ->singleton(self::class, fn () => $this)
             ->singleton(Container::class, fn () => $container)
             ->addInitializer(new RequestInitializer())
             ->addInitializer(new RouteBindingInitializer())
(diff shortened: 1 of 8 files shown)


src/Application/Kernel.php
```diff
@@ -9,46 +9,51 @@
...
     ) {
     }
 
     public function init(): Container
     {
-        $container = $this->initContainer();
+        $container = $this->createContainer();
 
-        $this->initDiscoveryLocations();
+        $bootstraps = [
```
brendt, line 29: There's no need to make bootstrappers configurable from the outside for now. Let's keep it internal until we have an actualy need for making them configurable.

src/Application/Kernel.php
```diff
@@ -4,184 +4,55 @@
...
-    private function initContainer(): Container
+    private function createContainer(): Container
     {
         $container = new GenericContainer();
 
         GenericContainer::setInstance($container);
 
         $container
-            ->singleton(Kernel::class, fn () => $this)
+            ->config($this->appConfig)
```
brendt, line 49: I moved appConfig registration to here, makes more sense

src/Bootstraps/DiscoveryBootstrap.php
```diff
@@ -0,0 +1,77 @@
...
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use ReflectionClass;
+use SplFileInfo;
+use Tempest\AppConfig;
+use Tempest\Container\Container;
+use Tempest\Discovery\Discovery;
+use Throwable;
+
+final readonly class DiscoveryBootstrap implements Bootstrap
```
brendt, line 16: I moved the `initDiscovery` method to a bootstrapper as well

src/Bootstraps/DiscoveryLocationBootstrap.php
```diff
@@ -0,0 +1,103 @@
...
+
+namespace Tempest\Bootstraps;
+
+use Tempest\AppConfig;
+use Tempest\Application\Kernel;
+use Tempest\Container\Container;
+use Tempest\Discovery\DiscoveryLocation;
+use function Tempest\path;
+
+final readonly class DiscoveryLocationBootstrap implements Bootstrap
```
brendt, line 13: I renamed this to DiscoverLocationBootstrap

src/Tempest.php
```diff
@@ -17,60 +16,60 @@
...
     {
         $dotenv = Dotenv::createUnsafeImmutable($root);
+
         $dotenv->safeLoad();
 
-        $createAppConfig ??= fn () => new AppConfig(
-            environment: Environment::from(env('ENVIRONMENT', Environment::LOCAL->value)),
-            discoveryCache: env('DISCOVERY_CACHE', false),
-            enableExceptionHandling: env('EXCEPTION_HANDLING', false),
+        $kernel = new Kernel(
```
brendt, line 28: The createAppConfig closure was only needed if we wanted to inject appconfig from the outside because dotenv had to load first. Without it, we can simply construct it

brendt (commented):
Great work 👍 

I made some changes to this branch so that I could merge them, I left some comments just so that you know what I changed and why.

src/Bootstraps/DiscoveryBootstrap.php
jtheuerkauf: It was right in front of me the whole time. 😫

PR #177, opened 2024-02-29
Container log

### LINKED ISSUE #165: Refactor Container Logging
Okay, per our discussion, below is the spec for container logging. @brendt will own the implementation of this according to the spec below and I will review the PR once it is submitted.

None of these classes should be self-formatting, but we should have the ability to get the first and last context so we could traverse up the dependency tree and format according to type, context step, etc.

All of these names can be changed if appropriate, it is the functionality we are after.

# Context Class
The **Context** class keeps track of the context of the dependency we are trying to resolve. For example:

- If we are attempting to resolve a method, the context is a class.
- If we are attempting to resolve a parameter, the context is a method.

The context class should keep a reference to the _previous context_, _next context_, and which context step it is (every step down the dependency tree increases that number). It should also provide some basic helpers to traverse that tree.

# Dependency Class
The **Dependency** class keeps track of the current dependency we are trying to resolve within the current context. For example:

- If we are attempting to resolve a DateTime type in the parameter context, our dependency is `DateTime`.

# Logging Class
The **Logging** class handles all the coordination needed to keep track of where we are in the dependency tree. Previously this has been called `ContainerLog`, but the name can stay or go based on whether it makes sense.

The container only has _one_ instance of this class and it is set on an internal property on creation. I don't think the implementation will necessarily be inefficient, but this _may_ or _may not_ be an interface so you could swap out for a throwaway option in production or something.

The method names below can be revised as appropriate, but we should be able to do the following:

## `startResolving(Context $context)`
Marks that we have started resolving a new context.

## `finishResolving()`
Marks that we have finished resolving the current context.

## `addDependency(Dependency $dependency)`
Adds a dependency to the current context.

## `classResolutionStackContains(string $className): bool`
Returns true if the class resolution stack contains the passed class name. This is used by the container to try and return a singleton OR throw a circular dependency exception if a singleton doesn't exist.

## `getFirstContext(): Context`
Gets the very first context we tried resolving in the stack. When we throw an exception from the container, it can use this to traverse _down_ the dependency tree since each context keeps a reference of the next context.

## `getLastContext(): Context`
Gets the very last context we tried resolving in the stack. When we throw an exception from the container, it can use this to traverse _up_ the dependency tree since each context keeps a reference of the previous context.

brendt:
closes #165

### DIFF
--- a/src/Container/Context.php
+++ b/src/Container/Context.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Tempest\Container;
 
 use ReflectionClass;
@@ -12,7 +14,8 @@ final class Context
         public ReflectionClass|ReflectionMethod|ReflectionFunction $reflector,
         /** @var \Tempest\Container\Dependency[] $dependencies */
         public array $dependencies = [],
-    ) {}
+    ) {
+    }
 
     public function addDependency(Dependency $dependency): self
     {
@@ -37,9 +40,9 @@ final class Context
     public function __toString(): string
     {
         return match($this->reflector::class) {
-            ReflectionClass::class => $this->reflector->getShortName(),
-            ReflectionMethod::class => $this->reflector->getDeclaringClass()->getShortName() . '::' . $this->reflector->getShortName() . '()',
-            ReflectionFunction::class => $this->reflector->getShortName() . '()',
+            ReflectionClass::class => $this->classToString($this->reflector),
+            ReflectionMethod::class => $this->methodToString($this->reflector),
+            ReflectionFunction::class => $this->functionToString($this->reflector),
         };
     }
 
@@ -50,11 +53,22 @@ final class Context
 
     private function methodToString(ReflectionMethod $reflector): string
     {
-        return $reflector->getDeclaringClass()->getShortName() . '::' . $reflector->getName() . '()';
+        return $reflector->getDeclaringClass()->getShortName() . '::' . $reflector->getName() . '(' . $this->dependenciesToString() . ')';
     }
 
     private function functionToString(ReflectionFunction $reflector): string
     {
-        return $reflector->getShortName() . '()';
+        return $reflector->getShortName() . '(' . $this->dependenciesToString() . ')';
+    }
+
+    private function dependenciesToString(): string
+    {
+        return implode(
+            ', ',
+            array_map(
+                fn (Dependency $dependency) => (string) $dependency,
+                $this->dependencies,
+            )
+        );
     }
 }
--- a/src/Container/Context.php
+++ b/src/Container/Context.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Tempest\Container;
+
+use ReflectionClass;
+use ReflectionFunction;
+use ReflectionMethod;
+
+final class Context
+{
+    public function __construct(
+        public ReflectionClass|ReflectionMethod|ReflectionFunction $reflector,
+        /** @var \Tempest\Container\Dependency[] $dependencies */
+        public array $dependencies = [],
+    ) {}
+
+    public function addDependency(Dependency $dependency): self
+    {
+        $this->dependencies[] = $dependency;
+
+        return $this;
+    }
+
+    public function currentDependency(): ?Dependency
+    {
+        return $this->dependencies[array_key_last($this->dependencies)] ?? null;
+    }
+
+    public function getId(): string
+    {
+        return match($this->reflector::class) {
+            ReflectionClass::class, ReflectionFunction::class => $this->reflector->getName(),
+            ReflectionMethod::class => $this->reflector->getDeclaringClass()->getName(),
+        };
+    }
+
+    public function __toString(): string
+    {
+        return match($this->reflector::class) {
+            ReflectionClass::class => $this->reflector->getShortName(),
+            ReflectionMethod::class => $this->reflector->getDeclaringClass()->getShortName() . '::' . $this->reflector->getShortName() . '()',
+            ReflectionFunction::class => $this->reflector->getShortName() . '()',
+        };
+    }
+
+    private function classToString(ReflectionClass $reflector): string
+    {
+        return $reflector->getShortName();
+    }
+
+    private function methodToString(ReflectionMethod $reflector): string
+    {
+        return $reflector->getDeclaringClass()->getShortName() . '::' . $reflector->getName() . '()';
+    }
+
+    private function functionToString(ReflectionFunction $reflector): string
+    {
+        return $reflector->getShortName() . '()';
+    }
+}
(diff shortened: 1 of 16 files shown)


brendt:
I didn't add the next and previous references to contexts. The only time you need it is for rendering exceptions, and I find `array_key_first/last` to be sufficient.

brendt:
This is the output of `CircularDependencyException`:

```txt
Tempest\Container\Exceptions\CircularDependencyException: 

Cannot autowire Tests\Tempest\Unit\Container\Fixtures\CircularA because it is a circular dependency:

	┌─► CircularA::__construct(ContainerObjectA $other, CircularB $b)
	│   CircularB::__construct(CircularC $c)
	│   CircularC::__construct(ContainerObjectA $other, CircularA $a)
	└───────────────────────────────────────────────────▒▒▒▒▒▒▒▒▒

Originally called in /Users/brentroose/Dev/tempest-framework/tests/Unit/Container/Exceptions/CircularDependencyExceptionTest.php:22
```

And this for `CannotAutowireException`:

```txt
Cannot autowire Tests\Tempest\Unit\Container\Fixtures\AutowireA because Tests\Tempest\Unit\Container\Fixtures\AutowireC cannot be resolved

	┌── AutowireA::__construct(AutowireB $b)
	├── AutowireB::__construct(AutowireC $c)
	└── AutowireC::__construct(ContainerObjectA $other, string $unknown)
	                                                    ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒

Originally called in /Users/brentroose/Dev/tempest-framework/tests/Unit/Container/Exceptions/CannotAutowireExceptionTest.php:22
```

We can improve on this further in the future. What's most important to me is that these exceptions show what's wrong, why, and where.

brendt:
Also, there will be a lot of edge cases when it comes to the rendering of these exception messages. I say we just fix them as they pop up.

src/Container/Exceptions/CannotAutowireException.php
```diff
@@ -0,0 +1,50 @@
...
+
+            $message .= PHP_EOL . "\t{$pipe} " . $currentContext;
+
+            $i++;
+        }
+
+        $currentDependency = $lastContext->currentDependency();
+        $currentDependencyName = (string) $currentDependency;
+        $firstPart = explode($currentDependencyName, (string) $lastContext)[0] ?? null;
+        $fillerLines = str_repeat(' ', strlen($firstPart) + 3);
```
aidan-casey, line 38: Shouldn't we rename this to something like `$fillerSpaces` for clarity?

src/Container/Context.php
```diff
@@ -0,0 +1,82 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Container;
+
+use ReflectionClass;
+use ReflectionFunction;
+use ReflectionMethod;
+
+final class Context
```
aidan-casey, line 11: Can this be `readonly` to prevent the modification of the class properties (if they need to be exposed)?

src/Container/GenericContainer.php
```diff
@@ -54,17 +57,15 @@ public function config(object $config): self
 
     public function get(string $className): object
     {
-        try {
-            return $this->resolve($className);
-        } catch (Throwable $throwable) {
-            // TODO: Logging
+        $this->log->startResolving();
```
aidan-casey, line 60: Probably not something to fix before merging this PR, but something to keep in mind...

If I have `ClassA` with a dependency on `ClassB` that has a dependency on `ClassC` and if `ClassC` looks like the following:

```php
class ClassC
{
    private SomeDependency $blah;

    public function __construct(Container $container)
    {
        $this->blah = $container->get(SomeDependency::class);
    }
}
```

Won't that reset my log for the autowiring of `ClassA`? I think the fix for this would be to store the originator information in a context as well, so that it is not tied to the master log class.

There's a chance this is what was affecting initializers as well?

src/Container/InMemoryContainerLog.php
```diff
@@ -0,0 +1,71 @@
...
+final class InMemoryContainerLog implements ContainerLog
+{
+    private string $origin = '';
+
+    public function __construct(
+        /** @var \Tempest\Container\Context[] $stack */
+        private array $stack = [],
+
+        /** @var \Tempest\Container\Dependency[] $stack */
+        private array $dependencies = [],
```
aidan-casey, line 19: Is there any real reason to have these exposed?

src/Container/InMemoryContainerLog.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function getStack(): array
+    {
+        return $this->stack;
+    }
+
+    public function currentContext(): Context
+    {
+        return $this->stack[array_key_last($this->stack)]
+            ?? throw new Exception("No current context found. That shoudn't happen. Aidan probably wrote a bug somewhere.");
```
aidan-casey, line 59: 🤣 

I recommend refactoring this to use an `AidanProbablyWroteIt` exception so we can do some reporting on how often these come up.

aidan-casey (commented):
Looks pretty good! Just a few comments.

src/Container/Context.php
brendt: Technically, it can, because you can add items to readonly arrays. But I find that confusing. A property is readonly, but you can still change its contents. That's why I prefer non-readonly properties in those cases.

However, these should just be private, I think

src/Container/Exceptions/CannotAutowireException.php
brendt: Definitely! I really don't think this is the greatest code btw. But it works for now, and is contained in one class, so I'm fine with it.

It would be interesting to build some kind of "highlighting component", but that's not for today 😁

src/Container/GenericContainer.php
brendt: That's exactly the problem initializers also have, yeah. You'd need to be able to distinguish whether a `container->get` call is made by the user, or originates from a nested container call.

I actually solved this problem in untitled by implementing two containers. They were the same (with a trait), except that one implementation didn't reset the log on `get`. I called this the `ContainerForInitializers` implementation, and they shared the same log instance.

Whenever I passed the container to an initializer from within the container, I'd give it that other implementation that didn't reset the log on `get`, we could do the same for injected container instances to solve this problem.

However, maybe it's easier to check `debug_backtrace` and determine whether we reset the log or not based on that info?

> I think the fix for this would be to store the originator information in a context as well, so that it is not tied to the master log class.

Not sure I understand, maybe this is an even better solution.

It would be nice if we could fix this, but let's first merge this PR 👍

brendt: I made an issue for it: https://github.com/tempestphp/tempest-framework/issues/180

src/Container/InMemoryContainerLog.php
brendt: No, this is a remnant of my first implementation where I did circular checks on dependencies instead of contexts, which didn't make any sense. I'll remove it

brendt:
Let us merge.

Issue #113, opened 2024-02-22
Add Cookie Support


aidan-casey:
We need to add support for cookies to the router. _Technically_ this could be done manually using a header right now, but we should give it first class support.

Issue #14, opened 2023-07-19
Catch validation exceptions in the router


brendt:

brendt:
I don't remember what I thought when I created this issue, but I'm pretty sure this isn't the way I want to do it anymore. Will look into it later

aidan-casey:
<img width="402" alt="Screenshot 2024-02-23 at 2 52 57 PM" src="https://github.com/tempestphp/tempest-framework/assets/6686277/8528c09a-ae01-4381-8353-130cdce5dfbc">

PR #178, opened 2024-03-01
Adds basic session implementation

### LINKED ISSUE #123: Add Session Support
It would be good to support sessions in Tempest. The way I see it, we have two options for this:

1. Session interface that different session storage types can implement.
2. Session storage interface that gets passed a session class owned by Tempest. This solution allows us to offer some "ease of access" items, such as dot notation, if we desire.

Any thoughts, @brendt?

aidan-casey:
Will (eventually) close #123.


aidan-casey:
Just a few thoughts as I work on a basic implementation here:

1. We should determine what our approach is to attempting to `get` or `set` a value when the session isn't started. Do we attempt to start it or fail fast (throw an exception)?
2. We should figure out more of the API we want to support. Do we want to add support for dot notation, etc.?

brendt:
> We should determine what our approach is to attempting to get or set a value when the session isn't started. Do we attempt to start it or fail fast (throw an exception)?

Definitely start it 👍

> We should figure out more of the API we want to support. Do we want to add support for dot notation, etc.?

Could be useful, there's btw already a class `ArrayHelper` that can unwrap dot notation, I don't think it can handle to other way around yet. I needed it for request validation.

src/Clock/Clock.php
```diff
@@ -0,0 +1,16 @@
...
+
+use DateTimeInterface;
+
+interface Clock
+{
+    public function now(): DateTimeInterface;
+
+    public function time(): int;
+
+    public function sleep(int $seconds): void;
```
brendt, line 15: What's with this `sleep` method?

src/Clock/GenericClock.php
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public function time(): int
+    {
+        return time();
+    }
+
+    public function sleep(int $seconds): void
+    {
+        sleep($seconds);
```
brendt, line 24: Isn't sleep only useful for testing? That's an indication it shouldn't be on the interface but rather on `MockClock`

src/Http/Session/ArraySessionHandler.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Session;
+
+use SessionHandlerInterface;
+use Tempest\Clock\Clock;
+
+final class ArraySessionHandler implements SessionHandlerInterface
```
brendt, line 10: I'm curious because I've never really looked into session handling. When would an in-memory array session handler without persistence be useful? During testing?

src/Http/Session/Session.php
```diff
@@ -0,0 +1,119 @@
...
+
+    // TODO: allow the passing of options
+    public function __construct(private SessionHandlerInterface $sessionHandler)
+    {
+        // Here we just make sure that if the application ever
+        // prematurely stops (e.g., we call exit or die),
+        // the session data will still be saved.
+        session_register_shutdown();
+
+        $this->setSessionHandler($this->sessionHandler);
```
brendt, line 23: Is setting the sessions handler something that needs to be accessed from the outside? When will people actually want to change their session handler while handling a request?

src/Http/Session/Session.php
```diff
@@ -0,0 +1,119 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Session;
+
+use const PHP_SESSION_ACTIVE;
+use RuntimeException;
+use SessionHandlerInterface;
+
+final class Session
```
brendt, line 11: We need an initializer here I think

src/Clock/GenericClock.php
aidan-casey: I've definitely used `sleep` for non-testing things before in CLI. Most clock implementations I've seen tend to include it (e.g., [Symfony](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Clock/Clock.php)).

src/Clock/Clock.php
aidan-casey: Continued below.

src/Http/Session/ArraySessionHandler.php
aidan-casey: Primarily during testing, yeah. 🙂

src/Http/Session/Session.php
aidan-casey: When they want to seriously mess with their users. 🙃

For me, that was more of a consistency thing. The one scenario I can think of is if you ever wanted to migrate existing sessions to a new session handler (e.g., array to database).

src/Http/Session/Session.php
aidan-casey: Yep. Usually the last item in my workflow, just in case API changes.

src/Clock/GenericClock.php
brendt: Oh, TIL. Then let's keep it 👍

src/Http/Session/Session.php
brendt: Ok. It's kind of an edge case, but I think a good enough reason to keep it as is

brendt:
No don't add PSR-20 🫣

brendt:
@aidan-casey I say we merge this, since it's blocking multiple issues:

- #113 
- #14

brendt:
We can improve on it later 👍

PR #184, opened 2024-03-02
Rework initializers

### LINKED ISSUE #183: Re-think initializers
One thing that has always bothered me about the `#[InitializedBy]` attribute is that it makes the interface somewhat aware of the implementation. I am proposing the following as a solution to this problem. I believe this is a serious issue for Tempest and would love to help drive a solution to that.

Below are some of my initial thoughts here, though I am sure I am missing some edge cases and things you have thought about @brendt, so feel free to school me. 🙂 

# Make Interfaces Unaware of Implementations

The first step to making interfaces unaware of implementations is moving the logic that tells the container how to create an object away from the interface. The initializer itself will inform the container what it provides and, for now, the container will register these in the order they are scanned. In the long run we might add a priority property or something.

To do this, I propose that we replace the `#[InitializedBy()]` with the `#[Initializes()]` attribute which is only available on _methods_.

For example, we might have the following initializer class for our http client:

```php
final class TempestHttpClientInitializer
{
    // Shorthand that wraps this initializer call and makes
    // the result a singleton in the container.
    #[Singleton]

    // Informs the container that this initializer handles
    // this class.
    #[Initializes(HttpClientDriver::class)]

    // Gets called when it's time to initialize HttpClientDriver
    public function initializePsr18Driver(): HttpClientDriver
    {
        return new Psr18Driver(...);
    }

    #[Initializes(HttpClient::class)]
    public function initializeHttpClient(): HttpClient
    {
        return new GenericHttpClient(...);
    }
}
```

Behind the scenes, this basically becomes a register call. The initializer method will be wrapped by a `Callable` class that allows for a reflection method _or_ function to be passed. When it comes time to initialize, we just grab the key for that class and run the `call` method. So something like this:

```php
if (isset($this->singletons[$class]) !== false) {
    return $this->singletons[$class];
}

if (isset($this->definitions[$class]) !== false) {
    // This will actually autowire the initializer,
    // which means we can add our dependencies
    // to the construct if we need to. That said,
    // this may be an initializer, or it may be
    // a callable.
    return $this->definitions[$class]->call();
}

return $this->autowire($class);
```

This has the bonus side effect of enforcing separation of concerns and some good contracts/dependencies.

brendt:
Closes #183

TODO:

- [x] Add `#[Singleton]`


brendt:
Eventually I decided to not go with `ContainerConfig`, since it would be a second config that's loaded before config bootstrap. Instead I just keep track of definitions on the container itself, as used to happen anyway.

src/Bootstraps/DiscoveryLocationBootstrap.php
```diff
@@ -6,18 +6,15 @@
...
 use Tempest\Discovery\DiscoveryLocation;
 use function Tempest\path;
 
 final readonly class DiscoveryLocationBootstrap implements Bootstrap
 {
-    private string $root;
-
     public function __construct(
-        private Container $container,
+        private AppConfig $appConfig,
```
brendt, line 15: I did some code cleanup in this file, it's not entirely related to this PR, but I stumbled across it when debugging some issues for this PR, so I include it here.

src/Container/Container.php
```diff
@@ -17,9 +19,14 @@ public function config(object $config): self;
      * @param class-string<TClassName> $className
      * @return TClassName
      */
-    public function get(string $className): object;
+    public function get(string $className, mixed ...$params): object;
```
brendt, line 22: I added support for passing extra params in `get`. I did this when I was debugging some issues during implementation with circular dependencies.

In the end, it turns out I didn't need it anyway. So I might remove it still. On the other hand, it might be a nice feature, and we already support it via `call`. So why not include it here?

src/Container/Context.php
```diff
@@ -32,41 +32,43 @@ public function currentDependency(): ?Dependency
     public function getName(): string
     {
         return match($this->reflector::class) {
-            ReflectionClass::class, ReflectionFunction::class => $this->reflector->getName(),
+            ReflectionClass::class => $this->reflector->getName(),
```
brendt, line 35: I ran into some bugs with context naming when working on this PR. I fixed them here, but they aren't really related to this PR. There's room for further improvement and I made an issue for that here: https://github.com/tempestphp/tempest-framework/issues/185

src/Container/GenericContainer.php
```diff
@@ -12,19 +12,27 @@
...
 {
     use HasInstance;
 
     public function __construct(
-        private array $definitions = [],
-        private array $singletons = [],
-        /** @var (Initializer&CanInitialize)[] */
-        private array $initializers = [],
-        private ContainerLog $log = new InMemoryContainerLog(),
+        public array $definitions = [],
```
brendt, line 22: Going back and forth to and from a `ContainerConfig`; I accidentally made these properties public. I'll need to change them back to private, but I'll also need to add methods to get this data, because I need it during testing.

src/Container/GenericContainer.php
```diff
@@ -55,11 +63,11 @@ public function config(object $config): self
         return $this;
     }
 
-    public function get(string $className): object
+    public function get(string $className, mixed ...$params): object
     {
         $this->log->startResolving();
 
-        return $this->resolve($className);
+        return $this->resolve($className, ...$params);
```
brendt, line 70: Added `...$params` support, not sure whether to keep it.

src/Container/GenericContainer.php
```diff
@@ -75,14 +83,34 @@ public function call(string|object $object, string $methodName, ...$params): mix
         return $reflectionMethod->invokeArgs($object, $parameters);
     }
 
-    public function addInitializer(CanInitialize $initializer): Container
+    public function addInitializer(ReflectionClass|string $initializerClass): Container
```
brendt, line 86: Todo: add explaining comments

src/Container/GenericContainer.php
```diff
@@ -108,39 +136,41 @@ private function resolve(string $className): object
                 $initializer->setClassName($className);
             }
 
-            return $initializer->initialize($this);
+            $object = $initializer->initialize($this);
```
brendt, line 139: Todo: refactor into separate methods, since this one is growing too large now.

src/Discovery/EventBusDiscovery.php
```diff
@@ -14,7 +14,7 @@
 
 final readonly class EventBusDiscovery implements Discovery
 {
-    private const CACHE_PATH = __DIR__ . '/event-bus-discovery.cache.php';
+    private const string CACHE_PATH = __DIR__ . '/event-bus-discovery.cache.php';
```
brendt, line 17: We should add a CSFixer rule for typing constants.

src/Container/Container.php
aidan-casey: I think we should totally leave this here. It'd be required anytime you need primitives added anyway.

src/Container/GenericContainer.php
aidan-casey: Not sure if we need a container config, but I'll reserve the right to change my mind. 😉 

Rather than add a getter for definitions, what about adding a `has` method?

src/Application/Kernel.php
```diff
@@ -49,9 +53,9 @@ private function createContainer(): Container
             ->config($this->appConfig)
             ->singleton(self::class, fn () => $this)
             ->singleton(Container::class, fn () => $container)
-            ->addInitializer(new RequestInitializer())
-            ->addInitializer(new RouteBindingInitializer())
-            ->addInitializer(new PDOInitializer());
+            ->addInitializer(RequestInitializer::class)
+            ->addInitializer(RouteBindingInitializer::class)
+            ->addInitializer(PDOInitializer::class);
```
aidan-casey, line 58: Somewhat unrelated, but is there a reason we don't move these to a bootstrapper?

src/Bootstraps/DiscoveryLocationBootstrap.php
```diff
@@ -32,7 +29,7 @@ public function boot(): void
 
     private function discoverInstalledPackageLocations(): array
     {
-        $composerPath = path($this->root, 'vendor/composer');
+        $composerPath = path($this->kernel->root, 'vendor/composer');
```
aidan-casey, line 32: I wish we could refactor this somehow. Does this belong in `AppConfig`?

src/Console/Console.php
```diff
@@ -4,9 +4,6 @@
 
 namespace Tempest\Console;
 
-use Tempest\Container\InitializedBy;
-
-#[InitializedBy(ConsoleInitializer::class)]
 interface Console extends ConsoleInput, ConsoleOutput
```
aidan-casey, line 7: Kinda unrelated to this PR, but do we actually need this interface?

src/Console/ConsoleInitializer.php
```diff
@@ -6,18 +6,16 @@
...
 class ConsoleInitializer implements Initializer
 {
-    public function initialize(Container $container): object
+    public function initialize(Container $container): Console
     {
-        $console = new GenericConsole(
+        return new GenericConsole(
             $container->get(ConsoleInput::class),
             $container->get(ConsoleOutput::class),
         );
```
aidan-casey, line 19: To me, this is a good example of where having initializer methods make sense. For example:

```php
class ConsoleInitializer
{
    #[Singleton]
    #[Initializes(ConsoleInput::class)]
    #[Initializes(GenericConsoleInput::class)]
    public function initializeConsoleInput(): ConsoleInput
    {
        return new GenericConsoleInput(
            output: $this->container->get(ConsoleOutput::class)
        );
    }

    #[Singleton]
    #[Initializes(
        ConsoleOutput::class,
        GenericConsoleOutput::class
    )] // Alternative syntax.
    public function initializeConsoleOutput(): ConsoleOutput
    {
        return new GenericConsoleOutput();
    }
}
```

src/Container/DynamicInitializer.php
```diff
@@ -4,7 +4,9 @@
 
 namespace Tempest\Container;
 
-interface CanInitialize
+interface DynamicInitializer
 {
     public function canInitialize(string $className): bool;
+
+    public function initialize(string $className, Container $container): object;
 }
```
aidan-casey, line 12: To me, something like an `#[InitializesAny(Request::class)]` attribute makes sense to me here, but I may just be unduly biased toward attributes. 😛

src/Container/GenericContainer.php
aidan-casey: Yes. 😉

src/Container/GenericContainer.php
```diff
@@ -12,21 +12,28 @@
...
+         * @var class-string<T> $initializers
+         */
+        public array $initializers = [],
+
+        /**
+         * @template T of \Tempest\Container\DynamicInitializer
+         * @var class-string<T> $dynamicInitializers
+         */
+        public array $dynamicInitializers = [],
+        private readonly ContainerLog $log = new InMemoryContainerLog(),
```
aidan-casey, line 35: Seems like we could clean this up. Not sure why we need to expose some of this stuff.

src/Discovery/EventBusDiscovery.php
aidan-casey: I gotchu.

src/Container/GenericContainer.php
brendt: Not sure about this one actually. Gonna keep it as is for now.

src/Console/Console.php
brendt: Yes, for convenience. Most of the time, you're dealing with both input and output, and having one interface that handles both is much nicer.

src/Container/DynamicInitializer.php
brendt: No that's actually a good alternative. I wanna consider that after merging this PR.

I tried to get rid of the interface and do `is_a` mapping on all dependencies, but that quickly lead to many recusive edge cases…

src/Bootstraps/DiscoveryLocationBootstrap.php
brendt: https://github.com/tempestphp/tempest-framework/issues/187

src/Container/DynamicInitializer.php
brendt: https://github.com/tempestphp/tempest-framework/issues/188

src/Application/Kernel.php
brendt: https://github.com/tempestphp/tempest-framework/issues/189

src/Console/ConsoleInitializer.php
brendt: Right now I prefer the other approach better, but maybe we should reconsider it closer to 1.0

PR #196, opened 2024-03-04
Request refactoring

brendt:


src/Http/GenericRouter.php
```diff
@@ -48,6 +49,8 @@ public function dispatch(Request $request): Response
 
         $callable = $this->getCallable($matchedRoute);
 
+        $request = $this->prepareRequest($request, $matchedRoute);
```
brendt, line 52: Ok so, this is the part that needed serious refactoring.

Here's what happened. I wanted to test request validation, which is handled by the ObjectMapper — it's how all validation is done in tempest.

However, our router is set up in a way that it expects an already created and valid request class. In real life apps, this would happen in `RequestInitializer`, which was called actually in two places:

- Once in the `HttpApplication`, which would always create a `GenericRequest`
- A second time when calling the controller action via the container within the `Router`

So far so good, if there were any validation errors, that container call to the controller action would throw that error (via the initializer, which used the mapper). 

However, we skipped all of that during testing, since we manually registered a singleton for our concrete request object in the `HttpRouterTester`.

So how to solve that? I could reimplement all the validation logic in the `HttpRouterTester` as well, but that's just stupid because now I'm not testing real validation anymore.

This made me realise that I was on the wrong path trying to use Initializers for request. What needs to be done is that the `Router` gets a generic request object (yes, this _could_ be a PSR-7 object), and the router should take care of transforming it into the right request object and validating it. 

This leads to a number of issues solved:

- The whole request mapping in `HttpRouterTester` is gone, and we don't need to worry about it duplicating code between the `RequestInitializer` and the test implementation.
- Validation errors are thrown in a place where the router can always catch it and send a response with validation error data — still need to do that. 
- We don't initialize the same request twice via the container, it's already registered when the router has found a matching route for the current URI
- And… yes, we _could_ support PSR-7 now without the user ever noticing. Maybe that makes me more open to the idea.

src/Http/RequestFactory.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http;
+
+use Tempest\Support\ArrayHelper;
+
+final readonly class RequestFactory
```
brendt, line 9: This is just a quick PoC. Yes… we could refactor this to use PSR

src/Mapper/RequestToRequestMapper.php
```diff
@@ -39,5 +40,10 @@ public function map(object|string $objectOrClass, mixed $data): array|object
             'query' => $origin->getQuery(),
             ...$body,
         ])->to($requestClass);
+
+        $validator = new Validator();
```
brendt, line 44: Will need to write some tests here.

src/Validation/Validator.php
```diff
@@ -11,7 +11,7 @@
 
 final readonly class Validator
 {
-    public function validate(IsValidated $object): void
+    public function validate(object $object): void
```
brendt, line 14: Will need to get rid of `IsValidated` entirely. Doesn't make sense for objects to implement an empty interface, just to be able to get validated.

src/Http/RequestFactory.php
aidan-casey: Huh! That's a thought. 🤔 🙃

brendt:
@aidan-casey I'm working on it, but there's one thing I don't agree with: the request body should be an array. For Tempest Requests, of course. I'm ok changing the name to `data` if that makes it more clear, but I want the API of requests to look like this:

```php
Request::getBody() // array with post body, if present
Request::getQuery() // array with query params, if present
Request::getFile // stream or string with file data (tbd), if present
```

I don't care that the underlying component treats everything as a string, which makes sense from the HTTP layer's point of view, but Tempest should provide a more convenient API.

aidan-casey:
> I'm working on it, but there's one thing I don't agree with: the request body should be an array.

No, it really shouldn't. 😜 

Our primary method of dealing with the request body has got to be a string, otherwise we start creating all kinds of headaches around serialization/unserialization. It's a terrible idea to assume the request is JSON and $_POST doesn't work the moment it's not a multipart form request. Never mind the fact that when we _send_ a request it has to be streamable and an array is not. 

Now, should we provide helpers for accessing the $_GET and $_POST data? Sure thing. I just don't think it's anywhere near useful to have that be the primary method in which we interact with the body.

Also, we really probably need to _not_ merge those into a single array. I could have a query parameter of `query` and a post parameter of `query` and that's totally valid. Not sure it's a good practice, but it is valid.

src/Mapper/ObjectMapper.php
```diff
@@ -21,7 +21,8 @@ final class ObjectMapper
     public function __construct()
     {
         $this->mappers = [
-            new RequestToRequestMapper(),
+            new PsrRequestToRequestMapper(),
```
brendt, line 24: We might consider doing discovery on mappers in the future. Just a thought

src/Mapper/Mapper.php
```diff
@@ -6,13 +6,13 @@
 
 interface Mapper
 {
-    public function canMap(object|string $objectOrClass, mixed $data): bool;
+    public function canMap(mixed $from, object|string $to): bool;
```
brendt, line 9: I did a bunch of cleanup for mappers because it was a mess. Technically not related to this PR, but it needed to be done.

src/Http/IsRequest.php
```diff
@@ -6,29 +6,17 @@
 
 trait IsRequest
 {
-    public Method $method;
-    public string $uri;
-    public array $body;
-    public array $headers;
     public string $path;
-    public ?string $query = null;
+    public array $query;
```
brendt, line 10: Refactored the query string to an array within the Tempest request

src/functions.php
```diff
@@ -47,9 +47,9 @@ function view(string $path): View
         return new GenericView($path);
     }
 
-    function request(string $uri, array $body = [], array $headers = []): Request
+    function request(string $uri, array $body = [], array $headers = []): PsrRequest
```
brendt, line 50: We don't really need this method anymore since we have proper testers. But we'll need to do some refactorings, so I just made it work with PSR for now.

tests/Integration/Http/ValidationResponseTest.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Tests\Tempest\Integration\Http;
+
+use App\Controllers\ValidationController;
+use Tempest\Testing\IntegrationTest;
+use function Tempest\uri;
+
+/**
+ * @internal
```
brendt, line 12: @aidan-casey what are these annotations?

brendt:
This PR turned into something different than planned. I can't really finish the validation part, because we'll need session support for it. 

That's ok, since we now have PSR request support. Responses will be another PR: https://github.com/tempestphp/tempest-framework/issues/200

Support is still basic. For example, we'll need to add file support, as well as cookie support. 

- https://github.com/tempestphp/tempest-framework/issues/201
- https://github.com/tempestphp/tempest-framework/issues/113

But that'll be for another PR. Let's get this one merged, so that we can continue work without worrying about keeping up to date with main

src/Mapper/ObjectMapper.php
aidan-casey: A good thought. 👍

tests/Integration/Http/ValidationResponseTest.php
aidan-casey: I added those in #198. They can be reverted just as easily if you don't like them.

`@internal` is technically not a huge deal anymore but basically says that it's not a class with a public API that should be used (helps IDE).

The size (`@small`, etc.) indicates how long the test should run to PHPunit and will fail if it goes over that threshold. I added after some seg fault errors that just hung up tests for a while.

src/Http/RequestFactory.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http;
+
+use Laminas\Diactoros\ServerRequestFactory;
+use Psr\Http\Message\ServerRequestInterface as PsrRequest;
+
+final readonly class RequestFactory
```
aidan-casey, line 10: Should we just make this PSR as well? We could swap this from Laminas pretty easily in the future.

src/Testing/Http/HttpRouterTester.php
```diff
@@ -47,40 +47,8 @@ public function sendRequest(Request $request): TestResponseHelper
...
-        // Finally, we register this newly created request object in the container
-        // This makes it so that RequestInitializer is bypassed entirely when the controller action needs the request class
-        // Making it so that we don't need to set any $_SERVER variables and stuff like that
-        $this->container->singleton(Request::class, fn () => $request);
-        $this->container->singleton($request::class, fn () => $request);
-
-        // Ok, now finally for real, we dispatch our request and return the response
-        return new TestResponseHelper($router->dispatch($request));
+        return new TestResponseHelper(
+            $router->dispatch(map($request)->to(PsrRequest::class))
```
aidan-casey, line 51: Why not just pass a PSR request to this method to begin with?

Also, don't we need to setup globals?

src/Mapper/PsrRequestToRequestMapper.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /** @var PsrRequest $from */
+        /** @var class-string<\Tempest\Http\Request> $requestClass */
+        $requestClass = is_object($to) ? $to::class : $to;
+
+        if ($requestClass === Request::class) {
+            $requestClass = GenericRequest::class;
+        }
+
+        $data = (array) $from->getParsedBody();
```
aidan-casey, line 31: What happens with, say, an XML request here?

brendt: Currently unsupported :)

src/Http/RequestFactory.php
brendt: Do you mean remove this class and use laminas directly? Or something else?

src/Testing/Http/HttpRouterTester.php
brendt: Well my idea was that it's pretty convenient being able to call `$this->http->sendRequest(new MyRequest)` from within tests. You don't want to deal with PSRs there either. 

That's why I plan on removing the `request` function, in favor of this approach

src/Http/RequestFactory.php
brendt: Do you mean `RequestFactory implements ServerRequestFactoryInterface` (or whatever it's called?)

aidan-casey: Yep.

PR #206, opened 2024-03-07
Sessions

### LINKED ISSUE #113: Add Cookie Support

### LINKED ISSUE #123: Add Session Support

brendt:
- Closes #123 
- Closes #113


src/Http/IsRequest.php
```diff
@@ -68,9 +68,9 @@ public function getQuery(): array
         return $this->query;
     }
 
-    public function session(): SessionManager
+    public function session(): Session
     {
-        return get(SessionManager::class);
+        return get(Session::class);
```
aidan-casey, line 73: Why don't we inject this?

src/Http/IsResponse.php
```diff
@@ -33,7 +34,7 @@ public function getHeaders(): array
         return $this->headers;
     }
 
-    public function session(): SessionManager
+    public function session(): Session
     {
         return get(SessionManager::class);
```
aidan-casey, line 39: Does it make sense for a response to have access to the session?

src/Clock/MockClock.php
```diff
@@ -35,4 +35,14 @@ public function sleep(int $seconds): void
             new DateInterval("PT{$seconds}S")
         );
     }
+
+    public function changeTime(int $seconds): void
```
aidan-casey, line 39: I wonder if we want to refactor this to allow for broader time travel, similar to Laravel?

src/Container/InMemoryContainerLog.php
```diff
@@ -13,7 +14,7 @@ final class InMemoryContainerLog implements ContainerLog
 
     public function __construct(
         /** @var \Tempest\Container\Context[] $stack */
-        private array $stack = [],
+        public array $stack = [],
```
aidan-casey, line 17: Do we need this to be public?

src/Http/Cookie/CookieManager.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Cookie;
+
+use DateTimeImmutable;
+use Tempest\Clock\Clock;
+
+final class CookieManager
```
aidan-casey, line 10: I wish we had a better name for this, but I'm also not sure what that would be. 😛

src/Http/Cookie/SetCookieMiddleware.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+    }
+
+    public function __invoke(Request $request, callable $next): Response
+    {
+        /** @var Response $response */
+        $response = $next($request);
+
+        foreach ($response->cookies()->all() as $cookie) {
+            $response->header('set-cookie', (string) $cookie);
```
aidan-casey, line 23: I wonder if there are any edge cases we need to be aware about using this over `setcookie`?

src/Http/RouterInitializer.php
```diff
@@ -7,12 +7,17 @@
...
 
 #[Singleton]
 final readonly class RouterInitializer implements Initializer
 {
     public function initialize(Container $container): Router
     {
-        return $container->get(GenericRouter::class);
+        $router = $container->get(GenericRouter::class);
+
+        $router->addMiddleware(SetCookieMiddleware::class);
```
aidan-casey, line 19: We should probably move this to a bootstrapper down the road.

src/Http/Session/Resolvers/CookieSessionIdResolver.php
```diff
@@ -0,0 +1,39 @@
...
+        private Clock $clock,
+    ) {
+    }
+
+    public function resolve(): SessionId
+    {
+        $id = $this->cookies->get(Session::ID)?->value ?? null;
+
+        if (! $id) {
+            $id = uniqid();
```
aidan-casey, line 28: I'm thinking we should probably refactor this. There may be some security concerns with this method?

src/Http/Session/Resolvers/HeaderSessionIdResolver.php
```diff
@@ -0,0 +1,25 @@
...
+    public function __construct(
+        private Request $request,
+    ) {
+    }
+
+    public function resolve(): SessionId
+    {
+        $id = $this->request->getHeaders()[Session::ID] ?? null;
+
+        return new SessionId($id ?? uniqid());
```
aidan-casey, line 23: I'm thinking we might want a different method for generating the ID.

src/Http/Session/Session.php
```diff
@@ -4,121 +4,48 @@
...
     {
-        return $_SESSION;
+        return $this->getSessionManager()->isValid($this->id);
     }
 
-    public function save(): void
+    private function getSessionManager(): SessionManager
     {
-        session_write_close();
+        return get(SessionManager::class);
```
aidan-casey, line 49: Why not inject this?

src/Http/Session/SessionConfig.php
```diff
@@ -0,0 +1,29 @@
...
+         * @template SessionManager of \Tempest\Http\Session\SessionManager
+         * @var class-string<SessionManager>
+         */
+        public string $managerClass = FileSessionManager::class,
+
+        /**
+         * @template SessionIdResolver of \Tempest\Http\Session\SessionIdResolver
+         * @var class-string<SessionIdResolver>
+         */
+        public string $idResolverClass = CookieSessionIdResolver::class,
```
aidan-casey, line 26: I think we should probably split this type of config between the session manager and the session classes. Not every session manager will need a path.

src/Http/Session/SessionManager.php
aidan-casey, line 1: What is the value add of giving each manager the flexibility of defining all these methods vs. making this `SessionStorage` that is only concerned about reading and writing the stored data?

src/Testing/IntegrationTest.php
```diff
@@ -52,4 +62,11 @@ protected function migrate(string ...$migrationClasses): void
             $migrationManager->executeUp($this->container->get($migrationClass));
         }
     }
+
+    protected function clock(DateTimeInterface|string $now): MockClock
```
aidan-casey, line 66: This might even be worth having as a global helper in the long run. 🤔

src/Http/Session/SessionDestroyed.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Tempest\Http\Session;
+
+final readonly class SessionDestroyed
```
aidan-casey, line 5: Is this an event?

tests/Integration/Testing/Http/TestResponseHelperTest.php
```diff
@@ -54,11 +54,11 @@ public function test_assert_header_value_equals()
         $helper = new TestResponseHelper(
             new GenericResponse(
                 status: Status::OK,
-                headers: ['Content-Type' => 'application/json']
+                headers: ['Content-Type' => ['application/json']]
```
aidan-casey, line 57: This is looking more and more like PSR. 🙃

brendt:
Ok so there's an important decision to make, API wise.

Ideally, I want to be able to read session and cookie data from the request. And to be able to set session and cookie data within the response.

There are many ways to design this API. Right now, I've just exposed `Session` and `CookieManager` within both the request and response, which honestly isn't ideal, because you can get and set on both the request and response.

I think it'd be better to make the request API readonly, something like this:

```php
$request->getSessionValue('name');
$request->getCookie('name');
```

And the response writeonly, like this:

```php
return response()
    ->addSession('name', 'value')
    ->removeCookie('name')
    ->…
```

I think I'm happy with that API for responses, I think that makes sense. Responses will also need to rely on service location, I don't think there's a way around that, and I think that's ok.

However, for the requests, I'm not super happy with that API. That's because there now is inconsistency within the request class:

```php
$request->getHeaders['name'];
$request->getSessionValue('name');
```

I think this means we'll need to refactor headers as well to follow the same pattern:

```php
$request->addHeader('name', 'value');
$request->removeHeader('name');
```

Ideally though, we'd have property hooks, which would make the API much smoother — imo.

src/Http/IsRequest.php
brendt: I'm gonna keep it as is for now, I think. The alternative is that we make session nullable, which isn't the most user friendly, because we can assume that if you request the session, you actually need it (and it needs to be started if it doesn't exist)

src/Http/Cookie/CookieManager.php
brendt: `Bakery`

src/Http/Session/Session.php
brendt: Oh I know why, because we serialize the session itself.

src/Http/Session/SessionManager.php
brendt: Well it's actually not that simple. Cleanup, for example, is also dependent on the specific storage system, it's more than read/write.

I would suggest not to optimize this prematurely. I think it makes sense to add Redis and Database support in the long run, we can optimise then if needed

app/Modules/Form/form.view.php
```diff
@@ -13,12 +13,26 @@
 <?php } ?>
 
 <form action="<?= uri([FormController::class, 'store']) ?>" method="post">
-    <label for="name">Name:</label>
-    <input type="text" name="name" id="name">
+    <div>
+        <label for="name">Name:</label>
+        <input type="text" name="name" id="name" value="<?= $this->original('name') ?>">
```
brendt, line 18: Two considerations to make on the long run:

- Should we add functions that allow us to skip `$this` everywhere?
- PHP as a templating engine isn't ideal, we should probably consider supporting another engine that has IDE support so that people can choose

src/Http/Session/Resolvers/HeaderSessionIdResolver.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Tempest\Http\Session\Resolvers;
+
+use Ramsey\Uuid\Uuid;
+use Tempest\Http\Request;
+use Tempest\Http\Session\Session;
+use Tempest\Http\Session\SessionId;
+use Tempest\Http\Session\SessionIdResolver;
+
+final readonly class HeaderSessionIdResolver implements SessionIdResolver
```
brendt, line 13: I need to write a test for this one

src/View/IsView.php
```diff
@@ -158,14 +162,46 @@ private function determineSlotName(string $content): string
      */
     public function getErrorsFor(string $name): array
     {
-        return $this->getSession()->get('validation_errors')[$name] ?? [];
+        $errors = $this->resolveValidationErrors();
+
+        return $errors[$name] ?? [];
+    }
+
+    public function hasErrorsFor(string $name): bool
```
brendt, line 170: Should we consider adding standalone functions as shorthands?

tests/Integration/Http/ValidationResponseTest.php
```diff
@@ -16,8 +16,16 @@ final class ValidationResponseTest extends IntegrationTest
...
+            ->assertHasValidationError('number');
 
-        $this->http->post(uri(ValidationController::class), ['number' => 11, 'item.number' => 11]);
+        $this->http
+            ->get(uri([ValidationController::class, 'get']))
+            ->assertOk()
+            ->assertHasNoValidationsErrors();
     }
+
+    // TODO: test with original values
```
brendt, line 30: TODO

src/Http/Responses/InvalidResponse.php
```diff
@@ -15,11 +15,14 @@ final class InvalidResponse implements Response
...
     public function __construct(
-        private PsrRequest $request,
-        private ValidationException $exception,
+        private readonly PsrRequest $request,
+        ValidationException $exception,
     ) {
         $this->status = Status::BAD_REQUEST;
         $this->redirect((string) $this->request->getUri());
-        $this->getSession()->set('validation_errors', $this->exception->failingRules);
+        // TODO: add these names as constants somewhere
```
brendt, line 23: TODO

src/Http/Responses/InvalidResponse.php
```diff
@@ -15,11 +15,14 @@ final class InvalidResponse implements Response
...
+        // TODO: mapping empty field to number results in type/validation error
```
brendt, line 24: TODO

brendt: #209 209

brendt:
I'm gonna merge, we can finetune further where needed on main

PR #211, opened 2024-03-10
Cache Router matching regex

N-Silbernagel:
As previously discussed in #175, this PR adds caching of the routing RegEx and caches RouteConfig to PHP files.

The first change here ist that I've now split the routes in RouteConfig into static- and dynamic-routes. This allows for the deletion of some code.
Also, I'm now computing the routing RegEx (actually one RegEx per HTTP-Method) in the RouteConfig when adding routes.

For the caching I'm now using files, which enables caching via OPcache. Unfortunately, this is not a easy as a "var_export", especially exporting the ReflectionMethod of the Route class is tricky. This is why I have included symfony/var-exporter.
This should also easily enable using this kind of caching for the remaining cache files.

The results of this are pretty good.

Where this is the performance on current main:
image

This is the performance for this PR:
image

A caveat here is the higher memory usage, which might be a blocker.

Let me know what you think.


src/Http/RouteConfig.php
```diff
@@ -8,18 +8,55 @@
...
+    private string $regexMark = 'a';
+
     public function __construct(
         /** @var array<string, array<string, \Tempest\Http\Route>> */
-        public array $routes = [],
+        public array  $staticRoutes = [],
+        /** @var array<string, array<string, \Tempest\Http\Route>> */
+        public array  $dynamicRoutes = [],
+        /** @var array<string, string> */
+        public array $matchingRegexes = []
```
brendt, line 20: Could we make this one private? Would that still work with the var exporter?

Regardless of whether that's possible, I'd move it outside of the constructor, as this is a variable we won't ever set from the outside — as far as I understand. I want to avoid confusion where possible.

src/Http/RouteConfig.php
```diff
@@ -8,18 +8,55 @@
 
 final class RouteConfig
 {
+    /** @var string The mark to give the next route in the matching Regex */
+    private string $regexMark = 'a';
+
     public function __construct(
         /** @var array<string, array<string, \Tempest\Http\Route>> */
-        public array $routes = [],
+        public array  $staticRoutes = [],
```
brendt, line 16: I wonder if we could rewrite it like so?

```php
private array $staticRoutes = [];
private array $dynamicRoutes = [];

__construct(array $routes = [])
{
    foreach ($routes as $route) {
        $this->addRoute($route);
    }
}
```

So that from the outside, it's all still one array?

src/Http/RouteConfig.php
```diff
@@ -8,18 +8,55 @@
...
     ) {
     }
 
     public function addRoute(ReflectionMethod $handler, Route $route): self
     {
         $route->setHandler($handler);
 
-        $this->routes[$route->method->value][$route->uri] = $route;
+        if ($route->isDynamic) {
+            $routeMark = $this->regexMark++;
```
brendt, line 29: Out of curiosity, what happens when you reach `z`?

brendt:
Once again a great PR. Left a couple of comments!

A couple general questions:

Why do you think the memory usage is doubled? I assume it's not because of the var exporting stuff, because that only happens once. Apart from that, we already built the whole regex string in memory before, so I would assume it wouldn't make that big a difference — but apparently it is. It's not a blocker necessarily, but I would like to better understand.

Also, hoe many routes did you do these benchmark with?

src/Http/RouteConfig.php
N-Silbernagel: Making it private is not possible because is needed in GenericRouter. Should I add a getter and make the property private?

I have moved it outside of the constructor.

src/Http/RouteConfig.php
N-Silbernagel: How would you imagine accessing the properties in the GenericRouter would look like? I am currently accessing these directly there.

src/Http/RouteConfig.php
N-Silbernagel: It will generate "aa", but researching this, I found out it is soft deprecated in PHP8.3.0! https://www.php.net/manual/en/language.operators.increment.php#language.operators.increment.string

So I replaced it with str_increment.

N-Silbernagel:
Thank you!

I have looked for some time but cannot exactly pinpoint the reason. I had initially tried writing the VarExporter myself but that had even higher Memory usage.

XDebug outputs this for the serialize:
image


and this for the PHP version:
image

So it contradicts PHPBench's output which uses memory_get_peak_usage. I'll need to invest some time into finding out more about this.

I am using 743 routes.

src/Http/RouteConfig.php
brendt: You could wrap them in a generator:

```php
function all(): Generator
{
    foreach ($this->staticRoutes as $route) {
        yield $route;
    }
    
    foreach ($this->dynamicRoutes as $route) {
        yield $route;
    }
}
```

N-Silbernagel: I had also thought about that. This would mean that the "all" method does not return them grouped by http-method though.

N-Silbernagel:
I've looked into the mem_peak and still cant quite pinpoint what is casing this as xdebug does not seem to report it.
What I can say is that it is probably not as important. I've I "warum up" the process, so require the file before running the benchmark, memory is as low as 1,5mb. The current serialize version does not benefit from that. It's mem_peak stays at ~2mb.

brendt:
Let's not worry about the memory then.

Are you ok to merge this? Or did you have anything else on your mind regarding this PR?

src/Http/RouteConfig.php
brendt: It's fine for now. Let's just hope the property hooks RFC passes so that we can get a bit more control over public properties :)

N-Silbernagel:
No, I like the scope of the PR. More than okay with a merge!

brendt:
Let's merge! Thanks again for the efforts :)

N-Silbernagel:
Thank you too 😊

PR #215, opened 2024-03-12
Added additional architecture tests

Plytas:
As per https://github.com/tempestphp/tempest-framework/issues/116#issuecomment-1970395938

* Added a test to make sure there are no abstract classes.
* From bullet point 2 and 3 we can extract a test to make sure that all classes are final.
* Added a test to make sure that classes that implement anything are readonly.

I've added these tests exactly as they were described in Brent's comment. There are 88 errors currently that would need to be addressed to satisfy those rules. I do believe that there will be exceptions to these rules so I'm not sure how feasible it will be to implement them.

### DIFF
--- a/tests/Architecture/ArchitectureTest.php
+++ b/tests/Architecture/ArchitectureTest.php
@@ -73,9 +73,9 @@ class ArchitectureTest
     {
         return PHPat::rule()
             ->classes(Selector::AND(
-				Selector::inNamespace('Tempest'),
-				Selector::NOT(Selector::isInterface()),
-			))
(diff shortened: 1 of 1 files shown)


brendt:
I think we'll need to finetune this a bit more, since it's now saying interfaces should be final, and enums should be readonly, which isn't needed.

Plytas:
Adjusted selectors to ignore interfaces and enums. Down to 45 errors. Let me know if you see more exceptions.

brendt:
Gonna merge this and fix the errors 👍

PR #217, opened 2024-03-17
feat: Add Date Handling & Validation filters

przemyslaw-przylucki:
Hi there!

I'm a big fan of the project and want to step up my open source contributions a bit - this is one of my first PRs to OS, so would appreciate any feedback.

The goal of this PR is to add a basic support for handling and validating dates:
- extended the mapper to handle DateTime and DateTimeImmutable casts, not sure if this is the right solution though
- introduced "Before" and "After" rules for working with dates

Things to consider:
- do you plan to introduce a first-class date-time wrapper?
- how do we handle pre-validation (see comment on DateTimeImmutableCaster

### DIFF
--- a/src/Mapper/ArrayToObjectMapper.php
+++ b/src/Mapper/ArrayToObjectMapper.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Tempest\Mapper;
 
+use DateTime;
+use DateTimeImmutable;
+use DateTimeInterface;
 use ReflectionClass;
 use ReflectionException;
 use ReflectionProperty;
@@ -12,11 +15,15 @@ use function Tempest\get;
 use Tempest\ORM\Attributes\CastWith;
 use Tempest\ORM\Caster;
 use Tempest\ORM\Casters\BooleanCaster;
+use Tempest\ORM\Casters\DateTimeCaster;
+use Tempest\ORM\Casters\DateTimeImmutableCaster;
 use Tempest\ORM\Casters\FloatCaster;
 use Tempest\ORM\Casters\IntegerCaster;
 use Tempest\ORM\Exceptions\MissingValuesException;
 use Tempest\Support\ArrayHelper;
+use Tempest\Validation\Rules\DateTimeFormat;
 use Tempest\Validation\Validator;
+use Throwable;
 
 final readonly class ArrayToObjectMapper implements Mapper
 {
@@ -94,10 +101,11 @@ final readonly class ArrayToObjectMapper implements Mapper
         }
 
         if (! $castWith) {
-            return match($property->getType()->getName()) {
+            return match($type = $property->getType()->getName()) {
                 'int' => new IntegerCaster(),
                 'float' => new FloatCaster(),
                 'bool' => new BooleanCaster(),
+                DateTimeImmutable::class, DateTime::class, DateTimeInterface::class => $this->createDateCaster($type, $property),
                 default => null,
             };
         }
@@ -235,5 +243,29 @@ final readonly class ArrayToObjectMapper implements Mapper
         $validator = new Validator();
 
         $validator->validate($object);
+    }
+
+    /**
+     * @param class-string<DateTime|DateTimeImmutable|DateTimeInterface> $type
+     * @param ReflectionProperty $property
+     *
+     * @return DateTimeImmutableCaster
+     */
+    private function createDateCaster(string $type, ReflectionProperty $property): Caster
+    {
+        $format = null;
+
+        try {
+            $format = attribute(DateTimeFormat::class)
+                ->in($property)
+                ->first();
+        } catch (Throwable) {
+
+        }
+
+        return match ($type) {
+            DateTime::class => new DateTimeCaster($format?->format),
+            default => new DateTimeImmutableCaster($format->format),
+        };
     }
 }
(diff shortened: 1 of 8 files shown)


src/Mapper/ArrayToObjectMapper.php
```diff
@@ -94,10 +101,11 @@ private function getCaster(ReflectionProperty $property): ?Caster
         }
 
         if (! $castWith) {
-            return match($property->getType()->getName()) {
+            return match($type = $property->getType()->getName()) {
                 'int' => new IntegerCaster(),
                 'float' => new FloatCaster(),
                 'bool' => new BooleanCaster(),
+                DateTimeImmutable::class, DateTime::class, DateTimeInterface::class => $this->createDateCaster($type, $property),
```
brendt, line 108: I don't think a hardcoded mapping will get us very far in the long run. But I'm also unsure about how to handle it yet.

I think I want all mappers (more than Date mappers) to be discovered automatically, and determine their mapping types on their return type.

I'm writing it here for some background info, it's totally fine to have it like this in the PR, we can refactor later.

src/ORM/Casters/DateTimeImmutableCaster.php
```diff
@@ -0,0 +1,45 @@
...
+         * Applies to this and DateTimeCaster.
+         *
+         * Handles the case when the date is not valid.
+         * This should be preventable from user land by adding a date time format validation rule.
+         *
+         * This is not ideal, as preferably, this would be handled by the format validation class
+         * I'm not sure how you want to handle this, so in order to not blow the scope up, I'll leave it here for now.
+         *
+         * One of the ideas would be a deferred cast to the date AFTER the validation has passed,
+         * this way we could avoid throwing exceptions and handle the error in a more controlled manner.
```
brendt, line 35: I think an exception is ok for now, validation is a separate concern and I think we can assume the user has taken care of it before we end up here.

src/Validation/Rules/After.php
```diff
@@ -0,0 +1,63 @@
...
+            ? DateTimeImmutable::createFromInterface($date)
+            : new DateTimeImmutable($date);
+    }
+
+    public function isValid(mixed $value): bool
+    {
+        /**
+         * @todo
+         *
+         * We should implement a better date handling, either a support date class that can be created
```
brendt, line 32: Good point, outside the scope for this PR though 👍

src/Validation/Rules/After.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Tempest\Validation\Rules;
+
+use Attribute;
+use DateTimeImmutable;
+use DateTimeInterface;
+use Tempest\Validation\Rule;
+use Throwable;
+
+#[Attribute]
+final readonly class After implements Rule
```
brendt, line 14: I think `AfterDate` might be a clearer name, but also sounds a bit… clunky. Unsure about it

src/Validation/Rules/Before.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Tempest\Validation\Rules;
+
+use Attribute;
+use DateTimeImmutable;
+use DateTimeInterface;
+use Tempest\Validation\Rule;
+
+#[Attribute]
+final readonly class Before implements Rule
```
brendt, line 13: Same for `BeforeDate`

src/Validation/Rules/DateTimeFormat.php
```diff
@@ -6,17 +6,23 @@
...
 final readonly class DateTimeFormat implements Rule
 {
-    public function __construct(private string $format = 'Y-m-d')
+    public function __construct(public string $format = 'Y-m-d')
     {
     }
 
     public function isValid(mixed $value): bool
     {
+        $value = match ($value instanceof DateTimeInterface) {
```
brendt, line 21: In general I prefer to write it like this:

```php
match ($value::class)
{
    DateTimeInterface::class => …
    default =>
```

That's because PhpStorm then knows which class it's dealing with within the match arms.

brendt (changes requested):

src/Mapper/ArrayToObjectMapper.php
przemyslaw-przylucki: Completely agree - wanted to keep the scope small for now

src/Validation/Rules/DateTimeFormat.php
przemyslaw-przylucki: We can't do $value::class here because we are unsure if it's an object, we'd need an additional check

src/Validation/Rules/After.php
przemyslaw-przylucki: Yeah I wasn't a fan of either After or AfterDate, but I don't really have a better idea. Future or FutureDate was also something that I've considered but it feels more confusing for some reason

src/Validation/Rules/DateTimeFormat.php
brendt: Good point!

src/Validation/Rules/After.php
brendt: Let's rename it to `AfterDate` and `BeforeDate` for now, and I'll try to come up with a better name once this is merged :)

brendt:
Thanks!

PR #218, opened 2024-03-18
feat: Migration Rollbacks

przemyslaw-przylucki:
The goal of this PR is to primarily introduce a migration rollback command.

Additional changes:
- added `delete` to model
- added `firstWhere` to model
- added a check to prevent running migrations in prod. Maybe it makes sense to add a "dangerous" property to ConsoleCommand and handle it before the execution of command, so user-land commands could also benefit from it and we have higher reusability but that's a bigger API change that I'm not sure is wanted
- added count to migration & migration rollback that indicates affected migrations count

Things to consider:
- Operator enum may not be the smartest choice as different databases will have different operators
- It could be helpful to create a wrapper on PDO exceptions to handle certain behaviour better (i.e migration rollbacks)


src/Database/Migrations/MigrationManager.php
```diff
@@ -63,4 +93,24 @@ public function executeUp(MigrationInterface $migration): void
...
+            return;
+        }
+
+        $this->database->execute($query);
+
+        try {
+            Migration::firstWhere('name', Operator::Equals, $migration->getName())
+                ->delete();
+        } catch (PDOException) {
+            // @todo should be handled better as PDO exception doesn't necessarily mean that the migrations table doesn't exist
```
brendt, line 111: Also dispatch an event so that the user knows a migration didn't run. Also: shouldn't we stop migrations as soon as an error occurs?

brendt (changes requested):
I say let's get rid of the whole querybuilder part for now, as it's unclear what we want to do with it in the future.

Apart form that and a couple of small details: great PR 👍

src/Database/Migrations/MigrationManager.php
przemyslaw-przylucki: I've added a comment in the code to better explain this case, lemme know if it makes sense!

przemyslaw-przylucki:
Thanks for the feedback!

What do you think about extending the ConsoleCommand attribute to include "dangerous" flag or something like that, that'd do the environment check?

I think that most of the framework-provided commands will want to include that, and I find myself using that in my own projects all the time, so could be fun if it'd happen automatically, seems very convenient =)

brendt:
>  What do you think about extending the ConsoleCommand attribute to include "dangerous" flag or something like that, that'd do the environment check?
>
> I think that most of the framework-provided commands will want to include that, and I find myself using that in my own projects all the time, so could be fun if it'd happen automatically, seems very convenient =) That should probably happen outside of this PR

I was thinking exactly the same going through this PR 😁 Also a global `-v` option to show the full stack trace of exceptions would be good. I'll make a separate issue for it: #225

Let's keep it in another PR though :) Are you ok with merging this one?

brendt:
Thanks!

PR #222, opened 2024-03-23
Starts to add collections.

aidan-casey:
This pull requests begins to add collections to Tempest along with a couple of helper classes. I was debating whether to separate out collections into `ArrayList` (an ordered list) and `Dictionary` (a keyed dictionary). There is value to that, in my opinion, since it can enforce a collection to be ordered or keyed respectively. Speaking with @brendt, we will hold off on that for now.

I have included `Queue` (first in, first out) and `Stack` (last in, first out) and may add a few other helpers as well.

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Collection;
+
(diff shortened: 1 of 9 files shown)


brendt:
Closing due to inactivity and refactoring to tempest/core

PR #223, opened 2024-03-23
Make container comply with PSR-11

przemyslaw-przylucki:
Goal of this PR is to make container implementation comply with [PSR-11](https://www.php-fig.org/psr/psr-11/) specification.

### DIFF
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -4,9 +4,10 @@ declare(strict_types=1);
 
 namespace Tempest\Container;
 
+use Psr\Container\ContainerInterface;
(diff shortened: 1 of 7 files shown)


aidan-casey:
Hey, @przemyslaw-przylucki! There was a PR for this previously [here](https://github.com/tempestphp/tempest-framework/pull/156). Can you please review and expound on how this might be helpful to you?

brendt:
So for now we don't have the goal of making tempest/container as a separate package. Until we do, there's no need to make it PSR-11 compliant.

Issue #225, opened 2024-03-26
Global console options


brendt:
It'd be nice to have some global console options available for all commands:

- [ ] A way to mark a command as "dangerous", which will add an automatic confirmation check and a `--force` flag in production environments when enabled
- [ ] A `-v` flag that will print out verbose exception output

przemyslaw-przylucki:
I think I have some cool ideas about this - will send a PR later this week

przemyslaw-przylucki:
Hey @brendt, I created a quick proof of concept PR - lemme know what you think!

PR #228, opened 2024-03-30
RFC: Console Application improvements (arguments bag, command validation, aliases, argument injectors, mistyped commands suggestions, themes)

przemyslaw-przylucki:
This PR aims to enhance the capabilities of the CLI layer 🎉 

image

Related to #225, it has a lot of rough edges - I've rushed a proof of concept in a few hours. So it's more like an RFC :)

Introduced changes:
- Global ArgumentBag that holds arguments passed to commands
- ConsoleCommands now get validated - annotations may be added to validate the argument (we may think about adding type-rules to, either to use manually or for validator to infer from reflection)
- Adds predefined flags (eventually arguments) - that are easily used in user-land code (i.e by passing isDangerous to ConsoleCommand attribute)
- These flags kinda act like a middleware for commands - maybe it's worth to add a "pipeline" support class?
- The "listing command" will also discover injected arguments
- Added "--help (-h)", "--force (-f)" "--silent (-s)" "--no-interaction (-n)" flags to all commands
- Added ability to alias commands

image

Left to do:
- tests 🥶


src/Console/ConsoleCommand.php
```diff
@@ -55,9 +66,68 @@ public function __unserialize(array $data): void
...
+        return new CommandArguments(
+            arguments: $arguments,
+            injectedArguments: $injected,
+        );
+    }
+
+    /**
+     * @return InjectedArgument[]
+     */
+    protected function flagList(): array
```
przemyslaw-przylucki, line 121: That should probably be discovered by the framework instead

przemyslaw-przylucki:
Also, would you be open to making the ConsoleCommand attribute target the entire class and not a handler?

The negative is that we would have to infer the default argument order based on the properties present in the class, but I think it'd be easier to add additional information to command parameters.

I've started exploring what's out there after building this and I mostly love the api this package provides - https://github.com/thettler/laravel-console-toolkit

aidan-casey:
Hey, @przemyslaw-przylucki! I haven't had the time to review everything here, but I think there is some stuff here that we should definitely think about. I hope to have some more helpful comments here soon. 🙂 

I'm going to defer to @brendt on the question about class-based commands. My guess is that will be a no, however.

przemyslaw-przylucki:
I didn't intend this to grow in scope so much but I went coo-coo and couldn't stop polishing this - still a lot of things that can be improved but I like how it stands from features perspective


Also, pretty rad UI if you ask me:

image

przemyslaw-przylucki:
I've pushed some additional code, they contains:
- console output builder
- suggested/similar commands, if user mistypes command and levenstein distance is <= 3
- ConsoleTheme - I think it'd be nice if we had a layer of abstraction on top of ConsoleOutput that can format strings, this would be nice for defining a style guide for some package authors or internal commands

image

brendt:
Ok this is an impressive PR 💪

It'll take some time to read through it! Please hold :)

src/Support/Reflection/Attributes.php
```diff
@@ -7,12 +7,13 @@
...
 use ReflectionClass;
 use ReflectionMethod;
+use ReflectionParameter;
 use ReflectionProperty;
 
 /** @template T */
 final class Attributes
 {
-    private ReflectionClass|ReflectionMethod|ReflectionProperty $reflector;
+    private ReflectionClass|ReflectionMethod|ReflectionProperty|ReflectionParameter $reflector;
```
brendt, line 16: Outside the scope of this PR, but we'll need to clean this part up :p

src/Validation/NestedValidator.php
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Tempest\Validation;
+
+use ReflectionClass;
+use ReflectionProperty;
+use function Tempest\attribute;
+use Tempest\Console\Argument;
+use Tempest\Validation\Exceptions\ValidationException;
+
+// todo: just a proof of concept.
```
brendt, line 13: Can you explain the concept and why it's needed?

src/Console/Commands/MigrateCommand.php
```diff
@@ -19,23 +18,21 @@ final class MigrateCommand
     public function __construct(
         private readonly Console $console,
         private readonly MigrationManager $migrationManager,
-        private readonly AppConfig $config,
     ) {
     }
 
     #[ConsoleCommand(
         name: 'migrate:up',
+        aliases: ['haha'],
```
brendt, line 26: This will need to change :)

src/Console/Commands/MigrateCommand.php
```diff
@@ -19,23 +18,21 @@ final class MigrateCommand
...
         description: 'Run all new migrations',
+        isDangerous: true,
+        help: [
```
brendt, line 29: Why does help text need to be an array? Wouldn't a string suffice?

src/Console/ConsoleCommand.php
```diff
@@ -6,16 +6,28 @@
...
 use Attribute;
 use ReflectionMethod;
+use Tempest\Console\Arguments\ForceArgument;
+use Tempest\Console\Arguments\HelpArgument;
+use Tempest\Console\Arguments\NoInteractionArgument;
+use Tempest\Console\Arguments\QuietArgument;
 
 #[Attribute]
-final class ConsoleCommand
+final class ConsoleCommand implements HasHelpLines
```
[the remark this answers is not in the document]
brendt: To add: why not a string instead of an array?

src/Console/HandlesConsoleInput.php
```diff
@@ -27,22 +29,16 @@ public function ask(
...
-                    ),
-                ),
-            )
-                . '] ';
-        }
-
-        $this->output->write($questionString);
+        $message = ConsoleOutputBuilder::new(" ")
+            ->error(" $question ")
+            ->when(! ! $options, function (ConsoleOutputBuilder $builder) use ($options, $default) {
```
brendt, line 34: I guess the double `!!` is to convert it to booleans? Why not `$options !== null` ?

src/Console/HandlesConsoleInput.php
```diff
@@ -27,22 +29,16 @@ public function ask(
...
-                        fn (string $option) => $option === $default ? strtoupper($option) : $option,
-                        $options,
-                    ),
-                ),
-            )
-                . '] ';
-        }
-
-        $this->output->write($questionString);
+        $message = ConsoleOutputBuilder::new(" ")
```
brendt, line 32: I like this concept!

src/Console/HandlesConsoleInput.php
```diff
@@ -27,22 +29,16 @@ public function ask(
...
+            ->error(" $question ")
+            ->when(! ! $options, function (ConsoleOutputBuilder $builder) use ($options, $default) {
+                $builder->formatted("[" . ConsoleStyle::FG_BLUE(implode(', ', $options)) . "]")
+                    ->formatted($default ? ConsoleStyle::FG_GRAY(" (default: $default) ") : " ");
+            });
+
+        $this->output->write(
```
brendt, line 39: Maybe add a convenience method on the builder class? `write(to: $output)`

src/Console/InjectedArgument.php
```diff
@@ -0,0 +1,32 @@
...
+     * @throws ExitException
+     */
+    public function handle(ConsoleCommand $command): void
+    {
+
+    }
+
+    public function shouldInject(): bool
+    {
+        return ! ! $this->value;
```
brendt, line 21: Double `!!`

src/Console/InjectedArgument.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console;
+
+use ReflectionParameter;
+
+readonly class InjectedArgument extends Argument
```
brendt, line 9: Deep inheritance chain should be refactored

src/Console/NullArgumentBag.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console;
+
+use InvalidArgumentException;
+
+final class NullArgumentBag implements ArgumentBag
```
brendt, line 9: Is this for testing? We should probably move it to `Tempest\Testing` if that's the case

src/Console/Styling/ConsoleOutputThemeInitializer.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Console\Styling;
+
+use Tempest\Container\Container;
+use Tempest\Container\Initializer;
+use Tempest\Container\Singleton;
+
+#[Singleton]
+final readonly class ConsoleOutputThemeInitializer implements Initializer
```
brendt, line 12: Should a theme really be registered in the container? I figure it's just a config value somewhere that's used when constructing the console?

src/Console/Styling/OutputType.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console\Styling;
+
+enum OutputType
```
brendt, line 7: `ConsoleOutputType` is probably clearer

src/Console/Styling/OutputLine.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console\Styling;
+
+final class OutputLine
```
brendt, line 7: `ConsoleOutputLine` is probably clearer

src/Console/Styling/RenderCommandNotFound.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public function __invoke(string $command): string
+    {
+        $similarCommands = $this->getSimilarCommands($command);
+
+        return ConsoleOutputBuilder::new()
+            ->error("Command `{$command}` not found")
+            ->blank()
+            ->when(! ! $similarCommands, function (ConsoleOutputBuilder $response) use ($similarCommands) {
```
brendt, line 27: `!!`

src/Console/Styling/RenderCommandNotFound.php
```diff
@@ -0,0 +1,55 @@
...
+                $response->warning('Did you mean one of these commands?')
+                    ->blank();
+
```
brendt, line 30: Something we could add along the way is to allow the user to pick the command automatically from the error output without having to restart the console session. But doesn't need to be added now.

src/Console/Styling/RenderConsoleCommandOverview.php
```diff
@@ -0,0 +1,87 @@
...
+        private RenderConsoleCommand $renderrer,
+    ) {
+    }
+
+    public function __invoke(): string
+    {
+        $commands = $this->getCommands();
+
+        $response = ConsoleOutputBuilder::new()
+            ->brand(' Tempest Console ')
```
brendt, line 26: The surrounding spaces should probably be added within the `brand` method?

src/Console/Styling/RenderConsoleCommandOverview.php
```diff
@@ -0,0 +1,87 @@
...
+        private ConsoleConfig $consoleConfig,
+        private RenderConsoleCommand $renderrer,
+    ) {
+    }
+
+    public function __invoke(): string
+    {
+        $commands = $this->getCommands();
+
+        $response = ConsoleOutputBuilder::new()
```
brendt, line 25: Just once again: I really love this output builder!

przemyslaw-przylucki:
Thanks for all the feedback - I agree with everything from the design perspective, I prefer to go wild and then refactor =)

I'd love to get this PR finalized - I think breaking this into multiple smaller PRs def makes sense - wanted to cram this bad boy with a lot of convenience features. I think there's a lot of potential in the console layer of the project and I'd love to continue working on it if it's alright with y'all!

I'll respond to individual comments in the evening :D

przemyslaw-przylucki:
> * Are you committed to finishing this PR? It'll probably take a while before we can merge it, as it's pretty huge. I'm totally fine if you don't have the capacity to do so, but I'd just like some clarity up front, so that I know how what's expected from my side.

One of the things that this PR needs is the type inference and changes in validation that were added in #229, could you let me know if that's something you'd want to go forward with?

brendt:
> One of the things that this PR needs is the type inference and changes in validation that were added in https://github.com/tempestphp/tempest-framework/pull/229, could you let me know if that's something you'd want to go forward with?

I still need to go through that one, probably won't happen today though. What I can tell from a quick scan is that we definitely need this, we might just need to finetune the details.

src/Console/Commands/MigrateCommand.php
przemyslaw-przylucki: I'd go with string | string[]. If user wants to create a longer description, or add some white space, this allows you for an easier control while still being convenient

src/Console/Styling/ConsoleOutputThemeInitializer.php
przemyslaw-przylucki: I don't remember why I went with that design.

I think you are right - the ConsoleApplication should probably pull the base theme from the ConsoleConfig. 

When it comes to user-defined themes:

We could add a property to ConsoleCommand there to overwrite the default theme. 

If we want to go crazy and drop final from ConsoleCommand, both 1st and 3rd party packages or organizations could create their own styling, which is pretty neat from package author "branding" perspective.

This way they could create a simple wrapper around ConsoleCommand and use their own brand i.e PlatypusCorpCommand

src/Console/ArgumentBagInitializer.php
```diff
@@ -0,0 +1,23 @@
...
+use Tempest\Container\Singleton;
+
+#[Singleton]
+final readonly class ArgumentBagInitializer implements Initializer
+{
+    public function initialize(Container $container): ArgumentBag
+    {
+        $arguments = $_SERVER['argv'];
+
+        return match (! ! $arguments) {
```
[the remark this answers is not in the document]
przemyslaw-przylucki: I love !! as a quick boolean cast - works in most languages

src/Support/BaseBuilder.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Support;
+
+class BaseBuilder
```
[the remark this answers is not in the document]
przemyslaw-przylucki: I'll go with a for now ;p

src/Console/ArgumentBagInitializer.php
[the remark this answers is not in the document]
brendt: Maybe it needs some getting used to 🤔

src/Support/BaseBuilder.php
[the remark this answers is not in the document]
brendt: Yeah I like the trait approach better 😁 thanks!

src/Console/Commands/MigrateCommand.php
brendt: Wouldn't a heredoc be sufficient in that case? I'm just trying to imagine when an array would really be helpful, and I can't really come up with a good example.

src/Console/Styling/ConsoleOutputThemeInitializer.php
brendt: > If we want to go crazy and drop final from ConsoleCommand, both 1st and 3rd party packages or organizations could create their own styling, which is pretty neat from package author "branding" perspective.

We don't want to go crazy now, do we? 😛

Sidenote: let me actually share some background info while we're at it: I'm considering to extract the console code into a separate package that people can use as standalone console replacement. I figure it's about time that `symfony/console` gets some proper competition 😛

That won't happen before we merge all these features though, and we shouldn't worry too much about it for now. I'll take care of extracting stuff all at once. I might need to make some changes down the line for things to work in a standalone context.

/sidenote

So regarding personalization, I think registering the output builder within the container is the way to go, and have the initializer take care of setting up branding. To me, that feels like the approach that's most in line with the "Tempest way"

PR #229, opened 2024-03-31
RFC: Mapper, Inferrers, Casters discovery

przemyslaw-przylucki:
Top of the morning to you beautiful folks!

This PR continues me going coo-coo on a couple of fun improvements, briefly mentioned in #217:
- discovery for mappers
- discovery for rule inferrers
- discovery for casters

Not sure if I'm happy with the implementation yet.

This PR also changes how validation is run - previously we've allowed casting arbitrary values to any types - now it'll cause a validation exception. We have to implement rules for rest of the types and add inferrers.

This will play nicely with #228 too =)


src/Mapper/ObjectMapper.php
```diff
@@ -101,12 +93,29 @@ private function map(
...
 
         throw new CannotMapDataException($from, $to);
     }
+
+    /**
+     * @return Mapper[]
+     */
+    protected function getMappers(): array
+    {
+        return [
```
przemyslaw-przylucki, line 111: One of the issues I've run into with dynamic mappers is that the order sometimes has to be enforced. I don't have good ideas on solving it easily

src/Config/casters.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+declare(strict_types=1);
+
+use Tempest\Mapper\CasterConfig;
+
+return new CasterConfig();
```
brendt, line 7: I think it makes more sense to have one `mapperConfig` and have several arrays in that class.

src/Mapper/ArrayToObjectMapper.php
```diff
@@ -101,13 +97,16 @@ private function getCaster(ReflectionProperty $property): ?Caster
...
 
         if (! $castWith) {
-            return match($type = $property->getType()->getName()) {
-                'int' => new IntegerCaster(),
-                'float' => new FloatCaster(),
-                'bool' => new BooleanCaster(),
-                DateTimeImmutable::class, DateTime::class, DateTimeInterface::class => $this->createDateCaster($type, $property),
-                default => null,
-            };
+            $config = get(CasterConfig::class);
```
brendt, line 100: This should probably be injected

src/Mapper/ArrayToObjectMapper.php
```diff
@@ -245,27 +246,20 @@ private function validate(object|string $object): void
...
-        $format = null;
-
-        try {
-            $format = attribute(DateTimeFormat::class)
-                ->in($property)
-                ->first();
-        } catch (Throwable) {
+        $failingRules = [];
 
+        foreach (get(InferrerConfig::class)->inferrers as $inferrer) {
```
brendt, line 253: Should be injected

src/Mapper/ArrayToObjectMapper.php
```diff
@@ -245,27 +246,20 @@ private function validate(object|string $object): void
...
+            foreach ($inferrer->infer($property, $value) as $rule) {
```
brendt, line 254: Can you explain the goal of inferrers?

brendt (changes requested):

src/Mapper/ArrayToObjectMapper.php
przemyslaw-przylucki: Sure thing!

The goal of inferrers is to provide a convenient way to register automatic validation rules for types. It's inspired by my fav feature of laravel-data =)

For example - in this PR we have ModelId, which is a simple vo, and in our system it has to start with 5 for some reason - we have registered the inferrer, now any time a new ModelId is created from a request, we are going to run that validation automatically.

This is especially useful in big apps where you use value objects. It saves tons of time and encourages developers to create stricter types, cuz you get the validation *everywhere* for free - people don't even have to know that it happens which is great if you add new people to the project and you want to ensure safety.

If you don't like the callback style, we could possibly just register inferrers as a key => value in the config, so it's slightly faster to infer the rules, we'd lose some flexibility though.

przemyslaw-przylucki: On top of that - before this change, if we sent a JSON request to an endpoint:

```
{
abc: [[[]]]
}
```

that has request class with

public bool $abc;

It'd just be casted to bool without any problems - which I think is safe to assume is not what we'd want to happen.

brendt:
Closing due to inactivity and refactoring to tempest/core

PR #232, opened 2024-04-03
Feat: Console Output Builder

przemyslaw-przylucki:
Hiya - first extracted PR out of #228 

Includes console output builder implementation and replaces commands to use the builder instead


src/Console/ConsoleOutputBuilder.php
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public function withDefaultBranding(): self
+    {
+        return $this->blank()
+            ->brand('Tempest')
+            ->blank();
+    }
+
+    public function write(ConsoleOutput $to): self
```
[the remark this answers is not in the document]
brendt: And like I said in a previous comment: I wonder if we can pass in the Console via the initializer, and get rid of this write method altogether? Maybe it's not as trivial as I think it is. Let me know your thoughts

src/Console/ConsoleOutputBuilder.php
```diff
@@ -0,0 +1,159 @@
...
+
+        return $this;
+    }
+
+    public function __toString(): string
+    {
+        return $this->toString();
+    }
+
+    public function toString(bool $format = true): string
```
brendt, line 139: Why would `format` be toggleable?

src/Console/RenderConsoleCommandOverviewMessage.php
```diff
@@ -16,15 +16,12 @@ public function __construct(
...
-        if ($this->appConfig->discoveryCache) {
-            $lines[] = ConsoleStyle::BG_RED(' Discovery cache is enabled! ');
-        }
-
-        $lines[] = '';
+        $builder = ConsoleOutputBuilder::new()
+            ->withDefaultBranding()
+            ->when(
+                $this->appConfig->discoveryCache,
+                fn (ConsoleOutputBuilder $builder) => $builder->error(' Discovery cache is enabled! ')->blank()
```
brendt, line 23: Should probably be a warning instead of an error

src/Console/RenderConsoleCommandOverviewMessage.php
```diff
@@ -40,16 +37,16 @@ public function __invoke(): string
         ksort($commands);
 
         foreach ($commands as $group => $commandsForGroup) {
-            $lines[] = ConsoleStyle::BOLD(ConsoleStyle::BG_BLUE(' ' . ucfirst($group) . ' '));
+            $builder->formatted(ConsoleStyle::BOLD(ConsoleStyle::BG_BLUE(' ' . ucfirst($group) . ' ')));
```
brendt, line 40: Why don't we provide a method for this style? I'm ok if it changes a bit, I just put random something together initially.

brendt (changes requested):
Great PR! Splitting it into smaller parts does indeed help me wrap my head around things :D 

The only major question I have is about the container, would love your input on that!

src/Console/ConsoleOutputBuilder.php
przemyslaw-przylucki: Lemme know what you wanna do here!

przemyslaw-przylucki:
@brendt I've pushed some stuff, lemme know what you think now - I like the current solution quite a lot 😎

src/Console/Commands/RoutesCommand.php
```diff
@@ -41,13 +42,13 @@ public function list(): void
         ksort($sortedRoutes);
 
 
-        ConsoleOutputBuilder::new()
-            ->withDefaultBranding()
+        $this->builder
```
brendt, line 43: Consistent naming: $this->outputBuilder

brendt (changes requested):

brendt:
> I like the current solution quite a lot

Same! I'll take care of that inconsistent naming issue, and merge this!

brendt:
Hi @przemyslaw-przylucki 

Some important things to mention about this PR. I'm in the process of extracting all console-related code into a separate package that's used by the framework. The reason being that I want `tempest/console` to work as a standalone package.

This impacts your work in two ways:

- All subsequent PRs regarding console should be sent to https://github.com/tempestphp/tempest-console
- In porting your ConsoleOutputBuilder to this package, I wondered: why not add these methods on the `ConsoleOutput` itself? Did you consider that?

PR #233, opened 2024-04-04
feat: Arguments bag but nicer

przemyslaw-przylucki:


src/Application/ConsoleApplication.php
```diff
@@ -59,41 +63,33 @@ private function handleCommand(string $commandName): void
 
         $handler = $consoleCommandConfig->handler;
 
-        $params = $this->resolveParameters($handler);
+        try {
```
przemyslaw-przylucki, line 68: We have two failing tests at the moment - this is because we're missing type-inferred validation (#229)

brendt:
Could you target this towards https://github.com/tempestphp/tempest-console ?

I did rework the console output builder quite significatly btw. We might need to make more changes to it down the line.

PR #237, opened 2024-04-25
Adds basic transaction manager

przemyslaw-przylucki:


src/Database/Database.php
```diff
@@ -13,4 +15,6 @@ public function getLastInsertId(): Id;
     public function fetch(Query $query): array;
 
     public function fetchFirst(Query $query): ?array;
+
+    public function transaction(): TransactionManager;
```
przemyslaw-przylucki, line 19: idk about this name, but I think this would be a nice helper to have

src/Database/Transactions/GenericTransactionManager.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    public function __construct(private PDO $pdo)
+    {
+    }
+
+    public function begin(): void
+    {
+        $transactionBegun = $this->pdo->beginTransaction();
+
+        if (! $transactionBegun) {
```
przemyslaw-przylucki, line 21: I decided to go with exceptions because the underlying methods already can throw, so it seems more convenient if both kinds of failures would throw

src/Database/Transactions/GenericTransactionManager.php
```diff
@@ -0,0 +1,60 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Transactions;
+
+use PDO;
+use Tempest\Database\Exceptions\TransactionException;
+use Throwable;
+
+final class GenericTransactionManager implements TransactionManager
```
przemyslaw-przylucki, line 11: this is a bit oversimplified as it doesn't handle nested transactions, but it's a start :)

src/Database/DatabaseInitializer.php
```diff
@@ -8,12 +8,16 @@
...
 
 #[Singleton]
 final readonly class DatabaseInitializer implements Initializer
 {
     public function initialize(Container $container): Database
     {
-        return new GenericDatabase($container->get(PDO::class));
+        return new GenericDatabase(
+            $container->get(PDO::class),
+            $container->get(TransactionManager::class),
```
aidan-casey, line 20: Is this something we can really resolve from the container? What happens if we are interacting with multiple databases and one of them is not supported by PDO?

brendt: Well right now, we only support PDO for databases, and I don't plan on adding anything else soon. So I'd say it's fine for now

brendt:
Still need to read through this, will do so next week

src/Database/DatabaseInitializer.php
aidan-casey: @brendt - that's not a very useful mentality for the users of the framework. 😛 

You might not support it out of the box, but interfaces exist so I could add the implementation for an API based database, for example. I have an application in the wild that has two different connections and one is not supported by PDO.

src/Database/Transactions/TransactionManager.php
```diff
@@ -0,0 +1,16 @@
...
+
+interface TransactionManager
+{
+    public function begin(): void;
+
+    public function commit(): void;
+
+    public function rollback(): void;
+
+    public function execute(callable $callback): bool;
```
brendt, line 15: Am I correct in saying that `execute` is a shorthand for `begin`, `commit`, and `rollback`?

Maybe it would make more sense to have this method on the database interface instead of `Database::transaction()`. So you'd be able to do something like this

```php
$database->withinTransaction(function (TransactionManager $transaction) {

});
```

It would keep the `TransactionManager` interface a bit more cleaner, and it would make it so that you can immediately start a transaction from the database interface without another method in between.

tests/Integration/Database/GenericTransactionManagerTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $manager->begin();
+
+        $this->console
+            ->call('migrate:up');
+
+        $this->assertCount(3, Migration::all());
+
+        $manager->rollback();
+
+        $this->expectExceptionMessage("no such table");
```
brendt, line 32: I would assert on the exception message, but rather the exception class. If there's no dedicated class for this exception yet then feel free to add it.

src/Database/Transactions/GenericTransactionManager.php
brendt: Seems reasonable :)

src/Database/Exceptions/TransactionException.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Exceptions;
+
+use Exception;
+
+final class TransactionException extends Exception
+{
+    public static function transactionFailed(): self
```
brendt, line 11: I'd rather have three separate exception classes instead

src/Database/Database.php
brendt: I made a comment about it somewhere further down below :)

src/Database/DatabaseInitializer.php
brendt: In that case I'd say the framework doesn't support it for now and you'll have to manage that other connection manually. 

In an ideal world we support everything from the get-go, but trying to do so would prevent us from ever reaching 1.0. While I agree with you in theory, I think extending the design of the database class outside of PDO is something we'll have to consider after 1.0.

brendt (changes requested):

src/Database/Transactions/TransactionManager.php
przemyslaw-przylucki: > Am I correct in saying that execute is a shorthand for begin, commit, and rollback?

Yuppers!

Hmmm, I kinda like this, but I'm not sure if I understand why this callback is getting an instance of `TransactionManager` - is `withinTransaction` supposed to be a replacement for `execute`, or just a callback that allows you to quickly get an instance of the `TransactionManager`? Meaning you'd still be responsible for calling commit?

przemyslaw-przylucki:
@brendt I've adjusted some things - lemme know if that's what you had in mind 💪

src/Database/Exceptions/DatabaseException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Exceptions;
+
+use Exception;
+
+abstract class DatabaseException extends Exception
```
przemyslaw-przylucki, line 9: While not required now, it could be useful for grouping in try-catches

Alternatively, this could be an interface that just extends Throwable, but idk class felt more natural here

brendt: Well you can also group catches like so:

```php
catch(ExceptionA|ExceptionB|ExceptionC $exception)
```

So I think separate exceptions will be better

src/Database/GenericDatabase.php
```diff
@@ -39,6 +42,23 @@ public function fetchFirst(Query $query): ?array
         return $this->fetch($query)[0] ?? null;
     }
 
+    public function withinTransaction(callable $callback): bool
+    {
+        $this->transactionManager->begin();
+
+        try {
+            $callback();
```
brendt, line 50: Should we pass in the transaction manager and/or the database into this closure? Not sure. We can keep it as is for now, but I'd like your input

src/Database/Transactions/GenericTransactionManager.php
```diff
@@ -0,0 +1,44 @@
...
+    public function __construct(private PDO $pdo)
+    {
+    }
+
+    public function begin(): void
+    {
+        $transactionBegun = $this->pdo->beginTransaction();
+
+        if (! $transactionBegun) {
+            throw new CannotBeginTransaction();
```
brendt, line 23: Technically, this exception should be called `CouldNotBeginTransaction`, since we've tried to start it, and it didn't work. The current name implies that we didn't even try to start it, because of some conditions unknown.

src/Database/Transactions/GenericTransactionManager.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    public function commit(): void
+    {
+        $transactionCommitted = $this->pdo->commit();
+
+        if (! $transactionCommitted) {
+            throw new CannotCommitTransaction();
```
brendt, line 32: Same for the other two

brendt (approved):
Cool! I'll merge this and make some small adjustments on main

src/Database/Exceptions/DatabaseException.php
przemyslaw-przylucki: Of course you can! But if you want to catch any problem with DB, especially as we add new ones, having a catch-all for db would be nice imho

PR #240, opened 2024-05-05
feat: Adds Timebox utility class and extends Clock

przemyslaw-przylucki:
Motivation:
Timebox is useful in:
- authentication, to prevent certain types of attacks (https://ephort.dk/blog/laravel-timing-attack-vulnerability/?ref=securinglaravel.com)
- sub-minute scheduler, where we want every iteration of scheduler to run for exactly one second

After this gets merged, I'll adjust https://github.com/tempestphp/tempest-console/pull/24 to utilize it


src/Clock/Clock.php
```diff
@@ -14,4 +14,8 @@ public function now(): DateTimeImmutable;
     public function time(): int;
 
     public function sleep(int $seconds): void;
+
+    public function utime(): int;
+
+    public function usleep(int $microseconds): void;
```
aidan-casey, line 20: This feels like an implementation detail to me. What's the harm in just updating the implementation to make `time`/`sleep` microseconds?

src/Clock/Timebox.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Clock;
+
+use Throwable;
+
+final readonly class Timebox
```
aidan-casey, line 9: This really feels like an early abstraction to me. You mentioned HTTP (I'd recommend a middleware) and scheduling in the console. What are the other use cases for this utility?

src/Clock/Clock.php
przemyslaw-przylucki: I agree with that, I have a better idea, will adjust tomorrow

src/Clock/Timebox.php
przemyslaw-przylucki: There are no other current use-cases, both of the things I mentioned could be solved locally in said places, but it felt like a good place to start the conversation about handling dates overall and unifying how we approach them, as well as a reminder to handle that while implementing the auth.

If this gets merged we can simplify how `Every` and `Interval` are implemented in tempest/console

src/Clock/GenericClock.php
```diff
@@ -13,13 +13,28 @@ public function now(): DateTimeImmutable
...
+     * Returns the unix timestamp of the current time in the given unit.
+     *
+     * @param TimeUnit $unit
+     *
+     * @return int
+     */
+    public function time(TimeUnit $unit = TimeUnit::SECOND): int
     {
-        return time();
+        return (int) floor((microtime(true) * 1_000_000) / $unit->toMicroseconds());
```
przemyslaw-przylucki, line 25: I don't really like this; Maybe we should default to microseconds instead and then add some utility class to convert any unit to microseconds?

brendt:
Since we're not going for sub-minute scheduling, I don't think we should add this PR for now. 

Maybe when we're further working on Auth we can reconsider it, but I'm gonna close it for now. Thanks for thinking along though! Really appreciate all your work 😁

PR #241, opened 2024-05-10
Adopt monorepo structure 😉 

aidan-casey:
@brendt - this is very much proof of concept, but I wanted to get approval before going too far down this road.


aidan-casey:
@brendt - A few notes on what this PR looks to accomplish and what I have done so far.

- [x] Adds `symplify/monorepo-builder` as a dev dependency to handle _most_ of the mono-repo tasks.
- [x] Moves everything to `src/Tempest` as the directory for `symplify/monorepo-builder` to watch over.
- [x] Adds license files to each sub-component.
- [x] Adds composer files to each sub-component.
- [ ] Adds cross dependencies to composer files. Going forward these will be managed by `monorepo-builder` and automatically bumped when a release is tagged on the framework repo.
- [ ] Adds cross dependency validation to the QA script.
- [ ] Moves certain namespaces to better package names (e.g., `Commands` becomes `CommandBus`).
- [ ] Moves `console` back in-house. 👀 
- [ ] Automatically push to read-only sub-splits.

brendt:
Could you explain the day-to-day flow? How would I tag specific releases for one package, for example?

aidan-casey:
@brendt - If you look at any other framework (Symfony or Laravel, for example) each component is bumped at the same time. If you do not do this, it very quickly becomes hard to map dependency requirements as you grow components and just plain old bizarre when you have an overarching "framework" package that can be required.

This is pretty standard practice for a monorepo and also adopted here via monorepo-builder.

brendt:
This confuses me though. Let's say we add a patch in tempest/console, will all other package get the version bump as well then? Even when they don't have any changes?

aidan-casey:
@brendt - think of it more like this: `framework` (which has all the components) receives a patch that is tagged and `console` is split off to its own repository + package and given the same version so someone could include it individually.

It used to bother me too, but if you think of it from the perspective of the framework version, it makes sense.

The alternative is handling individual release cycles and I think that will be more of a burden in the long run.

brendt:
Yeah it's not a blocker. A bit confusing at first but I'll get used to it.

How does tagging work? Should I always use the CLI tool? I reckon it won't suffice to tag a version via GitHub. What happens if I tag a version on GitHub? Does that mess things up?

aidan-casey:
My goals is to setup two different actions:

1. On push to master, readonly sub-splits are created (i.e., `tempestphp/container` will be updated with the latest container code).

2. On tag (from GitHub) each sub-split will be tagged.

So while the CLI tool will be used, it will be via actions that are triggered through the normal GitHub flow on the `tempestphp/framework` repo.

Issue #10, opened 2023-07-19
Authentication & Authorization


brendt:
We need an abstraction to handle user logins. I don't want it to be tied directly to the database though, so we probably need to carefully think about how we design it.

WendellAdriel:
Hey @brendt, since you don't want to tie the authentication to the DB, I was thinking about and I had an idea.

This is just a high-overview idea, but if you feel it's something we can explore I can think more about it and start drafting a PR.

What if we had something like this?

```php
final class AuthConfig
{
    public function __construct(
        public readonly CredentialsResolver $credentials,
        public readonly SessionDriver $driver,
    ) {
    }
}
```

```php
interface CredentialsResolver
{
    public function identifier(): string; // This would return for example an username or email

    public function secret(): string; // This would return for example a password or API key
}
```

Then we could have some implementations for the `CredentialsResolver` like `DatabaseCredentials` that would receive an `Authenticable` parameter that could be applied to any model class.

```php
final readonly DatabaseCredentials implements CredentialsResolver
{
    public function __construct(
        Authenticable $authenticable,
    ) {
    }
}
```

We could have an `OAuthCredentials` that would wrap the common logic of an OAuth flow.
Other examples would be:

- `SocialCredentials` that would provide out-of-the-box some social login logic.
- `LdapCredentials` that would provide an LDAP login logic.

brendt:
I think I like where this is going. What's still missing is the public facing API. How would framework users interact with this auth layer? Did you have some thoughts about that?

WendellAdriel:
Yes, but I think I sent some wrong names in my first comment, the idea would be to have something like:

```php
interface Authenticable
{
    public function identifier(): string; // This would return for example an username or email

    public function secret(): string; // This would return for example a password or API key
}
```

Then in the configuration, it would have something like this:

```php
final class AuthConfig
{
    public function __construct(
        public readonly Authenticator $authenticator,
        public readonly SessionDriver $driver,
    ) {
    }
}
```

Then we would have a common contract on how to handle authentication and that would be the public one.
I thought about a really minimalist public API.

```php
interface Authenticator
{
    public function authenticate(Authenticable $authenticable): void;  // It would throw an exception if the authentication fails

    public function invalidate(): void;

    public function user(): Authenticable;
}
```

Then we would have the different authenticator like `DatabaseAuthenticator`, `OAuthAuthenticator`, `SocialAuthenticator`, `LdapAuthenticator`.

[...]

PR #242, opened 2024-05-12
Authentication Layer Implementation

WendellAdriel:
Following the discussion on #10 this PR will address the implementation for the Authentication Layer of the framework.
This is still in an early stage, so I'll be updating the description or adding more comments as I go.
I created it just to get initial feedback if I'm going in the right direction.

Can you take a look and see if I should change anything on this initial approach @brendt?


src/Auth/AuthConfig.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Tempest\Auth;
+
+use Tempest\Auth\Authenticators\DatabaseAuthenticator;
+use Tempest\Auth\Contracts\Authenticator;
+
+final class AuthConfig
+{
+    public function __construct(
+        /** @var class-string<Authenticator> */
+        public string $authenticatorClass = DatabaseAuthenticator::class,
```
brendt, line 14: I'm still uncertain about naming conventions, just a note to myself that I need to make sure all of this is consistent (`authenticatorClass` vs `authenticator`)

src/Auth/Contracts/Authenticable.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth\Contracts;
+
+interface Authenticable
```
brendt, line 7: We need to find a better name

brendt: I think `HasIdentity` was my preference, but I'm open for other suggestions

src/Auth/Contracts/Authenticator.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Tempest\Auth\Contracts;
+
+interface Authenticator
+{
+    public function login(Authenticable $authenticable): void;
+
+    public function logout(): void;
+
+    public function user(): ?Authenticable;
```
brendt, line 13: This will need to be renamed depending on what we name `Authenticable`

src/Auth/AuthConfig.php
WendellAdriel: Should I update to authenticator for now?

src/Auth/Contracts/Authenticable.php
WendellAdriel: I'll update to `HasIdentity` for now then

src/Auth/Contracts/Authenticator.php
WendellAdriel: Got it!

src/Auth/Contracts/Authenticable.php
aidan-casey: @brendt - My only complaint with `HasIdentity` is that it seems to conflict with our trait naming convention.

src/Auth/AuthConfig.php
sneycampos: vote for authenticator

src/Auth/Contracts/Authenticable.php
sneycampos: AuthenticatableInterface ?

src/Auth/AuthConfig.php
brendt: Let's use `authenticator` for now 👍

src/Auth/Contracts/Authenticable.php
brendt: @aidan-casey good point. Did you suggest `Identifiable`? That's fine for me as well. I don't love it, but I don't hate it

src/Auth/AuthConfig.php
WendellAdriel: Got it!

src/Auth/Contracts/Authenticable.php
WendellAdriel: I'll be using `Identifiable` for the interface and I'm going to create a trait for Models.
Should I name it `HasIdentity` then? Or `IsIdentifiable`?
Should I put this trait in the `Database` folder or leave it in the `Auth` one @brendt?

Another thing I'd like to suggest is to start using something like I'm doing to structure the code in the folders, moving interfaces into a `Contracts` folder (or another name), doing the same for `Traits`, I'm thinking on creating them into a `Concerns` folder (or another name).

What do you think, @brendt?

brendt: `HasIdentity` and leave it in the `Auth` folder 👍

I was thinking about the same topic when it comes to structuring, but approach it from a different angle. My intuition says that all classes/interfaces/attributes/enums/… that are accessed by external users should live in the root namespace. 

I don't know if that'll work in the long run though. For now, I'd keep things together in places where it makes sense, and review this topic again at a later point

WendellAdriel: Nice, I'll use this name then.

Ok, I'll follow this approach then, the classes that are going be used/referenced/imported in the user code I'll move them to the root namespace for now.

Thanks for the help on this, @brendt! 💪

src/Auth/DatabaseAuthenticator.php
```diff
@@ -0,0 +1,63 @@
...
+            throw new InvalidLoginException();
+        }
+
+        $this->createSession($identifiable);
+    }
+
+    /**
+     * @return Identifiable|null
+     */
+    public function user()
```
WendellAdriel, line 27: @brendt the idea here is to return the class that the user implements the `Identifiable` interface, but IDK the best approach to map the query result to the target class, do you have any ideas on this?

WendellAdriel:
@brendt I worked on the changes from the previous comments and created a really simple and minimalist implementation for the `DatabaseAuthenticator`, can you take a look and let me know other things that you want the auth layer to provide? Also I created a `HasIdentity` trait that can be used in a model, for example in a User class:

```php
final class User implements Identifiable, Model
{
    use HasIdentity;
}
```

That would already set this as the model used for authentication and by default it would use the `email` and `password` fields for the authentication.

To customize this, it would need to simply provide new values for the `IDENTIFIER` and `SECRET` constants:

```php
final class User implements Identifiable, Model
{
    use HasIdentity;

    protected const string IDENTIFIER = 'username';

    protected const string SECRET = 'secret';
}
```

I'm not encrypting anything here, do you think we should offer encrypting the password to be transparent for the devs?

yassiNebeL (commented):
duplicate folder, move `DatabaseAuthenticator` , `GenericAuthenticator` and drop the "Authenticators" as it is confusing with the parent folder.

Also I don't think that's abstraction is encouraged specially with sensitve datas.

src/Auth/HasIdentity.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Tempest\Auth;
+
+use Tempest\Database\IsModel;
+
+trait HasIdentity
+{
+    use IsModel;
+
+    protected const string IDENTIFIER = 'email';
```
brendt, line 13: I don't think this is the way to go, as these can't be enforced with the interface. Instead provide two getters for it in the interface, and implement default implementations in the trait.

`getIdentifierFieldName` `getSecretFieldName`, something like that

src/Auth/DatabaseAuthenticator.php
brendt: I'm not sure I understand the question? Are you wondering about how to store it in the session?

src/Auth/DatabaseAuthenticator.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+use Tempest\Auth\Exceptions\InvalidLoginException;
+use Tempest\Database\Query;
+
+final class DatabaseAuthenticator extends GenericAuthenticator
```
brendt, line 10: I don't like having to extend from other classes. It's better to extract common functionality to a trait and have this class implement the interface directly.

I think, for now, we only need a database authenticator

src/Auth/GenericAuthenticator.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+use Tempest\Http\Session\Session;
+
+abstract class GenericAuthenticator implements Authenticator
```
brendt, line 9: Let's refactor to a trait `AuthenticatorWithSession` or something alike

src/Auth/GenericAuthenticator.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public const string SESSION_USER_KEY = 'tempest_session_user';
+
+    public function __construct(protected Session $session)
+    {
+    }
+
+    public function logout(): void
```
brendt, line 17: Let's call this method `destroySession`

src/Auth/HasIdentity.php
```diff
@@ -0,0 +1,41 @@
...
+    protected const string SECRET = 'password';
+
+    public function source(): string
+    {
+        return (string) static::table();
+    }
+
+    public function identifier(): string
+    {
+        return static::IDENTIFIER;
```
brendt, line 24: Oh, we already have a method for it. Let's then get rid of the constant alltogether.

src/Auth/Identifiable.php
```diff
@@ -0,0 +1,18 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+interface Identifiable
+{
+    public function source(): string;
+
+    public function identifier(): string;
```
brendt, line 11: Maybe rename to `identifierField`, but I don't have strong opinion about it.

brendt:
> I'm not encrypting anything here, do you think we should offer encrypting the password to be transparent for the devs?

Definitely, passwords should always be encrypted and devs shouldn't worry about it

src/Auth/DatabaseAuthenticator.php
WendellAdriel: No, I was thinking how I could return the model here instead of the array but I have an ideia, I’ll implement it so you can review

src/Auth/HasIdentity.php
WendellAdriel: We have the method, I’ll remove the constants

src/Auth/DatabaseAuthenticator.php
WendellAdriel: Oh ok, so you’re aiming for Composition instead. Got it, I’ll refactor it

src/Auth/GenericAuthenticator.php
WendellAdriel: Thanks! I’ll do it!

src/Auth/GenericAuthenticator.php
WendellAdriel: Got it!

src/Auth/Identifiable.php
WendellAdriel: I think that’s a good idea, I’ll do it!

aidan-casey:
> > I'm not encrypting anything here, do you think we should offer encrypting the password to be transparent for the devs?
> 
> 
> 
> Definitely, passwords should always be encrypted and devs shouldn't worry about it

To clarify, passwords should be hashed + salted, not just encrypted. 🙂 

This is probably worth a separate PR for that functionality.

WendellAdriel:
@brendt I worked on the things that you mentioned and also added an example of how the usage of the Auth layer would look with the current implementation in the app folder.

Please let me know if things are looking good and what should be the next steps!

WendellAdriel:
> To clarify, passwords should be hashed + salted, not just encrypted. 🙂

The password_hash function already uses a salt.

brendt:
> To clarify, passwords should be hashed + salted, not just encrypted. 🙂
>
> This is probably worth a separate PR for that functionality.

True, sorry for the confusion!

I think we can use PHP's hashing library? I don't see a need to add a layer on top of it for now tbh

src/Auth/Authenticator.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Tempest\Auth;
+
+interface Authenticator
+{
+    public function login(Identifiable $identifiable): void;
+
+    public function logout(): void;
+
+    public function user(): Identifiable|array|null;
```
brendt, line 13: What is this method used for?

app/Modules/Auth/Models/User.php
```diff
@@ -0,0 +1,23 @@
...
+
+final class User implements Identifiable, Model
+{
+    use HasIdentity;
+    use IsModel;
+
+    public function __construct(
+        public string $name,
+        public string $email = '',
+        public string $password = '',
```
brendt, line 20: All these properties should have `#[SensitiveParameter]` added to them

app/Modules/Auth/AuthController.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public function __construct(
+        private Authenticator $authenticator,
+    ) {
+    }
+
+    #[Post('/login')]
+    public function login(Request $request): Response
+    {
+        $user = (new User(name: 'John Doe'))->setCredentials(
```
brendt, line 25: I think it's weird that we create a new User object, why don't we retrieve it from the database?

brendt: I understand that we're doing the retrieving somewhere else, but imo this is a code smell. Either we should do the retrieval in the controller, and `Authenticator::login` only sets up the session; or we should do it in `DatabaseAuthenticator`, which means the `login` signature should change to `Authenticator::login($name, $password)`

IMO, the first approach is better, because there will be ways we can log in identifiables without username/password (think oauth, login links, …)

So, I would say the `Authenticator`'s concern is only about managing the login session for a user. It'll be interesting to brainstorm ideas about being able to log in outside a web context (CLI), and to log in as an admin user acting as another user. Those are concerns that the authenticator should be able to handle, but it shouldn't care about where an identifiable came from.

I think that leads us to conclude that we need something else that can resolve an identifiable from some kind of user input (username/password, login link, oauth token, …). I think there's a gap to be filled there as well, but maybe it's better to first have this merged, and then work on the resolving part separately.

WDYT?

src/Auth/Identifiable.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+interface Identifiable
```
brendt, line 7: I think that, if we separate the two concerns of "resolving an identifiable" and "managing the session", we'll probably have to refactor this interface as well.

brendt (changes requested):
I think there's a lot of potential here! I do feel like we need to improve the concept a bit more before merging. Let me know your thoughts about the comments I left, it's totally up for debate, I'm just brainstorming as well :)

src/Auth/Authenticator.php
WendellAdriel: This is to return the identifiable data from the session

app/Modules/Auth/Models/User.php
WendellAdriel: Got it!

app/Modules/Auth/AuthController.php
WendellAdriel: That's true!
I think these should be two separate things.
Here's an idea:

1 - We update the `Identifiable` to have just one method: `identifier` that's going to be the unique identifier for it

2 - We create another interface called `CredentialsResolver` or something like this. This is going to be an empty interface and we would use it only to mark a class as the one used to resolve some credentials. Why should it be empty? Because getting credentials for web, CLI, OAuth, etc, are all different. Then we could use this to have implementations for different types of credentials, for example, the Database one would be something like this.

```php
final readonly class DatabaseCredentials implements CredentialsResolver
{
    public function __construct(
        private string $source,
        private string $identifierField,
        private string $secretField,
    ) {
    }

    public function resolve(string $identifier, string $password): Identifiable
    {
        // Get user from DB here
    }
}
```

For a login it would be something like this:

```php
final readonly class AuthController
{
    public function __construct(
        private CredentialsResolver $credentialsResolver,
        private Authenticator $authenticator,
    ) {
    }

    #[Post('/login')]
    public function login(Request $request): Response
    {
        $user = $this->credentialsResolver->resolve(
            identifier: $request->get('email'),
            password: $request->get('password')
        );

        $this->authenticator->login($user);

        return response()->ok();
    }
}
```

The only thing I'm not sure for this approach is the best way to configure the three parameters of the `DatabaseCredentials`

brendt: > Why should it be empty? Because getting credentials for web, CLI, OAuth, etc, are all different

I was thinking about this as well. Another approach is to pass the variable data to the constructor of each resolver, and have the method be something like this: `resolve(): Identifiable`.

If the constructor is already saturated with input and we don't want to add any more params (also because we might want to register it in the container etc), we could have some kind of object and do manual instanceof checks

```php
final readonly class DatabaseCredentials implements CredentialsResolver
{
    public function __construct(
        private string $source,
        private string $identifierField,
        private string $secretField,
    ) {}

    public function resolve(AuthenticationRequest $request): Identifiable
    {
        if (! $request instanceof DatabaseAuthenticationRequest) {
            throw new Exception;
        }
        
        // …
    }
}
```

Just FYI, these are the kinds of problems for which I wish we had generics 😁

brendt: Also: `AuthenticationRequest` is a sucky name that'll need to be changed

WendellAdriel: I like the idea, I'm going to make some changes in the PR to follow this approach!
I think that generics would be a HUGE improvement to the language.

WendellAdriel:
Oh damn...
I knew I had typed the wrong alias for some git commands
I screwed the git history for this branch 😢

WendellAdriel:
@brendt can you take a look and check if the implementation is going in the right direction now and what should we improve? I can fix the git history before the merge if needed.

app/Modules/Auth/AuthController.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    public function __construct(
+        #[Tag('database')]
+        private CredentialsResolver $credentialsResolver,
+        #[Tag('database')]
+        private Authenticator $authenticator,
+    ) {
+    }
+
+    #[Post('/login')]
```
brendt, line 28: Shouldn't we have Get routes as well?

brendt: To show the login and logout form*

brendt: FYI I can add these after this is merge, just wanted to check. Forms are still a mess, so I don't expect you to struggle with those :p

src/Auth/AuthenticationCall.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+interface AuthenticationCall
```
brendt, line 7: I generally don't like empty interfaces, they usually indicate a code smell of some sorts.

src/Auth/Authenticator.php
```diff
@@ -0,0 +1,19 @@
...
+interface Authenticator
+{
+    public function login(Identifiable $identifiable): void;
+
+    public function logout(): void;
+
+    /**
+     * @throws MissingIdentifiableException
+     */
+    public function user(): Identifiable|null;
```
brendt, line 18: Shouldn't this method be called `identifiable`?

brendt: Edit: maybe `getIdentifiable`. We'll refactor to use property hooks in 8.4, and it'll be easier to do search on `get*` methods in interfaces.

src/Auth/AuthenticatorWithSession.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+use Tempest\Http\Session\Session;
+
+trait AuthenticatorWithSession
```
brendt, line 9: I think we just need a `SessionAuthenticator` class with this code, no?

src/Auth/DatabaseAuthenticator.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function logout(): void
+    {
+        $this->destroySession();
+    }
+
+    /**
+     * @throws MissingIdentifiableException
+     */
+    public function user(): Identifiable|null
```
brendt, line 36: IMO this method doesn't belong here. We should split the concern of resolving an identifiable and logging them in. So, in my mind, we'd have `DatabaseIdentifier` and `SessionAuthenticator`.

We could add more identifiers in the future like, for example, `GithubOAuthIdentifier`, and more authenticators (I actually can't come up with a proper example besides session atm)

brendt (changes requested):
Still some remarks, but thanks already for all the work you're putting into this 💪

About the Git mess up: yeah we need to figure that out before merging. If you want to you can reopen this PR if that's the easiest.

app/Modules/Auth/AuthController.php
WendellAdriel: Got it!
I won't add these for now then.

src/Auth/AuthenticationCall.php
WendellAdriel: We started discussing about this here: https://github.com/tempestphp/tempest-framework/pull/242#discussion_r1609698847 because every `AuthenticationCall` implementation will need different parameters and this was going to be used just to type the call parameter in the `CredentialsResolver` interface:

```php
interface CredentialsResolver
{
    public function resolve(AuthenticationCall $call): Identifiable;
}
```

How should we approach this then?

WendellAdriel:
Hey @brendt I worked on some changes, please let me know which approach you want to follow for the `AuthenticationCall` interface. The other observations that you made I think I addressed all of them.

Regarding the Git History, when this looks good and no other changes are needed I'll create a new PR with it fixed.

brendt:
Gonna merge this in a feature branch, and finetune it on there 👍 Thanks for the help, and sorry it took so long to get merged

PR #255, opened 2024-06-02
Adds console subsplit.

aidan-casey:


src/Tempest/Console/Scheduler/NullShellExecutor.php
```diff
@@ -6,7 +6,7 @@
 
 use Tempest\Console\ShellExecutor;
 
-class NullShellExecutor implements ShellExecutor
+final class NullShellExecutor implements ShellExecutor
```
aidan-casey, line 9: @brendt - We can't do this quite yet. Mocks were used for the testing of this class, which then fails tests because it is final.

phpstan-baseline.neon
```diff
@@ -1,5 +1,15 @@
 parameters:
 	ignoreErrors:
+		-
```
aidan-casey, line 3: Ahhh... so we're taking the easy way out? 🙃

src/Tempest/Console/Scheduler/NullShellExecutor.php
brendt: We shouldn't use mocks then :p Keep it in, I'll fix it

phpstan-baseline.neon
brendt: For now :)

src/Tempest/Console/Scheduler/NullShellExecutor.php
aidan-casey: @brendt - Totally agreed that we shouldn't be using that approach. Was going to leave in for this PR and create an issue/fix in another PR.

PR #257, opened 2024-06-04
View renderer

brendt:
I'll need to change the base once the add-console-subsplit branch has been merged

- [x] Error handling for components with the same name
- [x] Fix old view files that still call extends, etc.
- [x] Move to other DOM parser that isn't deprecated 
- [ ] `<x-slot />` works but `<x-slot></x-slot>` doesn't for default slot in component
- [x] `<pre>` rendering is broken because of whitespace stripping. Need to find another DOM parser library


src/Tempest/View/IsView.php
```diff
@@ -4,25 +4,18 @@
...
-use function Tempest\path;
-use function Tempest\type;
 use function Tempest\view;
 
 trait IsView
 {
     public string $path;
-    public array $params = [];
-    private array $rawParams = [];
+    public array $data = [];
```
brendt, line 15: All of this will need to be refactored to property hooks

src/Tempest/View/IsView.php
```diff
@@ -50,52 +38,70 @@ public function path(string $path): self
...
-
-            $extendsData = [...$slots, ...$this->extendsParams];
-
-            return view($this->extendsPath)
-                ->data(...$extendsData)
-                ->render();
-        }
-
-        return $this->parseViewComponents($contents);
+        return get(ViewRenderer::class)->render($this);
```
brendt, line 104: Need to remove this method

src/Tempest/View/ViewRenderer.php
```diff
@@ -0,0 +1,216 @@
...
+use ReflectionException;
+use ReflectionParameter;
+use SimpleXMLElement;
+use Tempest\AppConfig;
+use function Tempest\get;
+use function Tempest\path;
+use function Tempest\type;
+use function Tempest\view;
+
+final class ViewRenderer
```
brendt, line 18: Need to introduce an interface so that we can have multiple view renderers

src/Tempest/View/View.php
```diff
@@ -4,18 +4,33 @@
 
 namespace Tempest\View;
 
+/**
+ * @method include(string $path, ...$params): string
```
brendt, line 8: This sucks, will need to refactor so that the view can include another view, and the renderer knows how to render it.

brendt:
I removed the ability to call include, and will probably do the same for extends. All this functionality is covered by view components, which is a much better API.

PR #260, opened 2024-06-04
Add core sub splits

aidan-casey:


aidan-casey:
@brendt - I am assuming these test errors about the migrations table is because we are manually calling it prior to running the rest of the migrations and then it is also being discovered?

brendt:
🥳

Issue #259, opened 2024-06-04
Finalize `NullShellExecutor`


aidan-casey:
The `NullShellExecutor` is currently tested using mocks. This prevents us from making the class final.

We should revise the tests for that class and finalize the class.

Qupiter:
Hey there, can you assign this issue to me? 
I would like to work with you. This seems like a good issue to start with.

alibori:
Hi @aidan-casey !

This issue can be closed, isn't it? If I'm right the `NullShellExecutor` class is final.

Thank you! 😃

aidan-casey:
Hey, @alibori! It is final, but has no tests. We still need to refactor so we can test it. 🙂

brendt:
No no, I fixed that, just forgot to close the issue

PR #261, opened 2024-06-22
Finalize NullShellExecutor

andrewdevelopz:
**Issue:** https://github.com/tempestphp/tempest-framework/issues/259

We can not create a mock of final classes in phpunit.

**Solution:**

Tempest/Testing/BypassMock was created to filter out `final` token in each php file. It overrides php file stream wrapper and turns all final classes into normal classes that can be mocked.

### DIFF
--- a/src/Tempest/Console/Scheduler/NullShellExecutor.php
+++ b/src/Tempest/Console/Scheduler/NullShellExecutor.php
@@ -6,7 +6,7 @@ namespace Tempest\Console\Scheduler;
 
 use Tempest\Console\ShellExecutor;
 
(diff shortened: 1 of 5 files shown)


brendt:
While I appreciate the effort, I'm of the opinion that mocks lead to a bunch of overhead and workarounds (as this PR illustrates). The solution I'm aiming for is to not use mocks at all anymore :)

andrewdevelopz:
Hey @brendt, thanks for the feedback and I appreciate all that you do. I'll keep trying to find ways to contribute as I continue learning.

brendt:
Awesome! Appreciate that!

PR #262, opened 2024-07-11
Fix save method setting wrong id on model

### LINKED ISSUE #221: Wrong id after saving Model
I'm using an sqlite database and when I create two models and then save the first one again, it has the id of the second model.

```php
<?php

declare(strict_types=1);

namespace App\Models;

use Tempest\ORM\IsModel;
use Tempest\ORM\Model;

class Exercise implements Model
{
    use IsModel;

    public string $name;
}
```

```php
<?php
use App\Models\Exercise;

$a = Exercise::create(name: 'A');
$b = Exercise::create(name: 'B');
print_r($a);
$a->save();
print_r($a);
```

```
App\Models\Exercise Object
(
    [name] => A
    [id] => Tempest\Database\Id Object
        (
            [id] => 1
        )

)
App\Models\Exercise Object
(
    [name] => A
    [id] => Tempest\Database\Id Object
        (
            [id] => 2
        )

)
```

After reading through the code I think this happens because in [ModelToQueryMapper](https://github.com/tempestphp/tempest-framework/blob/main/src/Mapper/ModelToQueryMapper.php#L69) you remove the `id` from the `fields`. And then in [Query](https://github.com/tempestphp/tempest-framework/blob/main/src/Database/Query.php#L23-L25) you return the last inserted Id because `id` is missing in the bindings.

NeoIsRecursive:
Fixes #221

This PR changes how the `ModelToQueryMapper` sets the "where id" so that the `id` key can be kept in the `$fields` variable (->bindings) and thus mapping out correctly to the model.

I think this solution works (tests say so at least) but please let me know if I have missed something :)


brendt:
Thanks!

PR #293, opened 2024-07-29
[bug] fix MySQL driver DSN format

### LINKED ISSUE #292: [bug] MySQL driver does not use the correct DSN format
**Problem**

When using the MySQL driver it can not connect to the database.

The cause for this is the DSN format. It specifies the port as a separate option.

ℹ️ I'll make a PR for this.

**Environment**
- macOS 14.5
- Herd 1.9.1
- PHP 8.3.9
- MySQL 8.0.19
- Tempest SHA 8337b0b95040a2e9ad8151a054c632058d295c62

**Steps to reproduce**

1. create a php file in `app/Config` called `database.php`
2. return a database config with the MySQL driver and a different port (example below)
3. create a model and fetch a record or list its records
4. the results can not be displayed because the table can not be found

<details>
<summary>database config</summary>

```php
<?php

declare(strict_types=1);

use Tempest\Database\DatabaseConfig;
use Tempest\Database\Drivers\MySqlDriver;

return new DatabaseConfig(
    driver: new MySqlDriver(
        host: 'localhost',
        port: '3307',
        username: 'root',
        password: '',
        database: 'tempest'
    ),
);
```

</details>

**Solution**
The solution is to update the string that `getDsn()` returns.

<details>
<summary>solution diff</summary>

```diff
diff --git a/src/Tempest/Database/Drivers/MySqlDriver.php:28 b/src/Tempest/Database/Drivers/MySqlDriver.php:28
index 6789c54..bdb769a 100644
--- a/src/Tempest/Database/Drivers/MySqlDriver.php:28
+++ b/src/Tempest/Database/Drivers/MySqlDriver.php:28
@@ -25,7 +25,7 @@ final readonly class MySqlDriver implements DatabaseDriver
 
     public function getDsn(): string
     {
-        return "mysql:host={$this->host};port={$this->port};dbname={$this->database}";
+        return "mysql:host={$this->host}:{$this->port};dbname={$this->database}";
     }
 
     public function getUsername(): ?string


</details>

```

Treggats:
fixes #292

As outlined in the issue the format of the DSN that the MySQL creates did not have the right format.

This PR fixes the format and adds a very simple test to check the DSN both for MySQL and Sqlite. With the possibility to add PostgreSQL.

### DIFF
--- a/src/Tempest/Database/Drivers/MySqlDriver.php
+++ b/src/Tempest/Database/Drivers/MySqlDriver.php
@@ -25,7 +25,7 @@ final readonly class MySqlDriver implements DatabaseDriver
 
     public function getDsn(): string
     {
-        return "mysql:host={$this->host};port={$this->port};dbname={$this->database}";
+        return "mysql:host={$this->host}:{$this->port};dbname={$this->database}";
     }
 
     public function getUsername(): ?string
--- a/tests/Unit/Database/DatabaseDriverTest.php
+++ b/tests/Unit/Database/DatabaseDriverTest.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Tests\Tempest\Unit\Database;
+
+use Generator;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use Tempest\Database\DatabaseDriver;
+use Tempest\Database\Drivers\MySqlDriver;
+use Tempest\Database\Drivers\SQLiteDriver;
+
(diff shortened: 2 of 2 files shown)


brendt:
Thanks! Just FYI: you can run `composer qa` locally, which will fix codestyle issues for you (the one failing action), but it also runs automatically on main, so it's ok if you forget about it:)

Treggats:
> Thanks! Just FYI: you can run `composer qa` locally, which will fix codestyle issues for you (the one failing action), but it also runs automatically on main, so it's ok if you forget about it:)

Yeah I've seen that. Guess I forgot because I'm used to having them run locally on commit. (git hooks)

PR #296, opened 2024-07-31
add a driver for PostgreSql

### LINKED ISSUE #295: add database driver for PostgreSql
As mentioned in #291 Tempest should support the main database drivers.
- SQLite
- MySQL
- PostgreSQL

The latter isn't yet available within Tempest, so that needs to be added.

Treggats:
resolves #295

This PR adds a driver for using the PostgreSQL database. Pretty straightforward, mirrored after the MySQL driver.

I don't have PostgreSql locally, so I've not been able to test this.

### DIFF
--- a/src/Tempest/Database/Drivers/PostgreSqlDriver.php
+++ b/src/Tempest/Database/Drivers/PostgreSqlDriver.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Drivers;
+
+use SensitiveParameter;
+use Tempest\Database\DatabaseDriver;
+
+final class PostgreSqlDriver implements DatabaseDriver
+{
+    public function __construct(
+        #[SensitiveParameter]
+        public string $host = 'localhost',
+        #[SensitiveParameter]
+        public string $port = '5432',
+        #[SensitiveParameter]
+        public string $username = 'postgres',
+        #[SensitiveParameter]
+        public string $password = '',
+        #[SensitiveParameter]
+        public string $database = 'app',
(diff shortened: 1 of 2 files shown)


tests/Unit/Database/DatabaseDriverTest.php
```diff
@@ -28,8 +29,8 @@ public function driver_has_the_correct_dsn(DatabaseDriver $driver, string $dsn):
     public static function provide_database_drivers(): Generator
     {
         yield 'sqlite' => [
-            new SQLiteDriver(path: 'localhost'),
-            'sqlite:localhost',
+            new SQLiteDriver(path: '/usr/local/db.sqlite'),
```
Treggats, line 32: property name is `path`, so it should contain a path and not a hostname

src/Tempest/Database/Drivers/PostgreSqlDriver.php
```diff
@@ -0,0 +1,40 @@
...
+        public string $host = 'localhost',
+        #[SensitiveParameter]
+        public string $port = '5432',
+        #[SensitiveParameter]
+        public string $username = 'postgres',
+        #[SensitiveParameter]
+        public string $password = '',
+        #[SensitiveParameter]
+        public string $database = 'app',
+    ) {
```
Treggats, line 23: maybe it's because I'm working with Laravel for a long time. I always use an empty comment here. 🤷

brendt: Yeah not needed here :)

brendt:
Thanks!

PR #299, opened 2024-08-01
move cache files to a single directory

Treggats:
This might just be a personal preference, but this PR moves all the cache files to a single directory.
Instead of having them scattered throughout 

**Additional**
- enabled the cache for PHPStan
- enable parallel fixing for CS fixer

**Note**
Enabling cache on Github Actions is also possible :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.cache/
 .idea/
 build/
 sessions/
@@ -6,10 +7,6 @@ src/Tempest/database.sqlite
 tests/Fixtures/database.sqlite
 tests/Unit/Console/test-console.log
 .env
-.php-cs-fixer.cache
-.phpunit.result.cache
 composer.lock
 debug.log
 tempest.log
-*.cache.php
-tests/Integration/View/blade/cache
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,5 +1,7 @@
 <?php
 
+use PhpCsFixer\Runner\Parallel\ParallelConfigFactory;
+
 $finder = Symfony\Component\Finder\Finder::create()
     ->in([
         __DIR__ . '/src',
@@ -12,6 +14,7 @@ $finder = Symfony\Component\Finder\Finder::create()
 
 return (new PhpCsFixer\Config())
     ->setCacheFile('.cache/fixer/cs-fixer.cache')
+    ->setParallelConfig(ParallelConfigFactory::detect())
     ->setRules([
         '@PSR12' => true,
         'array_syntax' => ['syntax' => 'short'],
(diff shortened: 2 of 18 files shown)


.cache/tempest/.gitkeep
Treggats, line 1: adding to git because it isn't created automatically

.gitignore
```diff
@@ -6,10 +7,6 @@ src/Tempest/database.sqlite
 tests/Fixtures/database.sqlite
 tests/Unit/Console/test-console.log
 .env
-.php-cs-fixer.cache
```
Treggats, line 9: replaced by the `.cache` entry

.php-cs-fixer.dist.php
```diff
@@ -11,6 +13,8 @@
     ->ignoreVCS(true);
 
 return (new PhpCsFixer\Config())
+    ->setCacheFile('.cache/fixer/cs-fixer.cache')
```
Treggats, line 16: set the location of the cache

.php-cs-fixer.dist.php
```diff
@@ -11,6 +13,8 @@
...
+    ->setParallelConfig(ParallelConfigFactory::detect())
```
Treggats, line 17: enable parallel fixing to speedup the fixing

phpstan.neon.dist
```diff
@@ -10,6 +10,7 @@ services:
 parameters:
 	level: 5
 	reportUnmatchedIgnoredErrors: false
+	tmpDir: .cache/phpstan
```
Treggats, line 13: this was missing, thought it might be a good idea

phpunit.xml
```diff
@@ -8,6 +8,7 @@
         displayDetailsOnTestsThatTriggerNotices="true"
         displayDetailsOnTestsThatTriggerWarnings="true"
         bootstrap="./tests/bootstrap.php"
+        cacheDirectory=".cache/phpunit"
```
Treggats, line 11: updated the cache setting

src/Tempest/Console/Discovery/ScheduleDiscovery.php
```diff
@@ -13,9 +13,9 @@
 use Tempest\Discovery\Discovery;
 use Tempest\Support\Reflection\Attributes;
 
-final class ScheduleDisovery implements Discovery
+final class ScheduleDiscovery implements Discovery
```
Treggats, line 16: typo fix :)

.cache/tempest/.gitkeep
Treggats, line 1: added to git because it is not created automatically

brendt:
Looks good! One potential problem is that discovery cache paths will probably need to differ based on whether you're working on the Tempest codebase, or if you're using Tempest as a dependency within a project. But let's keep it for now. Thanks!

PR #304, opened 2024-08-05
Adds ElseIfAttribute feature

yassiNebeL:


brendt:
There are a lot of things missing from this PR:

- Tests
- Support for `elseif` in the `else` attribute
- Support for `elseif` after an `if` attribute

yassiNebeL:
I’m looking for a method that can help me to determine if a given statement was ok or wasn’t, in order to know if I should return an element or not

Thank you in advance!

brendt:
> I’m looking for a method that can help me to determine if a given statement was ok or wasn’t, in order to know if I should return an element or not

I'm not sure I fully understand? You might need to add a public method in `IfAttribute` and `IfElseAttribute` to determine whether their condition is matched or not.

brendt:
Gonna work on this some more in a separate branch

PR #306, opened 2024-08-07
support different SQL engines

### LINKED ISSUE #291: Migration SQL dialect
The `CreateMigrationsTable` class is written for SQLite, which isn't compatible with MySQL. Maybe we should provide three different implementations (SQLite, MySQL and postgres), the developer will need to publish the right migration file when setting up the database.

### LINKED ISSUE #297: [bug] sorting migrations are not done by name
**Problem**
It seems that the order in which migrations are run, are either not working as expected or I haven't got migration naming correct.

**Scenario**
When running two migrations, where the latter creates a constraint on the former.
The one with the constraint is run before the referenced table is created.

Running the two migrations separate is a work around in this case.

**Steps to reproduce**
- Create two migrations with the following statements
- run `php tempest migrate:up`
- the following error occurs
    `General error: 1824 Failed to open the referenced table 'authors`

<details>
<summary>Tabel 1 (authors)</summary>

```sql
# migration name: 1970-01-02_create_authors_table
create table if not exists authors (
    id bigint auto_increment primary key,
    name varchar(255) not null
);
```

</details>

<details>
<summary>Table 2 (books)</summary>

```sql
# migration name: 1970-01-03_create_books_table
create table books
(
    id        bigint auto_increment primary key,
    title     varchar(255) not null,
    author_id bigint       null,
    constraint books_authors_id_fk foreign key (author_id) references authors (id)
);

create index books__authors__index on books (author_id);
```

</details>

### LINKED ISSUE #298: [feature] add support for multiple database engines
resolves #179 #291

fixes #297

**Situation**
At the moment the initial migration is SQLite only. In order to support multiple database engines we need a way to load migrations with their own SQL dialect.

**Proposed changes**
Combine the above mentioned issues.

- [ ] create SQL files with the migration table creation, for every database driver we currently support
- [ ] the name of the migration should dictate the order that they are run in 
        (#297)
- [ ] let the `CreateMigrationTable` class decide which file to use, based on the currently configured database driver
- [ ] let user generated migrations use the same approach

Treggats:
resolves #298 
resolves #291
fixes #297

This pull request implements support for the supported database engines.
- MySQL
- PostgreSQL
- SQLite

It does so by having a "builder" class which only implements syntax that is different between the different engines. For example the primary key.

It also addresses the sorting of migrations. This caused a bug when dealing with migrations that have foreign keys defined.

**Todo:**
- [x] move binding the database driver to an initializer for the DatabaseDriver
- [x] move the `QueryStatement` init to the driver (`$this->driver->createQueryStatement()`)
- [x] improve the coverage for the `QueryStatement` class
- [x] move the `#[Singleton]` attribute to the `initialize` method (database driver initalizer)
- [x] move sorting the migrations to the manager
- [x] (optional) remove the need of the closure in the `QueryStatement` class
- [ ] (optional) make the database dialect configurable


src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -27,21 +27,21 @@ public function up(): void
         try {
             $existingMigrations = Migration::all();
         } catch (PDOException) {
-            $this->executeUp(new CreateMigrationsTable());
+            $this->executeUp(new CreateMigrationsTable($this->databaseConfig->driver));
```
Treggats, line 30: at the moment passing the driver to the migration, as I mentioned above I'm looking if I can skip this

src/Tempest/Database/QueryStatement.php
```diff
@@ -0,0 +1,102 @@
...
+            default => sprintf('%s INTEGER PRIMARY KEY AUTOINCREMENT', $key),
+        };
+
+        return $this;
+    }
+
+    /** @throws UnhandledMatchError */
+    public function constraint(string $localKey, string $table, string $key = 'id', string $onDelete = 'ON DELETE CASCADE', string $onUpdate = 'ON UPDATE NO ACTION'): self
+    {
+        $this->query[] = match ($this->driver::class) {
```
Treggats, line 76: this will throw a match exception if the driver is not in this list. Which is what we want because that driver might have a different syntax and if so it needs to be added to this list.

src/Tempest/Database/QueryStatement.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return $this;
+    }
+
+    public function __toString(): string
+    {
+        return implode(', ', $this->query);
+    }
+
+    public function toQuery(): Query
```
Treggats, line 98: it would be nice if we could somehow validate the syntax per driver. Haven't looked into it yet though

src/Tempest/Framework/Application/Kernel.php
```diff
@@ -35,6 +37,8 @@ public function init(): Container
             )->boot();
         }
 
+        $container->singleton(DatabaseDriver::class, $container->get(DatabaseConfig::class)->driver);
```
Treggats, line 40: binding the current driver to the driver interface/contract. Currently figuring out if I can make this work for the `QueryStatement` class

src/Tempest/Framework/Testing/IntegrationTest.php
```diff
@@ -49,10 +51,12 @@ protected function setUp(): void
         $request = new GenericRequest(Method::GET, '/', []);
         $this->container->singleton(Request::class, fn () => $request);
         $this->container->singleton(GenericRequest::class, fn () => $request);
+        $this->container->singleton(DatabaseDriver::class, fn () => new SQLiteDriver());
```
Treggats, line 54: same as above, but for tests

tests/Unit/Database/DatabaseQueryStatementTest.php
```diff
@@ -0,0 +1,78 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use Tempest\Database\DatabaseDriver;
+use Tempest\Database\Drivers\MySqlDriver;
+use Tempest\Database\Drivers\PostgreSqlDriver;
+use Tempest\Database\Drivers\SQLiteDriver;
+use Tempest\Database\QueryStatement;
+
+final class DatabaseQueryStatementTest extends TestCase
```
Treggats, line 15: test to check the syntax manually

src/Tempest/Database/MigrationDiscovery.php
```diff
@@ -37,7 +38,7 @@ public function createCachePayload(): string
 
     public function restoreCachePayload(Container $container, string $payload): void
     {
-        $migrations = unserialize($payload);
+        $migrations = unserialize($payload, ['allowed_classes' => [MigrationModel::class]]);
```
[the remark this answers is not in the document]
brendt: TIL!

src/Tempest/Database/Migrations/CreateMigrationsTable.php
```diff
@@ -4,22 +4,32 @@
...
 namespace Tempest\Database\Migrations;
 
+use Tempest\Database\DatabaseDriver;
 use Tempest\Database\Migration;
 use Tempest\Database\Query;
+use Tempest\Database\QueryStatement;
 
 final readonly class CreateMigrationsTable implements Migration
 {
+    public function __construct(
```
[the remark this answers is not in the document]
brendt: You could add something like `$this->driver->createQueryStatement()`, so that you don't have to manually pass it in L26. I wouldn't bother with trying to be more clever than that for now.

src/Tempest/Framework/Application/Kernel.php
brendt: You should move this logic outside the kernel and create an initializer class for `DatabaseDriver` 👍

src/Tempest/Framework/Testing/IntegrationTest.php
brendt: This is ok to hard-code for tests

brendt (changes requested):

src/Tempest/Database/DatabaseConfig.php
```diff
@@ -7,15 +7,35 @@
...
 
         return $this;
     }
+
+    public function set(array $migrations): void
+    {
+        $this->migrations = $migrations;
+    }
+
+    public function get(string $sort = 'asc'): array
```
Treggats, line 32: helper to get and sort the migrations.

src/Tempest/Database/DatabaseDriverInitializer.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use Tempest\Container\Container;
+use Tempest\Container\Initializer;
+use Tempest\Container\Singleton;
+
+#[Singleton]
+final class DatabaseDriverInitializer implements Initializer
```
Treggats, line 12: added wiring for the database driver, so I don't have to register it manually

src/Tempest/Database/Drivers/MySqlDriver.php
```diff
@@ -37,4 +38,9 @@ public function getPassword(): ?string
     {
         return $this->password;
     }
+
+    public function createQueryStatement(string $table): QueryStatement
```
Treggats, line 42: create the `QueryStatement` class through the driver

src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -27,21 +27,21 @@ public function up(): void
...
         }
 
         $existingMigrations = array_map(
-            fn (Migration $migration) => $migration->name,
+            static fn (Migration $migration) => $migration->name,
             $existingMigrations,
         );
 
-        foreach ($this->databaseConfig->migrations as $migrationClassName) {
+        foreach ($this->databaseConfig->get() as $migrationClassName) {
```
Treggats, line 40: get the migrations from start to end. e.g. 0 -> 100

src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -65,14 +65,14 @@ public function down(): void
             $existingMigrations,
         );
 
-        foreach ($this->databaseConfig->migrations as $migrationClassName) {
+        foreach ($this->databaseConfig->get('desc') as $migrationClassName) {
```
Treggats, line 68: when reversing the table creation, we need the migration but in reverse.

Otherwise, if you use foreign keys, you'd get an constraint exception. Because the referenced table got dropped first, but has a reference elsewhere. 💥

src/Tempest/Database/DatabaseConfig.php
```diff
@@ -7,15 +7,35 @@
...
+    {
+        match ($sort) {
```
Treggats, line 34: fixes #297

src/Tempest/Database/DatabaseConfig.php
```diff
@@ -7,15 +7,35 @@
 final class DatabaseConfig
 {
     public function __construct(
-        public readonly DatabaseDriver $driver,
-        public array $migrations = [],
+        private readonly DatabaseDriver $driver,
```
brendt, line 10: We make config properties public. 

In fact, we're actually gonna wait for PHP 8.4 until we release a v1 of Tempest, and we're going to refactor all interfaces to use property hooks instead of methods. It's not related to these properties, but I figured I'd let you know anyway :)

src/Tempest/Database/DatabaseConfig.php
brendt: I guess you need to sort them both ways for up and down? I don't really like hard-coded strings. But it also feels weird to make an enum for it. A boolean also feels icky… Would it make sense to do the sorting in place (where the get method is called)? I reckon it's only one place, but I could be wrong…

src/Tempest/Database/DatabaseDriverInitializer.php
brendt: FYI, you'll need to move the `#[Singleton]` attribute to the `initialize` method instead of the class. This was a change I made this morning :)

src/Tempest/Database/Drivers/MySqlDriver.php
brendt: 👍

src/Tempest/Database/DatabaseConfig.php
brendt: Also the naming is a bit weird, shouldn't it be `getMigrations`?

src/Tempest/Database/Migrations/MigrationManager.php
brendt: Circling back to my previous comment, I think I'd prefer to move the array sorting within the migration manager, so that we can get rid of that `get` method with its hard-coded string for asc/desc entirely.

src/Tempest/Database/QueryStatement.php
```diff
@@ -0,0 +1,110 @@
...
+use RuntimeException;
+use Stringable;
+use Tempest\Database\Drivers\MySqlDriver;
+use Tempest\Database\Drivers\PostgreSqlDriver;
+use Tempest\Database\Drivers\SQLiteDriver;
+use UnhandledMatchError;
+
+final class QueryStatement implements Stringable
+{
+    private array                   $query = [];
```
brendt, line 16: FYI, slightly different code style here, but CS Fixer will take care of it.

src/Tempest/Database/QueryStatement.php
```diff
@@ -0,0 +1,110 @@
...
+            throw new RuntimeException('alter statement should be the first statement');
+        }
+
+        $operation = match ($this->driver::class) {
+            MySqlDriver::class => sprintf('%s', strtoupper($action)),
+            PostgreSqlDriver::class,
+            SQLiteDriver::class => sprintf('%s COLUMN', strtoupper($action)),
+        };
+
+        $this->query[] = sprintf( // @coverage
```
brendt, line 52: What does `@coverage` do?

src/Tempest/Database/QueryStatement.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /** @throws UnhandledMatchError */
+    public function alterTable(string $action, callable $callback): self
+    {
+        if (! empty($this->query)) {
+            throw new RuntimeException('alter statement should be the first statement');
+        }
+
+        $operation = match ($this->driver::class) {
```
brendt, line 46: I think we might refactor at a later point to have dedicated grammar classes, though that's not necessary for now.

tests/Fixtures/Migrations/CreateAuthorTable.php
```diff
@@ -4,23 +4,33 @@
...
     }
 
     public function up(): Query|null
     {
-        return new Query("CREATE TABLE Author (
-            `id` INTEGER PRIMARY KEY AUTOINCREMENT,
-            `name` TEXT NOT NULL,
-            `type` TEXT
-        )");
+        return $this->driver->createQueryStatement('Author')
```
brendt, line 26: I wonder if we need the closure? 

```php
$this->driver
    ->createQueryStatement('Author')
    ->primary()
    ->statement('name TEXT NOT NULL')
    ->statement('type TEXT')
    ->toQuery();
```

brendt (changes requested):

src/Tempest/Database/DatabaseConfig.php
Treggats: Sounds interesting 😋 

for now I'll make it public.
```suggestion
        public readonly DatabaseDriver $driver,
```

src/Tempest/Database/DatabaseConfig.php
Treggats: >  I don't really like hard-coded strings. But it also feels weird to make an enum for it. A boolean also feels icky

I feel the same way 😅 this in combination with the match expression seemed like a good idea. But perhaps I'll move the sorting the the migration manager.

src/Tempest/Database/QueryStatement.php
Treggats: it was my part of remembering which part of the class needed tests :) forgot to remove it.
```suggestion
        $this->query[] = sprintf(
```

tests/Fixtures/Migrations/CreateAuthorTable.php
Treggats: the idea behind the closure is so that I can format the query easily. Maybe I can remove the need for it, but would require some work.

src/Tempest/Database/MigrationDiscovery.php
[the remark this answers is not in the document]
Treggats: I have a branch open where I do this for 99% of these cases

src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -53,9 +54,12 @@ public function down(): void
     {
         try {
             $existingMigrations = Migration::all();
-        } catch (PDOException) {
-            // @todo should be handled better as PDO exception doesn't necessarily mean that the migrations table doesn't exist
-            event(new MigrationFailed('Migration', MigrationException::noTable()));
+        } catch (PDOException $exception) {
+            /** @throw UnhandledMatchError */
+            match ((string) $exception->getCode()) {
+                'HY000' => event(new MigrationFailed('Migration', MigrationException::noTable())),
```
Treggats, line 60: this is the code that the table doesn't exist. Writing this I thought about getting the table name from the exception itself 🤔

src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -65,14 +69,16 @@ public function down(): void
             $existingMigrations,
         );
 
-        foreach ($this->databaseConfig->migrations as $migrationClassName) {
+        $migrations = $this->databaseConfig->getMigrations();
+        krsort($migrations);
```
Treggats, line 73: fixes #297

src/Tempest/Database/QueryStatement.php
Treggats: I like that idea

src/Tempest/Console/Commands/DiscoveryClearCommand.php
```diff
@@ -32,6 +32,10 @@ public function __invoke(): void
 
             $discovery->destroyCache();
 
+            if ($this->appConfig->environment->isTesting()) {
```
brendt, line 35: This is actually needed during testing, however, because we call it during testing bootstrap, it only triggers once per test run

src/Tempest/Database/DatabaseConfig.php
```diff
@@ -8,14 +8,29 @@ final class DatabaseConfig
 {
     public function __construct(
         public readonly DatabaseDriver $driver,
-        public array $migrations = [],
+        private array $migrations = [],
```
brendt, line 11: I guess this can be public again now that the getter and setter don't do anything? That means the getter and setter can also be removed

src/Tempest/Database/DatabaseDriver.php
```diff
@@ -11,4 +11,8 @@ public function getDsn(): string;
     public function getUsername(): ?string;
 
     public function getPassword(): ?string;
+
+    public function createQueryStatement(string $table): QueryStatement;
+
+    public function tableNotFoundCode(): string;
```
brendt, line 17: Maybe this can be moved to the Dialect enum?

src/Tempest/Database/DatabaseDriver.php
```diff
@@ -11,4 +11,8 @@ public function getDsn(): string;
     public function getUsername(): ?string;
 
     public function getPassword(): ?string;
+
+    public function createQueryStatement(string $table): QueryStatement;
```
brendt, line 15: Perhaps we can even move this to the `DatabaseDialect` enum, and simply have an `getDialect` on the driver class?

src/Tempest/Framework/Application/AppConfig.php
```diff
@@ -13,6 +14,7 @@ public function __construct(
         public Environment $environment = Environment::LOCAL,
         public bool $enableExceptionHandling = false,
         public bool $discoveryCache = false,
+        public DatabaseDialect $databaseDialect = DatabaseDialect::SQLITE,
```
brendt, line 17: I think it'll be better to move this to the database driver class, as I mentioned in a previous comment

src/Tempest/Framework/Config/database.php
```diff
@@ -2,11 +2,39 @@
...
 use Tempest\Database\Drivers\SQLiteDriver;
+use Tempest\Framework\Application\AppConfig;
 
-return new DatabaseConfig(
-    driver: new SQLiteDriver(
-        path: __DIR__ . '/../database.sqlite',
+/** @var AppConfig $app */
+$app = GenericContainer::instance()->get(AppConfig::class);
+
+return match ($app->databaseDialect) {
```
brendt, line 16: I don't think we need to provide this level of configurability: let's keep sqlite as the only option, if people want another driver, they should simply provide their own database config

tests/Fixtures/Migrations/CreateAuthorTable.php
brendt: Let's keep it for now, I'll probably look into refactoring it after the PR is merged :)

brendt (changes requested):

src/Tempest/Database/DatabaseDriver.php
Treggats: The `createQueryStatement(..)` needs a instance of the database driver. I'm not sure how to pass that through the enum.

brendt:
💪 I think I'm good to merge this! Anything else you wanted to tackle before I do?

Treggats:
> 💪 I think I'm good to merge this! Anything else you wanted to tackle before I do?

Nope, think I'm good. :)

src/Tempest/Database/Migrations/MigrationManager.php
```diff
@@ -27,21 +28,21 @@ public function up(): void
...
         }
 
         $existingMigrations = array_map(
-            fn (Migration $migration) => $migration->name,
+            static fn (Migration $migration) => $migration->name,
             $existingMigrations,
         );
 
-        foreach ($this->databaseConfig->migrations as $migrationClassName) {
+        foreach ($this->databaseConfig->getMigrations() as $migrationClassName) {
```
Treggats, line 41: normal sort, would otherwise cause errors when dealing with foreign keys #297 
```suggestion
        $migrations = $this->databaseConfig->getMigrations();
        ksort($migrations);
        foreach ($migrations as $migrationClassName) {
```

Treggats:
working on a additional test for the output given on the console. But that doesn't block this pull request.

So it's ready to merge 🎉

brendt:
Let's merge!

PR #310, opened 2024-08-09
[proposal] introduce Rector to keep code up to date

Treggats:
I propose to introduce [Rector](https://getrector.com/) to keep the codebase up to date and clean.

**Global changes**
- return types
- updated conditionals to be more specific
- early returns
- class name
- updated naming
- formatting


.php-cs-fixer.dist.php
```diff
@@ -14,7 +14,6 @@
 
 return (new PhpCsFixer\Config())
     ->setCacheFile('.cache/fixer/cs-fixer.cache')
-    ->setParallelConfig(ParallelConfigFactory::detect())
```
Treggats, line 17: this gave weird issues, so removed it

rector.php
```diff
@@ -0,0 +1,88 @@
...
+        ReadOnlyPropertyRector::class,
+        RemoveNullPropertyInitializationRector::class,
+        RemoveUnreachableStatementRector::class,
+        AddSensitiveParameterAttributeRector::class,
+        RemoveUnusedPublicMethodParameterRector::class,
+        RestoreDefaultNullToNullableTypePropertyRector::class,
+        ReturnNeverTypeRector::class,
+        StaticCallOnNonStaticToInstanceCallRector::class,
+        ClosureReturnTypeRector::class,
+        EncapsedStringsToSprintfRector::class
```
Treggats, line 71: skipping because it added `sprintf` throughout the codebase cluttering it

brendt:
The only thing I'm still uncertain about are return types for short closures. @aidan-casey what's your opinion?

aidan-casey:
First of all, I appreciate your work on this, @Treggats. Thank you!

When we first setup the repo, we did setup PHP-CS-Fixer to enforce our standards. My biggest concern here is having two tools fighting for the same role. Can you speak a little bit more to the additional functionality Rector is giving us here in comparison?

> The only thing I'm still uncertain about are return types for short closures. @aidan-casey what's your opinion?

I like it from a typing standpoint, I hate it from a syntax standpoint. I have mixed feelings and would probably favor syntax over typing here personally. That said, I think the answer probably depends on what our target is with PHPStan in the long run.

Treggats:
Did a rerun of Rector with the  `AddArrowFunctionReturnTypeRector` disabled.  Rector shouldn't have added the return types on short closures.

rector.php
```diff
@@ -0,0 +1,91 @@
...
+        AddSensitiveParameterAttributeRector::class,
+        RemoveUnusedPublicMethodParameterRector::class,
+        RestoreDefaultNullToNullableTypePropertyRector::class,
+        ReturnNeverTypeRector::class,
+        StaticCallOnNonStaticToInstanceCallRector::class,
+        ClosureReturnTypeRector::class,
+        EncapsedStringsToSprintfRector::class,
+        AddArrowFunctionReturnTypeRector::class,
+    ])
+    ->withSkipPath('src/Tempest/Framework/Exceptions/')
```
Treggats, line 75: Due to the weird structure of the controller and file in this directory Rector removes 90% of the controller.
Ignoring the directory so we should not see weird errors or phpunit failures

rector.php
```diff
@@ -0,0 +1,78 @@
...
+        RemoveNullPropertyInitializationRector::class,
+        RemoveUnreachableStatementRector::class,
+        AddSensitiveParameterAttributeRector::class,
+        RemoveUnusedPublicMethodParameterRector::class,
+        RestoreDefaultNullToNullableTypePropertyRector::class,
+        ReturnNeverTypeRector::class,
+        StaticCallOnNonStaticToInstanceCallRector::class,
+        ClosureReturnTypeRector::class,
+        EncapsedStringsToSprintfRector::class,
+        AddArrowFunctionReturnTypeRector::class,
```
Treggats, line 63: no return types for short closures please

tests/Unit/Container/ContainerTest.php
```diff
@@ -204,15 +205,16 @@ public function test_circular_with_initializer_log(): void
     {
         $container = new GenericContainer();
         $container->addInitializer(CircularWithInitializerBInitializer::class);
+        $this->assertContains(CircularWithInitializerBInitializer::class, $container->getInitializers());
```
Treggats, line 208: added this to at least have one assertion. Is this test outdated?

brendt:
> You mention PHPStan, how does that fit into this discussion?

I _think_ Aidan meant phpcs fixer.

To pitch: I feel like both tools serve other goals, but there is some overlap between them. I don't really mind having both of them myself. It's all about automations anyway, so I don't really care how many tools are run. If phpcs fixer can be completely replaced by Rector, then let's do that. But if it can't then let's keep what we have and simply add Rector on top of it 👍

Treggats:
> If phpcs fixer can be completely replaced by Rector

As far as my experience goes with both tools, I don't think that's the case.

For example, Rector can unwrap an else statement. But there won't be a empty line before the `return` statement. We need the fixer for that.

The order as I see it would be to run Rector first and then the fixer.

brendt:
Cool. Let's merge this and see where it goes? Quick question: can also take care of setting up the github action and updating the `composer qa` command?

Treggats:
> can also take care of setting up the github action and updating the composer qa command?

sure no problem.

brendt:
Those changes will go in a separate PR. Already merging this

Issue #273, opened 2024-05-23
Discovery on Windows doesn't work


brendt:
https://github.com/tempestphp/tempest-app/actions/runs/9203485022/job/25315100490

My guess is that it has to do with reading files, a `glob` or `is_file` or `file_exists` function somewhere within the bootstrappers is the most likely culprit. Maybe something with filesystem permissions?

kickofitall:
I stumbled upon the error myself and was able to narrow it down to the following line in [DiscoveryBootstrap.php](https://github.com/tempestphp/tempest-framework/blob/b4393ab89f13ad66b8607789b41babbfe88f1313/src/Tempest/Framework/Bootstraps/DiscoveryBootstrap.php#L41):

**not working**
`$directories = new RecursiveDirectoryIterator($discoveryLocation->path);`

**working**
`$directories = new RecursiveDirectoryIterator($discoveryLocation->path, RecursiveDirectoryIterator::UNIX_PATHS);`

Later in the code, `$discoveryLocation->path` ends with a "/".
`$file->getPathname()` uses "\\" from the same position until the end of the string.
This means that `str_replace` cannot replace the path [here](https://github.com/tempestphp/tempest-framework/blob/b4393ab89f13ad66b8607789b41babbfe88f1313/src/Tempest/Framework/Bootstraps/DiscoveryBootstrap.php#L59).

---

Just to mention, there is another flag `RecursiveDirectoryIterator::SKIP_DOTS.`
Maybe that would allow you to [omit the check in line 48](https://github.com/tempestphp/tempest-framework/blob/b4393ab89f13ad66b8607789b41babbfe88f1313/src/Tempest/Framework/Bootstraps/DiscoveryBootstrap.php#L48).

brendt:
Ah, awesome! Thanks for looking into this. Do you wanna submit a PR? I can do it, but if you'd like the contribution, I'm happy to merge it :) Let me know!

kickofitall:
Sure, I'd like to contribute.

There is another small code fragment that doesn't work under Windows.
I'd also include that in the PR.

brendt:
Ah, cool! I actually just pushed a WIP commit to main to test out the things you were talking about :D https://github.com/tempestphp/tempest-framework/commit/e08549753aecd634a156b540ab938d77346ede23

Feel free to try it out, send a PR if anything pops up :)

kickofitall:
Great! I'll take a look at it.

brendt:
[...]

PR #319, opened 2024-08-12
Adds Windows to the test suite

aidan-casey:
We should have Windows in the test suite so we can ensure there are no issues there (see #273).

### DIFF
--- a/src/Tempest/Support/PathHelper.php
+++ b/src/Tempest/Support/PathHelper.php
@@ -11,8 +11,8 @@ final readonly class PathHelper
         $path = implode('/', $parts);
 
         return str_replace(
-            ['//', '\\', '\\\\'],
-            ['/', '/', '/'],
+            ['//', '\\'],
+            DIRECTORY_SEPARATOR,
             $path,
         );
     }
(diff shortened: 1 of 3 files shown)


aidan-casey:
@brendt - It does look like we have some issues to address on Windows. I have a Windows box I can take a look at these on, but we should probably make sure this is on the roadmap.

src/Tempest/Support/PathHelper.php
```diff
@@ -11,8 +11,8 @@ public static function make(string ...$parts): string
         $path = implode('/', $parts);
 
         return str_replace(
-            ['//', '\\', '\\\\'],
-            ['/', '/', '/'],
+            ['//', '\\'],
+            DIRECTORY_SEPARATOR,
```
brendt, line 15: Sometimes I feel like such a noob :p

brendt:
@aidan-casey Sounds good, here's the issue regarding discovery: https://github.com/tempestphp/tempest-framework/issues/273

Feel free to create others and add them to the 1.0-alpha1 milestone. Did you want to merge this one with the failing actions? I'm fine if we can fix it short-term, but I don't want to have failing actions for the coming months

PR #320, opened 2024-08-12
Adds PHP 8.4 to tests

aidan-casey:


aidan-casey:
Looks like we need to wait on some dependencies for this.

brendt:
It'll be interesting to see how long it takes for dependencies to catch up

PR #321, opened 2024-08-12
[security] allow only known classes to unserialize

Treggats:
This PR allows us to be more strict with the `unserialize(..)` behaviour.

If a different payload is given than the one we allow, then the application will throw an exception.

This was already done in #306 for the Migration cache, and I think that it's good to be security focused when dealing with outside input. Wether human based, or computational based.

### DIFF
--- a/src/Tempest/Console/Scheduler/GenericScheduler.php
+++ b/src/Tempest/Console/Scheduler/GenericScheduler.php
@@ -76,7 +76,7 @@ final readonly class GenericScheduler implements Scheduler
             return [];
         }
 
-        return unserialize(file_get_contents(self::CACHE_PATH));
+        return unserialize(file_get_contents(self::CACHE_PATH), ['allowed_classes' => false]);
     }
 
     /**
--- a/src/Tempest/Container/InitializerDiscovery.php
+++ b/src/Tempest/Container/InitializerDiscovery.php
@@ -8,12 +8,15 @@ use Tempest\Discovery\Discovery;
 use Tempest\Discovery\HandlesDiscoveryCache;
 use Tempest\Support\Reflection\ClassReflector;
 
+/**
+ * @property GenericContainer $container
+ */
 final readonly class InitializerDiscovery implements Discovery
 {
     use HandlesDiscoveryCache;
 
     public function __construct(
-        private Container&GenericContainer $container,
+        private Container $container,
     ) {
     }
 
@@ -38,7 +41,10 @@ final readonly class InitializerDiscovery implements Discovery
 
     public function restoreCachePayload(Container $container, string $payload): void
     {
-        $data = unserialize($payload);
+        $data = unserialize($payload, ['allowed_classes' => [
+            Initializer::class,
+            DynamicInitializer::class,
+        ]]);
 
         $this->container->setInitializers($data['initializers'] ?? []);
         $this->container->setDynamicInitializers($data['dynamic_initializers'] ?? []);
--- a/src/Tempest/Discovery/DiscoveryDiscovery.php
+++ b/src/Tempest/Discovery/DiscoveryDiscovery.php
@@ -4,18 +4,9 @@ declare(strict_types=1);
 
 namespace Tempest\Discovery;
 
-use Tempest\CommandBus\CommandBusDiscovery;
-use Tempest\Console\Discovery\ConsoleCommandDiscovery;
-use Tempest\Console\Discovery\ScheduleDiscovery;
 use Tempest\Container\Container;
-use Tempest\Container\InitializerDiscovery;
-use Tempest\Database\MigrationDiscovery;
-use Tempest\EventBus\EventBusDiscovery;
 use Tempest\Framework\Application\AppConfig;
-use Tempest\Http\RouteDiscovery;
-use Tempest\Mapper\MapperDiscovery;
 use Tempest\Support\Reflection\ClassReflector;
-use Tempest\View\ViewComponentDiscovery;
 
 final readonly class DiscoveryDiscovery implements Discovery
 {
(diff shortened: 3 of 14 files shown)


src/Tempest/Console/Scheduler/GenericScheduler.php
```diff
@@ -76,7 +76,7 @@ private function getPreviousRuns(): array
             return [];
         }
 
-        return unserialize(file_get_contents(self::CACHE_PATH));
+        return unserialize(file_get_contents(self::CACHE_PATH), ['allowed_classes' => false]);
```
Treggats, line 79: no classes are allowed, as there is only a array item in the payload

src/Tempest/Container/InitializerDiscovery.php
```diff
@@ -8,12 +8,15 @@
...
+/**
+ * @property GenericContainer $container
+ */
 final readonly class InitializerDiscovery implements Discovery
 {
     use HandlesDiscoveryCache;
 
     public function __construct(
-        private Container&GenericContainer $container,
+        private Container $container,
```
Treggats, line 19: simplified to only the interface, and type hint to help the IDE see what methods are actually available.

[rector]

src/Tempest/Discovery/DiscoveryDiscovery.php
```diff
@@ -42,7 +51,20 @@ public function storeCache(): void
 
     public function restoreCache(Container $container): void
     {
-        $discoveryClasses = unserialize(file_get_contents(self::CACHE_PATH));
+        $discoveryClasses = unserialize(file_get_contents(self::CACHE_PATH), [
+            'allowed_classes' => [
+                CommandBusDiscovery::class,
```
Treggats, line 56: I believe these are all the discovery classes there are.

src/Tempest/Framework/Bootstraps/DiscoveryBootstrap.php
```diff
@@ -40,7 +40,7 @@ public function boot(): void
             }
 
             foreach ($this->appConfig->discoveryLocations as $discoveryLocation) {
-                $directories = new RecursiveDirectoryIterator($discoveryLocation->path, FilesystemIterator::UNIX_PATHS);
+                $directories = new RecursiveDirectoryIterator($discoveryLocation->path, FilesystemIterator::UNIX_PATHS | FilesystemIterator::SKIP_DOTS);
```
Treggats, line 43: I suppose it makes sense to skip dotted directories

[rector]

src/Tempest/Http/Session/Managers/FileSessionManager.php
```diff
@@ -63,11 +69,7 @@ public function isValid(SessionId $id): bool
 
         $validUntil = $session->createdAt->getTimestamp() + $this->sessionConfig->expirationInSeconds;
 
-        if ($validUntil - $this->clock->time() <= 0) {
-            return false;
-        }
-
-        return true;
+        return $validUntil - $this->clock->time() > 0;
```
Treggats, line 72: [rector]

src/Tempest/Http/Session/Managers/FileSessionManager.php
```diff
@@ -82,7 +84,15 @@ private function resolve(SessionId $id): ?Session
         try {
             $content = @file_get_contents($path);
 
-            return unserialize($content);
+            return unserialize($content, ['allowed_classes' => [
+                SessionId::class,
```
Treggats, line 88: these are the classes the tests revealed. but we probably need to add every rule to this?

I did not go as far (yet) as this is what I initially found

src/Tempest/Http/Session/Managers/FileSessionManager.php
```diff
@@ -114,8 +124,8 @@ private function persist(SessionId $id, ?array $data = null): Session
         $path = $this->getPath($id);
         $dir = dirname($path);
 
-        if (! is_dir($dir)) {
-            mkdir($dir, recursive: true);
+        if (! is_dir($dir) && ! mkdir($dir, recursive: true) && ! is_dir($dir)) {
+            throw new RuntimeException(sprintf('Directory "%s" was not created', $dir));
```
Treggats, line 128: Not sure if this was Rector or the PHP Inspections (EA extended) from my IDE.

But it adds an extra check if the directory can actually be created.

src/Tempest/Support/Reflection/TypeReflector.php
```diff
@@ -80,7 +80,7 @@ public function isBuiltIn(): bool
 
     public function isClass(): bool
     {
-        return class_exists($this->getName());
+        return class_exists($this->definition);
```
Treggats, line 83: we can use the property that the method returned as it's in the same class

src/Tempest/Discovery/DiscoveryDiscovery.php
brendt: Hmm yeah this will be annoying, since packages can provide their own discovery classes. This special case will need another solution.

src/Tempest/Http/Session/Managers/FileSessionManager.php
brendt: Hmm, technically it could be anything. I don't think we can add filtering here.

src/Tempest/Http/Session/Managers/FileSessionManager.php
brendt: Hmm I don't like these complex `if` statements. Let's revert this one. I plan on adding a proper path helper at some point which will take care of this.

brendt (changes requested):

src/Tempest/Discovery/DiscoveryDiscovery.php
```diff
@@ -42,7 +42,9 @@ public function storeCache(): void
 
     public function restoreCache(Container $container): void
     {
-        $discoveryClasses = unserialize(file_get_contents(self::CACHE_PATH));
+        $discoveryClasses = unserialize(file_get_contents(self::CACHE_PATH), [
```
Treggats, line 45: because the payload can be dynamic we can't really provide a list here. We'll allow all classes

brendt:
Thanks!

PR #324, opened 2024-08-13
[proposal] database factory

Treggats:
@aidan-casey does this match your idea?

### DIFF
--- a/src/Tempest/Database/DatabaseDriverFactory.php
+++ b/src/Tempest/Database/DatabaseDriverFactory.php
@@ -15,9 +15,9 @@ final class DatabaseDriverFactory
         $instance = new self();
 
         return match ($dialect) {
-            DatabaseDialect::MYSQL => new MySqlDriver(...$instance->formatOptions($dialect, $options)),
-            DatabaseDialect::POSTGRESQL => new PostgreSqlDriver(...$instance->formatOptions($dialect, $options)),
-            DatabaseDialect::SQLITE => new SQLiteDriver(...$instance->formatOptions($dialect, $options)),
+            DatabaseDialect::MYSQL => new MySqlDriver(...array_filter($instance->formatOptions($dialect, $options))),
+            DatabaseDialect::POSTGRESQL => new PostgreSqlDriver(...array_filter($instance->formatOptions($dialect, $options))),
+            DatabaseDialect::SQLITE => new SQLiteDriver(...array_filter($instance->formatOptions($dialect, $options))),
         };
     }
 
--- a/src/Tempest/Database/DatabaseDriverFactory.php
+++ b/src/Tempest/Database/DatabaseDriverFactory.php
@@ -8,19 +8,17 @@ use Tempest\Database\Drivers\MySqlDriver;
 use Tempest\Database\Drivers\PostgreSqlDriver;
 use Tempest\Database\Drivers\SQLiteDriver;
 
-final class DatabaseFactory
+final class DatabaseDriverFactory
 {
-    public static function make(DatabaseDialect $dialect, array $options): DatabaseConfig
+    public static function make(DatabaseDialect $dialect, array $options): DatabaseDriver
     {
         $instance = new self();
 
-        return new DatabaseConfig(
-            match ($dialect) {
-                DatabaseDialect::MYSQL => new MySqlDriver(...$instance->formatOptions($dialect, $options)),
-                DatabaseDialect::POSTGRESQL => new PostgreSqlDriver(...$instance->formatOptions($dialect, $options)),
-                DatabaseDialect::SQLITE => new SQLiteDriver(...$instance->formatOptions($dialect, $options)),
-            },
-        );
+        return match ($dialect) {
+            DatabaseDialect::MYSQL => new MySqlDriver(...$instance->formatOptions($dialect, $options)),
+            DatabaseDialect::POSTGRESQL => new PostgreSqlDriver(...$instance->formatOptions($dialect, $options)),
+            DatabaseDialect::SQLITE => new SQLiteDriver(...$instance->formatOptions($dialect, $options)),
+        };
     }
 
     private function formatOptions(DatabaseDialect $dialect, array $options): array
--- a/src/Tempest/Database/DatabaseDriverFactory.php
+++ b/src/Tempest/Database/DatabaseDriverFactory.php
@@ -26,13 +26,15 @@ final class DatabaseDriverFactory
         return match ($dialect) {
             DatabaseDialect::MYSQL,
             DatabaseDialect::POSTGRESQL => [
-                'host' => $options['host'],
-                'port' => $options['port'],
-                'username' => $options['username'],
-                'password' => $options['password'],
-                'database' => $options['database'],
+                'host' => $options['host'] ?? null,
+                'port' => $options['port'] ?? null,
+                'username' => $options['username'] ?? null,
+                'password' => $options['password'] ?? null,
+                'database' => $options['database'] ?? null,
             ],
-            DatabaseDialect::SQLITE => ['path' => $options['path']],
+            DatabaseDialect::SQLITE => [
+                'path' => $options['path'] ?? null,
+            ],
         };
     }
 }
--- a/src/Tempest/Database/DatabaseFactory.php
+++ b/src/Tempest/Database/DatabaseFactory.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use Tempest\Database\Drivers\MySqlDriver;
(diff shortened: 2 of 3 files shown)


tests/Unit/Database/DatabaseDriverTest.php
```diff
@@ -37,6 +39,15 @@ public static function provide_database_drivers(): Generator
             null,
         ];
 
+        yield 'sqlite factor' => [
```
Treggats, line 42: matches with the 'regular' SQLite driver test

src/Tempest/Database/DatabaseFactory.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use Tempest\Database\Drivers\MySqlDriver;
+use Tempest\Database\Drivers\PostgreSqlDriver;
+use Tempest\Database\Drivers\SQLiteDriver;
+
+final class DatabaseFactory
```
aidan-casey, line 11: Rename from `DatabaseFactory` to `DatabaseDriverFactory`.

src/Tempest/Database/DatabaseFactory.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public static function make(DatabaseDialect $dialect, array $options): DatabaseConfig
```
aidan-casey, line 13: Should return `DatabaseDriver`.

src/Tempest/Database/DatabaseFactory.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $instance = new self();
+
+        return new DatabaseConfig(
+            match ($dialect) {
```
aidan-casey, line 18: You could run an `array_filter` on the options and pass those to the class? That should leave the default if not added.

aidan-casey (commented):
Yeah, this is generally speaking what I was thinking of, though the driver factory should be called from the config file instead of returning the config.

I'm torn on validation/helper methods here. Curious to get @brendt's thoughts as well.

src/Tempest/Database/DatabaseFactory.php
Treggats: I just read your message in Discord again and was like wait a minute 😋

src/Tempest/Database/DatabaseDriverFactory.php
```diff
@@ -0,0 +1,40 @@
...
+            DatabaseDialect::SQLITE => new SQLiteDriver(...array_filter($instance->formatOptions($dialect, $options))),
+        };
+    }
+
+    private function formatOptions(DatabaseDialect $dialect, array $options): array
+    {
+        return match ($dialect) {
+            DatabaseDialect::MYSQL,
+            DatabaseDialect::POSTGRESQL => [
+                'host' => $options['host'] ?? null,
```
Treggats, line 29: default to null when the key is not provided

src/Tempest/Database/DatabaseDriverFactory.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Tempest\Database;
+
+use Tempest\Database\Drivers\MySqlDriver;
+use Tempest\Database\Drivers\PostgreSqlDriver;
+use Tempest\Database\Drivers\SQLiteDriver;
+
+final class DatabaseDriverFactory
+{
+    public static function make(DatabaseDialect $dialect, array $options): DatabaseDriver
```
Treggats, line 13: was thinking of adding a default to the `$options` parameter but argued against it as we want the user to pass options

Treggats:
small diff to use this factory in the test suite.

<details>
<summary>Changes</summary>

```diff
Index: tests/Fixtures/Config/database.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/Fixtures/Config/database.php b/tests/Fixtures/Config/database.php
--- a/tests/Fixtures/Config/database.php	(revision Staged)
+++ b/tests/Fixtures/Config/database.php	(date 1723571627337)
@@ -3,10 +3,15 @@
 declare(strict_types=1);
 
 use Tempest\Database\DatabaseConfig;
-use Tempest\Database\Drivers\SQLiteDriver;
+use Tempest\Database\DatabaseDialect;
+use Tempest\Database\DatabaseDriverFactory;
 
-return new DatabaseConfig(
-    driver: new SQLiteDriver(
-        path: __DIR__ . '/../database.sqlite',
-    ),
-);
+$driver = DatabaseDriverFactory::make(DatabaseDialect::from(env('DB_DRIVER')), [
+    'host' => env('DB_HOST', 'localhost'),
+    'port' => env('DB_PORT', 3306),
+    'username' => env('DB_USERNAME', 'root'),
+    'password' => env('DB_PASSWORD', ''),
+    'database' => env('DB_DATABASE', 'test'),
+]);
+
+return new DatabaseConfig($driver);
Index: phpunit.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/phpunit.xml b/phpunit.xml
--- a/phpunit.xml	(revision Staged)
+++ b/phpunit.xml	(date 1723571148132)
@@ -18,6 +18,14 @@
       <directory suffix="Test.php">./tests/Unit</directory>
     </testsuite>
   </testsuites>
+    <php>
+        <env name="ENVIRONMENT" value="testing"/>
+        <env name="DB_DRIVER" value="mysql"/>
+        <env name="DB_USERNAME" value="root"/>
+        <env name="DB_PASSWORD" value=""/>
+        <env name="DB_PORT" value="3307"/>
+        <env name="DB_DATABASE" value="tempest"/>
+    </php>
   <coverage />
   <source>
     <include>
```

</details>

brendt:
Folks, aren't we overcomplicating stuff? I feel like all of this comes from the problem of having a `Driver` object in config, which we can't directly adjust from env variables. So what if we change things up? 

This would be the config file:

```php
final class DatabaseConfig
{
    public function __construct(
        #[SensitiveParameter]
        public string $database = __DIR__ . '/database.sqlite',
        #[SensitiveParameter]
        public ?string $host = null,
        #[SensitiveParameter]
        public ?string $port = null
        #[SensitiveParameter]
        public ?string $username = null,
        #[SensitiveParameter]
        public ?string $password = null,
    ) {
    }
```

And this would be the driver's initializer:

```php
    public function initialize(Container $container): DatabaseDriver
    {
        $config = $container->get(DatabaseConfig::class);

        return match (true) {
            str_contains($config->database, '.sqlite') => new SQLiteDriver($config->database),
            $config->port === 3306 => new MySQLDriver($config->database),
            $config->port === 5432 => new PostgreSqlDriver($config->database),
        };
    }
```

Granted, we probably need better ways of detecting which driver to use; I don't oppose the idea of adding the `DatabaseDialect` enum to the database config as well. 

Anyway, I guess my biggest point is: we don't need a factory, we already have the initializer, and we can finetune the config if needed

aidan-casey:
@brendt I'm in agreement a factory is not ideal. Having the keys be directly in the config is not the answer though. We _have_ to be able to support multiple connections.

brendt:
> We have to be able to support multiple connections.

Why?

Also, I thought about it some more, and I don't agree on my previous comment. We should never adjust the framework in favour of CI. I was playing around with Tempest yesterday, and the driver config is super convenient from a user's point of view, so I wanna keep it.

To solve the CI problem, I'd simply have three config files somewhere in tests/fixtures, and copy over the right one per run: `cp Tests/fixtures/database.mysql.php App/Config` something like that?

That is, of course, unless I'm missing something with the multiple drivers Aidan mentioned

Treggats:
> We should never adjust the framework in favour of CI. 

I agree, or testing in general.

> To solve the CI problem, I'd simply have three config files somewhere in tests/fixtures, and copy over the right one per run: `cp Tests/fixtures/database.mysql.php App/Config` something like that?

I've mentioned this somewhere, can't remember where. This is something that's easily done.

However, we'd need to fix the testsuite that it can use the MySQL (or PostgresSQL) driver. Right now the database changes are not rolled back after a test.

brendt:
I think we had this conversation already a year ago, no? 🤔

Here are my thoughts:

- I'm not against supporting multiple connections, but it needs more thought, and is not needed for 1.0
- Either way, I think the default approach should keep working more or less the same

I propose the following:

- Let's create an issue to brainstorm multiple connections, and have it target alpha2, or post-1.0 if we can do it without breaking changes
- For the whole database testing setup, let's just keep three config files that we copy within a github action, no additional factories needed

brendt:
Connection issue: https://github.com/tempestphp/tempest-framework/issues/339

PR #335, opened 2024-08-15
removed redundant  conditionals

binkapS:
added src/Tempest/Framework/database.sqlite to .gitignore
ignored vscode configuration
made some changes to some syntax changes


brendt:
It feels like this PR does a lot of things all over the place. I'd prefer a PR to focus on one thing specifically, and also explaining what it's trying to do :)

Issue #340, opened 2024-08-16
database rollback support for the testsuite


Treggats:
**Description**
We've added support for multiple database dialects. However, this CI still only runs on SQLite.
Which is fine at first, but if we want to fully support multiple database dialects then we need to use them in the CI as well.

Setting up the CI to use those can be done pretty easily, but the testsuite at the moment isn't ready for this.

**The issue**
When running tests with SQLite I believe that the database is fully destroyed and created again. This is, however, not the case for the other dialects.
With which we run into test failures, because the table it is trying to create already exists from a previous test.

brendt:
Why don't we spin up an individual action using a matrix? That way we always start from a clean slate?

Treggats:
The different databases aren't the issue. Use the MySQL driver for the tests are run the testsuite. That bests demonstrates the issue

aidan-casey:
With [`FrameworkIntegrationTestCase`](https://github.com/tempestphp/tempest-framework/blob/main/tests/Integration/FrameworkIntegrationTestCase.php), the SQLite database is deleted and re-copied _every_ test run. Nothing is currently happening with other dialects like Laravel supports with the `RefreshDatasebase` trait.

I think our default stance should be to _not_ refresh the database on every test case, as that will impact performance. Obviously we do need to support this though, since you will need to do it for tests around migrations, etc.

This probably goes hand-in-hand with a refactor of the test helpers.

Treggats:
> I think our default stance should be to not refresh the database on every test case

I agree, only on the parts where the tables are being recreated. what data is stored inside them is of less concern I gather

brendt:
Gotcha! Ok let's take this into account for alpha2 :)

brendt:
I'm gonna look into this

brendt:
Rollback in itself works, although there are still several errors on MySQL and Postgres. Gonna make a new issue for it

PR #343, opened 2024-08-16
preparation for multiple database CI

Treggats:
This PR prepares the CI to be able run with different databases.

At the moment only sqlite is enabled, as the other databases have issues with using database tables. As detailed in #340


tests/Integration/FrameworkIntegrationTestCase.php
```diff
@@ -44,6 +45,11 @@ protected function setUp(): void
         @unlink($databasePath);
         copy($cleanDatabasePath, $databasePath);
 
+        $filename = __DIR__ . '/../Fixtures/Config/database.php';
```
Treggats, line 48: make sure the tests are run with a database configured

tests/Fixtures/Config/database.mysql.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+use Tempest\Database\DatabaseConfig;
+use Tempest\Database\Drivers\MySqlDriver;
+
+return new DatabaseConfig(
```
Treggats, line 8: using default config

.github/workflows/quality-assurance.yml
```diff
@@ -79,13 +87,18 @@ jobs:
...
+
       - name: Install dependencies
         run: |
           composer global require php-coveralls/php-coveralls
           composer update --${{ matrix.stability }} --prefer-dist --no-interaction
+
       - name: List Installed Dependencies
         run: composer show -D
 
+      - name: Set database config - ${{ matrix.database }}
```
Treggats, line 99: copy the config for the database dialect from the matrix

.github/workflows/quality-assurance.yml
```diff
@@ -58,11 +58,19 @@ jobs:
...
+        php:
+          - 8.3
+        database:
+          - sqlite
+#          - mysql
+#          - postgresql
+        stability:
+          - prefer-stable
+
+    name: Run Tests - P${{ matrix.php }} - ${{ matrix.database }} - ${{ matrix.stability }} - ${{ matrix.os }}
```
Treggats, line 73: update the name with the database dialect

.github/workflows/quality-assurance.yml
```diff
@@ -58,11 +58,19 @@ jobs:
...
-
-    name: Run Tests - P${{ matrix.php }} - ${{ matrix.stability }} - ${{ matrix.os }}
+        os:
+          - ubuntu-latest
+          - windows-latest
+        php:
+          - 8.3
+        database:
+          - sqlite
+#          - mysql
```
Treggats, line 68: disable for now

.github/workflows/quality-assurance.yml
```diff
@@ -58,11 +58,19 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        php: [8.3]
-        stability: [prefer-stable]
-
-    name: Run Tests - P${{ matrix.php }} - ${{ matrix.stability }} - ${{ matrix.os }}
+        os:
+          - ubuntu-latest
+          - windows-latest
+        php:
```
Treggats, line 64: should we already include [8.4](https://github.com/marketplace/actions/setup-php-action#tada-php-support)? (nightly build)

tests/Integration/FrameworkIntegrationTestCase.php
```diff
@@ -44,6 +45,11 @@ protected function setUp(): void
...
+        if (! file_exists($filename)) {
```
brendt, line 49: Do we want an additional filesystem I/O operation on every test? Wouldn't it fail somewhere else if we forgot to set up te database?

brendt:
Only one question, the rest seems fine :)

tests/Integration/FrameworkIntegrationTestCase.php
Treggats: We could just let it fail hard with stack trace and all. But I didn't think that was very user/developer friendly, and it would also be harder to see why the tests wouldn't run.

I realise that this also throws an exception, however the stack trace would be significantly smaller and there's only one point of origin. You wouldn't need to follow the trace to where it originated. Which again is not impossible, but it would take longer. Or at least that was my reasoning behind adding this.

We could also place it higher, above the unlinking or even before the `appConfig`.

Treggats: additionally we could also add the following. to reduce file operations with other database drivers.

```php
$driver = $this->container->get(DatabaseDriver::class);
if ($driver->dialect()->value === DatabaseDialect::SQLITE->value) {
    $databasePath = __DIR__ . '/../Fixtures/database.sqlite';
    $cleanDatabasePath = __DIR__ . '/../Fixtures/database-clean.sqlite';

    @unlink(DiscoveryDiscovery::CACHE_PATH);
    @unlink($databasePath);
    copy($cleanDatabasePath, $databasePath);
}
```

brendt: Yeah I like that :)

tests/Integration/FrameworkIntegrationTestCase.php
```diff
@@ -37,12 +45,16 @@ protected function setUp(): void
         );
 
         parent::setUp();
-        $databasePath = __DIR__ . '/../Fixtures/database.sqlite';
-        $cleanDatabasePath = __DIR__ . '/../Fixtures/database-clean.sqlite';
 
-        @unlink(DiscoveryDiscovery::CACHE_PATH);
-        @unlink($databasePath);
-        copy($cleanDatabasePath, $databasePath);
+        $driver = $this->container->get(DatabaseDriver::class);
```
Treggats, line 49: only do the file operations specific for SQLite, if that driver is configured

brendt:
Looks good, thanks!

PR #367, opened 2024-09-06
Kernel refactor

brendt:


src/Tempest/Http/HttpApplication.php
```diff
@@ -16,33 +17,29 @@
...
-        private AppConfig $appConfig,
-    ) {
-    }
+    public function __construct(private Container $container) {}
 
-    public static function boot(string $root, ?AppConfig $appConfig = null): self
+    /** @param \Tempest\Core\DiscoveryLocation[] $discoveryLocations */
+    public static function boot(
+        string $root,
+        array $discoveryLocations = [],
```
aidan-casey, line 25: In theory, this can move to a discovery configuration file. It's "framework dependency" so it's config will be in the application skeleton or autowired with defaults.

src/Tempest/Http/HttpApplication.php
```diff
@@ -16,33 +17,29 @@
...
+    ): self
     {
-        $container = Tempest::boot($root, $appConfig);
-
-        // TODO: check if we need this
-        $appConfig = $container->get(AppConfig::class);
+        $container = Tempest::boot($root, $discoveryLocations);
```
aidan-casey, line 28: We should probably allow the container to be passed in. If we are going to override core services (like the event bus) we're going to have to have the ability to control what we pass in.

src/Tempest/Http/HttpApplication.php
```diff
@@ -16,33 +17,29 @@
...
+
         // Application-specific setup
         $logConfig = $container->get(LogConfig::class);
-        $logConfig->debugLogPath = PathHelper::make($appConfig->root, '/log/debug.log');
+        $logConfig->debugLogPath = PathHelper::make($root, '/log/debug.log');
         $logConfig->serverLogPath = env('SERVER_LOG');
-        $logConfig->channels[] = new AppendLogChannel(PathHelper::make($appConfig->root, '/log/tempest.log'));
+        $logConfig->channels[] = new AppendLogChannel(PathHelper::make($root, '/log/tempest.log'));
 
         return $application;
```
aidan-casey, line 44: This feels like setup that could be in the kernel?

src/Tempest/Http/HttpApplication.php
brendt: I agree, it's also more than $discoveryLocations, it's locations, classes, and whether cache is enabled or not

src/Tempest/Http/HttpApplication.php
brendt: In that case people can construct the kernel themselves I think? This `Tempest::boot` method is only here for "the default flow"

src/Tempest/Http/HttpApplication.php
brendt: I'm not sure… wouldn't logging be application specific? OTOH, I'm fine moving it to kernel first and see whether it actually becomes necessary to complicate further

src/Tempest/Console/ConsoleApplication.php
```diff
@@ -25,11 +26,13 @@ public function __construct(
     ) {
     }
 
-    public static function boot(string $name = 'Tempest', ?AppConfig $appConfig = null): self
-    {
-        $root = $appConfig->root ?? getcwd();
-        $appConfig ??= new AppConfig(root: $root);
-        $container = Tempest::boot($root, $appConfig);
+    /** @param \Tempest\Core\DiscoveryLocation[] $discoveryLocations */
+    public static function boot(
```
brendt, line 30: TODO: we cannot enable/disable discovery cache with .env now anymore. We need to find a way to make that possible from with the kernel

src/Tempest/Core/Kernel/LoadDiscoveryLocations.php
```diff
@@ -2,36 +2,37 @@
...
+namespace Tempest\Core\Kernel;
 
-use Tempest\Core\AppConfig;
+use Tempest\Core\DiscoveryException;
 use Tempest\Core\DiscoveryLocation;
+use Tempest\Core\Kernel;
 use Tempest\Support\PathHelper;
 
-final readonly class DiscoveryLocationBootstrap implements Bootstrap
+final readonly class LoadDiscoveryLocations
```
brendt, line 12: TODO: mark these internal action classes as internal, they are just here so that Kernel isn't too crowded

brendt: Or we refactor even further towards a more decoupled system

src/Tempest/Core/Tempest.php
```diff
@@ -30,19 +33,20 @@ public static function boot(?string $root = null, ?AppConfig $appConfig = null):
         // Env
         $dotenv = Dotenv::createUnsafeImmutable($root);
```
brendt, line 34: We should move environment loading to the kernel now

src/Tempest/EventBus/EventBusDiscovery.php
```diff
@@ -28,21 +30,32 @@ public function discover(ClassReflector $class): void
                 continue;
             }
 
-            $parameters = iterator_to_array($method->getParameters());
+            $eventName = match (true) {
```
brendt, line 33: TODO: test

src/Tempest/Http/HttpApplication.php
brendt: I might disagree with my previous agreement, but I need to sleep on the matter

aidan-casey: That can all belong in a config file. 😉

src/Tempest/Http/HttpApplication.php
aidan-casey: Yeah, I think we would construct it manually in the entrypoint, maybe?

src/Tempest/Http/HttpApplication.php
aidan-casey: Come to think of it... do we really need to do any setup here or can that be in a config?

brendt: That's a good point, though it's outside the scope of this PR I think

src/Tempest/Http/HttpApplication.php
brendt: I've slept. I think there's a good argument to be made against a config file: config files are discovered by Tempest, except this one. That would cause a lot of confusion once again as before. I like the current approach better.

aidan-casey: So if I as a developer want to adjust default discovery locations, I'm doing this by modifying both the `tempest` and `public/index.php` files?

PR #368, opened 2024-09-06
File support in requests

yassiNebeL:


brendt:
I only realised yesterday that we already have PSR requests, so file support actually comes kind of built-in: https://github.com/tempestphp/tempest-framework/commit/5536306ce4731e0376b5745a0dd7fff530b320fc

PR #369, opened 2024-09-06
Isolate extract method

bYemma:
When the extract part is in an isolated method, variables from the `resolveContent` method can be reused in the view file.

Conflicts with variables like `$view` or `$path` e.g. are prevented. Only `$_path` and `$_data` are reserved as a minimum.


brendt:
Nice, thanks!

PR #373, opened 2024-09-11
Use correct session value to test destroySession

Bapawe:


brendt:
Oh 😅 Thank you so much!

PR #377, opened 2024-09-12
Refactors directory structure of packages.

aidan-casey:


aidan-casey:
@brendt - If you are good with this, let's just merge it and I will deal with the static analysis errors. This sets the groundwork for moving tests, etc. to the isolated package.

brendt:
> let's just merge it

You don't have to say that twice. MEEEEEERGE!

PR #378, opened 2024-09-12
Refactor unit tests

aidan-casey:


aidan-casey:
Still some work to do here with PHPUnit configuration. Dependent on #377.

PR #391, opened 2024-09-14
Updates contributing.

aidan-casey:


aidan-casey:
@brendt - WDYT? Assuming you like the label ideas, I will create those with the merge.

PR #414, opened 2024-09-17
Map upload files into request properties

yassiNebeL:


yassiNebeL:
@brendt, I think we need a caster for `Tempest\Http\Upload` that's what we got from `$type->getName()` within `CasterFactory` class.

src/Tempest/Mapper/src/Casters/CasterFactory.php
```diff
@@ -48,6 +49,7 @@ public function forProperty(PropertyReflector $property): ?Caster
             'float' => new FloatCaster(),
             'bool' => new BooleanCaster(),
             DateTimeImmutable::class, DateTimeInterface::class, DateTime::class => DateTimeCaster::fromProperty($property),
+            Upload::class => new FileUploadCaster(),
```
brendt, line 52: You should use the `#[CastWith]` attribute on the upload class. This mapping is only for classes and types that we don't control

src/Tempest/Http/src/Mappers/RequestToObjectMapper.php
```diff
@@ -18,6 +18,8 @@ public function canMap(mixed $from, mixed $to): bool
     public function map(mixed $from, mixed $to): array|object
     {
         /** @var Request $from */
-        return map($from->getBody())->to($to);
+        $data = array_merge($from->getBody(), $from->getFiles());
```
brendt, line 21: Is this the right place though? I thought this mapping should be done in `\Tempest\Http\Mappers\PsrRequestToRequestMapper`

src/Tempest/Mapper/src/Casters/FileUploadCaster.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Mapper\Casters;
+
+use InvalidArgumentException;
+use Tempest\Http\Upload;
+use Tempest\Mapper\Caster;
+
+final readonly class FileUploadCaster implements Caster
```
brendt, line 11: I don't see how this would work? The goal would be to map `\Psr\Http\Message\UploadedFileInterface` to `\Tempest\Http\Upload::__construct`.

On top of that, we're already doing that in `\Tempest\Http\Mappers\PsrRequestToRequestMapper:43`, so I don't think an additional caster is actually needed.

brendt (changes requested):
Overall, I'm left confused with what you're trying to do in this PR. I left some comments, but I don't think this is the right approach

aidan-casey:
Hey @yassiNebeL! I appreciate you taking a look at this. I am going to close this PR due to inactivity.

yassiNebeL:
No worries!

PR #418, opened 2024-09-18
feat(support): add `Pluralizer` implementation

innocenzi:
This pull request adds support for pluralizing (or singularizing) nouns through `doctrine/inflector`. The implementation is a simplified and cleaner version of Laravel's.

This can be used as a helper in application code through `LanguageHelpers::pluralize()`, but is also intended to be used in the framework itself, such as in database table names, exception and log messages, validation, etc.

Note: I didn't know if I should make a new package or put it in the support one, so I kept things simple and added it to `Support`. I'm assuming `Support` will grow over time and it's not a bad place to have it, but at the same time, `Clock` is a separate package, so I don't know. Happy to update the PR.

Related: https://discord.com/channels/1236153076688359495/1236153079603269766/1268515847073173606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `LanguageHelper::pluralize` and `LanguageHelper::singularize`
 - Unit tests for the CommandBus `dispatch` method.
 - Replaced Ignition with Whoops.
 - Properly detect environment from `.env` when present.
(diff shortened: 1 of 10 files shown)


aidan-casey:
Thanks, @innocenzi! This is definitely a good thing to have in the framework, I cannot tell you how many times I have reached for this in Laravel.

What does this implementation do with the words Laravel deems as "uncountable"?

innocenzi:
Currently, it ignores them because I wanted to keep things simple. I was thinking we could add them back later if needed.

I checked why Laravel had these exceptions and stumbled upon these—it seems they were added a long time ago as an edge case:
- https://github.com/doctrine/inflector/pull/194
- https://github.com/laravel/framework/pull/28749

I personally think we don't even need to add them, since words like that are quite common and we should instead manually override the table name through the framework.

aidan-casey:
Agreed. Thank you for the thorough discovery. 🙂

PR #425, opened 2024-09-18
ci: automatic changelog and releases

innocenzi:
As discussed on Discord, this pull request implements the GitHub Action that automatically generates `CHANGELOG.md` and creates GitHub release notes.

## Tool used

I settled on using [git-cliff](https://git-cliff.org/) for its flexibility and configurability. It's fast and a lot of things can be changed if needed.

Note that, since commits weren't following a convention until now, they cannot be properly included in the changelog, so I chose to ignore tags before `1.0-alpha1` (included).

&nbsp;

## Release workflow

The Action is triggered when new tags are pushed:

```bash
git tag v1.0.1
git push --tags
```

When triggered, the Action will re-generate the `CHANGELOG.md` file and commit the changes with the `chore: update changelog` message.

It will then use the latest tag to create a GitHub release with the same format.

### Side notes regarding versioning

The standard for version names on GitHub is to prefix them with `v`. I noted that Tempest didn't use that, I'm not sure how `git-cliff` behaves without them—but I would suggest to start using `v` anyway.

Also, apparently `1.0-alpha1` is not semver-compliant. The correct version would be `1.0.0-alpha.1` (patch version specified, and dot before the alpha number). I don't think this is a big deal either, but I still suggest trying to stick to semver.

&nbsp;

## New commit convention

From now on, commits must all respect the [conventional commit specification](https://www.conventionalcommits.org/en/).

Note that commit descriptions **should not** start with an uppercase letter and should use [imperative mood](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181):

```diff
- feat(support): Adds some cool feature
+ feat(support): add some cool feature
```

In the changelog, commit descriptions can be [processed to have their first letter uppercased](https://git-cliff.org/docs/templating/syntax) if you want.

Scopes (`support` in the example above) are not mandatory, but I highly recommend always using them for consistency and grouping purposes.

Generally, `chore` commits can omit them though:

```
chore: fix typo in `README.md`
```

&nbsp;

## Other notes

As `git-cliff` is very flexible, it's totally possible to adopt other workflows.

For example, instead of running everything in GitHub Actions, you could create a `release.php` or `release.sh` that does the same thing automatically. You would miss on the automatic release creation though.


aidan-casey:
Thanks, @innocenzi! Just a couple of thoughts:

1. The GitHub action testing that the changelog gets updated on each PR can go, correct?

2. It feels like we should update contributing with a section outlining the PR merge process and standard/allowed commit types. Is that best done here or in a separate PR?

aidan-casey:
Thanks, @innocenzi!

Issue #288, opened 2024-07-27
Primitive type helpers


brendt:
- Arr
- Str
- Path
- Int

Tresor-Kasenda:
Speaking of primitives, can I say that you want to create classes for each primitive and then add helper? If yes i can contribute to this issue

brendt:
ArrayHelper PR is here: https://github.com/tempestphp/tempest-framework/pull/434

aidan-casey:
@brendt - Does it make sense to move the path helper to a `Filesystem` component? I am re-looking at the filesystem component, so just thinking about that now.

brendt:
I don't think so. I want this path helper to be just that: a small wrapper around pathinfo. I might also want to add something like `path('…')->put($input)` (which ensures the directory exists), but nothing more than that. 

I think a proper filesystem with multiple drivers is still needed, but it's not what this helper is for.

Tresor-Kasenda:
it's a first for me to contribute on an open source project i'd like to have details on the path helper primitive in the case of path

brendt:
Hi @Tresor-Kasenda thanks for asking, I noticed you asked on stream yesterday and I forgot to answer, my bad! 🫣

Now that Aidan has been working on the filesystem, I might actually reconsider my stance on the matter, and maybe `path` should be a simple wrapper that returns a filesystem instance. We'll have to wait until Aidan finished #441 before we can continue with this one though

PR #430, opened 2024-09-18
feat(support): add `StringHelper`

innocenzi:
Related to https://github.com/tempestphp/tempest-framework/issues/288

This pull request adds a new `StringHelper` class with a few of the most used string helpers from Laravel ("most used" is totally arbitrary there, but I think these are the ones most likely to be used withing Tempest itself).

Most of the code is the same, except for `snake`—the Laravel version doesn't take some edge cases into account, which I've fixed here (like trimming or deduplicating `$delimiter`).

### DIFF
--- a/src/Tempest/Support/src/StringHelper.php
+++ b/src/Tempest/Support/src/StringHelper.php
@@ -1,0 +1,156 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Support;
+
+use Countable;
+
+final readonly class StringHelper
+{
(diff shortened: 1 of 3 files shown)


aidan-casey:
Nice PR. Thanks!

brendt:
Can I suggest an addition here? This is good, but what if we had this kind of API?

```php
$string = str(' blablablabla asdhkjashd asldkj ')
    ->trim()
    ->camelCase()
    ->finish('_')
```

PR #433, opened 2024-09-19
feat: refactor string helper to object

brendt:
Related to #288 and #430

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -94,7 +94,8 @@
             "src/Tempest/EventBus/src/functions.php",
             "src/Tempest/Http/src/functions.php",
             "src/Tempest/Mapper/src/functions.php",
-            "src/Tempest/Reflection/src/functions.php"
+            "src/Tempest/Reflection/src/functions.php",
+            "src/Tempest/Support/src/functions.php"
         ]
     },
     "autoload-dev": {
--- a/src/Tempest/Support/composer.json
+++ b/src/Tempest/Support/composer.json
@@ -11,7 +11,10 @@
     "autoload": {
         "psr-4": {
             "Tempest\\Support\\": "src"
-        }
+        },
+        "files": [
+            "src/functions.php"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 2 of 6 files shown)


brendt:
So I wanted to add this object notation, but then thought the static variant isn't _really_ needed: we don't need two ways to do the same thing. Wdyt @innocenzi ?

composer.json
```diff
@@ -94,7 +94,8 @@
             "src/Tempest/EventBus/src/functions.php",
             "src/Tempest/Http/src/functions.php",
             "src/Tempest/Mapper/src/functions.php",
-            "src/Tempest/Reflection/src/functions.php"
+            "src/Tempest/Reflection/src/functions.php",
+            "src/Tempest/Support/src/functions.php"
```
Treggats, line 98: reduce the diff and overal convention (as far as I know)
```suggestion
            "src/Tempest/Support/src/functions.php",
```

src/Tempest/Support/composer.json
```diff
@@ -11,7 +11,10 @@
     "autoload": {
         "psr-4": {
             "Tempest\\Support\\": "src"
-        }
+        },
+        "files": [
+            "src/functions.php"
```
Treggats, line 16:
```suggestion
            "src/functions.php",
```

src/Tempest/Support/src/StringHelper.php
```diff
@@ -1,79 +1,104 @@
 <?php
 
-declare(strict_types=1);
```
Treggats: Why'd you remove this?

src/Tempest/Support/src/StringHelper.php
```diff
@@ -5,75 +5,105 @@
...
+use function preg_replace;
+use function rtrim;
+use Stringable;
+use function trim;
+
+final readonly class StringHelper implements Stringable
 {
-    public static function title(string $value): string
+    public function __construct(private string $string = '')
+    {
```
Treggats, line 18: This is a personal preference, do with it what you will 😄 
```suggestion
    public function __construct(
        private string $string = '',
    ) {
```

Treggats (changes requested):

composer.json
brendt: Trailing commas aren't allowed in JSON :(

Treggats: Oh right 🤦

Treggats (approved):
Looks good to me

brendt:
Awesome, thanks!

PR #434, opened 2024-09-19
feat: add array helper

brendt:
TODO:

- [x] mapWithKeys syntax
- [x] reconsider pop syntax

### DIFF
--- a/src/Tempest/Mapper/src/Mappers/ArrayToObjectMapper.php
+++ b/src/Tempest/Mapper/src/Mappers/ArrayToObjectMapper.php
@@ -11,7 +11,7 @@ use Tempest\Mapper\Strict;
 use Tempest\Mapper\UnknownValue;
 use Tempest\Reflection\ClassReflector;
 use Tempest\Reflection\PropertyReflector;
-use Tempest\Support\ArrayHelper;
+use function Tempest\Support\arr;
 use Tempest\Validation\Validator;
 use Throwable;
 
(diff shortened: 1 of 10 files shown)


aidan-casey:
@brendt - It looks like you made some changes to the `Mapper` component in this PR. The isolated tests are installing dependencies from `dev-main` and those dependencies don't exist yet.

innocenzi:
I would love to keep `ArrayHelper` immutable honestly.

What do you think of separating the base `ArrayHelper` class, keeping it immutable, and having another class more similar to Laravel's `Collection`, that would implement `Iterator` and `ArrayAccess`?

Also, related to `pop`, I know PHP itself implements this paradigm in its own functions, but I don't find this pleasant to work with (eg. `preg_match` 😵). I personally prefer the destructuring API suggested MD on Discord:

```php
[$last, $array] = $array->pop();
```

aidan-casey:
> What do you think of separating the base ArrayHelper class, keeping it immutable, and having another class more similar to Laravel's Collection, that would implement Iterator and ArrayAccess?

I do agree with this. It makes sense to split these out to collection classes. In some sense, that may allow the array helper to just call those underlying classes. I'm a little biased, but it'd be cool to resurrect #222.

brendt:
I did some thinking about immutability:

- I don't feel like splitting the iterator from from base class makes sense, that's just extra steps for users to take
- I don't think making a distinction between `Dictionary` and `ArrayList` is the right way to go, that's not the goal of this component. The goal is to add a wrapper around PHP's native array implementation, and to stay as close as possible to it.
- The problem with mutability comes from the fact that I use PHPs built-in `current`, `next`, and `reset` functions. The reason I need to use those instead of manually keeping track of the current position is because a PHP array could have named keys. 
- I considered doing something with a generator instead, but there's nothing to be gained from it:

```php
foreach ($array->loop() as $key => $value) // generator approach
foreach ($array->toArray() as $key => $value) // already possible
foreach ($array as $key => $value) // this is what we want
```

For now, I don't see a solution, and I'm fine with the array being mutable only for iteration. All public methods are still immutable, which is what matters the most.

Issue #272, opened 2024-03-06
Filesystem


brendt:
Needed for 

- tempestphp/tempest-framework#123 
- tempestphp/tempest-framework#201

brendt:
Do we really need this before 1.0?

shaffe-fr:
I think Tempest needs file upload management for 1.0. It's pretty tedious with vanilla PHP. 
I guess local driver is required for that. Multiple disks support is bonus but should not be so tough once it's done. 
Are you considering wrapping flysystem?

aidan-casey:
@shaffe-fr I'm _thinking_ we should separate out local filesystem and third-party storage as two different packages in Tempest world, with priority on the former.

shaffe-fr:
I like your idea to use Filesystem for framework level stuff and Storage for application level. 👌
We could start with Filesystem and Storage with local driver only. That would not require to hook on Flysystem.

brendt:
I really like the difference between fs and storage as well 👍👍

PR #441, opened 2024-09-20
Begins to add filesystem

aidan-casey:


shaffe-fr:
Hey, it might be good to check how https://github.com/thephpleague/Flysystem/blob/3.x/src%2FLocal%2FLocalFilesystemAdapter.php uses `clearstatcache` to avoid edge cases.

aidan-casey:
Yes, for sure we will want to utilize `clearstatcache`. 🙂

innocenzi:
Do we want to avoid the dependency on Flysystem? I think it would be less work to wrap them than to rewrite and maintain an equivalent. This would also allow people to use third-party Flysystem adapters, as this is the de-facto storage implementation in PHP land

aidan-casey:
Hey, @innocenzi! See #272 where we discuss this a little bit. I'd like to keep this portion lightweight and independent of Flysystem, but potentially make use of it for the "storage" component.

innocenzi:
Ah sorry, didn't check the linked issue. Makes total sense 👌

brendt:
One thing I want to be possible is to create a file WITH the directory if it doesn't exist, so basically a oneliner for this

```php
        if (! is_dir(dirname($path))) {
            mkdir(dirname($path), recursive: true);
        }

        file_put_contents(…

// or 

        copy(__DIR__ . '/index.php', $path);
```

aidan-casey:
Definitely a good thought. I'll think about the API a bit. I'd like to also allow the disabling of that functionality if the developer determines they _don't_ want that to happen.

aidan-casey:
@brendt - What do you think about the API with the `Permission` class here? I have a couple of thoughts...

We could simplify the combinations by allowing an integer to the `with` method. So then you could do:

```php
Permission::allow(
    Permission::ownerReadWrite(),
    Permission::groupRead(),
    Permission::othersRead(),
);
```

There's not an easy way to validate that though.

brendt:
@aidan-casey instead of providing methods that combine permissions, I'd simply provide values:

```
    case OWNER_ALL = 0o007;
    case GROUP_ALL = 0o007;
    case OTHERS_ALL = 0o007;
```

brendt:
I'd also add a `without` method on the enum which starts from `777` and subtracts

brendt:
Sidenote: this is a genius idea and I already don't want anything else anymore

aidan-casey:
@brendt - Good thought. Instead of counting back from 777, I am counting back from the current value. It feels safe to me than something like:

```php
// 577
Permission::OWNER_FULL->without(Permission::OWNER_WRITE);
```

Let me know if you want to change that though.

brendt:
Totally agree, this is awesome

src/Tempest/Filesystem/src/Exceptions/UnableToCopyFile.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Tempest\Filesystem\Exceptions;
+
+use RuntimeException;
+
+final class UnableToCopyFile extends RuntimeException
+{
+    public static function fromSourceToDestination(string $source, string $destination): self
+    {
+        return new self("Unable to copy file from '{$source}' to '{$destination}'");
```
sebsobseb, line 13:
```suggestion
        return new self("Unable to copy file from '{$source}' to '{$destination}'.");
```

src/Tempest/Filesystem/src/ErrorContext.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Tempest\Filesystem;
+
+final class ErrorContext
```
aidan-casey, line 5: @innocenzi - Can I get a quick sanity check here? Is this helpful or an unnecessary abstraction?

innocenzi: If the functions were `try`/`catch`able, I'd prefer that, but since you need to call `error_get_last()`, you might as well have a nice experience with that

However, I'm wondering if `error_clear_last()` is really needed? Since you check for the result of warning-emitting functions, `error_get_last()` will always be the relevant error.

So maybe simplify `ErrorContext` to have a sane way of retrieving the last error message, without needing for the reset part?

brendt: Yeah this is good 👍

aidan-casey:
Mental note:

- Exception messages are a little inconsistent right now. Clean that up.
- Method behavior is a little inconsistent right now. Clean that up.
- Determine a good API for directory traversal _maybe_ use in methods like `deleteDirectory`.
- Add some helpers for file metadata.
- Add some helpers for file streams.

brendt:
I think the best way forward with this PR is to rebase and merge it. We can always improve on it later, but right now it's in limbo and no one else is able to help with it because it's not on main but a separate branch. 

@aidan-casey wdyt?

PR #442, opened 2024-09-20
fix(view/tempest-renderer): join generic elements with an empty string

innocenzi:
On the live docs, I noticed there's a space between inline elements and text: 

CleanShot 2024-09-20 at 20 36 59@2x

This happens because the Tempest view renderer separates elements with `PHP_EOL`, so the generated HTML has a new line. In this case, the generated HTML is the following:

```html
<p><strong>Tempest is a PHP framework that gets out of your way</strong>

. Its design philosophy is that developers should write as little framework-related code as possible, so that they can 

<strong>focus on application code</strong>

 instead.</p>
```

This pull request changes this behavior by separating generic elements with an empty string, resulting in the following HTML:

```html
<p><strong>Tempest is a PHP framework that gets out of your way</strong>. Its design philosophy is that developers should write as little framework-related code as possible, so that they can <strong>focus on application code</strong> instead.</p>
```

I didn't touch the other parts where elements are separated by `PHP_EOL` as I didn't notice other bugs, but for consistency we might want to do that.

### DIFF
--- a/src/Tempest/View/src/Renderers/TempestViewRenderer.php
+++ b/src/Tempest/View/src/Renderers/TempestViewRenderer.php
@@ -285,7 +285,7 @@ final class TempestViewRenderer implements ViewRenderer
             $content[] = $this->renderElement($view, $child);
         }
 
-        $content = implode(PHP_EOL, $content);
+        $content = implode('', $content);
 
         $attributes = [];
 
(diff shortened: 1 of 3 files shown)


aidan-casey:
Thanks!

PR #453, opened 2024-09-21
feat(database): implement table naming strategies

innocenzi:
This pull request introduces the concept of configurable "table naming strategies". The goal of this implementation is to offer a sensible default, but allow configuration with as less friction as possible.

## New default

The new default naming for tables is pluralized `snake_case`. 

As far as I have worked in web development, this is the naming style that has best suited my brain, as my reasoning for database tables is something like "a box of \<object\>", so, for instance, "a box of users".

As for `PascalCase` and `snake_case`, it's about readability.

## Configurability

I personally don't mind if you'd prefer to keep the currently-implemented naming style (`PascalCase`), as this pull request implements an easy way to configure it:

```php
use Tempest\Database\Tables\PascalCaseStrategy;

// app/Config/database.php
return new DatabaseConfig(
    connection: new SQLiteConnection(
        path: __DIR__ . '/../database.sqlite',
        namingStrategy: new PluralizedSnakeCaseStrategy()
    ),
);
```

For now, the following two strategies are implemented:
- `PascalCaseStrategy` (eg. `PersonalAccessToken`)
- `PluralizedSnakeCaseStrategy` (eg. `personal_access_tokens`)

It would be fairly easy for userland to implement their own strategy, simply by implementing `Tempest\Database\Tables\NamingStrategy`.

## Notes

- I wasn't very inspired by the namespace, so if you have a better idea than `Tempest\Database\Tables`, I'm open for it.

- I'm not-so-sure about this part of the implementation. I don't think there's a huge performance cost, as it's a simple table lookup (the configuration is a singleton), but it somehow feels wrong: https://github.com/tempestphp/tempest-framework/pull/453/files#diff-71139a1d93468f4b2a4d83c6b70374139c48a5ffb42252794913fedb7ebdae39R58-R61

- Barely related, but I have seen that multiple `QueryStatement` had `forModel` method. We could discuss this in another PR, but I think we should remove that, because when users will refactor their models at some point, their migrations will stop working properly because the table name inference change. This is like editing an existing migration, it will inevitably break environments when migrations will run from scratch.


aidan-casey:
I'm gonna leave the review of this one to @brendt because there is a lot subjective about all of this. :smile: 

Table names, for example. When I was learning, we were always taught that it should be singular, lower, and camelCase like a bag or collection (apple bag vs. apples bag, user collection vs. users collection). I had never used snake_case and plural until Laravel, but since then I've now worked on numerous projects with numerous opinions.

Regardless of the default position taken by the framework, I do think the naming strategy class and approach is a phenomenal idea and well implemented here.

innocenzi:
@aidan-casey The CI is failing because I added a `classBasename` to `tempest/support`, but I think it is trying to pull the latest from the `main` branch instead of the local version. Is that expected?

That aside, the CI is awesome, it caught a lot of small inconsistencies. Super good work on it.

aidan-casey:
> but I think it is trying to pull the latest from the main branch instead of the local version. Is that expected?

I was just talking to Brent about this the other day! Yes, it is to be expected as we want to test that an individual (isolated) component can install the lowest version of its dependencies and highest version of its dependencies from composer and work. Once we have an official major release out, this will also help us keep an eye on breaking changes.

As you rightly note, this proves a little bit of a challenge when making changes to two packages in the same PR and I haven't decided how to tackle that yet. 😅 

> That aside, the CI is awesome, it caught a lot of small inconsistencies. Super good work on it.

That's awesome to hear! It was a collective effort. 🙂

src/Tempest/Database/src/Connections/PostgresConnection.php
```diff
@@ -21,6 +23,7 @@ public function __construct(
         public string $password = '',
         #[SensitiveParameter]
         public string $database = 'postgres',
+        public NamingStrategy $namingStrategy = new PluralizedSnakeCaseStrategy()
```
brendt, line 26: Haven't read through the rest of the code, but my first thought is that this property belongs in `DatabaseConfig` and not the connection

src/Tempest/Database/src/Connections/PostgresConnection.php
```diff
@@ -52,4 +55,9 @@ public function dialect(): DatabaseDialect
     {
         return DatabaseDialect::POSTGRESQL;
     }
+
+    public function tableNamingStrategy(): NamingStrategy
```
brendt, line 59: We usually rely on public properties directly instead of getters. We plan on removing most getters from Tempest when PHP 8.4 arrives with property hooks

src/Tempest/Database/src/Migrations/CreateMigrationsTable.php
```diff
@@ -18,13 +19,13 @@ public function getName(): string
 
     public function up(): QueryStatement|null
     {
-        return (new CreateTableStatement('Migration'))
+        return (new CreateTableStatement(Model::table()->tableName))
```
brendt, line 22: Couldn't you use `CreateTableStatement::forModel(Model::class)`?

Is there a reason you import `Migration` as `Model`?

src/Tempest/Database/src/Migrations/CreateMigrationsTable.php
```diff
@@ -18,13 +19,13 @@ public function getName(): string
...
             ->primary()
             ->text('name');
     }
 
     public function down(): QueryStatement|null
     {
-        return new DropTableStatement('Migration');
+        return new DropTableStatement(Model::table()->tableName);
```
brendt, line 29: Same remark here

src/Tempest/Support/src/StringHelper.php
```diff
@@ -211,4 +211,9 @@ public function length(): int
     {
         return mb_strlen($this->string);
     }
+
+    public function classBasename(): self
```
brendt, line 215: Oooooh I like this 🤩

src/Tempest/Database/src/IsDatabaseModel.php
```diff
@@ -54,7 +55,12 @@ public function getId(): Id
 
     public static function table(): TableName
     {
-        return new TableName(pathinfo(str_replace('\\', '/', static::class), PATHINFO_FILENAME));
+        $name = get(DatabaseConfig::class)
```
brendt, line 58: Yeah I'm pretty sure it should be on `DatabaseConfig` and not the connection, unless you had a reason for it?

brendt (changes requested):
Great PR, a couple of remarks and one question, but I'm happy to merge this when we've tackled all remarks

src/Tempest/Database/src/Connections/PostgresConnection.php
aidan-casey: [This](https://discord.com/channels/1236153076688359495/1236153079603269766/1287067156835602588) might be relevant.

brendt: Nope, I don't see where it got discussed?

brendt: Yeah I'm not quite sure yet how multi-database support would work. I think keeping this on database config makes most sense for now :)

src/Tempest/Database/src/Migrations/CreateMigrationsTable.php
brendt: > Yes, it's because the Migration interface is already imported. I chose Model since the Migration class is a model, but we can rename it to something else

Aaah yes, I forgot about that. I really hate the name collision, but don't know what to do about it. If you have suggestions, I welcome them! Not blocking this PR though.

brendt:
> Barely related, but I have seen that multiple QueryStatement had forModel method. We could discuss this in another PR, but I think we should remove that, because when users will refactor their models at some point, their migrations will stop working properly because the table name inference change. This is like editing an existing migration, it will inevitably break environments when migrations will run from scratch.

Let's indeed move this to a separate issue, do you want to make one for it @innocenzi ?

brendt:
@innocenzi Do you want to rebase unto main so that we can merge?

innocenzi:
I rebased 👍

brendt:
@aidan-casey how should I merge this one? It's blocked by the failing tests, but I think those fail because the changes aren't merged yet.

aidan-casey:
@brendt it's not blocked by the tests. It's blocked by an unresolved change that was requested. 😉

brendt (approved):

brendt:
Oh that was confusing, GitHub showed everything as resolved, but there was still a requested change when I clicked through. Dumb of me :p

brendt:
Thanks @innocenzi , great addition!

PR #464, opened 2024-09-23
refactor(database)!: remove `forModel` static constructor

innocenzi:
This pull request follows-up on the note in https://github.com/tempestphp/tempest-framework/pull/453:

> I think we should remove that, because when users will refactor their models at some point, their migrations will stop working properly because the table name inference change. This is like editing an existing migration, it will inevitably break environments when migrations will run from scratch.

### DIFF
--- a/src/Tempest/Database/src/QueryStatements/AlterTableStatement.php
+++ b/src/Tempest/Database/src/QueryStatements/AlterTableStatement.php
@@ -16,12 +16,6 @@ final class AlterTableStatement implements QueryStatement
     ) {
     }
 
-    /** @param class-string<\Tempest\Database\DatabaseModel> $modelClass */
-    public static function forModel(string $modelClass): self
-    {
-        return new self($modelClass::table()->tableName);
-    }
-
     public function add(QueryStatement $statement): self
     {
         $this->statements[] = new AlterStatement(Alter::ADD, $statement);
--- a/src/Tempest/Database/src/QueryStatements/CreateTableStatement.php
+++ b/src/Tempest/Database/src/QueryStatements/CreateTableStatement.php
@@ -16,12 +16,6 @@ final class CreateTableStatement implements QueryStatement
     ) {
     }
 
-    /** @param class-string<\Tempest\Database\DatabaseModel> $modelClass */
-    public static function forModel(string $modelClass): self
-    {
-        return new self($modelClass::table()->tableName);
-    }
-
     public function primary(string $name = 'id'): self
     {
         $this->statements[] = new PrimaryKeyStatement($name);
--- a/src/Tempest/Database/src/QueryStatements/DropTableStatement.php
+++ b/src/Tempest/Database/src/QueryStatements/DropTableStatement.php
@@ -25,12 +25,6 @@ final class DropTableStatement implements QueryStatement
         return $this;
     }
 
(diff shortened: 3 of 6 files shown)


aidan-casey:
@innocenzi - A couple questions about the conventional commit format here...

1. I see the linter is failing. I suspect this has something to do with the breaking change (!) format or something. Any insight?
2. Should the component (`database`) be the scope here?

innocenzi:
Fixed it—I misnamed the commit and forgot to change the PR name after that. The bang goes at the end if there's a scope before, and you're right about the scope itself, it should be either `database` or `database/query-statements`

I see you're getting the grasps of it hehe

aidan-casey:
Awesome! I'm definitely liking the structure conventional commits give us. 🙂 

@brendt - These failing tests are because of #466 which I just noticed was a bug now. 😅  I'm ready to go ahead and merge this, but waiting your review as you added this initially.

brendt:
I don't agree that this is actually fixing the underlying problem. 

Whether you write this in your migrations:

```php
return (new CreateTableStatement(Book::table()->tableName))
```

Or this:


```php
return CreateTableStatement::forModel(Book::class)
```

Is the same: if the model's table name changes, it breaks. The only way around this is to hard-code the table name in the migration. 

I'm not convinced this is a problem that needs fixing. Sure, if people decide to change table names, they'll have to update migrations for backwards compatibility. This is something that rarely happens in my experience. I've worked on pretty large applications and only remember having done this once.

For sure, it will happen _sometimes_, but do we really need to worsen the user-experience for the 5% edge case? I personally don't think so. I think the better approach here is to make a note in the docs about migration naming, and how to deal with it when you decide to change table names instead.

So I'm going to close this PR, make a new issue for the docs change, but feel free to counter my arguments and I might reconsider it.

innocenzi:
In userland codebases, this matters and developers should never use `(new CreateTableStatement(Book::table()->tableName))` (nor the `forModel` equivalent). They should use `(new CreateTableStatement('books'))`. 

Migration files should NEVER contain references to code that may be refactored. They will break at some point down the line.

I think this is a footgun; documenting it is fine, I personally know I'll never gonna use this, but I had people in my teams do that a few times more than I'd like.

aidan-casey:
I'm technically agreed with @innocenzi here and I _think_ we had that conversation when this first got added, so I'll try not to be redundant.

This to me is another "bad habit by default" that feels like it comes out of the Laravel way. It only takes one third-party package or team member using this and then updating the model name to entirely brick my app (has happened with Laravel 🥲). Tests won't catch it either, since they are doing fresh migrations.

Only in simple CRUD apps are database tables a 1-for-1 with your models. Anything beyond that requires acknowledging they serve different purposes. I'd rather not continue to confuse people with the former idea.

brendt:
I gave it some more thought. 

_if_ we had a `make:migration` command, the problem would go away and I'd have no problem removing this method.

I made an issue for it, I'll see when I or someone else can tackle it: https://github.com/tempestphp/tempest-framework/issues/472

aidan-casey:
@brendt - I suggest we build out the migration command in `tempest/cli` and have it detect the project. It seems like a very Tempest thing to do. 😎

Wulfheart:
For generating commands you might also want to create the class with https://github.com/nette/php-generator instead of Laravel's approach with a template file.

aidan-casey:
@Wulfheart - Agreed!

PR #475, opened 2024-09-24
fix: dependency loop between kernel and event-bus

brendt:


brendt:
This is the most pragmatic solution I can come up with to allow the kernel to dispatch events, but still not have a dependency on the event bus. 

I'm going to merge this no matter what, because all CI in PRs are failing right now because of it. We might need to further improve in the future, but this works for now

PR #479, opened 2024-09-24
feat: use directories instead of file names to allow default server config setup

### LINKED ISSUE #478: Static file generation for servers that don't allow custom server config
### Description

An event would be emitted for each static page that is generated. The event would include at least the path to the page.

### Benefits

I’d like to use “pretty URLs” with my static content, e.g. `/contact-us` instead of `/contact-us.html`. This can be achieved by placing a `/contact-us/index.html` page. 

Instead of adding this as an option to the framework if there was an event I could create the “pretty” directories and move the generated file as they are generated.

brendt:
This would fix #478 

I wonder if there are any downsides to this approach?


brendt:
One "downside" is that local development setups will now also prioritize the rendered HTML if it's present. So whenever you make a change locally, you need to make sure you clear the statically generated pages. I don't know if that's really an issue though.

PR #493, opened 2024-09-27
feat: authenticator

brendt:


brendt:
After this PR, we'll look into an authorization API (preferably with user roles built-in). All front-end stuff and user-management is left alone for now.

src/Tempest/Auth/.gitattributes
```diff
@@ -0,0 +1,10 @@
+# Exclude build/test files from the release
+.github/                  export-ignore
+tests/                    export-ignore
+.gitattributes            export-ignore
+.gitignore                export-ignore
+phpunit.xml               export-ignore
+README.md                 export-ignore
+
+# Configure diff output for .php and .phar files.
+*.php diff=php
```
Treggats, line 10:
```suggestion
*.php diff=php

```

src/Tempest/Auth/composer.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "tempest/auth",
+    "require": {
```
Treggats, line 3: You'd probable have a better description than this 😅 

```suggestion
    "name": "tempest/auth",
    "description": "Tempest authentication",
    "require": {
```

src/Tempest/Database/src/database.sqlite
Treggats, line 1: I'm not a fan having this in the git tree, why did you include it?

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -177,9 +180,35 @@ public function assertHasNoValidationsErrors(): self
             sprintf(
                 "There should be no validation errors, but there were: %s",
                 implode(', ', array_keys($validationErrors)),
-            )
+            ),
```
Treggats, line 183: did you add this purely because of the possible `message` argument?

Treggats (changes requested):

innocenzi:
I think this is great, there shouldn't need to be more built-in (besides authorization) 👍

However, I wouldn't let Tempest take care of the user model itself and the migration, I would rather have Tempest use an interface that has to be implemented in userland

src/Tempest/Database/src/database.sqlite
brendt: IDK actually 🤔

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
brendt: This was my autoformatter, I don't know why php cs fixer didn't pick up on it though

brendt:
> However, I wouldn't let Tempest take care of the user model itself and the migration, I would rather have Tempest use an interface that has to be implemented in userland

Yeah, that's a fair point.

brendt:
While I want the user model to be configurable, I do want to provide a default implementation — I might reconsider that though. 

Also, I'm ok with "the user" being tied to the database, I don't really care about other ways of storing user data for now.

innocenzi:
Looking at the current code, it's unclear to me—is it possible to ignore the provided migration? Or would you recommend to add another one that would run after it and modify the table as we see fit?

That'd be one reason I was suggesting not to provide a default model/migration, but it seems low enough friction

aidan-casey:
I'd love to see this use a repository to decouple retrieving the user details. If it were me personally, my criteria would be for this to be decoupled so that it could work entirely outside the framework. I understand that's a bit of an extreme view, though. 🙂

brendt:
> Looking at the current code, it's unclear to me—is it possible to ignore the provided migration? Or would you recommend to add another one that would run after it and modify the table as we see fit?

Right now it's not possible, but this is something I already planned on figuring out before merging this PR :)

brendt:
> I understand that's a bit of an extreme view, though. 🙂

Yes, not within my scope for now :)

brendt:
> We have to think about the people storing passwords in the filesystem, @brendt!

I thought about it: they are free to roll their own auth layer separate from Tempest's default implementation 👍

src/Tempest/Auth/phpunit.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.4/phpunit.xsd" bootstrap="vendor/autoload.php" executionOrder="depends,defects" beStrictAboutOutputDuringTests="true" displayDetailsOnPhpunitDeprecations="true" failOnPhpunitDeprecation="false" failOnRisky="true" failOnWarning="true">
+    <testsuites>
+        <testsuite name="Tempest Clock">
```
aidan-casey, line 4: Update the package name. 😉

src/Tempest/Auth/composer.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "tempest/auth",
+    "description": "Tempest authentication",
```
aidan-casey, line 3: Update to something like:

> A flexible authentication package for the Tempest PHP framework, providing user authentication, session management, and token-based access control.

src/Tempest/Auth/src/User.php
```diff
@@ -0,0 +1,59 @@
...
+use BackedEnum;
+use SensitiveParameter;
+use Tempest\Database\DatabaseModel;
+use Tempest\Database\IsDatabaseModel;
+use function Tempest\Support\arr;
+use UnitEnum;
+
+final class User implements DatabaseModel, CanAuthenticate, CanAuthorize
+{
+    use IsDatabaseModel;
```
aidan-casey, line 16: :sob:

src/Tempest/Database/src/Id.php
```diff
@@ -15,11 +15,18 @@
 
     public function __construct(string|int|self $id)
```
aidan-casey, line 16: Is there a reason we don't separate these out?

src/Tempest/Database/src/Mappers/QueryToModelMapper.php
```diff
@@ -51,18 +51,40 @@ private function parse(ClassReflector $class, DatabaseModel $model, array $row):
 
             $count = count($keyParts);
 
+            // TODO: clean up and document
```
aidan-casey, line 54: Do et. 🙃

src/Tempest/Auth/src/CanAuthorize.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Auth;
+
+use UnitEnum;
+
+interface CanAuthorize
+{
+    public function grantPermission(string|UnitEnum $permission): self;
```
aidan-casey, line 11: Should the Authorizer really have this?

brendt: Not sure yet, I'm in the sketching phase right now :)

src/Tempest/Database/src/Mappers/QueryToModelMapper.php
brendt: I will before merging!

src/Tempest/Database/src/Id.php
brendt: The idea is that all IDs are represented by a single value object, and that people don't have to worry about using UUIDs or normal ids. BUT that's in theory, I haven't really touched this code in ages. But outside the scope of this PR

src/Tempest/Auth/src/User.php
brendt: I might refactor it! Step by step

brendt:
How to deal with this circular dependency: auth requires http because of session management, but http also requires auth because of the middleware. 

Since this auth layer is only concerned with http authentication, would it make sense to keep it all within the same component?

I know @aidan-casey will tell me otherwise :)

innocenzi:
Is the circular dependency itself a technical issue? If not, I think keeping the `tempest/auth` package separate is a good idea

brendt:
Apparently it is, composer can't deal with it (at least not in our CI setup). There's no issue in this branch because I haven't added the circular dependency yet, but we had the same issue once before between core and event-bus. 

Maybe there's a solution for it, I'll do some digging.

aidan-casey:
@brendt - I haven't looked through the code yet, but probably I would have said use a driver/repository for storing the auth data. 🙈 🙃

brendt:
I solved the issue, it was pretty straight forward: just need to hook into the Kernel::BOOTED event and add the middleware like so

src/Tempest/Database/src/Mappers/QueryToModelMapper.php
brendt: Ok I won't before merging 🫣

I don't really like this part, and I need to think some more about how to improve it. I made an issue for it though: #507

brendt:
Ok so, this branch contains so much code that I want to merge it rather sooner than later. I'm still going to write an implementation for `revokePermission`, but after that I plan on merging it.

PR #503, opened 2024-09-30
chore: run tests in random order

vsergiu93:


vsergiu93:
Hey @aidan-casey there is this `migrade:fresh` command test for which I can't for the life of me figure out why is failing, the MemoryOutputBuffer gets some values set but, by the time we get to actually assert the output buffer, its empty.

Maybe you have some idea?

aidan-casey:
@vsergiu93 - I honestly think the fibers tester is too clever. 😛 

It looks to me like something is happening with the output buffer singletons that shouldn't be, because the object IDs are not matching between the `ConsoleTester` class and the `MigrateFreshCommand` class. I haven't done enough digging to figure out why.

vsergiu93:
@aidan-casey Hmm, yeah is very strange, I followed a bit the call stack
image
Here basically things start to diverge, as you can see $consoleCommandClass holds an instance of console with output buffer having a different id that the one found in $this->container. 

For context: I added and id property to MemoryOutputBuffer which I'm setting in the constructor like this ` $this->id = spl_object_id($this);` and when we get to this method `getCallable` from ExecuteConsoleCommand class I start to see that what I have in the container is not the same as what we have in the resolved object in `$consoleCommandClass`

vsergiu93:
@brendt That makes sense, it was just an idea.

aidan-casey:
For what it's worth, it was a good idea, @vsergiu93 and clearly shows we have a bug to work out. I'll create an issue for this so we can resolve before v1.

PR #517, opened 2024-10-02
feat(string-helper): add replacement and concatenation methods

innocenzi:
## New additions

While working on other parts of Tempest, I came across the need to use methods added in this pull request.

- Multiple `replace` methods
- `append` and `prepend` for chaining

## Changes to existing functions

I also slightly updated the definitions of `after` and `before` to support searching with arrays instead of just strings.

## Debug method

I added a `dd` method, as it was tedious to keep wrapping the whole chain when working on it—however, I could not use the `never` return type because the underlying `dd` function does not use it and thus triggers phpstan.

I also bypassed phpstan's warning on the usage of `dd` by using the underlying `ld` function instead of adding an exception for this warning.

### DIFF
--- a/src/Tempest/Support/src/StringHelper.php
+++ b/src/Tempest/Support/src/StringHelper.php
@@ -275,4 +275,80 @@ final readonly class StringHelper implements Stringable
     {
         return str_ends_with($this->string, $needle);
     }
+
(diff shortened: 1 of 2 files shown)


brendt:
Looks good, although I'd rename `dd` to `ld` to be consistent. I'll do for you

PR #518, opened 2024-10-02
fix(console): support `default` parameter on textbox component

innocenzi:
This pull request adds support for the `default` parameter of `Console#ask`, which wasn't passed down nor taken into account in `TextBoxComponent`.

Note: my approach was the easiest I could think of—reusing the existing methods instead of adding a new cursor manipulation, at the cost of introducing a loop, which I believe is fast enough to not care about it.


brendt:
Yeah, looks good

PR #520, opened 2024-10-02
ci(lint-pr-title): fix subject case rule

innocenzi:
This pull request decreases the strictness of the linting or pull request titles. The following commit name is now allowed:

```
feat(core): add `Composer` util 
```

It was previously not allowed due to the uppercase `C`, which should not be forbidden.


brendt:
Agree

PR #521, opened 2024-10-02
feat(discovery): allow exceptions on `DoNotDiscover` classes

innocenzi:
This pull request is a follow-up of #512—it adds the ability for classes that should not be discovered by Discovery to still be seen by the specified Discovery classes.

For instance, publishable classes that generally should not be discovered may allow the `tempest publish` command to show them as publishable classes.

Related to #513

### DIFF
--- a/src/Tempest/Core/src/Kernel/LoadDiscoveryClasses.php
+++ b/src/Tempest/Core/src/Kernel/LoadDiscoveryClasses.php
@@ -95,12 +95,10 @@ final readonly class LoadDiscoveryClasses
                         }
                     }
 
-                    if ($input instanceof ClassReflector && $input->hasAttribute(DoNotDiscover::class)) {
-                        continue;
-                    }
-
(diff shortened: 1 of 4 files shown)


src/Tempest/Core/src/Kernel/LoadDiscoveryClasses.php
```diff
@@ -95,12 +95,10 @@ public function __invoke(): void
                         }
                     }
 
-                    if ($input instanceof ClassReflector && $input->hasAttribute(DoNotDiscover::class)) {
-                        continue;
-                    }
-
                     if ($input instanceof ClassReflector) {
-                        $discovery->discover($input);
+                        if ($this->shouldDiscover($discovery, $input)) {
```
innocenzi, line 99: Not a fan of the nesting here, but it also avoids repeating the condition above 🫠

brendt:
Nice!

PR #522, opened 2024-10-02
feat: add tagged iterators to container resolution

blackshadev:
Hi, I really like what Tempest brings to the table! When digging around I missed something: tagged iterators [1](https://symfony.com/doc/current/service_container/tags.html) , [2](https://laravel.com/docs/11.x/container#tagging). So after some trying I came up with this one: `#[Tagged("tag-name")]` on both the classes to inject and the parameters where to use them. 

The main use case is to easily append dependencies with the same interface. Very useful when implementing the [chain of responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility) design pattern.

What do you think? Happy to receive any feedback. My "main" gripe is the word "tagged" is already used internally for something different, you also already use the `#[Tag("x")]`  for  something different which can be confusing. Any suggestions?

# Usage

```
interface Inspirator {
    public function inspire(): string;
}

#[Tagged("inspirator")]
final class FirstInspire implements Inspirator
{
    public function inspire(): string
    {
        return "One";
    }
}

#[Tagged("inspirator")]
final class TwoInspire implements Inspirator
{
    public function inspire(): string
    {
        return "Two";
    }
}

final class Inspire
{
    public function __construct(
        private Console $console,
        #[Tagged("inspirator")] private iterable $inspirators
    ) {}

    #[ConsoleCommand]
    public function inspire(): void
    {
        $quotes = array_map(static fn (Inspirator $inspirator) => $inspirator->inspire(), (array)$this->inspirators);

        $this->console->writeln(implode("\n", $quotes));
    }
}
```

### DIFF
--- a/src/Tempest/Container/src/GenericContainer.php
+++ b/src/Tempest/Container/src/GenericContainer.php
@@ -30,6 +30,9 @@ final class GenericContainer implements Container
         /** @var ArrayIterator<array-key, class-string> $initializers */
         private ArrayIterator $initializers = new ArrayIterator(),
 
+        /** @var ArrayIterator<array-key, class-string[]> */
+        private ArrayIterator $taggedDefinitions = new ArrayIterator(),
+
         /** @var ArrayIterator<array-key, class-string> $dynamicInitializers */
         private ArrayIterator $dynamicInitializers = new ArrayIterator(),
         private ?DependencyChain $chain = null,
@@ -39,6 +42,13 @@ final class GenericContainer implements Container
     public function setDefinitions(array $definitions): self
     {
         $this->definitions = new ArrayIterator($definitions);
+
+        return $this;
+    }
+
+    public function setTaggedDefinitions(array $definitions): self
+    {
+        $this->taggedDefinitions = new ArrayIterator($definitions);
 
         return $this;
     }
@@ -62,6 +72,11 @@ final class GenericContainer implements Container
         return $this->definitions->getArrayCopy();
     }
 
+    public function getTaggedDefinitions(): array
+    {
+        return $this->taggedDefinitions->getArrayCopy();
+    }
+
     public function getInitializers(): array
     {
         return $this->initializers->getArrayCopy();
@@ -78,6 +93,16 @@ final class GenericContainer implements Container
 
         return $this;
     }
+
+    /** @param class-string $class */
+    public function tag(string $tag, string $class): self
+    {
+        $this->taggedDefinitions[$tag] = $this->taggedDefinitions[$tag] ?? [];
+        $this->taggedDefinitions[$tag][] = $class;
+
+        return $this;
+    }
+
 
     public function singleton(string $className, object|callable $definition, ?string $tag = null): self
     {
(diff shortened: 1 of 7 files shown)


aidan-casey:
Hey @blackshadev! I appreciate the PR.

We actually have a variant of this feature in the form of tagged singletons (see [tagged singletons](https://tempestphp.com/docs/framework/02-the-container/) in the container documentation). This is exactly what the `Tag` attribute is for. 🙂 

I recommend reading through that portion of the docs and then would be curious to get your feedback.

brendt:
Ok so if I understand correctly, you're grouping multiple dependencies into one array, and inject them as a whole — correct?

I'm not against the idea! But, we'll need to come up with another name. It's true that `#[Tag]` is already used for something else, and I think adding `#[Tagged]` would be rather confusing. 

Another concern is that I how this "grouped configuration" is spread across several files: say you open `FirstInspire`, you don't know which other classes are also part of this group. Now, full disclosure: I've never used this feature in Laravel or Symfony before, so maybe it's not as big as a concern as I'm making it out to be.

Personally, I would prefer to keep this group config together in a single place:

```php
final readonly class InspiratorsInitializer implements Initializer
{
    #[Singleton('inspirators')]
    public function initialize(Container $container): iterable
    {
        return [
            new FirstInspire(),
            new SecondInspire(),
        ];
    }
}
```

In fact, the above example already works if you change the `iterable` return type to `ArrayHelper`:

```php
final readonly class InspiratorInitializer implements Initializer
{
    #[Singleton('inspirators')]
    public function initialize(Container $container): ArrayHelper
    {
        return arr([
            new FirstInspire(),
            new SecondInspire(),
        ]);
    }
}
```

We could also make it so that the container is able to handle tagged primitive types so that you could use iterable or array directly. I'm not against that.

So, tl;dr:

- If we go with an attribute, we'd need something with a different name. Maybe something like `#[DependencyGroup]` or `#[Group]`
- I'm not convinced that using attributes is the right approach here, since it spreads configuration that belongs together across files. But I'm open to further discuss it and would like to hear your thoughts.

blackshadev:
Thanks for your swift en thorough reply!

Using an Initializer in between your dependencies is actually quite a good solution. I didn't thing of that. Personally I would introduce a new class to bind these groups of dependencies together. Most of the time you would have multiple different "types" of tagged-iterators for different usecases. For instance multiple groups of form field validators. Bot that would still work with initializers.

I do think it would be nicer to support native return types, I also hit some troubles with the native typing when I was working on this tagged solution.

Regarding whenever we should merge my approach or not (al be it with different naming): I am unsure, I get and understand your main argument and I could live with your solution. Some references in the docs about it would be nice, but that's about it. Let's leave it for now, I will close this on Friday if nobody (including me) is interested. In the meantime I will try  and find something else to add in your pearl of a framework (like support for native values).

brendt:
> like support for native values

Yeah I'd welcome a PR for that! I'd say native values could only be registered in the container if they are tagged, otherwise there would be no way to differentiate between them. Go for it! 

I'll keep the PR open for another day and think about it some more from my side, though I lean towards not merging it for now

brendt:
Gonna close it for now :) Thanks for thinking along though! I've made an issue to explain what's currently possible in the docs (#534), and I'm definitely open for the primitive type PR :)

Issue #472, opened 2024-09-24
[Feature Request]: make:migration command


brendt:
### Description

A command that generates a migration file

### Benefits

Convenience, see discussion at https://github.com/tempestphp/tempest-framework/pull/464

innocenzi:
Something to think about, a good way to generate classes where we actually want them. It's a hard problem to tackle; Laravel does it wrong in my opinion (tried to work around it [here](https://github.com/laravel/framework/pull/48571)), but I'm sure Tempest could have a clean way of handling that.

aidan-casey:
I think @wulfheart is on to something with `nette/php-generator`. I've used it in the past and it's a joy to use.

brendt:
I agree, especially with how Tempest doesn't require you to follow a fixed structure, it's a feature that shouldn't be taken lightly. I'm a total noob with code generation, so I appreciate all the input people can give. Let's actually widen the scope of this issue and use it as a starting point for a proper code generator.

aidan-casey:
One time I built a code generator using `nette/php-generator` to create an API client for 280 endpoints after some guy on the internet who now tells me he's a noob with code generation that it would probably be the best approach.

gturpin-dev:
Hi there 👋
To be sure to understand everything.
Now we have a tempest/generation package that helps generate php code easily from the ClassManipulator class.
Then is the vision for the future is to take an opinion about a way to give users access to something like an attribute #[AsGeneratorCommand] with some parameters to use the ClassManipulator under the hood ?
We could also have some basic generator class in the CLI to build classes like Commands, Controllers, Models, Views and so on based on stub files which can be overriden.
Then it will be easier to let user create its own generation command based on custom stub files aswell.

innocenzi:
@gturpin-dev I see it that way:

- If possible, have a high-level `#[GeneratorCommand]` that makes it super easy to create a generator command—the underlying method could return a `ClassManipulator` or `ClassGenerator` instance

- Have some kind of lower-level tool (abstract class?) for more advanced use cases

- Have built-in `make:` commands for classes, models, controllers, migration files, and everything that could make sense in Tempest

- Ideally, the `make:` commands would behave just like `tempest publish` (as of #513) and would prompt the path to the file, defaulting to their main namespace.

gturpin-dev:
I totally agree with all of this.

Thinking about how to implement a first step of `#[GeneratorConsoleCommand]` because at the moment a classic command can be made on a class method itself.
[...]

PR #531, opened 2024-10-03
feat(generation): add `ClassManipulator`

innocenzi:
This pull request introduces a new package, `tempest/generation`. It should be the place to implement anything related to PHP file generation and manipulation.

For now, I only implemented a `ClassManipulator`, which is used as follows:

```php
$class = new ClassManipulator(CreateMigrationsTable::class); // also accepts a `ReflectionClass` instance

$class->updateNamespace('App');
$class->removeClassAttribute(CanBePublished::class);
$class->setFinal();
$class->setReadOnly();
$class->setStrictTypes(false);

$code = $class->print();
```

Related to #472
Related to #513


brendt:
Nice! I'm just gonna merge this, we can of course improve on it over time, but what's there makes sense :)

PR #532, opened 2024-10-03
feat: allow object to define how they are mapped to array

Jeroen-G:
I wrote an object that I wanted to be mapped to array differently then just property -> value. PHP supports this with the JsonSerialize interface. So I wrote a simple check (with test) for that!

For the example, I think the test speaks for itself.


brendt:
Makes sense 👍

PR #535, opened 2024-10-04
feat(container): add ability to invoke arbitrary closures

innocenzi:
This pull request adds the ability to invoke closures from the container. Previously, the container only had the ability to invoke class methods.

```php
$this->container->invoke(function (Kernel $kernel) {});
$this->container->invoke(function (string $foo) {}, foo: 'bar');
$this->container->invoke(InvokableClass::class);
$this->container->invoke([ServiceClass::class, 'handle']);
\Tempest\invoke(function (Service $service) {});
```


brendt:
Thanks! Awesome :)

PR #538, opened 2024-10-04
feat(support): improve helpers

innocenzi:
This pull request adds some improvements to `ArrayHelper` and `StringHelper`:

## ArrayHelper

- Added a `explode` constructor for better chaining (I wanted to use that in #513)
- Added support for no callback on the `filter` method—by default, it filters out `false` and `null` values (strictly)
- Added a `dd` method to help with debugging

## StringHelper

- Added `Stringable` support where it made sense
- Added `toString` for better chaining (`__toString` is ugly, `(string)` feels bad)
- Renamed `ld` to `dd` (I learned that we could ignore a line with phpstan, and `dd` is more intuitive to reach for than `ld`)

### DIFF
--- a/src/Tempest/Support/src/ArrayHelper.php
+++ b/src/Tempest/Support/src/ArrayHelper.php
@@ -10,6 +10,7 @@ use Countable;
 use Generator;
 use Iterator;
 use Serializable;
+use Stringable;
 use function Tempest\map;
 
 /**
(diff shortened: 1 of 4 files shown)


aidan-casey:
Nice PR! I'm going to give @brendt the honor of reviewing this one because there some stylistic choices that had his influence here. 🙂

brendt:
Looks good! I don't know about `dd` though, I've grown used to `ld` by now, but I don't have a strong opinion on it. Maybe we can simply provide both? Anyway, happy to merge this, and feel free to share your opinions on it afterwards

PR #539, opened 2024-10-04
refactor(reflection): replace `eval` call in `TypeReflector`

Phil-Venter:
Implemented TypeValidator class to remove the use of eval in the `TypeReflector::accepts()` function. This refactor improves security, adds explicit type validation, and handles built-in, class, and compound types.


aidan-casey:
I've been waiting for this PR, THANK YOU! I'm a little on the fence about having this be in a separate class though. What are the pros and cons?

Phil-Venter:
At first, I used a private function for type-checking, but ran into trouble when dealing with Union and Intersection types. To avoid repeating code, I reorganised things so that validation is handled separately. 

**Pros**
- Makes type validation testable and allows it to be used elsewhere
  - think adding to collection instead of doing a reflection check every time we basically do a string comparison
- avoids making an internal `isBuiltIn` to handle any definition and public user facing `isBuiltIn`

**Cons**
- Increased complexity especially because it's only used in one place right now
- Adds performance overhead because creating a `TypeReflector` is basically creating 2 objects

Granted this can be either abstracted to a trait or Union and Intersection types can create a new `TypeReflector` and validate based on that... Now that I'm typing this out, the latter option sounds a lot cleaner.

Phil-Venter:
Jip, a lot cleaner, and argueably simpler

innocenzi:
May I suggest this PR name?
```
refactor(reflection): replace `eval` call in `TypeReflector`
```

aidan-casey:
Thank you, @Phil-Venter!

PR #540, opened 2024-10-04
feat(event-bus): support closure-based listeners

innocenzi:
> [!IMPORTANT]
> This pull request depends on #535, and it contains its commit. I suggest against reviewing it yet because the diff is inaccurate right now, I will have to rebase on `main` when #535 is merged.

---

This pull request adds support for closure listeners on the `EventBus` implementation.

Right now, only class methods with the `EventHandler` attribute work. This is a good pattern, but sometimes we need one-off listeners that are better inlined in the current code. One example of that is a console command:

```php
#[ConsoleCommand(name: 'migrate:up')]
public function __invoke(): void
{
	$this->migrationManager->up();
}

#[EventHandler]
public function onMigrationFailed(MigrationFailed $event): void
{
	$this->console->error($event->exception->getMessage());
}
```

This code works well, until I want the same thing in a `MigrationDown` command:

```php
#[ConsoleCommand(name: 'migrate:down')]
public function __invoke(): void
{
	$this->migrationManager->down();
}

#[EventHandler]
public function onMigrationFailed(MigrationFailed $event): void
{
	$this->console->error($event->exception->getMessage());
}
```

With these two files, Tempest will discover both `EventHandler` methods and register them—but I don't want it to call both of them when running `migrate:up` or `migrate:down`, just the one in the same file.

This is fixed by adding a closure-based event listener:

```php
#[ConsoleCommand(name: 'migrate:down')]
 public function __invoke(): void
{
	$this->eventBus->listen(MigrationFailed::class, function (MigrationFailed $event) {
		$this->console->error($event->exception->getMessage());
	});

	$this->migrationManager->up();
}
```

Note: I added a `listen` function, but I'm not so sure about it. I personally don't think I'd ever reach for it, but since there was already the `event` function, I added it for consistency. Let me know if you think it should be removed.


aidan-casey:
So... I think this goes to a broader conversation on something we need to figure out.

Consider for example, a microservices setup. I dispatch some event from some other Tempest application or even some other programming language. This shouldn't require a dependency on the Tempest object and I should be able to do something with it.

Another consideration would be a scenario where I want to listen to multiple events. For example, maybe we have an audit log mechanism that takes the event and writes to a read model information about what happened and who did it (we're getting _close_ to event sourcing at this point, I know).

Point being, I'm not saying this addition is a bad one, but I'm not sure it gets to the heart of the issue either.

innocenzi:
> @aidan-casey 
> Another consideration would be a scenario where I want to listen to multiple events. For example, maybe we have an audit log mechanism that takes the event and writes to a read model information.

I'm guessing a middleware would do the trick in the current implementation. If not, are you thinking of a way to listen to multiple events, like so?

```php
#[EventHandler(event: [SomeEvent::class, OtherEvent::class])]
public function handle(SomeEvent|OtherEvent $event): void
{
    // ...
}
```

> @aidan-casey
> Consider for example, a microservices setup. I dispatch some event from some other Tempest application or even some other programming language. This shouldn't require a dependency on the Tempest object and I should be able to do something with it.

I admit I don't see what you mean there. Could you share a snippet so I can understand what you mean?

> @aidan-casey
> Point being, I'm not saying this addition is a bad one, but I'm not sure it gets to the heart of the issue either.

So if I understand correctly, there's another fundamental issue with the current event bus implementation, and this pull request is a kind of workaround to that fundamental issue?

aidan-casey:
> I'm guessing a middleware would do the trick in the current implementation. If not, are you thinking of a way to listen to multiple events, like so?

Yeah, essentially. For example, what if I want to perform some random action (like clear the cache or something) anytime a handful of events are emitted. Right now, that requires a different method for each event we might handle.

Example:

```php
final class ClearTheCacheOrSomething
{
    #[EventHandler]
    public function onDiscoveryStarted(DiscoveryStarted $event): void
    {
        $this->cache->clearOrSomething();
    }

    #[EventHandler]
    public function onPackageInstalled(PackageInstalled $event): void
    {
        $this->cache->clearOrSomething();
    }
}
```

These are just random examples, obviously, but definitely use cases for it.

> I admit I don't see what you mean there. Could you share a snippet so I can understand what you mean?

Really I'm talking about integrating with a message broker. Tempest should allow the picking up/handling of events from a message broker like RabbitMQ, etc. where the event payloads may not be under the control or understanding of Tempest directly.

Ecotone solves this by accepting objects and arrays for deserialized payloads. I don't necessarily love that approach, but it's something to be considering here, I think.

> So if I understand correctly, there's another fundamental issue with the current event bus implementation, and this pull request is a kind of workaround to that fundamental issue?

Potentially, yes.

brendt:
@aidan-casey what you're talking about should be fixed by supporting union types. I know you have a bigger vision for the event bus, but it's really out of scope for 1.0 😁

I think this PR adds a good solution for a real issue, I'm fine merging it. @innocenzi could you rebase with the container changes so that I can review afterwards?

innocenzi:
@brendt rebase done!

src/Tempest/EventBus/src/CallableEventHandler.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if ($this->handler instanceof Closure) {
+            return function (string|object $event) use ($container): void {
+                $container->invoke($this->handler, event: $event);
+            };
+        }
+
+        return function (string|object $event) use ($container): void {
+            $this->handler->handler->invokeArgs(
+                $container->get($this->handler->handler->getDeclaringClass()->getName()),
```
brendt, line 28: Do I get it right that from now on we can also inject any dependency into event handler methods because of this? That was actually on my TODO somewhere, but I never got to it :p

brendt:
Yeah this looks great! Do you also wanna PR some docs afterwards?

src/Tempest/EventBus/src/CallableEventHandler.php
innocenzi: Actually, not right now—only on `EventBus#listen`... but this would be fairly easy to implement, I just forgot about this. The change is as simple as this:

```diff
return function (string|object $event) use ($container): void {
+     $container->invoke($this->handler->handler, event: $event);
-     $this->handler->handler->invokeArgs(
-         $container->get($this->handler->handler->getDeclaringClass()->getName()),
-        [$event],
-     );
```

PR #542, opened 2024-10-04
refactor: improve matchAll string helper

vsergiu93:
I think I did a bad job on the first PR about this helper, this aims to improve the method a bit, and I simplify the no match result.


aidan-casey:
Thanks!

PR #544, opened 2024-10-05
feat(generation): add `ClassGenerator`

innocenzi:
As a follow-up of https://github.com/tempestphp/tempest-framework/pull/531, and as a requirement for https://github.com/tempestphp/tempest-framework/issues/472, this pull request introduces a new `ClassGenerator` in the `tempest/generation` package.

Most of the logic is common to both `ClassGenerator` and `ClassManipulator`, so I extracted it in traits. The main difference is how the underlying structures are initialized.

```php
# New `ClassGenerator`
$class = new ClassGenerator('CreateUsersTable', namespace: 'App');

$class->addImplement(Migration::class);
$class->setFinal();
$class->setReadOnly();

$class->addMethod('up', body: <<<PHP
    return (new \Tempest\Database\QueryStatements\CreateTableStatement('migrations'))
        ->primary()
        ->text('name');
PHP, returnType: QueryStatement::class);

// $class->print();
```

```php
# Existing `ClassManipulator`
$class = new ClassManipulator(CreateMigrationsTable::class);
$class->setNamespace('App');
$class->removeClassAttribute(CanBePublished::class);
$class->removeClassAttribute(DoNotDiscover::class);

// $class->print();
```


src/Tempest/Generation/src/SimplifiesClassNames.php
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private function extractFqcnFromBody(string $body): array
+    {
+        preg_match_all('/(?:\\\\?[A-Za-z_][\w\d_]*\\\\)+[A-Za-z_][\w\d_]*/', $body, $matches);
+
+        return array_filter(array_unique(
+            array_map(fn (string $fqcn) => rtrim(ltrim($fqcn, '\\'), ':'), $matches[0])
+        ));
+    }
```
aidan-casey, line 130: I've done some research on this - this happens because of how Reflection loads the class, converting all the class names to FQCN. Are we able to avoid this by loading the code as a string?

aidan-casey:
I'll admit, I'm on the fence about this one. I think it's a clever idea, but what are the pros and cons over simply providing factory classes that return `nette/php-generator` objects?

innocenzi:
I'm not 100% sure what you mean, so I'll develop more:

I found `nette/php-generator` to have a bit of a learning curve. I had to understand what objects contained the API needed for me to manipulate what I wanted (eg. the concept of "file" and "namespace"). When you get it, it makes sense, but it's still a lot of jumping around different objects.

The "wrapper" approach taken here allows for a controlled and simple API. We could also change the underlying tech if needs be in the future. That being said, maybe we should add interfaces?

Hope I answered the question.

brendt:
> I found nette/php-generator to have a bit of a learning curve. I had to understand what objects contained the API needed for me to manipulate what I wanted (eg. the concept of "file" and "namespace"). When you get it, it makes sense, but it's still a lot of jumping around different objects.
>
> The "wrapper" approach taken here allows for a controlled and simple API. We could also change the underlying tech if needs be in the future. That being said, maybe we should add interfaces?

Personally, I find that a good argument @aidan-casey 😇 I also offers the possibility to easily add more stuff in the future, I reckon we have a lot more code generation features we could add, and I'm all for making it as easy as possible.

brendt:
So @innocenzi , @aidan-casey should we merge this? We can always improve on it more afterwards

Jeroen-G:
I've been toying with this, one thing I am missing is being able to set the type of parameters.

innocenzi:
> @Jeroen-G
> I've been toying with this, one thing I am missing is being able to set the type of parameters.

Thanks! I've added support for this in the form of passing a `name => type` array to `parameters`.

@gturpin-dev I fixed the issue you encountered with the importing of functions. I also added support for manipulating the generated code with `->manipulate(fn (StringHelper $str) => $str)`

brendt:
@innocenzi what's the status here? Can we merge?

innocenzi:
@brendt In my opinion, yes! It's pretty solid already, and improvements can be PR'd later

brendt:
Exactly my thinking as well, merged. 🌊

PR #545, opened 2024-10-05
build: add release script

innocenzi:
This pull request adds a `./bin/release` script that interactively creates a tag and pushes it to Git.

## Flow

The script:

- Checks if the repository is in a clean state
- Check if the current origin and branch is `origin/main` (this could be changed or made dynamic)
- Checks if the user has write access by calling `git push`
- Suggests new versions based on the current version
- Creates and pushes the tag for the chosen version

This will trigger the "Changelog and release" action which will write to `CHANGELOG.md` and create a GitHub release.

> [!WARNING]
> The current version scheme, `1.0-alpha2`, is not semver-compliant. The pre-release identifiers should be separated by dots, and the patch version should be included. It should be `1.0.0-alpha.2`.
>
> However, consistency is better, so I suggest <ins>keeping the current scheme</ins> until we reach `1.0.0`. After that, we can start using this script, which will use semver-compliant version numbers (eg. `1.0.0`, `1.0.1`, `1.1.0`, `2.0.0-alpha.1`, etc)

## Examples

**With the current version** (`1.0-alpha2`)

https://github.com/user-attachments/assets/b7ebc0c3-fb3c-46a8-826a-3a4692e58797

**With a beta version** (`1.0.0-beta.3`)

https://github.com/user-attachments/assets/54549765-ece0-40c5-a964-1c9ae2784209

**With a non-prerelease version** (`1.1.0`)

https://github.com/user-attachments/assets/8375475f-410c-4747-be91-c19355766d8b

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -73,7 +73,7 @@ function releaseTag(string $tag): void
 {
     $commands = [
         "git tag {$tag}",
-        "git push --tags",
+        "git push origin tag {$tag}",
     ];
 
     foreach ($commands as $command) {
--- a/bin/release
+++ b/bin/release
@@ -1,0 +1,136 @@
+#!/usr/bin/env php
+<?php
+
+declare(strict_types=1);
+
+use Composer\Semver\VersionParser;
+use Tempest\Console\Console;
+use Tempest\Console\ConsoleApplication;
+use Tempest\Console\Exceptions\InterruptException;
+use function Tempest\get;
+use function Tempest\Support\arr;
+use function Tempest\Support\str;
+
+require_once getcwd() . '/vendor/autoload.php';
+
+function ensureAccess(string $remote, string $branch): void
+{
+    if (! empty(shell_exec('git status --porcelain 2>&1'))) {
+        throw new Exception('Repository must be in a clean state to release.');
+    }
+
+    if (! str_starts_with(shell_exec('git rev-parse --abbrev-ref --symbolic-full-name @{u}'), "$remote/$branch")) {
+        throw new Exception("You must be on the $remote/$branch branch to release.");
+    }
+}
+
+function getCurrentVersion(): string
+{
+    return exec('git describe --tags --abbrev=0');
+}
+
+function normalizeVersion(string $version): string
+{
+    return preg_replace('/^(\d+\.\d+\.\d+)\.0(-|$|\+)/', '$1$2', (new VersionParser())->normalize($version));
+}
+
+function suggestNextVersions(string $current): array
+{
+    $version = normalizeVersion($current);
+
+    if (! preg_match('/^(\d+)\.(\d+)\.(\d+)(?:-(?:alpha|beta)\.?(\d+))?$/', $version, $matches)) {
+        throw new InvalidArgumentException('Version must be in format X.Y.Z or X.Y.Z-alpha.N or X.Y.Z-beta.N');
+    }
+
+    $isStable = ! isset($matches[4]);
+    $major = (int) $matches[1];
+    $minor = (int) $matches[2];
+    $patch = (int) $matches[3];
+
+    // Current version is stable
+    if ($isStable) {
+        return [
+            'next_patch' => sprintf('%d.%d.%d', $major, $minor, $patch + 1),
+            'next_minor' => sprintf('%d.%d.0', $major, $minor + 1),
+            'next_major' => sprintf('%d.0.0', $major + 1),
+            'next_alpha' => sprintf('%d.0.0-alpha.1', $major + 1),
+            'next_beta' => sprintf('%d.0.0-beta.1', $major + 1),
+        ];
+    }
+
+    // Current version is pre-release
+    $preRelease = $matches[4] ?? 0;
+    $isBeta = str_contains($version, 'beta');
+
+    return array_filter([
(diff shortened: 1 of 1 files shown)


aidan-casey:
Niiiiiiiice. 😍 

A couple thoughts:

First, can you add some support for the [items that monorepo-builder checks](https://github.com/symplify/monorepo-builder?tab=readme-ov-file#7-release-flow) by way of interdependencies, etc.? I'm not saying we should use their script directly, I like this flow better, but it'd be worth incorporating.

Second, I'm going to disagree with your proposal about consistency. This is alpha, we can break what we want (jk, kinda). 🙃 

I think the right approach is to correct the tagging going forward and add a corrected tag in retrospect for the previous releases. We can maintain the existing (incorrect) tags along side the corrected ones.

aidan-casey:
Side note: I really love this. Well done.

innocenzi:
> @aidan-casey 
> First, can you add some support for the items that monorepo-builder checks by way of interdependencies, etc.? I'm not saying we should use their script directly, I like this flow better, but it'd be worth incorporating.

Right now, none of these are used, right? Do you think calling the `qa` script from composer would be enough?

> @aidan-casey 
> I think the right approach is to correct the tagging going forward and add a corrected tag in retrospect for the previous releases. We can maintain the existing (incorrect) tags along side the corrected ones.

That seems great to me as well!

aidan-casey:
> Right now, none of these are used, right?

Right now bumping the inter-dependencies is done manually before release.

bin/release
```diff
@@ -0,0 +1,136 @@
...
+        'next_beta' => sprintf('%d.%d.%d-beta.%d', $major, $minor, $patch, $isBeta ? ($preRelease + 1) : 1),
+        'next_major' => sprintf('%d.0.0', $major + 1),
+    ]);
+}
+
+function releaseTag(string $tag): void
+{
+    $commands = [
+        "git tag {$tag}",
+        "git push --tags",
```
aidan-casey, line 76: Can we update this to `git push origin tag {$tag}`? I'd for this to inadvertently push some other tag.

bin/release
```diff
@@ -0,0 +1,136 @@
...
+    $patch = (int) $matches[3];
+
+    // Current version is stable
+    if ($isStable) {
+        return [
+            'next_patch' => sprintf('%d.%d.%d', $major, $minor, $patch + 1),
+            'next_minor' => sprintf('%d.%d.0', $major, $minor + 1),
+            'next_major' => sprintf('%d.0.0', $major + 1),
+            'next_alpha' => sprintf('%d.0.0-alpha.1', $major + 1),
+            'next_beta' => sprintf('%d.0.0-beta.1', $major + 1),
```
aidan-casey, line 57: Can we update these to include the `v` prefix?

bin/release
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    // Current version is pre-release
+    $preRelease = $matches[4] ?? 0;
+    $isBeta = str_contains($version, 'beta');
+
+    return array_filter([
+        'next_alpha' => ! $isBeta ? sprintf('%d.%d.%d-alpha.%d', $major, $minor, $patch, $preRelease + 1) : null,
+        'next_beta' => sprintf('%d.%d.%d-beta.%d', $major, $minor, $patch, $isBeta ? ($preRelease + 1) : 1),
+        'next_major' => sprintf('%d.0.0', $major + 1),
```
aidan-casey, line 68: > Can we update these to include the `v` prefix?

Same here.

aidan-casey:
> I think the right approach is to correct the tagging going forward and add a corrected tag in retrospect for the previous releases. We can maintain the existing (incorrect) tags along side the corrected ones.

@innocenzi this has been done. I've utilized the semantic versioning scheme and prefixed with `v`. If we can sort out the comments above, we can flesh out the other release steps after merging.

CC: @brendt for visibility.

aidan-casey:
@brendt yes. There is a PR in `docs` let me know if there was something else you wanted updated.

I'll have the rest of the tags cleaned up today. With the revisions requested, I'll get this merged.

@innocenzi can you add a little comment at the end of the contributing file about how releases are handled?

brendt:
> @brendt yes. There is a PR in docs let me know if there was something else you wanted updated.

Go ahead and merge it :) You can deploy by triggering the deploy action: https://github.com/tempestphp/tempest-docs/actions/workflows/deploy.yml

bin/release
innocenzi: Good catch!

aidan-casey:
> Go ahead and merge it

I'll go ahead and do that. Just didn't want to deploy and break the site on your weekend away. 🙃

bin/release
innocenzi: This is [already the case](https://github.com/tempestphp/tempest-framework/pull/545/files#diff-ba4e74f7fdeafe09fab0dd8bd69b66578025911d8ea9403bf70182183fdd6600R115), or do you mean something else?

innocenzi:
> can you add a little comment at the end of the contributing file about how releases are handled?

I'll create another PR for the bumping of inter-dependencies and that 👍

bin/release
aidan-casey: Missed this. Nice!

Issue #537, opened 2024-10-04
[Bug]: tempest serve without port


brendt:
### Tempest Version

1.0-alpha2

### PHP Version

8.3

### Operating System

MacOS

### Description

`./tempest serve 0.0.0.0` gives an error because the port is required. We should add `:8000` if no port is set.

### Steps to Reproduce

/

aidan-casey:
@brendt - I wonder if we should have these be separate parameters to the command/method?

hanieas:
@brendt 
Could we consider adding $port as an argument to the `ServeCommand::__invoke` function? This would automatically handle the port definition, allowing the command to be run with or without specifying a port.

I suggest modifying the ServeCommand like this:
``` php
public function __invoke(string $host = 'localhost', int $port = 8000, string $publicDir = 'public/'): void
{
    passthru("php -S $host:$port -t $publicDir");
}
```
This would allow the command to work as follows:

Without specifying a port:
./tempest serve --host=localhost

With a specific port:
./tempest serve --host=localhost --port=8082

This approach provides more flexibility and handles the port argument dynamically. What do you think?

aidan-casey:
@hanieas - This is definitely the way to do it. Are you willing to submit a PR?

hanieas:
@aidan-casey - Sure! I'll submit PR.

aidan-casey:
Thank you, @hanieas!

hanieas:
@aidan-casey - The [PR](#546 ) is submitted.
I'll send another PR on docs shortly.

aidan-casey:
Thank you!

PR #546, opened 2024-10-05
fix: add optional port argument to serve command(#537)

hanieas:
Tempest serves without specifying a port:
```bash
./tempest serve --host=localhost 
```

With a specific port:

```bash
 ./tempest serve --host=localhost --port=8082 
 ```


aidan-casey:
Thanks, @hanieas! Can you run `composer qa` and re-push before we merge?

Issue #275, opened 2024-05-27
Clean debug log on a daily/weekly basis


brendt:
So that it doesn't grow too big. It's only used as a communication channel between the debug functions and `tempest tail`, so it can be truncated often.

innocenzi:
I'm assigning this to myself, I plan on working on the logger alongside the debug UI, which is currently blocked by proper multiple database support.

PR #548, opened 2024-10-06
fix(#275): implement weekly log rotation

Phil-Venter:
Added weekly log rotation.

This fix creates a `RotatingFileHandler` that extends the `Mono/RotatingFileHandler` and overwrites only the methods required to make this work. Might be considered hacky by some.

### DIFF
--- a/src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
+++ b/src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Log\FileHandlers;
+
+use DateTimeImmutable;
+use InvalidArgumentException;
+use Monolog\Handler\RotatingFileHandler as MonoRotatingFileHandler;
+
+final class RotatingFileHandler extends MonoRotatingFileHandler
+{
+    public const string FILE_PER_WEEK = 'Y-W';
+
+    protected function setDateFormat(string $dateFormat): void
+    {
+        if (0 === preg_match('{^[Yy](([/_.-]?m([/_.-]?d)?)|([/_.-]?W))?$}', $dateFormat)) {
+            throw new InvalidArgumentException(
+                'Invalid date format - format must be one of '.
+                'RotatingFileHandler::FILE_PER_DAY ("Y-m-d"), RotatingFileHandler::FILE_PER_WEEK ("Y-W"), '.
+                'RotatingFileHandler::FILE_PER_MONTH ("Y-m") or RotatingFileHandler::FILE_PER_YEAR ("Y"), '.
+                'or you can set one of the date formats using slashes, underscores and/or dots instead of dashes.'
+            );
+        }
+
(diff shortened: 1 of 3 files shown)


src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        $this->dateFormat = $dateFormat;
+    }
+
+    protected function getNextRotation(): DateTimeImmutable
+    {
+        return match (str_replace(['/','_','.'], '-', $this->dateFormat)) {
+            self::FILE_PER_WEEK => (new DateTimeImmutable('first day of next week'))->setTime(0, 0, 0),
+            self::FILE_PER_MONTH => (new DateTimeImmutable('first day of next month'))->setTime(0, 0, 0),
```
brendt, line 33: Where are these other constants?

src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Log\FileHandlers;
+
+use DateTimeImmutable;
+use InvalidArgumentException;
+use Monolog\Handler\RotatingFileHandler as MonoRotatingFileHandler;
+
+final class RotatingFileHandler extends MonoRotatingFileHandler
```
brendt, line 11: I think this could be refactored to an enum, wouldn't that be cleaner?

brendt (changes requested):

src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
Phil-Venter: They are defined at the top of the base `Monolog\Handler\RotatingFileHandler`

I wanted to create a minimal invasive "wrapper" over `Monolog\Handler\RotatingFileHandler` that just adds support for weekly

image

src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
Phil-Venter: You want to convert `RotatingFileHandler::FILE_PER_WEEK` etc to enums... I can do this, are you sure we want to introduce that much changes to essentially internal monolog functionality, if it ever changes more will have to be updated in that case. Or am I misunderstanding something??

brendt: Oh I totally missed the extend part 😂 No I get it now :)

src/Tempest/Log/src/FileHandlers/RotatingFileHandler.php
brendt: Yeah, totally missed that this class was extending monolog 😁

brendt (approved):

brendt:
Sorry for missing the extension part :) Thanks for contributing!

Issue #379, opened 2024-09-13
Fix open PHPStan failures


aidan-casey:
There are several PHPStan failures that should be fixed prior to alpha next week.

PR #556, opened 2024-10-08
fix(phpstan): fix phpstan issues

blackshadev:
## What 

Related to #379

I resolved some issues related to phpstan. I moved some ignore patterns around, removed unused ignore patterns, fixed some issues.

Here is the full breakdown:

Some are ignored in the "normal" phpstan config file like others. 
- forbidden `exec` in `src/Tempest/Console/src/Terminal/Terminal.php` . Which seems as expected and other exec calls are also ignored this way
- `src/Tempest/Http/src/Exceptions/exception.php` seems to have been moved, so I changed the path for the ignore rule

Others where ignore patterns which were no longer required:
- `NullShellExecutor.php`
- `CircularWithInitializer[ABC].php`
- `ModelDefinition.php`
- I excluded the whole `tests/Integration/View/blade/cache/**.php` path instead of ignoring errors from it.

Lastly I fixed some actual  issues:
- issues in `Status.php`:
  -  Some cases were missing in description, this would cause issues, so jeeej for phpstan spotting it.
  - In some condition a "sanity" check was done on the status value, like `status >= 100` and  `status < 600` . But phpstan doesn't like sanity, it likes strict typing ;) 
- `Route.php`: Fixed some logic you cannot apply a template in the middle of any class. In this case I didn't think it was needed anyway.
- `HttpExceptionHandler.php` :  added a "hack" to ensure `getCodeSample` get's referenced. If you don't like it I can always re-add it to the ignore list
- `HttpExceptionHandler.php`: str_pad got used incorrectly, this actually seemed like a bug? so jeeej for phpstan spotting it.

## Pending issues
- `HttpExceptionHandler.php` :  added a "hack" to ensure `getCodeSample` get's referenced. If you don't like it I can always re-add it to the ignore list
- The whole `A should be final` thing seems weird. I get that you would want to enforce final, but something that is explicitly marked as abstract should be the exception right. From my knowledge phpstan should not report this on abstract classes.

### DIFF
--- a/src/Tempest/Http/src/Status.php
+++ b/src/Tempest/Http/src/Status.php
@@ -109,6 +109,7 @@ enum Status: int
             303 => 'See Other',
             304 => 'Not Modified',
             305 => 'Use Proxy',
+            306 => 'Unused',
             307 => 'Temporary Redirect',
             308 => 'Permanent Redirect',
 
@@ -130,6 +131,7 @@ enum Status: int
             415 => 'Unsupported Media Type',
             416 => 'Range Not Satisfiable',
             417 => 'Expectation Failed',
+            418 => 'I Am A Teapot',
             421 => 'Misdirected Request',
             422 => 'Unprocessable Content',
             423 => 'Locked',
@@ -157,7 +159,7 @@ enum Status: int
 
     public function isInformational(): bool
     {
-        return $this->value >= 100 && $this->value < 200;
+        return $this->value < 200;
     }
 
     public function isSuccessful(): bool
@@ -177,6 +179,6 @@ enum Status: int
 
     public function isServerError(): bool
     {
-        return $this->value >= 500 && $this->value < 600;
+        return $this->value >= 500;
     }
 }
--- a/src/Tempest/HttpClient/src/Driver/Psr18Driver.php
+++ b/src/Tempest/HttpClient/src/Driver/Psr18Driver.php
@@ -28,17 +28,15 @@ final class Psr18Driver implements ClientInterface, HttpClientDriver
 
     public function send(Request $request): Response
     {
+        $psrRequest = $this->convertTempestRequestToPsrRequest($request);
+
         return $this->convertPsrResponseToTempestResponse(
-            $this->sendRequest($request)
+            $this->sendRequest($psrRequest)
         );
     }
 
-    public function sendRequest(Request|RequestInterface $request): ResponseInterface
+    public function sendRequest(RequestInterface $request): ResponseInterface
     {
-        if ($request instanceof Request) {
-            $request = $this->convertTempestRequestToPsrRequest($request);
-        }
-
         return $this->client->sendRequest($request);
     }
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,76 +6,12 @@ parameters:
 			path: src/Tempest/Console/src/Exceptions/ConsoleException.php
 
 		-
-			message: "#^Calling exec\\(\\) is forbidden\\.$#"
-			count: 3
-			path: src/Tempest/Console/src/Terminal/Terminal.php
-
-		-
 			message: "#^Tempest\\\\Database\\\\Exceptions\\\\DatabaseException should be final$#"
 			count: 1
 			path: src/Tempest/Database/src/Exceptions/DatabaseException.php
 
 		-
-			message: "#^Method Tempest\\\\Http\\\\Exceptions\\\\HttpExceptionHandler\\:\\:getCodeSample\\(\\) is unused\\.$#"
+			message: "#^Tempest\\\\Generation\\\\Tests\\\\TestCase should be final$#"
 			count: 1
-			path: src/Tempest/Http/src/Exceptions/HttpExceptionHandler.php
+			path: src/Tempest/Generation/tests/TestCase.php
 
-		-
-			message: "#^Parameter \\$string of function str_pad expects string, float\\|int given\\.$#"
-			count: 1
-			path: src/Tempest/Http/src/Exceptions/HttpExceptionHandler.php
-
-		-
-			message: "#^PHPDoc tag @var contains unresolvable type\\.$#"
-			count: 1
-			path: src/Tempest/Http/src/Exceptions/exception.php
-
-		-
-			message: "#^PHPDoc tag @var does not specify variable name\\.$#"
-			count: 1
-			path: src/Tempest/Http/src/Exceptions/exception.php
-
-		-
-			message: "#^Variable \\$this might not be defined\\.$#"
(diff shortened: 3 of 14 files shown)


src/Tempest/Http/src/Status.php
aidan-casey, line 1: I like having the additions of the descriptions, but I think we should put this in a separate PR.

src/Tempest/Http/src/Status.php
```diff
@@ -157,7 +159,7 @@ public function description(): string
 
     public function isInformational(): bool
     {
-        return $this->value >= 100 && $this->value < 200;
+        return $this->value < 200;
```
aidan-casey, line 162: What's the advantage of this change?

src/Tempest/HttpClient/src/Driver/Psr18Driver.php
```diff
@@ -28,17 +28,15 @@ public function __construct(
...
+        $psrRequest = $this->convertTempestRequestToPsrRequest($request);
+
         return $this->convertPsrResponseToTempestResponse(
-            $this->sendRequest($request)
+            $this->sendRequest($psrRequest)
         );
     }
 
-    public function sendRequest(Request|RequestInterface $request): ResponseInterface
+    public function sendRequest(RequestInterface $request): ResponseInterface
```
aidan-casey, line 38: `Request|RequestInterface` is necessary to be PSR compliant.

tests/Fixtures/Views/rawAndEscaping.php
```diff
@@ -2,7 +2,7 @@
 
 use Tempest\View\GenericView;
 
-/** @var GenericView $this */ ?>
+/** @var GenericView|stdClass $this */ ?>
```
aidan-casey, line 5: @brendt - Can you just confirm this one?

aidan-casey (changes requested):
Thanks, @blackshadev! Just a few comments.

src/Tempest/Http/src/Status.php
blackshadev: These added descriptions are required for phpstan, phpstan actually complained we were missing some cases ;) so why a separate PR?

src/Tempest/Http/src/Status.php
blackshadev: This is the part about phpstan doesn't like sanity checks
> In some condition a "sanity" check was done on the status value, like status >= 100 and status < 600 . But phpstan doesn't like sanity, it likes strict typing ;)

The reason why phpstan errors on this is because it knows $value can be a certain set of values, which start above 100. Meaning `$this->value >= 100`  will always be true. And that is the phpstan error we got.

src/Tempest/HttpClient/src/Driver/Psr18Driver.php
blackshadev: I am very certain that is not true. `Psr\Http\Client\ClientInterface`  has an send method with this signature 
`public function sendRequest(RequestInterface $request): ResponseInterface;`. The `Request` is a `Tempest\Http\Request` which doesn't seem like it has anything to do with PSR.

The phpstan error actually said this, the method was required to be `sendRequest(RequestInterface $request): ResponseInterface` and this actually shouldn't accept `Tempest\Http\Request` which we did provide and this caused an error.

tests/Fixtures/Views/rawAndEscaping.php
blackshadev: for clarification it solves the error caused by `<?= $this->property ?>` . `property` is a property of `GenericView`  thus this will error. By unioning `stdClass` we can add / access any property without phpstan complaining.

brendt: Wouldn't PHPStan know that there's a magic getter on GenericView?

blackshadev: It knows, but you will still get the `Access to an undefined property Tempest\View\GenericView::$property.` error because you should than still define which properties one can get (instead of just any). "normally" you whouldn't be able to get just about anything, there would still be a limit. But we do not know this limit in this case.

brendt: Gotcha, we were ignoring the error before — right? I'm fine with this docblock, seems like a good solution :)

src/Tempest/HttpClient/src/Driver/Psr18Driver.php
brendt: It could actually be that we allowed `Request|RequestInterface` before, then refactored some stuff, and now only `RequestInterface` is needed. Part of the alpha phase 😅 If all tests are green and PHPStan is as well, then I'm fairly certain we're ok to drop `Request`

src/Tempest/Http/src/Status.php
brendt: But how would this be correct? `Status::CONTINUE->isInformational()` will now return `true` while it should return `false`?

src/Tempest/HttpClient/src/Driver/Psr18Driver.php
aidan-casey: Naw, we have to keep this in unless we are going to change some design decisions.

Originally I had it as a PSR only client, but @brendt wanted us to be able to send Tempest request objects as well. This change was made per his request [here](https://github.com/tempestphp/tempest-framework/pull/157#pullrequestreview-1912845989).

brendt:
FYI 

> tests/Fixtures/Views/rawAndEscaping.php

That one can be removed :)

src/Tempest/Http/src/Status.php
blackshadev: I want to believe you , but... `Status::CONTINUE = 100` correct? meaning the old code `$this->value >= 100 && $this->value < 200`  wil evaluate to `true` . The new code of `$this->value < 200` also returns `true`. If it shouldn't return `true` I can change the expression. But functionally my code is the same as the old one. 

PHPStan noted that because Status starts at `100` the expression `$this->value >= 100` is always true. That is why we can remove it. If we should is another topic. But in my eyes, we would never have a status sub 100, so why not.

blackshadev:
> > tests/Fixtures/Views/rawAndEscaping.php

Rebased and removed, thx!

blackshadev:
I will fix the pipeline later today. It seems that some merged fixed (or moved) some phpstan issues which causes it to fail

src/Tempest/Http/src/Status.php
brendt: Ooooh I'm a total fool… I read it as `>=200 and <300` no idea why 😅

Yeah, totally fine this change. Sorry for the noise

brendt:
Thanks!

PR #558, opened 2024-10-08
refactor(http): move request resolution to middleware

NeoIsRecursive:
Saw the `todo` to refactor the request resolution to a middleware so I took a shot at it :)

Got som errors with the invalid response not getting "referer" header from `Request` instances so fixed that as well.

This solution will do one "unnecessary" mapping (if you don't want a generic request), and I don't really know how to get around that since the middleware handler needs an instance of `Request`.

Very new to routing and "framework code" so I understand if it's not the way to do it, just fun to try to contribute!

Thanks and have a nice day :D

### DIFF
--- a/src/Tempest/Http/src/GenericRouter.php
+++ b/src/Tempest/Http/src/GenericRouter.php
@@ -13,13 +13,11 @@ use Tempest\Http\Exceptions\ControllerActionHasNoReturn;
 use Tempest\Http\Exceptions\InvalidRouteException;
 use Tempest\Http\Exceptions\MissingControllerOutputException;
 use Tempest\Http\Mappers\RequestToPsrRequestMapper;
-use Tempest\Http\Responses\Invalid;
 use Tempest\Http\Responses\NotFound;
 use Tempest\Http\Responses\Ok;
 use function Tempest\map;
 use Tempest\Reflection\ClassReflector;
 use function Tempest\Support\str;
-use Tempest\Validation\Exceptions\ValidationException;
 use Tempest\View\View;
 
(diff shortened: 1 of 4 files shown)


brendt:
Thanks! Really appreciate it :) 

One thing I'm wondering: I think my original TODO was only about handling failing requests via a middleware, but keep request resolution itself in the router for now. I haven't looked at this code in ages though, so I'm happy to hear your thoughts on that.

NeoIsRecursive:
> Thanks! Really appreciate it :)
> 
> One thing I'm wondering: I think my original TODO was only about handling failing requests via a middleware, but keep request resolution itself in the router for now. I haven't looked at this code in ages though, so I'm happy to hear your thoughts on that.

My pleasure :D

I think that was my original interpretation as well, but the request validation seems to be done during the mapping so I figured it would be necessary to move the "resolution" to the middleware.

The first thing I tried was to keep the router code as is, but move the `$request->validate()` and try catch to a `ValidateRequestMiddleware`.

But I guess that would mean changes to how requests are mapped and/or validated? Is that an option? If so I could try to get something like this to work instead?

```php
try {
$request->validate(); // throw validation exception here instead of during mapping(????)
return $next($request);
} catch {
// return invalid response
}
```

brendt:
I see… that's indeed a very good point. Hmmm. 

In that case, I don't think we should do it: moving everything to a middleware isn't the right approach. 

Thanks for looking into it though, really appreciate it! Unfortunately I'll close the PR but I hope that doesn't discourage you from contributing in the future 😁

PR #557, opened 2024-10-08
feat(support): improve array helper with additional methods

gturpin-dev:
As discussed in #543 This pull request add some additional methods to the array helper.

As seen in other PR/Issues the goal for this helper isn't to be a clone of laravel collections but a much simpler wrapper around native php's array functions and some other useful methods such as mapWithKeys, pluck etc..

Here's my current work progression :
- [x] put()
- [x] combine()
- [x] keys()
- [x] dump()
- [x] diff()
- [x] diffKeys()
- [x] merge()
- [x] intersect()
- [x] intersectKeys()
- [x] unique()
- [x] flip()
- [x] pad()
- [x] add()
- [x] push()
- [ ] reduce()
- [x] pluck()
- [ ] chunk() ?
- [x] random()
- [ ] pull() ?
- [ ] search() ?
- [ ] slice()
- [ ] sort() => and probably its variations
- [ ] flatten() => I must check if unwrap isn't doing the same before
- [ ] wrap() ?
- [ ] remove()
- [ ] forget() => alias of remove()
- [ ] shuffle

If i'm doing something wrong please tell me 🙏

PR #563, opened 2024-10-11
feat(support): refactor dd() method to use logger in string helper and add the dump() method

gturpin-dev:
Changes `dd()` method in `str()` helper to use the Logger to be consistent with the changes made in array helper.
Also added the `dump()` one for same purposes.

## Benefits
Have better debugging and be tracked in logs

### DIFF
--- a/src/Tempest/Support/src/StringHelper.php
+++ b/src/Tempest/Support/src/StringHelper.php
@@ -413,7 +413,8 @@ final readonly class StringHelper implements Stringable
         ld($this->string, ...$dd); // @phpstan-ignore-line
     }
 
-    public function dump(mixed ...$dumps): self {
+    public function dump(mixed ...$dumps): self
+    {
         lw($this->string, ...$dumps); // @phpstan-ignore-line
 
         return $this;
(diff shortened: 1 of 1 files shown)


src/Tempest/Support/src/StringHelper.php
```diff
@@ -410,7 +410,14 @@ public function matches(string $regex): bool
 
     public function dd(mixed ...$dd): void
     {
-        dd($this->string, ...$dd); // @phpstan-ignore-line
+        ld($this->string, ...$dd); // @phpstan-ignore-line
```
innocenzi, line 413: I'm confused—doesn't `dd` use `ld` under the hood? 

https://github.com/tempestphp/tempest-framework/blob/f54a0e0f30fcc9a7ea7b041d4a38e20181161aa2/src/Tempest/Debug/src/functions.php#L31-L37

gturpin-dev: In theory yes.
But check this discuss here https://github.com/tempestphp/tempest-framework/pull/557#issuecomment-2405043570
In fact the `if (! function_exists())` comes after composer has load the `symfony/var-dumper` and so it doesn't run actually

innocenzi: Oh, right! Thank you

gturpin-dev: You're welcome 😉

brendt:
Thanks!

PR #564, opened 2024-10-11
feat(support): add implode()/explode() methods in string helper

gturpin-dev:
This PR add both `implode()` and `explode()` method in string helper with consistency regarding the ones in array helper.

It also update the return type of `implode()` in array helper to `StringHelper` object for consistency


brendt:
Nice :)

PR #567, opened 2024-10-11
feat(console): support string keys in `MultipleChoiceComponent`

innocenzi:
This pull request adds support for using string keys in the options array of `MultipleChoiceComponent`. 

This is useful to have the component display something but return a different value (see https://github.com/tempestphp/tempest-framework/pull/513).


brendt:
Nice :)

PR #568, opened 2024-10-11
feat(generator/console): add make:controller, make:model commands and the generator attribute

gturpin-dev:
This PR is a starting point of creating the `make commands` for the framework using the internal Generation Component.
Here's the related issue #472

It add a make:controller command 

- it take a classname as an argument like `Test`
- it will suffix the class with `Controller` like `TestController.php`
- it will prepend the main autoloaded namespace/path from the composer file to it
- it will then glue them with a `Controllers` directory
- it will generate the right namespace PSR-4 compatible
- it support adding subfolders directly in the `classname` arg from the command

So, for example if I have a project with a default autoloaded namespace from composer like :
`my-project/app` => `App`
And I run `make:controller Admin/Dashboard`

It will generate a class named `DashboardController` at `my-project/app/Controllers/Admin/DashboardController.php`
And with the following namespace `App\Controllers\Admin`

---
Next steps before duplicate this into others make commands are :

- [x] Replace the way to retrieve the root path of the final user project
- [ ] Find a way to test the file is properly generated and at the right location, with the right classname and namespace
- [x] Extract the main generation logic to the Generator component
- [ ] ( later ) refactor the file creation logic by the Filesystem component
- [x] Handle Dummy replacements in Stubs file
- [x] Add `MakeControllerCommand` specifics params like the route path and the view path

For the 2 first TODOs I'm looking for ideas, if someone have any.


gturpin-dev:
@innocenzi For now, I'm not sure of the purpose of the `ClassGenerator` but I think the main logic could go there right ?

aidan-casey:
@brendt are we good moving this outside the framework to the CLI project?

innocenzi (commented):
I haven't tested this locally, but as I understand it, it will generate a controller in a predefined directory, right?

I don't really like this, personally, and I don't think it follows the philosophy of Tempest, where everything can be anywhere. 

My implementation of [`tempest publish`](https://github.com/tempestphp/tempest-framework/pull/513) prompts the user every time for the path at which the file would be published. If the file already exists, the user is asked if the file should be replaced. What do you think of this approach?

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,73 @@
...
+    )]
+    public function __invoke(
+        #[ConsoleArgument(
+            help: 'The name of the controller class to create ( "Controller" will be suffixed )',
+        )]
+        string $classname
+    ): void {
+
+        // @TODO replace '' with the project root path ( maybe from the .ENV or something like that )
+        $project_namespace = (new Composer(''))->mainNamespace;
```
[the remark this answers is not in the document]
gturpin-dev: Cool, I'll check how this is implemented then, thanks !

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,73 @@
...
+
+        // @TODO Need to Extract the whole logic to a service/helper ( maybe the ClassGenerator ) to be able to test/use it separately for other commands
+        // anyways, this should be moved to the Generation Component
+        // The Console component should only be responsible for the input/output around the command
+
+        $pathPrefix    = 'Controllers';
+        $classSuffix   = 'Controller';
+        $fullNamespace = NamespaceHelper::make($pathPrefix . DIRECTORY_SEPARATOR . $classname);
```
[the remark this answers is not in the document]
gturpin-dev: I probably miss it then, I only find the `static function make` in the `PathHelper` which ones are you talking about ?

src/Tempest/Generation/tests/ClassGeneratorTest.php
```diff
@@ -50,8 +50,8 @@ public function creates_methods_with_parameters(): void
             //
         PHP, parameters: ['id' => 'int'], returnType: '?App\\Models\\User');
 
-        $this->assertMatchesSnapshot($class->print());
-    }
+        $class->simplifyImplements();
+        $class->addImplement(FakeMigration::class);
```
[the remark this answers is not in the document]
gturpin-dev: I think it's a mistake from my own.
To work with your updated `ClassManipulator` i've merged your waiting PR fork and then rebase on the main this day, I think i've done something wrong because I didn't change any of your files.

I'll fix that later, thanks for notice it 🙏

gturpin-dev:
> I haven't tested this locally, but as I understand it, it will generate a controller in a predefined directory, right?
> 
> I don't really like this, personally, and I don't think it follows the philosophy of Tempest, where everything can be anywhere.
> 
> My implementation of [`tempest publish`](https://github.com/tempestphp/tempest-framework/pull/513) prompts the user every time for the path at which the file would be published. If the file already exists, the user is asked if the file should be replaced. What do you think of this approach?

No problem for me, I can follow your already approved work, I'll check how your command work !
Then, we assume that every `make` command will always ask for the dist folder, we could imagine to give the main namespace path as an indication in the prompt if the user has no idea.
I'll try to go in this way ( must think about how to generate the right namespace for the file then, maybe show a select for all composer.json autoload or ? )

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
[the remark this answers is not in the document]
innocenzi: My bad! This is something I added in a PR that wasn't merged yet. 🤦 

Sorry about this. Here is what I was referring to: https://github.com/tempestphp/tempest-framework/pull/513/files#diff-cf35c13fa02d83f45cd40817d4baa907e175d3be1c9de9678db4c8399e984917

brendt:
> are we good moving this outside the framework to the CLI project?

Will discuss on GitHub

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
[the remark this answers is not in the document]
gturpin-dev: This looks definitely good !

src/Tempest/Generation/tests/ClassGeneratorTest.php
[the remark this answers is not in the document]
gturpin-dev: I can't find a real version of this file, the entire method has no assertions so it break my tests, I've commented it for now.
Maybe an unmerged work 🤷‍♂️

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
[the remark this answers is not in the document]
gturpin-dev: I've merged it to my branch to work with it, so it's updated 👌

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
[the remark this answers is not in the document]
gturpin-dev: I've used it like your recommmandation !

gturpin-dev:
Don't worry about commits, I've pulled the unmerged work of Enzo to refactor the code with its helpers on namespaces and `ClassManipulator`.

Then I tried to stick as much as possible to the implementation of the `publish` command.
So actually there is lot of duplicate code between both which is good, because it means we can now abstract those mechanism to create commands that generate files faster.
I'll look into that in the followings days.

gturpin-dev:
As discussed in Discord here https://discord.com/channels/1236153076688359495/1295433932027990027/1298306566335893525
This PR is "ready", we are waiting to see the organization of the "roadmap" for the feature

composer.json
```diff
@@ -150,5 +150,10 @@
             "composer phpunit",
             "composer phpstan"
         ]
+    },
+    "config": {
```
brendt, line 154: Could you remove this? To make sure it doesn't get added again, rebase or merge main into your branch, then run `composer up` once more.

src/Tempest/Console/src/Discovery/ConsoleCommandDiscovery.php
```diff
@@ -41,7 +42,7 @@ public function createCachePayload(): string
 
     public function restoreCachePayload(Container $container, string $payload): void
     {
-        $commands = unserialize($payload, ['allowed_classes' => [ConsoleCommand::class, MethodReflector::class]]);
+        $commands = unserialize($payload, ['allowed_classes' => [GeneratorCommand::class, ConsoleCommand::class, MethodReflector::class]]);
```
brendt, line 45: This is one of the reasons I'm hesitant about "a special kind of command class": we need several changes all over the place to make sure they work. That leads to a brittle and complicated design.

src/Tempest/Generation/tests/ClassGeneratorTest.php
```diff
@@ -37,22 +37,6 @@ public function creates_class_from_scratch(): void
         $this->assertMatchesSnapshot($class->print());
     }
 
-    #[Test]
```
brendt, line 40: Why the removal?

src/Tempest/Generation/src/StubFileGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @param string $targetPath The path where the generated file will be saved including the filename and extension.
+     * @param class-string $stubFile The stub file to use for the generation.
+     * @param array<string, string> $replacements An array of key-value pairs to replace in the stub file.
+     *     The keys are the placeholders in the stub file (e.g. 'DummyNamespace')
+     *     The values are the replacements for the placeholders (e.g. 'App\Models')
+     * @param bool $shouldOverride Whether the generator should override the file if it already exists.
+     */
+    public function __construct(
```
brendt, line 28: Is there a reason why this data is injected into the constructor instead of being passed to the generate method? Tempest tries to have all classes injectable, and aims to have a clear separation of data and functionality. When I see a "Stub File Generator", my assumption is that it's a class that I can inject and reuse, while its `generate` method would require me to pass in context-specific data.

Something like this:

```php
$this->stubGenerator->generate($stub, $target, $replacements, $shouldOverwrite);
```

src/Tempest/Reflection/src/ClassReflector.php
```diff
@@ -72,6 +72,11 @@ public function getName(): string
         return $this->reflectionClass->getName();
     }
 
+    public function getFilePath(): string|false
```
brendt, line 75: Let's change this to `string`. It'll only return `false` if we're reflecting over internal PHP classes. I'd rather throw an exception in that case:

```php
$path = $this->reflectionClass->getFileName();

if ($path === false)
{
    // throw
}

return $path;
```

src/Tempest/Reflection/src/TypeReflector.php
```diff
@@ -42,7 +42,7 @@
     private bool $isNullable;
 
     public function __construct(
-        private PHPReflector|PHPReflectionType|string $reflector,
+        private PHPReflector|PHPReflectionType|string|null $reflector,
```
brendt, line 45: Out of curiosity: when does this happen?

src/Tempest/Support/src/PathHelper.php
```diff
@@ -4,8 +4,67 @@
...
 namespace Tempest\Support;
 
+use Exception;
+use Tempest\Core\Composer;
+use Tempest\Core\Kernel;
+use function Tempest\get;
+
 final readonly class PathHelper
 {
+    public static function root(string ...$paths): string
```
brendt, line 14: I don't think we need this one anymore. Use `root_path()` instead

brendt (changes requested):
My main concern is all the hoops we're jumping through to be able to return a `StubFileGenerator` from a command. There's a lot of complexity that comes with it, which could be avoided if we decide to rework the generator a bit, and instead inject it into the console command.

src/Tempest/Support/src/PathHelper.php
gturpin-dev: You're right, i'll correct this ASAP

src/Tempest/Reflection/src/ClassReflector.php
gturpin-dev: I'll check this out because I don't remember tweaking on this, so maybe a vestige of the `feat/publish`

src/Tempest/Generation/src/StubFileGenerator.php
gturpin-dev: This is because atm the generator isn't a service but something like a DTO and the generate method is expected to be run once.
The instance is created and returned by the `GeneratorCommand` method handler.
In case, if we decide to change the overall approch this could be changed as you mention I guess 😄

brendt:
Let's close this one in favor of #647

PR #570, opened 2024-10-11
feat(validation): support validating by closure

innocenzi:
This pull request adds support for validating values using closures:

```php
$validator->validateValue('foo+1@gmail.com', function (mixed $value) {
    if ((str_contains('@gmail.com', $value) && str_contains('+', $value)) {
        return 'Gmail aliases are not allowed';
    }

    return true;
});
```

I've had this need when writing the release script and when working on the publish command. Writing a class was quite cumbersome in these situations.

I've opened a discussion about this, among other things, here: https://discord.com/channels/1236153076688359495/1294321824498323547

### DIFF
--- a/src/Tempest/Validation/src/Validator.php
+++ b/src/Tempest/Validation/src/Validator.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace Tempest\Validation;
 
+use Closure;
 use Tempest\Reflection\ClassReflector;
+use function Tempest\Support\arr;
 use Tempest\Validation\Exceptions\InvalidValueException;
 use Tempest\Validation\Exceptions\ValidationException;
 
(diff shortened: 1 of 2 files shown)


Phil-Venter:
Adding another attribute that takes a closure might also not be a bad idea...

```php
final class Book
{
    public function __construct(
        #[Custom(
            fn ($val) => str($val)->pluralize(2)->equals($val),
            'Book names should be plural'
        )]
        public string $name,
    ) {
    }
}
```

The implementation also does not need to be complex

```php
<?php

declare(strict_types=1);

namespace Tempest\Validation\Rules;

use Attribute;
use Closure;
use Tempest\Validation\Rule;

#[Attribute(Attribute::IS_REPEATABLE)]
final readonly class Custom implements Rule
{
    public function __construct(
        private Closure $callable,
        private string $message = '',
    ) {
    }

    public function isValid(mixed $value): bool
    {
        return call_user_func($this->callable, $value);
    }

    public function message(): string
    {
        return $this->message;
    }
}
```

innocenzi:
I like the idea, and it would be better than what we have now, but between those two I prefer the version with less boilerplate:

```php
$validator->validateValue('foo+1@gmail.com', function (mixed $value) {
    if (str_contains('@gmail.com', $value) && str_contains('+', $value)) {
        return 'Gmail aliases are not allowed';
    }

    return true;
});
```

```php
$validator->validateValue(
    value: 'foo+1@gmail.com',
    rules: new Custom(
        callback: function (mixed $value) {
            return str_contains('@gmail.com', $value) && str_contains('+', $value);
        },
        message: 'Gmail aliases are not allowed'
   )
);
```

Also, side-note, related to the discussion I linked above, in order to have more context about the message, the second argument would also need to be a callback

brendt:
I think both should be possible: 

```php
$validator->validateValue('foo+1@gmail.com', function (mixed $value) {
    if (str_contains('@gmail.com', $value) && str_contains('+', $value)) {
        return 'Gmail aliases are not allowed';
    }

    return true;
});
```

Definitely has its use cases, but it would also nice to have an attribute that takes a closure. That could be another PR though

brendt:
Also: thanks for the PR :)

PR #571, opened 2024-10-11
feat(validation): add closure driven rule

Phil-Venter:
> I think we should also support closure-based rules. I've wanted that a few times, once when writing the release script, and another time when working on the publish command. Writing a class is a bit cumbersome in that situation.
> Enzo

### DIFF
--- a/src/Tempest/Validation/src/Rules/Custom.php
+++ b/src/Tempest/Validation/src/Rules/Custom.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Validation\Rules;
+
+use Attribute;
+use Closure;
+use Tempest\Validation\Rule;
+
+#[Attribute(Attribute::IS_REPEATABLE)]
+final readonly class Custom implements Rule
+{
+    public function __construct(
+        private Closure $callable,
+        private string $message = '',
+    ) {
+    }
+
+    public function isValid(mixed $value): bool
+    {
+        return call_user_func($this->callable, $value);
+    }
+
+    public function message(): string
+    {
+        return $this->message;
+    }
+}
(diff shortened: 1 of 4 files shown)


Phil-Venter:
RE-opened because of https://github.com/tempestphp/tempest-framework/pull/570#issuecomment-2408387865

src/Tempest/Validation/src/Rules/Custom.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Validation\Rules;
+
+use Attribute;
+use Closure;
+use Tempest\Validation\Rule;
+
+#[Attribute(Attribute::IS_REPEATABLE)]
+final readonly class Custom implements Rule
```
brendt, line 12: Nice PR! The only thing I'm not convinced about it the name. Any ideas for something better? Maybe `Validate` ? Maybe something else?

brendt (changes requested):

src/Tempest/Validation/src/Rules/Custom.php
innocenzi: Maybe `CallbackRule` or `Callback`?

Phil-Venter: @innocenzi @brendt you apposed to `LambdaValidation` or is that too obscure??

innocenzi: Lambda is not a term used much in PHP-land to refer to closures, so I'm personally not a fan of it 😞

Phil-Venter: Alright, point taken, will be using `Callback` and based on the namespace it should be obvious...

brendt:
Uh, I just had a realisation: is it even possible to pass closures to an attribute? I don't think it is? https://3v4l.org/kCoEl

PR #577, opened 2024-10-12
fix: low terminal frame rate causing keystrokes to be dropped

brendt:


brendt:
@innocenzi was there an issue somewhere for dropped keystrokes? This PR fixes it :)

PR #582, opened 2024-10-13
chore: cache refactor

### LINKED ISSUE #395: Discovery cache improvements
Let's look into whether we can improve discovery cache for local development

- [x] ~Partial discovery cache: only discover files that have been changed or added according to local git history~ won't do this for now
- [ ] Always cache vendor discovery: vendor files can only be changed on composer update/install. Let's add a post-install script that will re-store discovery cache, but let's not re-scan vendor files when not needed
- [x] Have a discovery cache manager so that paths don't have to be hard-coded in each discovery class
- [x] ~Store discovery cache as plain PHP instead of manually serializing it~

### LINKED ISSUE #561: [Feature Request]: Disabled caches
### Description

Currently there are two env variables `DISCOVERY_CACHE` and `CACHE`, and they both work differently:

- Discovery cache isn't using `Cache` yet
- `CACHE` disabled both `GenericCache` and `ViewCache`

I think we need a better approach: a way to toggle individual caches, as well as disable cache altogether via env

### Benefits

/

brendt:
closes #561 

TODO:

- [x] `ViewCache` should be _disabled_ by default
- [x] `DiscoveryCache` should use `Cache`
- [x] `DiscoveryCache` should be disabled by default
- [x] We need individual properties to enable or disabled cache: `DISCOVERY_CACHE=`, `VIEW_CACHE=`, `GENERIC_CACHE=` — OR we need one flag to enable/disable everything — to be determined
- [ ] Update docs (cache and view pages)
- [ ] Update docs for removed `discovery:cache` command
- [ ] Update docs for added `cache:status` command


brendt:
I need a sanity check here, does this make sense?

```env
# The CACHE key is used as a global override to turn all caches on or off
# Should be true in production, but null or false in local development
CACHE=null

# Enable or disable discovery cache
DISCOVERY_CACHE=false

# Enable or disable view cache
VIEW_CACHE=false

# Enable or disable project cache (allround cache)
PROJECT_CACHE=false
```

aidan-casey:
@brendt to me that feels a little misleading/convoluted. If anything, I could see doing something like:

```env
CACHE=true

DISCOVERY_CACHE=${CACHE}
```

But even that feels slightly icky to me.

innocenzi:
I can see value in only `VIEW_CACHE` and `DISCOVERY_CACHE` honestly. Having more than that becomes kind of confusing

brendt:
So, for the record, I'm totally fine with only having one toggle for all caches. But I'm a bit afraid that it won't be flexible enough. That being said, we can always make ik more flexible down the road if needed. 

In that case though, I don't think we need any separate cache toggles, and only have `CACHE=true/false`

brendt:
Did some more testing, it's actually pretty crucial being able to toggle individual caches on and off for local development (testing comes to mind: you want discovery cache on, but view cache off most of the time; while project cache will depend on what you're testing)

PR #586, opened 2024-10-13
refactor(validation): improve readability of `validateValue`

innocenzi:
Slight refactor of my previous PR, for readability. I was going to continue the work on the validation messages, but I've decided I didn't like the API I came up with, so I'll do it later and just push this change in the meantime.


brendt:
Nice :)

PR #587, opened 2024-10-13
build: improve release script

innocenzi:
This pull request improves on https://github.com/tempestphp/tempest-framework/pull/545 by adding a few steps when a major release happens:

- Inter-dependencies will be bumped through monorepo-builder
- Dependencies will be checked through monorepo-builder
- A release commit with the new dependencies will happen

The rest of the flow is unchanged. I added a `composer release` script as well—maybe `composer qa` should be called directly from within the release script, but we'd make it dependent on composer script that could be changed anytime, then.

Notes:
- I'm not sure if the dependency versions should go back to `dev-main` after the actual release
- `monorepo-builder validate` fails because `tempest/highlighter` has version mismatches inside a few packages in the repository

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -69,13 +69,8 @@ function suggestNextVersions(string $current): array
     ]);
 }
 
-function releaseTag(string $tag): void
+function executeCommands(array $commands): void
 {
-    $commands = [
-        "git tag {$tag}",
-        "git push origin tag {$tag}",
-    ];
-
     foreach ($commands as $command) {
         exec($command, result_code: $code);
 
@@ -83,7 +78,7 @@ function releaseTag(string $tag): void
             continue;
         }
 
-        throw new Exception("Pushing of git tag failed.");
+        throw new Exception("Command failed: {$command}");
     }
 }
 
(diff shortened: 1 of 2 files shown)


brendt:
Who should review this? @aidan-casey is probably better than me?

aidan-casey:
Yeah, this one should get assigned to me. I'll take a look today.

aidan-casey:
@innocenzi - for the time being, what do you think about checking whether the version is correct and throwing an error/exception if not? Then bumping needs to be done in an intentional commit and we can ignore the highlighter errors for now.

brendt:
Folks, we're releasing alpha.3 in three drays on October 31st. Do we need to bump this?

innocenzi:
I missed the notification somehow!

> @aidan-casey
> For the time being, what do you think about checking whether the version is correct and throwing an error/exception if not?

I'm not sure what this implies. You mean ensuring `./vendor/bin/monorepo-builder validate` passes, but ignoring the `tempest/highlight` errors?

> @brendt
> Folks, we're releasing alpha.3 in three drays on October 31st. Do we need to bump this?

I don't think so, since it won't be a major version

innocenzi:
For now, I'll close this. Let's continue tracking it in #676.

PR #589, opened 2024-10-14
fix(phpstan): fix last phpstan issues

blackshadev:
I found a solution for the last "class should be final" phpstan errors.

Normally you would either have a class which is known abstract or final. So I baked that rule in the existing architecture tests.

I also removed the exception for "integrationtest" since it already is abstract.


brendt:
Fair enough!

PR #591, opened 2024-10-17
fix: nullable properties not seen as nullable by TypeReflector

### LINKED ISSUE #594: [Bug]: Nullable properties not seen as nullable by TypeReflector
### Tempest Version

1.0-alpha2

### PHP Version

8.3

### Operating System

MacOS

### Description

After updating from alpha1 to alpha2 in a basic todo app I started getting `MissingValue` exceptions when accessing null database values in a model where the property is nullable and had a default value of null. 

```php
final class Todo implements DatabaseModel
{
    use IsDatabaseModel;

    public function __construct(
        public string        $title,
        public ?string $description = null,
    )
    {}
}
```

In tracing from `QueryToModelMapper` > `ClassReflector` > `TypeReflector` I found that `TypeReflector` would only set the `definition` for my description field as `string` and not `?string` or the union `null|string`. 

I submitted #591  to confirm the issue. I have a PR to submit with a fix but Im waiting for this to get confirmed.

### Steps to Reproduce

```php

final class Todo implements DatabaseModel
{
    use IsDatabaseModel;

    public function __construct(
        public string        $title,
        public ?string $description = null,
    )
    {}
}

Todo::create(title: "test");
ll(Todo::query()->first()->description); //throws MissingValue
```php

ainesophaur:
This contains a failing test to highlight the issue when checking a class with a nullable property that was mapped by TypeReflector

closes #594

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,17 +1,2 @@
 parameters:
-	ignoreErrors:
-		-
-			message: "#^Tempest\\\\Console\\\\Exceptions\\\\ConsoleException should be final$#"
-			count: 1
-			path: src/Tempest/Console/src/Exceptions/ConsoleException.php
-
-		-
-			message: "#^Tempest\\\\Database\\\\Exceptions\\\\DatabaseException should be final$#"
-			count: 1
-			path: src/Tempest/Database/src/Exceptions/DatabaseException.php
-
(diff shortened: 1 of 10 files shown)


brendt:
Thanks! I added a fix for it in your branch before merging :)

PR #604, opened 2024-10-20
fix(http): fix so referer header is resolved depending on request class in invalid response

NeoIsRecursive:
This PR fixes `Call to undefined method Tempest\Http\GenericRequest::getHeader()` error that happens when the request is not an instance of `PsrRequest` when you return a `Invalid` response.

This also happens when you throw a `ValidationException` from a controller. (and according to the signature it should work 😬)


brendt:
Perfect, thanks!

PR #606, opened 2024-10-21
refactor(http): resolve `HttpApplication` via the container

yassiNebeL:
resolving `HttpApplication` via the container, and setting-up the `Singleton` attribute.


brendt:
Cool, thanks!

PR #607, opened 2024-10-21
feat(core): add root_path helper

NeoIsRecursive:
When working with tempest I found myself missing something like laravels `base_path` ([here](https://laravel.com/docs/11.x/helpers#method-base-path)).

This PR adds that to Tempest 😄

One question if this is something Tempest should provide: should functions tempest "exports" be snake_case?


innocenzi:
> @NeoIsRecursive 
> One question if this is something Tempest should provide: should functions tempest "exports" be snake_case?

In my opinion, yes—just like PHP's own functions.

I have two comments:
- What do you think of keeping this in `PathHelper`, instead of as a function?
- What do you think of naming this `root` instead of `base`? To stay consistent with the naming already used in the kernel

NeoIsRecursive:
> What do you think of keeping this in `PathHelper`, instead of as a function?

I thought about that (if you meant something like this `PathHelper::root()`), but that would add a new dependency to `tempest/support` (?).

And maybe its just me, but I could imagine `PathHelper::root('/path/to/somewhere')` being a method to get the root of the path passed in, instead of building a path starting from project root. If that makes any sense?

> What do you think of naming this `root` instead of `base`? To stay consistent with the naming already used in the kernel
This I agree with :)

NeoIsRecursive:
> Two fair points, honestly. The intent of `PathHelper::root` can be solved by docblocks, but the dependency concern is valid.
> 
> Since `Tempest\path` already exists, I think `Tempest\root_path` is fine to have, then 👍

Since you mentioned docblocks, should I add a docblock to the function?

"Get the fully qualified path or a path relative to your application's root directory" or something like that.

I kinda like it when you get a description while hovering, even if its self explanatory. What do you think?

innocenzi:
I definitely think that—see https://github.com/tempestphp/tempest-framework/pull/611 😄 

That being said, maybe that should be in another PR, with all other functions documented as well for consistency

NeoIsRecursive:
Nice!

I'll leave it like it is then 😄 

Thanks for taking the time!

PR #611, opened 2024-10-22
feat(support): add inline documentation on helper classes

innocenzi:
This pull request adds inline documentation on all methods of `StringHelper`, `ArrayHelper` and `LanguageHelper`.

Not all method names are explicit enough to understand the behavior of the method, so inline documentation helps a lot.


brendt:
Thanks!

PR #612, opened 2024-10-22
refactor(support): move `LanguageHelper#join` to `ArrayHelper` and `StringHelper`

innocenzi:
This pull request moves `LanguageHelper#join` to `ArrayHelper`, and adds a passthrough static method to `StringHelper` as well.

This makes it more consistent, since `implode` already exists in both helper classes. More flexibility has been added to `join` as well (`and` is no longer hard-coded).


brendt:
Nice :)

PR #616, opened 2024-10-23
feat: add inline documentation on all namespaced functions

innocenzi:
For some of them, I also added example snippets. 

Generally speaking, I think we should aim towards adding inline documentation on all public APIs—this really, really improves the developer experience when we don't already know about something.


src/Tempest/CommandBus/src/functions.php
```diff
@@ -6,6 +6,9 @@
 
     use Tempest\CommandBus\CommandBus;
 
+    /**
+     * Dispatches the given `$command` to the {@see CommandBus}, triggering all associated command handlers.
```
aidan-casey, line 10: Does @see need to be fully qualified?

innocenzi: I think so! At least in VSC. But php-cs-fixer isn't happy with that, it simplifies the FQCN. Can you tell me if that works in PHPStorm?

aidan-casey: I'll double check PHPStorm. CS-fixer is probably something we could change if necessary/there is a comment specific rule.

brendt:
I'm good with this, really nice!

brendt:
@aidan-casey did you plan on figuring out that PHP CS Fixer rule? Just wanna make sure this can get merged soon :)

brendt:
I'm just going to merge this, we can finetune with new PRs later

PR #617, opened 2024-10-24
feat(console): add `name` parameter to `#[ConsoleArgument]`

### LINKED ISSUE #588: [Feature Request]: Maybe add a name to the ConsoleArgument attribute to override the parameter name
### Description

Actually on console commands, the arguments are named from the parameters name of the method itself.
Because we have a `#[ConsoleArgument]` I wonder why there is no `name` param in the attribute to override this behavior.
Is there a reason for this ?
I think it could be nice to take the param name by default but let the user override it if he wants. It can help to have different variable name inside the method and the ones in the console.

### Benefits

In case you want to be more precise in your method param you could do it without alterate the simplicity of the command description in console like when running `tempest` we have all commands and their params named as the method.

gturpin-dev:
This PR allow `ConsoleArgument` Attribute to have a `name` parameter to override the method parameter name if wanted.

This Fix #588

### DIFF
--- a/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
+++ b/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
@@ -7,7 +7,6 @@ namespace Tempest\Console\Actions;
 use Tempest\Console\Console;
 use Tempest\Console\ConsoleCommand;
 use Tempest\Console\Input\ConsoleArgumentDefinition;
-use Tempest\Reflection\ParameterReflector;
 
 final readonly class RenderConsoleCommand
 {
@@ -35,12 +34,12 @@ final readonly class RenderConsoleCommand
         $name = str($argument->name)
             ->prepend('<em>')
             ->append('</em>');
-        
+
         $asString = match($argument->type) {
             'bool' => "<em>--</em>{$name}",
             default => $name,
         };
-            
+
         if (! $argument->hasDefault) {
             return "<{$asString}>";
         }
--- a/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
+++ b/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
@@ -7,6 +7,7 @@ namespace Tempest\Console\Actions;
 use Tempest\Console\Console;
 use Tempest\Console\ConsoleCommand;
 use Tempest\Console\Input\ConsoleArgumentDefinition;
+use function Tempest\Support\str;
 
 final readonly class RenderConsoleCommand
 {
--- a/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
+++ b/src/Tempest/Console/src/Actions/RenderConsoleCommand.php
@@ -6,6 +6,7 @@ namespace Tempest\Console\Actions;
 
 use Tempest\Console\Console;
 use Tempest\Console\ConsoleCommand;
+use Tempest\Console\Input\ConsoleArgumentDefinition;
 use Tempest\Reflection\ParameterReflector;
 
 final readonly class RenderConsoleCommand
@@ -18,8 +19,8 @@ final readonly class RenderConsoleCommand
     {
         $parts = ["<em><strong>{$consoleCommand->getName()}</strong></em>"];
 
-        foreach ($consoleCommand->handler->getParameters() as $parameter) {
-            $parts[] = $this->renderParameter($parameter);
+        foreach ($consoleCommand->getArgumentDefinitions() as $argument) {
+            $parts[] = $this->renderArgument($argument);
         }
 
         if ($consoleCommand->description !== null && $consoleCommand->description !== '') {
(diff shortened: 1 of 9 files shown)


tests/Fixtures/Console/Hello.php
```diff
@@ -23,8 +24,14 @@ public function world(string $input): void
     }
 
     #[ConsoleCommand]
-    public function test(?int $optionalValue = null, bool $flag = false): void
-    {
+    public function test(
+        #[ConsoleArgument]
+        ?int $optionalValue = null,
+        #[ConsoleArgument(
+            name: 'custom-flag',
```
brendt, line 31: Do we already have a test for this? That's still missing as far as I can tell?

src/Tempest/Console/src/Input/ConsoleArgumentDefinition.php
```diff
@@ -24,18 +24,17 @@ public function __construct(
...
-
         $type = $parameter->getType();
 
         return new ConsoleArgumentDefinition(
-            name: $parameter->getName(),
-            type: $type->getName(),
-            default: $parameter->isDefaultValueAvailable() ? $parameter->getDefaultValue() : null,
-            hasDefault: $parameter->isDefaultValueAvailable(),
-            position: $parameter->getPosition(),
+            name       : $attribute?->name ?? $parameter->getName(),
```
brendt, line 30: We don't use colon alignment, but apparently PHP CS fixer doesn't fix it :/

src/Tempest/Console/src/Actions/RenderConsoleCommand.php
```diff
@@ -18,8 +18,8 @@ public function __invoke(ConsoleCommand $consoleCommand): void
     {
         $parts = ["<em><strong>{$consoleCommand->getName()}</strong></em>"];
 
-        foreach ($consoleCommand->handler->getParameters() as $parameter) {
-            $parts[] = $this->renderParameter($parameter);
+        foreach ($consoleCommand->getArgumentDefinitions() as $argument) {
```
brendt, line 21: What confuses me: where do you map the custom name to the actual input provided in the console? Eg. `./tempest my:command --custom-input=abc`

brendt (dismissed):

src/Tempest/Console/src/Input/ConsoleArgumentDefinition.php
gturpin-dev: Ok good to know
I'll fix it ASAP 👍

src/Tempest/Console/src/Actions/RenderConsoleCommand.php
gturpin-dev: The mapping is already done with the `ConsoleArgumentDefinition` so this is build using the `fromParameter()` method, then I've updated the name constructuon to give the attribute name ( added in this PR ) instead of the parameter and still fallback to it if not set.

So here we need to update the input from parameter to the argument definition (  which still can fallback to it )

Tell me if it's not clear 😅

tests/Fixtures/Console/Hello.php
gturpin-dev: This seems to be a console command fixture I've updated it to test.
But I didn't find where it is used as test in phpunit  (  but I'm really bad at testing )

src/Tempest/Console/src/Input/ConsoleArgumentDefinition.php
gturpin-dev: done

brendt:
I think there's still something missing here. I've added a failing test into your branch, I'd expect this one to work. (`\Tests\Tempest\Integration\Console\ConsoleArgumentBagTest::test_name_mapping`)

gturpin-dev:
> I think there's still something missing here. I've added a failing test into your branch, I'd expect this one to work. (`\Tests\Tempest\Integration\Console\ConsoleArgumentBagTest::test_name_mapping`)


I don't think this should work if I refer to https://tempestphp.com/docs/console/building-console-commands/

If I'm not wrong the `CommandWithArgumentName` fixture have a `string input` and a `bool flag` so the flag could be update using `--new-flag` statement in the console input ( which is ok and works for me )

But, the input is a positional parameter, so to update its value we can't name it with `--new-name=foo` I guess.
I think this test is valid using `command-with-argument-name foo --new-flag` atm.

But are you asking to update the actual console arg matching to also support the named param ?
If yes I've just misunderstand the request ! ( but does this wouldn't be better in a separate PR ? )
As this one can help us with flag named params and for the default tempest command + help command

brendt:
Well in Tempest, all arguments can be passed by name, including positional ones:

```php
    #[ConsoleCommand]
    public function __invoke(string $foo, string $bar): void
    {
        $this->writeln($foo);
        $this->writeln($bar);
    }
```

```php
./vendor/bin/tempest my --bar=a --foo=b
b
a
```

So I would say that should be supported with this PR before we can merge.

innocenzi:
I just pushed a few more changes:
- fixed a `str` usage that wasn't imported (you were using the Laravel one locally)
- fixed the mixing between kebab and camel case, this wasn't coming from your PR specifically but it was confusing before and when trying the PR it felt like it needed immediate fixing
- added more test coverage for the changes

innocenzi (approved):

PR #626, opened 2024-10-25
feat: optimize routing

### LINKED ISSUE #554: Improve Router Functionality
There are a few things we should do to improve the functionality of the current router. Each of these should be a smaller PR.

## Chunk the REGEX

The current implementation has a bug that will crash the application in the event that the expression is too long. We can solve this by breaking it down into smaller chunks.

## Group Prefixes

We can improve the performance of the expression by [grouping by route prefix](https://nicolas-grekas.medium.com/making-symfony-router-lightning-fast-2-2-19281dcd245b).

## Route Caching

We may (or may not) benefit from route caching. We should run some benchmarks and do some tests.

**Edit**: I was mistaken and we are caching (see #211). We should just dust this off and make sure everything is still looking good here.

## Resolve Params REGEX

In #181, @brendt noted:

> I think we can refactor resolveParams to either not use regex matches, or a more simplified version; since we're now sure we already have a match

I'm not sure if this is still possible with the latest PR #486, but we should take a look.

## Abstract parts of route matching

In #181, @brendt noted:

>  Add a RouteMatcher class that matches a Route based on a Request and cleans up parts of the Router

We should revisit this. I think it is worth looking to the structure of [FastRoute](https://github.com/nikic/fastroute) for some inspiration.

## Cleanup Custom REGEX

We need to make some design decisions about custom REGEX and implement accordingly. Specifically:

1. Should we support spaces around parameter names/REGEX (e.g., `/users/{ username: \w }`).
2. How do we want to handle weird scenarios like: `/{lang:(en|de)}` or `/foo/{test}/{test:\d+}`?

---

CC: @vsergiu93, @N-Silbernagel

blackshadev:
Fixes #554

# What

Add a new way to build the routing regex for dynamic routes using a tree. Using the tree it is was easier to introduce regex prefix grouping while still being performant during route discovery.

Also added a way to benchmark code more thoroughly using phpbench. All benchmarks can be run using

```
./vendor/bin/phpbench run tests/Benchmark --report=aggregate --iterations=5
```

# One thing stil left unsolved

> Chunk the REGEX

I did not implement it yet, I found that the current changes are already pretty impact-full (in size and changing the internals). So before continue-ing with chunking the routing regexes I wanted to supply my changes as an PR

# Benchmark comparison

I composed a benchmark to test between the `GenericRouter::matchRoute` method and the newly introduced `GenericRouteMatcher::match`. With these I can conclude the following:

- Setting-up routes (and this discovery) has become twice as slow. This is a path which can be optimized more
- Matching dynamic routes has become 50% faster  which scales proportionately to the path length / complexity.
- Static route matching has not changed

## This branch

```
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
| benchmark          | subject          | set                | revs | its | mem_peak | mode     | rstdev |
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
| GenericRouterBench | benchMatchMethod | Dynamic            | 1000 | 5   | 10.555mb | 16.233μs | ±1.29% |
| GenericRouterBench | benchMatchMethod | Non existing long  | 1000 | 5   | 10.555mb | 15.494μs | ±4.01% |
| GenericRouterBench | benchMatchMethod | Non existing short | 1000 | 5   | 10.555mb | 14.208μs | ±1.80% |
| GenericRouterBench | benchMatchMethod | Static route       | 1000 | 5   | 10.539mb | 12.437μs | ±2.64% |
| GenericRouterBench | benchRouterSetup |                    | 1000 | 5   | 10.235mb | 4.319ms  | ±1.69% |
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
```

## Main

```
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
| benchmark          | subject          | set                | revs | its | mem_peak | mode     | rstdev |
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
| GenericRouterBench | benchMatchMethod | Dynamic            | 1000 | 5   | 10.331mb | 23.685μs | ±1.26% |
| GenericRouterBench | benchMatchMethod | Non existing long  | 1000 | 5   | 10.331mb | 28.738μs | ±0.84% |
| GenericRouterBench | benchMatchMethod | Non existing short | 1000 | 5   | 10.331mb | 12.261μs | ±2.76% |
| GenericRouterBench | benchMatchMethod | Static route       | 1000 | 5   | 10.331mb | 12.347μs | ±1.74% |
| GenericRouterBench | benchRouterSetup |                    | 1000 | 5   | 10.040mb | 2.466ms  | ±1.46% |
+--------------------+------------------+--------------------+------+-----+----------+----------+--------+
```

[GenericRouterBench.php.zip](https://github.com/user-attachments/files/17530212/GenericRouterBench.php.zip)


brendt:
Well this is an impressive PR! Could you help me out though @blackshadev : could you go through the PR yourself, and leave some review comments explaining what changed and why? That would help me out a lot with the review :)

src/Tempest/Http/src/GenericRouter.php
```diff
@@ -45,7 +43,7 @@ public function dispatch(Request|PsrRequest $request): Response
             $request = map($request)->with(RequestToPsrRequestMapper::class);
         }
 
-        $matchedRoute = $this->matchRoute($request);
+        $matchedRoute = $this->routeMatcher->match($request);
```
blackshadev, line 46: I extracted all matching logic to a new service "RouteMatcher". So  any logic for matching, static routes, dynamic routes and resolving parameters is moved to this new service.

src/Tempest/Http/src/Route.php
```diff
@@ -55,4 +44,24 @@ public function setHandler(MethodReflector $handler): self
...
+    {
+        $regex = '#\{'. self::ROUTE_PARAM_NAME_REGEX . self::ROUTE_PARAM_CUSTOM_REGEX .'\}#';
+
+        preg_match_all($regex, $uriPart, $matches);
+
+        return $matches[1] ?? [];
+    }
+
+    /** @return string[] */
+    public function routeParts(): array
```
blackshadev, line 59: This is a utility function used during the Route construction.  it will split the URI on the `/` and thus return all "parts" of the route. These parts are used to build up a routing tree.

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -29,7 +34,14 @@ public function addRoute(MethodReflector $handler, Route $route): self
         if ($route->isDynamic) {
             $this->regexMark = str_increment($this->regexMark);
             $this->dynamicRoutes[$route->method->value][$this->regexMark] = $route;
-            $this->addToMatchingRegex($route, $this->regexMark);
+
+            $this->routingTree->add(
```
blackshadev, line 38: So this is the actual start of the whole new routing structure. Let me explain what the whole idea is:

Instead of "just appending" route regexes one after the other. I want to build up a hierarchical regex, where groups of route with common prefixes are matches together with the same prefix. So if we got 2 routes `/test/edit` and `/test/delete` , previously we would create a regex `~^/test/edit(*MARK:b)$|^/test/delete(*MARK:c)$~` While we want to create `~^/test/(?edit(*MARK:b)$|delete(*MARK:c)$)~` (these regexes are written by hand, so if I made an small error, sorry, but you should get the idea.)

A good way to  get this route hierarchy is to create a routing tree first, and than create the regex as a product. After we created the matching regex we also do not need the tree anymore. And building up trees can be efficient enough.

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -40,25 +52,24 @@ public function addRoute(MethodReflector $handler, Route $route): self
...
+    {
+        $this->prepareMatchingRegexes();
+
+        return ['staticRoutes', 'dynamicRoutes', 'matchingRegexes'];
+    }
+
+    public function __wakeup(): void
+    {
+        $this->routingTree = new RoutingTree();
     }
```
blackshadev, line 74: These magic methods are actually used during `serialize` and `unserialize` I opted to use these instead of `__serialize` and `__unserialize` since those require you to return arrays and serialize everything yourself. This `__sleep` call instruct which properties to serialize (and to ensure `matchingRegexes` is filled by calling `prepareMatchingRegexes). The `__wakeup` call is called after `unserialize`  ensure your internal state is consistent.  In my case I just initialize an empty routing tree, without it this routing tree would be undefined since `unserialize` doesn't go through our normal constructor and since I do not serialize it in the `__sleep` call, I do not need it since we got the end product `matchingRegexes` already.

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -40,25 +52,24 @@ public function addRoute(MethodReflector $handler, Route $route): self
         return $this;
     }
 
-    /**
-     * Build one big regex for matching request URIs.
-     * See https://github.com/tempestphp/tempest-framework/pull/175 for the details
-     */
-    private function addToMatchingRegex(Route $route, string $routeMark): void
+    public function prepareMatchingRegexes(): void
```
blackshadev, line 55: During the Route discovery we do not yet create the matchingRegexes, doing so would be wastefull. That is why we expose a separate method to create these regexes right before we need them. Currently this is done:

- before route matching
- before serialization (in `__sleep`)

src/Tempest/Http/src/Routing/Construction/MarkedRoute.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+use Tempest\Http\Route;
+
+final readonly class MarkedRoute
```
blackshadev, line 9: I needed a clean way to carry the route `mark` used in the matching regex and the route into the tree.

src/Tempest/Http/src/Routing/Construction/RoutingTree.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+/**
+ * @internal
+ */
+final class RoutingTree
```
blackshadev, line 10: So your routing tree contains one RouteTreeNode per routing method. 

For each `add` call we check if we need to create a new Root Node. And add the route to it by means of recursion.

The tree can also return all routing regexes by means of recursion .

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+use Tempest\Http\Route;
+
+/**
+ * @internal
+ */
+final class RouteTreeNode
```
blackshadev, line 12: During route discovery I build up a tree (per HTTP method) representing all routes discovered. Each node of this tree can have multiple children and if this node is "the end" of a route, it will also have a `$leaf`  set to the `MarkedRoute`. 

For example, we got 3 routes `/`. `/{id}/edit` and `/new` 
The tree will look something like this:
tree-example

Where only the node `/` -> `{id}` doesn't have a `$leaf` and thus will not be a valid end route. While the others are.

The distinction between `staticPaths` and `dynamicPaths` is also rather useful since we can math the staticPaths before the dynamic ones.

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        return new self(RouteTreeNodeType::Parameter, $regex);
+    }
+
+    public static function createStaticRoute(string $name): self
+    {
+        return new self(RouteTreeNodeType::Static, $name);
+    }
+
+    public function addPath(array $pathSegments, MarkedRoute $markedRoute): void
```
blackshadev, line 43: Adds a new route to the Routing tree. it uses recursion to do so. so with each step we take the first element of `$pathSegments` , look up the appropriate node (or create it) and recurse into it's addPath with the remaining `pathSegments`.

If `$pathSegments` is empty, we mark said node as a valid "leaf" node and attach the `MarkedRoute` to it

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+            $uriPart,
+        );
+    }
+
+    private function regexSegment(): string
+    {
+        return match($this->type) {
+            RouteTreeNodeType::Root => '^',
+            RouteTreeNodeType::Static => "/{$this->segment}",
+            RouteTreeNodeType::Parameter => '/(' . $this->segment . ')',
```
blackshadev, line 84: We can capture the parameter routes in a capture group search that we can extract the routeParameters in match from it.

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+        };
+    }
+
+    public function toRegex(): string
```
blackshadev, line 88: This uses recursion to construct the regex.

It will create nesting for all `$this->staticPaths` and `$this->dynamicPaths`. And if it is a leaf it will add an regex marker with the route `mark` to it.  We add the staticPaths before the dynamicPaths to give them preference.

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Tempest\Http\Routing\Matching;
+
+use Psr\Http\Message\ServerRequestInterface as PsrRequest;
+use Tempest\Http\MatchedRoute;
+use Tempest\Http\Route;
+use Tempest\Http\RouteConfig;
+use Tempest\Http\Routing\Construction\MarkedRoute;
+
+final readonly class GenericRouteMatcher implements RouteMatcher
```
blackshadev, line 13: This roughly contains everything which the GenericRouter had initially. It translates the incoming request to a `MatchedRoute` including parameters. 

I changed two things:

1. Match using the new regexes and call `prepareMatchingRegexes` to ensure the routing tree is translated (once) to regexes
2. The parameters are extracted from the initial route match instead of matching them seperatly. An old todo in the GenericRouter also suggested this.

src/Tempest/Http/src/Route.php
brendt: Maybe rename to `getRouteParts`? Or `split` could also be an option.

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -14,12 +16,15 @@ final class RouteConfig
...
 
+    public RoutingTree $routingTree;
+
     public function __construct(
         /** @var array<string, array<string, \Tempest\Http\Route>> */
         public array $staticRoutes = [],
         /** @var array<string, array<string, \Tempest\Http\Route>> */
         public array $dynamicRoutes = [],
     ) {
+        $this->routingTree = new RoutingTree();
```
brendt, line 27: At this point I'm wondering whether the routing tree will be compiled and cached somewhere?

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -29,7 +34,14 @@ public function addRoute(MethodReflector $handler, Route $route): self
         if ($route->isDynamic) {
             $this->regexMark = str_increment($this->regexMark);
             $this->dynamicRoutes[$route->method->value][$this->regexMark] = $route;
```
brendt, line 36: Do we still need these two separate arrays if we have the routing tree? Don't worry about refactoring it in this PR if the answer is "no", but I did wanna check up on it

src/Tempest/Http/src/RouteConfig.php
brendt: Fair enough, maybe this explanation should be added as a comment, because I'll admit I was a bit confused why we created an empty routing tree at first.

src/Tempest/Http/src/Routing/Construction/MarkedRoute.php
brendt: Fair enough, though isn't this a code smell indicating that we're using route mark in two different places? My assumption would be that the tree should manage the mark internally as well, and that no other place should know about it. Maybe this circles back to my question about whether the dynamic and static routes arrays in `RouteConfig` are still necessary? 

Again, not saying we should refactor it in this PR, but worth discussing imo

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
brendt: If I followed my gut, my assumption would be that we should pass in the `string $path` instead of `array $pathSegments`, and split it internally. But there might be good reasons to use the part segments that I'm overlooking.

brendt: Another sidenote: in `Route` it's called `routeParts`, while over here it's called `pathSegments`, I think it'd be best to make the name consistent — or maybe I'm confused and mixing two different concepts 😁

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            $this->leaf = $markedRoute;
+
+            return;
+        }
+
+        $currentPathSegment = array_shift($pathSegments);
+
+        $regexPathSegment = self::convertDynamicSegmentToRegex($currentPathSegment);
```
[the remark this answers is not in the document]
brendt: Maybe add this explanation as a comment in the code as well, I think that's gonna be useful in the long run

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -0,0 +1,114 @@
...
+
+    public function toRegex(): string
+    {
+        $regexp = $this->regexSegment();
+
+        if ($this->staticPaths !== [] || $this->dynamicPaths !== []) {
+            $regexp .= "(?";
+
+            foreach ($this->staticPaths as $path) {
+                $regexp .= '|' . $path->toRegex();
```
brendt, line 96: Won't the leading `|` be a problem?

src/Tempest/Http/src/Routing/Construction/RouteTreeNodeType.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+enum RouteTreeNodeType
```
brendt, line 7: Pure on intuition, I'd assume here to be `Dynamic` as well? It's probably not necessary but I figured I'd ask to be sure

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
```diff
@@ -0,0 +1,86 @@
...
+    private function matchDynamicRoute(PsrRequest $request): ?MatchedRoute
+    {
+        // If there are no routes for the given request method, we immediately stop
+        $routesForMethod = $this->routeConfig->dynamicRoutes[$request->getMethod()] ?? null;
+        if ($routesForMethod === null) {
+            return null;
+        }
+
+        // Ensures matching regexes are available
+        $this->routeConfig->prepareMatchingRegexes();
```
brendt, line 50: I'm wondering about caching at this point. We already have config cache, so maybe we should simply do this in the constructor of `RouteCache`, and store the compiled regex there?

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
```diff
@@ -0,0 +1,86 @@
...
+            return $staticRoute;
+        }
+
+        // match dynamic routes
+        return $this->matchDynamicRoute($request);
+    }
+
+    private function matchStaticRoute(PsrRequest $request): ?MatchedRoute
+    {
+        $staticRoute = $this->routeConfig->staticRoutes[$request->getMethod()][$request->getUri()->getPath()] ?? null;
```
brendt, line 32: So, technically the static routes are in the tree as well, but we don't need them? That makes me wonder:

1. why do we keep them in the tree?
2. is it again a code smell indicating that we're doing two related things in separate places?
3. FTR: performance would be a good argument to keep them apart, but if that is the case, I think we can get rid of static routes in the tree alltogether?

src/Tempest/Http/src/GenericRouter.php
brendt, line 46: I don't have a strong opinion on it, but maybe we could move the registration of `MatchedRoute` in the container to `RouteMatcher` as well? Currently, the router still does it on line 52.

tests/Benchmark/README.md
```diff
@@ -0,0 +1,17 @@
+# Tempest benchmarks
```
brendt, line 1: Thank you for adding this 🎉

tests/Fixtures/Config/database.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
brendt, line 1: This file should be in gitignore, because our testsuite will generate it based on whether we're testing sqlite, mysql or postgres. Is there a reason you added it in VCS?

brendt (changes requested):
Super nice PR! I left some comments, but this is almost ready to merge! 

I am going to wait until after alpha.3 is tagged though, just so that we have some time to test it ourselves before outsiders try it out

brendt:
One thing I am a little worried about:

> Setting-up routes (and this discovery) has become twice as slow. This is a path which can be optimized more

I noticed myself how discovery without caching already has a noticeable impact for local development. I'm fairly confident that the solution is to have partial discovery cache (#395), but it requires a pretty big refactor. Nevertheless, it's something I'll focus on for alpha.4

tests/Fixtures/Config/database.php
blackshadev: No clue, I think this is a rebasing issue because I rebased onto the newer version which uses database.config.php. Will remove it.

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
blackshadev: I do not see `RouteCache`, do you mean `RouteConfig` ? 

The reason why I do this: if you do not have discovery caching enabled. And you hit an endpoint, the RouteDiscovery will first build up the routingTree in the config. After which it needs to build up the regex. I do not (yet) have a hook to execute some logic after the Discovery phase. So I added this here to ensure it has.

If you do have an hook "after" the discovery phase. We could separate the whole route discovery  with tree building and the routing config. Which would be really amazing. But maybe a bit to much for this PR?

src/Tempest/Http/src/RouteConfig.php
blackshadev: Yes. The routing tree only facilitates building up the routing regex for the dynamic routes.

src/Tempest/Http/src/RouteConfig.php
blackshadev: No , it will not. Because of `__sleep` code. I do not return "routingTree" so it shouldn't be serialized.

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
blackshadev: I will stick to `pathSegments` as it is more describtive. I renamed the Route method to `split` as suggested and put `segments` into the explaining doc block.

> my assumption would be that we should pass in the string $path instead of array $pathSegments, and split it internally

It would be possible, but realllly wasteful. The RouteTreeNode is already the internal of the tree, so normally people will not see / use this. The reason why I use an array is because use recursion for each element in the `$pathSegments` array. You can see on me using `array_shift` to get the first element (and mutate the existing array) and passing the remaining array into the addPath of the next node. 

So when I start with the segments: `['test', '{id}', 'edit']` 
- I first find or create the node for `test`, and call `addPath` with the remaining segments: `['{id}', 'edit'], 
- then I find or create the node for `{id}` in the `test` node and call `addPath` with the remaining segments:  `['edit']`, 
- then I find or create edit `edit` node in the `{id}` node and call `addPath` with the remaining segments `[]`
- Last I add the passed along `$markedRoute` into the `$leaf`

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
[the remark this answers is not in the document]
blackshadev: Done

src/Tempest/Http/src/Routing/Construction/MarkedRoute.php
blackshadev: Well, yes and no. 

Yes, currently "Route Construction" is done into the RouteConfig, and maybe it shouldn't . RouteConfig could also be used to only carry the resulting regexes and routes array, while not caring about the `MarkedRoute` concept and `RoutingTree`. But these regexes need to be build somewhere. If there was a function part of the `RouteDiscovery` class which was called after discovery was done. I could separate the discovery phase from the matching phase.


The  dynamic route array is a result of the routing tree, the routing tree is used to build up the matching regexes, not the routing itself. The static routes are not added to the routing tree, the current implementation is the most efficient (and easiest) one.

src/Tempest/Http/src/RouteConfig.php
blackshadev: done

src/Tempest/Http/src/Route.php
blackshadev: done, great suggestion

blackshadev:
> One thing I am a little worried about:
> 
> > Setting-up routes (and this discovery) has become twice as slow. This is a path which can be optimized more
> 
> I noticed myself how discovery without caching already has a noticeable impact for local development. I'm fairly confident that the solution is to have partial discovery cache (#395), but it requires a pretty big refactor. Nevertheless, it's something I'll focus on for alpha.4

Yeah , currently it is a tad slower. I do think I can increase it allot over time. But I didn't want to go full optimization in one PR. I wanted to have a chance to get feedback early and iterate on the choices made.

src/Tempest/Http/src/Route.php
```diff
@@ -55,4 +44,29 @@ public function setHandler(MethodReflector $handler): self
...
+     *
+     * @example '/test/{id}/edit' becomes ['test', '{id}', 'edit']
+     * @return string[]
+     */
+    public function split(): array
+    {
+        $parts = explode('/', $this->uri);
+
+        return array_values(
+            array_filter($parts, static fn (string $part) => ! empty($part))
```
homersimpsons, line 69: I think we should use `=== ''` instead of `empty`, because `empty('0')` is `true`. So if I understand correctly, a route like `api/0/foo` would become `['api', 'foo']` instead of `['api', '0', 'foo']`.

```suggestion
            array_filter($parts, static fn (string $part) => $part !== '')
```

This will most likely be a corner case but I think tis will happen in the wild for api versionning or similar things.

blackshadev: Yes, you correct! thank you! I will, use your suggestion as is to ensure your contribution and  add a quick test to proof it  in tests.

src/Tempest/Http/src/Routing/Construction/RouteTreeNodeType.php
blackshadev: I have renamed `Parameter` as dynamic. you were right I wasn't consistent in my naming

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
blackshadev: Ahh good question.
`(?|a|b|c)` is a valid regex. the `(?|` group is a "Branch Reset Group" Basically a switch statement in the regex. It also allows capture groups in the `|` with the same index as all others , which is really nice for us since all parameters will match on the same index everytime.

Read more about them here https://www.regular-expressions.info/branchreset.html

I didn't know about this this until I saw it in the sample implementation in the symfony router linked in #554

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
brendt: Fair enough, thanks for the clarification!

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
brendt: Ok, maybe add this explanation in the code itself as well for clarity 👍

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
brendt: That's a bit too much for this PR, indeed. Although, doesn't `KernelEvent::BOOTED` work for that? It's triggered after discovery: https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Core/src/Kernel.php#L42

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
brendt: Don't know why I thought they were 🫣 sorry for the noise

src/Tempest/Http/src/GenericRouter.php
brendt: Good point.

brendt:
> Yeah , currently it is a tad slower. I do think I can increase it allot over time. But I didn't want to go full optimization in one PR. I wanted to have a chance to get feedback early and iterate on the choices made.

That's totally fine :) It's also not a blocker for merging this PR, but we do need to further improve it. 

Although I assume partial discovery cache will have the biggest impact. Unfortunately for me, it'll become more difficult to implement with this routing tree. But that's fine!

src/Tempest/Http/src/Routing/Construction/MarkedRoute.php
brendt: Ok, let's see whether `KernelEvent::BOOTED` I mentioned in the other comment helps you. Apart from that, we don't need to worry about it in this PR. We can further improve after this has been merged.

src/Tempest/Http/src/Routing/Matching/GenericRouteMatcher.php
blackshadev: Yeah it would. I did not expect it to be a thing. Nice! after this PR I will create one to separate the route construction from the route matching

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
blackshadev: done

brendt (approved):

PR #630, opened 2024-10-26
fix(container): fix caching of autowire discovered classes

blackshadev:
# What

@Jeroen-G 

I found a bug using the `#[Autowire]` feature while enabled discovery cache. Instead of caching the container `definitions` I introduced a separate array which holds the autowire mapping, put that in the discovery cache and re-add these to the container during cache recovery. 

# How to test

- set `CACHE=true` in `.env`
- Run `php tempest cache:clear --all` for good measure
- Use my provided zip which includes which includes the steps below: [[autowire-example.zip](https://github.com/user-attachments/files/17530379/autowire-example.zip)]
  - Create an interface
  - Create a class which implements said class and add `#[Autowire]`
  - Create a controller using the autowired interface
- Run `php tempest serve`
- See no error

Performing the same steps on the `main` branch will result in the following error: `Serialization of 'Closure' is not allowed`  since it tries to serialize the container `definitions`

### DIFF
--- a/src/Tempest/Container/src/AutowireDiscovery.php
+++ b/src/Tempest/Container/src/AutowireDiscovery.php
@@ -14,6 +14,8 @@ use Tempest\Reflection\ClassReflector;
 final class AutowireDiscovery implements Discovery
 {
     use HandlesDiscoveryCache;
+    /** @var array<string, string> */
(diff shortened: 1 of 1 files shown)


brendt:
Nice! I like this approach 💪

PR #647, opened 2024-10-30
feat(console): add `make:controller` and `make:model` commands

gturpin-dev:
This PR replace the PR #568
I started from scratch because lot of work in the other PR is stale.
Here's a clean version of the real changes.

I've updated the overall "generator commands" as discussed in Discord and in the other PR reviews.

So this is a first step to generator commands MVP. It's a starting point for #472

If everything is ok, the #568 could be closed.

@brendt Could you please add @innocenzi as co-author of this one as a lot of work is also done by/with him 🙏


src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Tempest\Console\Commands\Generators;
+
+use Tempest\Console\ConsoleArgument;
+use Tempest\Console\ConsoleCommand;
+use Tempest\Console\Stubs\ControllerStub;
+use Tempest\Generation\HasGeneratorCommand;
+
+final class MakeControllerCommand
+{
+    use HasGeneratorCommand;
```
brendt, line 14: Let's rename to `IsGeneratorCommand`

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+    ): void {
+        $suggestedPath = $this->getSuggestedPath(
+            className  : $className,
+            pathPrefix : 'Controllers',
+            classSuffix: 'Controller',
+        );
+        $targetPath = $this->promptTargetPath($suggestedPath);
+        $shouldOverride = $this->askForOverride($targetPath);
+
+        $this->stubFileGenerator->generate(
```
brendt, line 43: Maybe not within the scope of this PR, but it might be nice to also make a view file together with the controller or something alike.

src/Tempest/Generation/src/StubFileGenerator.php
```diff
@@ -0,0 +1,119 @@
...
+
+use Tempest\Console\Console;
+use Tempest\Generation\Exceptions\StubFileGenerationFailedException;
+use Tempest\Support\PathHelper;
+use Tempest\Support\StringHelper;
+use Throwable;
+
+/**
+ * This class can generate a file from a stub file with additional useful methods.
+ * It only works with PHP class files.
```
brendt, line 15: This worries me, what if we want a `make:view-component --anonymous` command?

brendt (changes requested):
Looks really nice now! I left a review, but there's one thing I'm unsure about that I wanted to discuss with this PR.

We now also have the concept of "installers": https://tempestphp.com/docs/internals/package-development/

These installers aren't the same as stub generators, but there are some similarities. I wonder if we can make it so that these generators "feel" as close as possible to installers?

In particular, I'm thinking about the `PublishesFiles` trait, which is actually very similar to what `StubFileGenerator` does. It already has support for both files and classes, it will also take care of updating imports where necessary, and it will handle the override part automatically.

Here's an example of what that would look like:

```php

final class MakeControllerCommand
{
    use PublishesFiles;
    use HasGeneratorCommand;

    #[ConsoleCommand(
        name       : 'make:controller',
        description: 'Create a new controller class with a basic route',
        aliases    : ['controller:make', 'controller:create', 'create:controller'],
    )]
    public function __invoke(
        #[ConsoleArgument(
            help: 'The name of the controller class to create ( "Controller" will be suffixed )',
        )]
        string $className,
        #[ConsoleArgument(
            help: 'The route path inside the controller',
        )]
        ?string $controllerPath = null,
        #[ConsoleArgument(
            help: 'The view name to return in the controller',
        )]
        ?string $controllerView = null,
    ): void {
        $suggestedPath = $this->getSuggestedPath(
            className  : $className,
            pathPrefix : 'Controllers',
            classSuffix: 'Controller',
        );

        $targetPath = $this->promptTargetPath($suggestedPath);

        $this->publish(
            source: ControllerStub::class,
            destination: $targetPath,
            callback: function (string $source, string $destination) use ($controllerView, $controllerPath): void 
            {
                $content = str(file_get_contents($destination));
                
                $replacements = arr([
                    'dummy-path' => $controllerPath,
                    'dummy-view' => $controllerView,
                ]);
                
                file_put_contents(
                    $destination,
                    $content->replace(
                        $replacements->keys()->toArray(),
                        $replacements->values()->toArray()
                    )
                );
            }
        );
    }
}
```

We could of course improve the `PublishesFiles` trait to have a shorthand for string replacements as well, something like:


```php
        $this->publish(
            source: ControllerStub::class,
            destination: $targetPath,
            replace: [
                'dummy-path' => $controllerPath,
                'dummy-view' => $controllerView,
            ],
        );
```

I _think_ it would be nice if the `PublishesFiles` trait could be reused. I realise it goes against my first suggestion of extracting `StubFileGenerator`, but I only realised the similarities after reading through the simplified implementation just now.

It's an iterative process, I hope you'll indulge me 😅

Anyway, I'm totally open to discuss this, I just want to make sure we go with the absolute easiest, most consistent approach 👍 And, of course, really great job on the PR already, we're getting there! 💪

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
gturpin-dev: I've named it like so because it can be a class that holds multiple console commands on multiple methods.
I think this is also why we currently had the `HasConsole` trait and not a `IsConsoleCommand`.
What do you think ?

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
gturpin-dev: I'll dig this idea, but I think this could be a flag or something to additionally create the view 🤷 
And yes maybe in a separate PR as an improvement of the `make:controller` method.
As this one mainly aim as setup generator commands and with basic examples.

src/Tempest/Generation/src/StubFileGenerator.php
gturpin-dev: I'm not ultra familiar with the `nette/php-generator` package but in case we use the `ClassManipulator` to create the stub and manipulate it.
So currently I don't think it's possible to create a stub from another file of php classes.
If we wanted to, I think something needs to be refactor like using a `FileManipulator` or something like this.

Do you think this could comes with the PR that will add the `make:view-component` command ?
I think it could be the right scope, doing things/refactor when we need it, and then update this PHPDoc

brendt:
> I have another idea, what about keep the StubFileGenerator as a separate service from the tempest/generation package.
Then the PublishesFiles trait stay at the tempest/core and could use the StubFileGenerator under the hood to be DRY ( with some refactoring as the publish method works with files that are not classes ( this could solve this review https://github.com/tempestphp/tempest-framework/pull/647#discussion_r1824368137 )

The downside here is that now `core` has a dependency on `generation`. But that's not a problem: the trait can move to the generation package. We just need to make sure all components that depend on it, also require it.

Apart from that, I think it's a good idea to have the trait simple wrap an underlying service, just like we do with `HasConsole`.

> Another question, in general and for this PR, do you think the generators commands (e.g. make:controller ) should live in the tempest/console or as commands in tempest/generation ?

The third option, actually, it should live in `http` 😂

So, I realise it won't be available if people install `tempest/http`, but don't install `tempest/console`. But that's actually fine: it won't break anything, and it will be available as soon as people decide they want to pull in console as well 👍

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+
+final class MakeControllerCommand
+{
+    use HasGeneratorCommand;
+
+    #[ConsoleCommand(
+        name       : 'make:controller',
+        description: 'Create a new controller class with a basic route',
+        aliases    : ['controller:make', 'controller:create', 'create:controller'],
+    )]
```
innocenzi, line 20:
```suggestion
    #[ConsoleCommand(
        name: 'make:controller',
        description: 'Creates a new controller class with a route',
        aliases: ['controller:make', 'controller:create', 'create:controller'],
    )]
```

One of the reasons I wanted an abstraction initially was to avoid repeating descriptions and aliases. It's easy to become inconsistent about what aliases are available for specific `make` commands.

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+    public function __invoke(
+        #[ConsoleArgument(
+            help: 'The name of the controller class to create ( "Controller" will be suffixed )',
```
innocenzi, line 23:
```suggestion
            help: 'The name of the controller class to create',
```

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        )]
+        string $className,
+        #[ConsoleArgument(
+            help: 'The route path inside the controller',
+        )]
+        ?string $controllerPath = null,
+        #[ConsoleArgument(
+            help: 'The view name to return in the controller',
```
innocenzi, line 31:
```suggestion
            help: 'The name of the view returned from the controller',
```

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        description: 'Create a new controller class with a basic route',
+        aliases    : ['controller:make', 'controller:create', 'create:controller'],
+    )]
+    public function __invoke(
+        #[ConsoleArgument(
+            help: 'The name of the controller class to create ( "Controller" will be suffixed )',
+        )]
+        string $className,
+        #[ConsoleArgument(
+            help: 'The route path inside the controller',
```
innocenzi, line 27:
```suggestion
            help: 'The path of the route',
```

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        ?string $controllerPath = null,
+        #[ConsoleArgument(
+            help: 'The view name to return in the controller',
+        )]
+        ?string $controllerView = null,
+    ): void {
+        $suggestedPath = $this->getSuggestedPath(
+            className  : $className,
+            pathPrefix : 'Controllers',
+            classSuffix: 'Controller',
```
innocenzi, line 38:
```suggestion
            className: $className,
            pathPrefix: 'Controllers',
            classSuffix: 'Controller',
```

I am not sure if specifying a path prefix and a class suffix is a good idea. This enforces conventions that not everyone would want to follow.

For instance, I personally don't put controllers in a `Controllers` directory. But right now, this is hardcoded, and will always be suggested to the user when running `make:controller`.

This is why, in my initial implementation, I wanted to simply publish the file to the root of the main namespace, so there is no specific opinion about it. This is one of the core features of Tempest: everything can be wherever, there's no convention.

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        );
+        $targetPath = $this->promptTargetPath($suggestedPath);
+        $shouldOverride = $this->askForOverride($targetPath);
+
+        $this->stubFileGenerator->generate(
+            stubFile      : ControllerStub::class,
+            targetPath    : $targetPath,
+            shouldOverride: $shouldOverride,
+            replacements  : [
```
innocenzi, line 47:
```suggestion
            stubFile: ControllerStub::class,
            targetPath: $targetPath,
            shouldOverride: $shouldOverride,
            replacements: [
```

src/Tempest/Console/src/Commands/Generators/MakeModelCommand.php
```diff
@@ -0,0 +1,46 @@
...
+use Tempest\Generation\HasGeneratorCommand;
+
+final class MakeModelCommand
+{
+    use HasGeneratorCommand;
+
+    #[ConsoleCommand(
+        name       : 'make:model',
+        description: 'Create a new model class',
+        aliases    : ['model:make', 'model:create', 'create:model'],
```
innocenzi, line 20:
```suggestion
        name: 'make:model',
        description: 'Creates a new model class',
        aliases: ['model:make', 'model:create', 'create:model'],
```

src/Tempest/Console/src/Commands/Generators/MakeModelCommand.php
```diff
@@ -0,0 +1,46 @@
...
+    )]
+    public function __invoke(
+        #[ConsoleArgument(
+            help: 'The name of the model class to create ( "Model" will be suffixed )',
```
innocenzi, line 24:
```suggestion
            help: 'The name of the model class to create',
```

src/Tempest/Console/src/Commands/Generators/MakeModelCommand.php
```diff
@@ -0,0 +1,46 @@
...
+        string $className,
+        #[ConsoleArgument(
+            help: 'Whether the model is a database model',
+        )]
+        bool $isDatabaseModel = false,
+    ): void {
+        $suggestedPath = $this->getSuggestedPath(
+            className  : $className,
+            pathPrefix : 'Models',
+            classSuffix: 'Model',
```
innocenzi, line 35:
```suggestion
            className: $className,
            pathPrefix: 'Models',
            classSuffix: 'Model',
```

src/Tempest/Console/src/Commands/Generators/MakeModelCommand.php
```diff
@@ -0,0 +1,46 @@
...
+        );
+        $targetPath = $this->promptTargetPath($suggestedPath);
+        $shouldOverride = $this->askForOverride($targetPath);
+
+        $this->stubFileGenerator->generate(
+            stubFile      : $isDatabaseModel ? DatabaseModelStub::class : ModelStub::class,
+            targetPath    : $targetPath,
+            shouldOverride: $shouldOverride,
```
innocenzi, line 43:
```suggestion
            stubFile: $isDatabaseModel ? DatabaseModelStub::class : ModelStub::class,
            targetPath: $targetPath,
            shouldOverride: $shouldOverride,
```

src/Tempest/Generation/src/HasGeneratorCommand.php
```diff
@@ -0,0 +1,96 @@
...
+use Tempest\Core\Composer;
+use Tempest\Support\PathHelper;
+use function Tempest\Support\str;
+use Tempest\Validation\Rules\EndsWith;
+use Tempest\Validation\Rules\NotEmpty;
+
+/**
+ * This provide bunch of methods to generate files.
+ * Also methods to manipulate the user generic input like the class name.
+ */
```
innocenzi, line 17: I love docblocks that describe the purpose of a class. Well done ❤️ 
```suggestion
/**
 * Provides a bunch of methods to generate files and work with common user input.
 */
```

src/Tempest/Generation/src/HasGeneratorCommand.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Get a suggested path for the given class name.
+     * This will find the main autoloaded namespace in the project's composer file and use it as the base path.
+     *
+     * @param string $className The class name to generate the path for, can include path parts (e.g. 'Models/User').
+     * @param string|null $pathPrefix The prefix to add to the path (e.g. 'Models').
+     * @param string|null $classSuffix The suffix to add to the class name (e.g. 'Model').
+     *
+     * @return string The fully suggested path including the filename and extension.
```
innocenzi, line 35:
```suggestion
     * Gets a suggested path for the given class name.
     * This will use the user's main namespace as the base path.
     *
     * @param string $className The class name to generate the path for, can include path parts (e.g. 'Models/User').
     * @param string|null $pathPrefix The prefix to add to the path (e.g. 'Models').
     * @param string|null $classSuffix The suffix to add to the class name (e.g. 'Model').
     *
     * @return string The fully suggested path including the filename and extension.
```

src/Tempest/Support/src/PathHelper.php
```diff
@@ -36,4 +41,69 @@ public static function make(string ...$paths): string
...
+
+        throw new Exception(sprintf('No registered namespace matches the specified path [%s].', $path));
+    }
+
+    /**
+     * Convert a path to a class name.
+     *
+     * @param string $path The path to convert.
+     */
+    public static function toClassName(string $path): string
```
innocenzi, line 99: I think this is missing test coverage?

innocenzi (commented):
Here are a few style nitpicks, as well as a concern regarding conventions suggested by the implementation.

Also, the pull request title should be:
```
feat(console): add `make:controller` and `make:model` commands
```

gturpin-dev:
> The third option, actually, it should live in http 😂
>
> So, I realise it won't be available if people install tempest/http, but don't install tempest/console. But that's actually fine: it won't break anything, and it will be available as soon as people decide they want to pull in console as well 👍

@brendt So, we should provide the `make:controller` command in `tempest/http` package, the `make:model` in `tempest/database` ( or another one because a model isn't tied to database until we add the `IsDatabaseModel` trait ? )
and so on.. So in one word, the command should live in its related package right ?

Then how I must handle the fact that `ConsoleCommand` will be available only if the user require the `tempest/console` if the `make:controller` is in `tempest/http` I will use the command from the console package, aren't we getting a fatal or something like that ?

gturpin-dev:
@innocenzi Huge thanks for your review and requested changes 🙏 

> Also, the pull request title should be:
Thanks for that aswell 😄

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
gturpin-dev: That's true !

src/Tempest/Console/src/Commands/Generators/MakeControllerCommand.php
gturpin-dev: That's true, I was seeing that as an extra work, but in case I guess we could keep those options on the helpers because it can be useful.
But I'll remove it from the internal commands 👍

src/Tempest/Generation/src/HasGeneratorCommand.php
gturpin-dev: Thanks, but thank you for updating them from an external reading POV 🙏

innocenzi:
@gturpin-dev I saw you changed the PR title, but it's still wrong (missing the colon after the commit type, missing backticks around commands and wrong scope) 😅

You can copy paste this one:
```
feat(console): add `make:controller` and `make:model` commands
```

gturpin-dev:
@innocenzi Nevermind this is not my day 😂 
Thanks 🙏

brendt:
> So in one word, the command should live in its related package right ?

Correct

> Then how I must handle the fact that ConsoleCommand will be available only if the user require the tempest/console if the make:controller is in tempest/http I will use the command from the console package, aren't we getting a fatal or something like that ?

That's the beauty of discovery :) If you install one of the standalone components in a project that doesn't have `tempest/console` installed, then nothing will happen because there's no discovery class dedicated for console commands :) But as soon as you install `tempest/console`, it'll detect the package-specific commands and just work :)

gturpin-dev:
> That's the beauty of discovery :) If you install one of the standalone components in a project that doesn't have tempest/console installed, then nothing will happen because there's no discovery class dedicated for console commands :) But as soon as you install tempest/console, it'll detect the package-specific commands and just work :)

This is what I believed but I was wondering about the fact theoretically anyone can call itself and break things, but nobody have to do that, so I think it's ok 😅

brendt:
> This is what I believed but I was wondering about the fact theoretically anyone can call itself and break things, but nobody have to do that, so I think it's ok

Yeah, no reason to worry about it :)

Let me know when I need to review again?

src/Tempest/Generation/src/StubFileGenerator.php
gturpin-dev: So, now this is possible with the refactoring of the `StubFileGenerator` to support `PublishesFiles` features

gturpin-dev:
@brendt Sure !
There is this review still waiting your thoughts https://github.com/tempestphp/tempest-framework/pull/647#discussion_r1824410556

I'm still working on things we've discussed.
I hope I'll finish them soon.

Do you think some tests on those commands ( and the `StubFileGenerator` ) could come in another PR ?
I think I can get some inspiration from the current `Installers` tests but I'll make a long time to understand them ( it's new to me )
And there is some refactor in this PR, I'd like it to not become stale again because too much time has passed.

gturpin-dev:
> Hmm no tests are pretty crucial, but I'm happy to help you with them :) Here's what we can do: you tell me which parts you think should definitely be tested, and I'll write those tests. I think it would be a good opportunity for you to learn to identify testable parts, and then you'll get an example of how those tests can be written, which you can learn from again.

Absolutely, would love this 🙏
I'll let you know when I'm ready

src/Tempest/Support/src/PathHelper.php
gturpin-dev: Added with the latest commit 😄

src/Tempest/Core/src/PublishesFiles.php
```diff
@@ -5,65 +5,94 @@
...
-        } else {
-            if (! $this->confirm(
-                question: "Do you want to create {$destination}?",
+        try {
+            if (! $this->console->confirm(
+                question: sprintf('Do you want to create "%s"', $destination),
                 default: true,
             )) {
-                $this->writeln('Skipped');
+                throw new FileGenerationAbortedException('Skipped.');
```
brendt, line 42: Is this a good idea though? If we don't confirm one file, then the whole installation stop working. From the point of view of installers, that's not how it should work 🤔

brendt (changes requested):

src/Tempest/Core/src/PublishesFiles.php
gturpin-dev: This is catched at the end of the method reusing the exception message.
So there is no execution break for now 😄

src/Tempest/Generation/src/ClassManipulator.php
```diff
@@ -37,7 +45,11 @@ public function save(string $path): self
             mkdir($dir, recursive: true);
         }
 
-        file_put_contents($path, $this->print());
+        $is_success = (bool) file_put_contents($path, $this->print());
```
brendt, line 48: rename to `$isSuccess`

src/Tempest/Generation/src/DataObjects/StubFile.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+final class StubFile
+{
+    public function __construct(
+        public readonly string $filePath,
+        public readonly StubFileType $type,
+    ) {
+    }
+
+    public static function fromFilePath(string $filePath): self
```
brendt, line 26: Can't these two static constructors be combined into one? `::from($pathOrClass)`. You can check whether it's a path with `if(is_file($pathOrClass))`

It's just a minor detail, but makes the API a bit more convenient to use

src/Tempest/Generation/src/StubFileGenerator.php
```diff
@@ -0,0 +1,150 @@
...
+     * @param bool $shouldOverride Whether the generator should override the file if it already exists.
+     *
+     * @throws FileGenerationFailedException If the operation has been aborted.
+     */
+    private function prepareFilesystem(
+        string $targetPath,
+        bool $shouldOverride = false,
+    ): void {
+        // Delete the file if it exists and we force the override
+        if (file_exists($targetPath)) {
```
brendt, line 136: Shouldn't we better do this right before copying the file? 

Let's say something goes wrong in between `prepareFilesystem` and actually copying the file. Then we'll have removed the original file, but haven't been able to copy to new one.

src/Tempest/Http/src/Commands/MakeControllerCommand.php
```diff
@@ -0,0 +1,58 @@
...
+        )]
+        ?string $controllerPath = null,
+        #[ConsoleArgument(
+            help: 'The name of the view returned from the controller',
+        )]
+        ?string $controllerView = null,
+    ): void {
+        $suggestedPath = $this->getSuggestedPath($className);
+        $targetPath = $this->promptTargetPath($suggestedPath);
+        $shouldOverride = $this->askForOverride($targetPath);
```
brendt, line 40: Is there a reason we specifically _ask_ for the override? Why not a flag that's false by default? `bool $override = false`.

src/Tempest/Support/src/PathHelper.php
```diff
@@ -36,4 +43,79 @@ public static function make(string ...$paths): string
 
         return $path;
     }
+
+    private static function prepareStringForNamespace(string $path, string $root = ''): StringHelper
```
brendt, line 47: The PathHelper should follow the same structure as `StringHelper` and `ArrayHelper`. That means making an instance of it with chainable immutable methods.

I'll take care of refactoring this on your branch 👍

brendt (changes requested):

src/Tempest/Generation/src/ClassManipulator.php
gturpin-dev: Oops, interesting that Rector or CS Fixer not fixed it 🤔

src/Tempest/Support/src/PathHelper.php
gturpin-dev: As discussed in Discord, this is probably the scope of another PR so I close this 😄

src/Tempest/Http/src/Commands/MakeControllerCommand.php
gturpin-dev: I've done this to avoid another parameter to add on each individual make command.
This also trigger only when the file exists.
If I'm not wrong, the `--force` is global and can skip all confirm interactions right ?

src/Tempest/Generation/src/ClassManipulator.php
gturpin-dev: Fixed 😄

src/Tempest/Generation/src/DataObjects/StubFile.php
gturpin-dev: You're right, added 😄

src/Tempest/Generation/src/StubFileGenerator.php
gturpin-dev: I agree

I've changed this

src/Tempest/Http/src/Commands/MakeControllerCommand.php
brendt: > If I'm not wrong, the --force is global and can skip all confirm interactions right ?

Only if you add the `ForceMiddleware`. Let's keep it for now though, I'll give it some more thought after this PR has been merged :)

src/Tempest/Core/src/PublishesFiles.php
brendt: Gotcha! I understand now. I don't think I have a very strong opinion on it  for now

brendt (approved):

brendt:
Let's go!!

innocenzi:
Well done!

PR #642, opened 2024-10-28
feat (ArrayHelper): implement sort method.

aarcex3:
# Summary
Implemented `sort` method for `ArrayHelper`.
This sorts the array in ascending order.

# Closes
#631

PR #659, opened 2024-11-01
feat(support): add sorting methods to `ArrayHelper`

### LINKED ISSUE #631: `ArrayHelper::sort`
To sort items in an array

Bapawe:
This pull request improves on https://github.com/tempestphp/tempest-framework/pull/642 by adding options to the sort method to change sorting order and preserve keys, added extra method to sort values by callback and added methods to sort the array keys.

Closes #631


Bapawe:
@gturpin-dev can you take a look at the PR, not to sure about the PHPDoc return type.

src/Tempest/Support/src/ArrayHelper.php
```diff
@@ -768,12 +768,85 @@ public function mapTo(string $to): self
...
 
     /**
-     * Sorts the array in ascending order.
+     * Returns a new instance of this array sorted by its values.
+     *
+     * @param bool $desc Sorts in descending order if `true`; defaults to `false` (ascending).
+     * @param bool|null $preserveKeys Preserves array keys if `true`; reindexes numerically if `false`.
+     *                                Defaults to `null`, which auto-detects preservation based on array type  (associative or list).
+     * @param int $flags Sorting flags to define comparison behavior, defaulting to `SORT_REGULAR`.
+     * @return ($preserveKeys is true ? self<TKey, TValue> : self<int|TKey, TValue>)
```
gturpin-dev, line 777: I don't think IDEs can properly parse this syntax.

Then why not go with something like `@return self<array-key, TValue> Key type depends on $preserveKeys` or something similar ?
Like that, the IDE still works with `array-key` syntax, we don't have the `TKey` because we aren't sure that this is the same as the current ones ( since preserveKeys can change them ). But we make a text explanations for developers ( a more precise explanations can be done, but I think $preserveKeys is self-explanatory enough ).

However, If there is a syntax that allow to conditionally set return type based on a parameter, we could change them to it.
Maybe it's possible with Generics but I'm not experienced enough with them to know how.

src/Tempest/Support/src/ArrayHelper.php
```diff
@@ -768,12 +768,85 @@ public function mapTo(string $to): self
...
+    /**
+     * Returns a new instance of this array sorted by its values using a callback function.
+     *
+     * @param callable $callback      The function to use for comparing values. It should accept two parameters
+     *                                and return an integer less than, equal to, or greater than zero if the
+     *                                first argument is considered to be respectively less than, equal to, or
+     *                                greater than the second.
+     * @param bool|null $preserveKeys Preserves array keys if `true`; reindexes numerically if `false`.
+     *                                Defaults to `null`, which auto-detects preservation based on array type  (associative or list).
+     * @return ($preserveKeys is true ? self<TKey, TValue> : self<int|TKey, TValue>)
```
gturpin-dev, line 805: Maybe this could be changed according to the above review

gturpin-dev (changes requested):
Sorry for late 🙏 .

I left a review, I think the IDE support is more important than the conditional explanation.
So maybe we could change it to begin with. 

We still can make another PR if we find a better way to handle this conditional return type.
I think this is not a critical stuff because the variable is self-explanatory enough so it's not a big deal for now.

src/Tempest/Support/src/ArrayHelper.php
Bapawe: It should work, i used that syntax from: https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types
PHPstorm did not have any troubles with it, not sure about other IDE's though. 

But the problem is that i cannot rely on the `$preserveKeys` variable since the default is null and in that case the method will determine to preserve the keys based on the `isAssoc` method.

So yeah i think using  `@return self<array-key, TValue>` with a proper comment might be better.

gturpin-dev: Thanks for the link 🙌 

I see, so yes I'm not sure that can be based on in-method code but only on parameters.

Bapawe: Hehe it only works with parameters, i already tried that 😉

gturpin-dev (approved):

innocenzi (approved):

brendt:
Thanks, appreciate it!

PR #660, opened 2024-11-02
feat(console): support negative arguments

innocenzi:
This pull request adds support for converting boolean flags that start with `--no` to their non-negative version.

For instance, the flag `--no-interaction` will be parsed as `$interaction = false`. This is a common practice in JavaScript and Rust CLIs, and a practical feature, both user and developer wise, to handle boolean flags in multiple ways, whichever feels more natural.

Non-boolean flags will be handled as before, keeping the `no-` prefix.

Here are a few examples:

| Flag                     | Name          | Value   |
| ------------------------ | ------------- | ------- |
| `--no-interaction`       | `interaction` | `false` |
| `--noInteraction`        | `interaction` | `false` |
| `--interaction`          | `interaction` | `true`  |
| `--interaction=false`    | `interaction` | `false` |
| `--interaction=true`     | `interaction` | `true`  |
| `--no-interaction=true`  | `interaction` | `false` |
| `--no-interaction=false` | `interaction` | `true`  |

### DIFF
--- a/src/Tempest/Console/src/Input/ConsoleInputArgument.php
+++ b/src/Tempest/Console/src/Input/ConsoleInputArgument.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Tempest\Console\Input;
 
+use function Tempest\Support\str;
+
 final class ConsoleInputArgument
 {
     public function __construct(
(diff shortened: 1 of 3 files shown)


brendt:
Huh, very nice!

PR #661, opened 2024-11-02
feat(console): support "no prompt" mode

innocenzi:
This pull request implements a "no prompt" mode. This mode can be enabled by passing the `--no-interaction` flag to any command.

That mode disables all user interactions with the console. Instead, default values will be used when provided, otherwise `null` or empty arrays will be returned by input components.

To support this mode, some components and their static version needed a new `default` option, which this pull request includes.

Additional changes:
- The `Console` interface has a new `disablePrompting` and `disableTty` method to, well, disable prompting and TTY programmatically.
- The `interactiveSupported` is renamed to `supportsTty`, and `supportsPrompting` has been added. Both have been added to the interface and made public.
- The `ConsoleTester` class has a new `withoutPrompting` method to ease testing.
- I also added a `dd` method to `ConsoleTester`, which is quite unrelated but I always try to reach for it.

### DIFF
--- a/src/Tempest/Console/src/Console.php
+++ b/src/Tempest/Console/src/Console.php
@@ -34,7 +34,7 @@ interface Console
         bool $multiple = false,
         bool $asList = false,
         array $validation = [],
-    ): string|array;
+    ): null|string|array;
 
     public function confirm(string $question, bool $default = false): bool;
 
@@ -56,4 +56,12 @@ interface Console
     public function when(mixed $expression, callable $callback): self;
 
     public function withLabel(string $label): self;
+
+    public function supportsTty(): bool;
+
+    public function supportsPrompting(): bool;
+
+    public function disableTty(): self;
+
+    public function disablePrompting(): self;
 }
--- a/src/Tempest/Console/src/Console.php
+++ b/src/Tempest/Console/src/Console.php
@@ -45,7 +45,7 @@ interface Console
     /**
      * @param Closure(string $search): array $search
      */
-    public function search(string $label, Closure $search): mixed;
+    public function search(string $label, Closure $search, ?string $default = null): mixed;
 
     public function info(string $line): self;
 
--- a/src/Tempest/Console/src/GenericConsole.php
+++ b/src/Tempest/Console/src/GenericConsole.php
@@ -16,6 +16,7 @@ use Tempest\Console\Components\Interactive\TextBoxComponent;
 use Tempest\Console\Components\InteractiveComponentRenderer;
 use Tempest\Console\Exceptions\UnsupportedComponent;
 use Tempest\Console\Highlight\TempestConsoleLanguage\TempestConsoleLanguage;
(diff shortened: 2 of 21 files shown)


brendt:
I merged #660, feel free to rebase, I'll review afterwards, but I'm fairly confident this can be merged 👍

src/Tempest/Console/src/Console.php
```diff
@@ -56,4 +56,12 @@ public function success(string $line): self;
     public function when(mixed $expression, callable $callback): self;
 
     public function withLabel(string $label): self;
+
+    public function supportsTty(): bool;
```
brendt, line 60: Technically, the right place for TTY stuff would be in the `Terminal` class, not on the `Console`. Think of the `Terminal` is the car's engine, while the `Console` adds the features to interact with it.

src/Tempest/Console/src/GenericConsole.php
```diff
@@ -197,17 +226,38 @@ public function progressBar(iterable $data, Closure $handler): array
...
 
-    public function search(string $label, Closure $search): mixed
+    public function search(string $label, Closure $search, ?string $default = null): mixed
     {
-        return $this->component(new SearchComponent($label, $search));
+        return $this->component(new SearchComponent($label, $search, $default));
     }
 
-    private function interactiveSupported(): bool
+    public function supportsTty(): bool
```
brendt, line 234: Ah but I see it's just a rename of `interactiveSupported`. I don't remember why I built it like this originally… Probably because I wanted to take a shortcut while hacking on interactive support, and never remembered to properly refactor it. 

It's not a problem for now, let's keep it, but it might be good to refactor it at a later point.

src/Tempest/Console/src/GenericConsole.php
```diff
@@ -197,17 +226,38 @@ public function progressBar(iterable $data, Closure $handler): array
...
     {
+        if ($this->supportsTty === false) {
+            return false;
+        }
+
         if ($this->componentRenderer === null) {
             return false;
         }
 
+        if (! $this->supportsPrompting()) {
```
brendt, line 244: So if I get it right, prompting is available for both static and dynamic components?

brendt (changes requested):

src/Tempest/Console/src/GenericConsole.php
innocenzi: I'll try to refactor it in my upcoming prompt overhaul PR 👍

src/Tempest/Console/src/GenericConsole.php
innocenzi: Yes! Prompting is letting the user type on the terminal. Disabling prompting is useful when there is no visible terminal, so in automations for instance.

brendt (approved):

brendt:
Let's go!! 🌊

PR #666, opened 2024-11-02
refactor(routing): split route construction

blackshadev:
# What 

Split route construction from matching. As discussed in #626 , it would be nice if the routing construction is concentrated in the route discovery and clearly separated from the route matching. This means that all logic is moved away from the RouteConfig and it is used just as a storage container.

## Benchmarks

- Dynamic route matching remained roughtly the same. 
- Route adding has remained the same. 
- Serialization of the route config has become significantly faster (432.739μs to 364.133μs, small in absolute value, but relative to each other is an improvement of 15% ).
- Added a benchmark for the RouteConfig construction which includes the whole regex creation

### Main 
```
+--------------------------+--------------------+--------------------+------+-----+----------+-----------+--------+
| benchmark                | subject            | set                | revs | its | mem_peak | mode      | rstdev |
+--------------------------+--------------------+--------------------+------+-----+----------+-----------+--------+
| RouteConfigBench         | benchRoutingSetup  |                    | 1000 | 5   | 2.113mb  | 1.100ms   | ±0.70% |
| RouteConfigBench         | benchSerialization |                    | 1000 | 5   | 2.947mb  | 432.739μs | ±0.57% |
| GenericRouteMatcherBench | benchMatch         | Dynamic            | 1000 | 5   | 2.552mb  | 4.921μs   | ±1.36% |
| GenericRouteMatcherBench | benchMatch         | Non existing long  | 1000 | 5   | 2.552mb  | 5.308μs   | ±1.01% |
| GenericRouteMatcherBench | benchMatch         | Non existing short | 1000 | 5   | 2.552mb  | 3.070μs   | ±1.50% |
| GenericRouteMatcherBench | benchMatch         | Static route       | 1000 | 5   | 2.552mb  | 2.881μs   | ±1.44% |
+--------------------------+--------------------+--------------------+------+-----+----------+-----------+--------+
```

### This branch

```
+-----------------------------+-----------------------------------------+--------------------+------+-----+----------+-----------+--------+
| RouteConfigBench            | benchSerialization                      |                    | 1000 | 5   | 2.721mb  | 364.133μs | ±0.80% |
| GenericRouteMatcherBench    | benchMatch                              | Dynamic            | 1000 | 5   | 2.434mb  | 4.982μs   | ±2.19% |
| GenericRouteMatcherBench    | benchMatch                              | Non existing long  | 1000 | 5   | 2.434mb  | 5.324μs   | ±2.52% |
| GenericRouteMatcherBench    | benchMatch                              | Non existing short | 1000 | 5   | 2.434mb  | 3.085μs   | ±1.07% |
| GenericRouteMatcherBench    | benchMatch                              | Static route       | 1000 | 5   | 2.434mb  | 3.049μs   | ±1.46% |
| RouteConfigConstructorBench | benchRouteConfigConstructionToConfig    |                    | 1000 | 5   | 2.131mb  | 65.156μs  | ±0.30% |
| RouteConfigConstructorBench | benchRouteConfigConstructionRouteAdding |                    | 1000 | 5   | 2.101mb  | 1.097ms   | ±1.12% |
+-----------------------------+-----------------------------------------+--------------------+------+-----+----------+-----------+--------+
```

### DIFF
--- a/src/Tempest/Http/src/RouteConfig.php
+++ b/src/Tempest/Http/src/RouteConfig.php
@@ -4,81 +4,22 @@ declare(strict_types=1);
 
 namespace Tempest\Http;
 
-use Tempest\Http\Routing\Construction\MarkedRoute;
-use Tempest\Http\Routing\Construction\RoutingTree;
-use Tempest\Reflection\MethodReflector;
-
 final class RouteConfig
 {
-    /** @var string The mark to give the next route in the matching Regex */
-    private string $regexMark = 'a';
-
-    /** @var array<string, string> */
-    public array $matchingRegexes = [];
-
-    public RoutingTree $routingTree;
-
     public function __construct(
-        /** @var array<string, array<string, \Tempest\Http\Route>> */
+        /** @var array<string, array<string, Route>> */
         public array $staticRoutes = [],
-        /** @var array<string, array<string, \Tempest\Http\Route>> */
+        /** @var array<string, array<string, Route>> */
         public array $dynamicRoutes = [],
+        /** @var array<string, string> */
+        public array $matchingRegexes = [],
     ) {
-        $this->routingTree = new RoutingTree();
     }
 
-    public function addRoute(MethodReflector $handler, Route $route): self
+    public function apply(RouteConfig $newConfig): void
     {
-        $route->setHandler($handler);
-
-        if ($route->isDynamic) {
-            $this->regexMark = str_increment($this->regexMark);
-            $this->dynamicRoutes[$route->method->value][$this->regexMark] = $route;
-
-            $this->routingTree->add(
-                new MarkedRoute(
-                    mark: $this->regexMark,
-                    route: $route,
-                )
-            );
-
-        } else {
-            $uriWithTrailingSlash = rtrim($route->uri, '/');
-
-            $this->staticRoutes[$route->method->value][$uriWithTrailingSlash] = $route;
-            $this->staticRoutes[$route->method->value][$uriWithTrailingSlash . '/'] = $route;
-        }
-
-        return $this;
-    }
-
-    public function prepareMatchingRegexes(): void
-    {
-        if (! empty($this->matchingRegexes)) {
-            return;
-        }
-
-        $this->matchingRegexes = $this->routingTree->toMatchingRegexes();
-    }
-
-    /**
-     * __sleep is called before serialize and returns the public properties to serialize. We do not want the routingTree
-     * to be serialized, but we do want the result to be serialized. Thus prepareMatchingRegexes is called and the
-     * resulting matchingRegexes are stored.
-     */
-    public function __sleep(): array
-    {
-        $this->prepareMatchingRegexes();
-
-        return ['staticRoutes', 'dynamicRoutes', 'matchingRegexes'];
-    }
-
-    /**
-     * __wakeup is called after unserialize. We do not serialize the routingTree thus we need to provide some default
-     *  for it. Otherwise, it will be uninitialized and cause issues when tempest expects it to be defined.
-     */
-    public function __wakeup(): void
-    {
-        $this->routingTree = new RoutingTree();
+        $this->staticRoutes = $newConfig->staticRoutes;
+        $this->dynamicRoutes = $newConfig->dynamicRoutes;
+        $this->matchingRegexes = $newConfig->matchingRegexes;
     }
 }
--- a/src/Tempest/Http/src/RouteDiscovery.php
+++ b/src/Tempest/Http/src/RouteDiscovery.php
@@ -8,13 +8,13 @@ use Tempest\Container\Container;
 use Tempest\Core\Discovery;
 use Tempest\Core\KernelEvent;
 use Tempest\EventBus\EventHandler;
-use Tempest\Http\Routing\Construction\RouteConfigConstructor;
+use Tempest\Http\Routing\Construction\RouteConfigurator;
 use Tempest\Reflection\ClassReflector;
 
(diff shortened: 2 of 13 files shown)


src/Tempest/Http/src/RouteDiscovery.php
```diff
@@ -24,21 +28,31 @@ public function discover(ClassReflector $class): void
                 continue;
             }
 
-            $this->routeConfig->addRoute($method, $routeAttribute);
+            $routeAttribute->setHandler($method);
```
blackshadev, line 31: Moved the handler setting from the RouteConfig to the Discovery .

It seemed really odd to me to move state mutations (setHandler) down (more method calls down) instead of up (more to the surface). I like to have to think as less as possible about the different variations a object has to be and at what time (I really like to use immutable values for this reason).

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -4,81 +4,22 @@
...
         public array $staticRoutes = [],
         /** @var array<string, array<string, \Tempest\Http\Route>> */
         public array $dynamicRoutes = [],
+        /** @var array<string, string> */
+        public array $matchingRegexes = [],
     ) {
-        $this->routingTree = new RoutingTree();
-    }
-
-    public function addRoute(MethodReflector $handler, Route $route): self
```
blackshadev, line 30: This all is moved to `Routing/Construction/RouteConfigConstructor`. This RouteConfig doesn't have any behavior any more, it has become a plain old data carrying object. "unfortunatly" I couldn't make it immutable since this is treaded as a singleton we serialize, thus we need to mutate state. (I think, I could look into it more)

src/Tempest/Http/src/RouteConfig.php
```diff
@@ -4,81 +4,22 @@
...
-     *  for it. Otherwise, it will be uninitialized and cause issues when tempest expects it to be defined.
-     */
-    public function __wakeup(): void
+    public function apply(RouteConfig $newConfig): void
     {
-        $this->routingTree = new RoutingTree();
+        $this->staticRoutes = $newConfig->staticRoutes;
+        $this->dynamicRoutes = $newConfig->dynamicRoutes;
+        $this->matchingRegexes = $newConfig->matchingRegexes;
     }
```
blackshadev, line 24: I didn´t like how we mutate these properties externally. That is why I opt-ed to have a method to update the config. Since all mutation paths potentially mutate all these properties I passed a new RouteConfig to apply instead. Otherwise I would either: copy call properties into this method, Or have a separate value object te carry the exact same properties.

src/Tempest/Http/src/RouteDiscovery.php
```diff
@@ -24,21 +28,31 @@ public function discover(ClassReflector $class): void
...
+        }
+    }
+
+    #[EventHandler(KernelEvent::BOOTED)]
+    public function finishDiscovery(): void
+    {
+        if ($this->routeConfigConstructor->hasRoutes()) {
+            $this->routeConfig->apply($this->routeConfigConstructor->toRouteConfig());
         }
     }
```
blackshadev, line 43: Previously the method `prepareMatchingRegexes` would ensure the tree is build and applied into the routeConfig. This method needed to be called before creating the cache and before route matching. 

I moved the latter to the KernelEvent Booted to seperate the route matching from the route discovery as much as possible.

src/Tempest/Http/src/RouteDiscovery.php
```diff
@@ -24,21 +28,31 @@ public function discover(ClassReflector $class): void
...
+            $routeAttribute->setHandler($method);
+
+            $this->routeConfigConstructor->addRoute($routeAttribute);
+        }
+    }
+
+    #[EventHandler(KernelEvent::BOOTED)]
+    public function finishDiscovery(): void
+    {
+        if ($this->routeConfigConstructor->hasRoutes()) {
```
blackshadev, line 40: We do need to check if we have any new routes to apply because we have two paths coming here, when caching is not enabled it will be hit once by the KernelEvent::Booted. But with caching enabled it can be it two times in one run since we need the RouteConfig fully constructed when creating the cache as well, which is called before the booted event. Thus it will be called twice.

src/Tempest/Http/src/Routing/Construction/RouteConfigConstructor.php
```diff
@@ -0,0 +1,81 @@
...
+
+use Tempest\Container\Singleton;
+use Tempest\Http\Route;
+use Tempest\Http\RouteConfig;
+
+/**
+ * @internal
+ */
+#[Singleton]
+final class RouteConfigConstructor
```
blackshadev, line 15: All logic for building up the routing tree during discovery is moved here, it was originally in the route config. This singleton service will keep the state during construction. And it will result in a `RouteConfig` as output. This can be applied to the DI-ed RouteConfig

src/Tempest/Http/src/Routing/Construction/RouteConfigConstructor.php
```diff
@@ -0,0 +1,88 @@
...
```
brendt, line 15: This is great, I'm only a little in doubt about the name. Not a dealbreaker btw, but maybe something like `RouteConfigurator` would fit?

brendt (commented):
Only one question, this is a great PR :)

src/Tempest/Http/src/Routing/Construction/RouteConfigConstructor.php
blackshadev: Fixed, great naming suggestion

brendt:
Let's go 🎉

PR #670, opened 2024-11-03
feat: exception handler improvements

brendt:
TODO:

- [x] Update HTTP exception handler for all cases
    - [x] Use whoops for non-production (for now)
    - [x] Display generic 500 page
- [x] ~Handle kernel exceptions before exception handlers are loaded~
- [x] ~Tests~


brendt:
About tests: I don't know what to test, because PHPUnit complains whenever you add custom exception handlers…

brendt:
I decided it was better to have exception handlers being resolved via the container, which means that they have to be registered _after_ discovery. I've made an issue for it: https://github.com/tempestphp/tempest-framework/issues/671

Issue #435, opened 2024-09-19
Add GitHub actions to close issues/pull requests


aidan-casey:
In our [contributing](https://github.com/tempestphp/tempest-framework/blob/main/.github/CONTRIBUTING.md#report-an-error-or-bug) guide, we articulate that issues will be closed after 30 days of no response. We should setup some GitHub actions to enforce this across issues and pull requests.

ericsizemore:
Could probably come up with something using [actions/stale](https://github.com/marketplace/actions/close-stale-issues). 

If you all are good with using this particular action, the workflow would be rather simple:

```yaml
name: "Close stale issues/pull requests."
on:
  schedule:
    - cron: "30 1 * * *"

jobs:
  close-issues:
    runs-on: ubuntu-latest
    permissions:
      issues: write
      pull-requests: write
    steps:
      - uses: actions/stale@v9
        with:
          days-before-stale: 30
          days-before-close: 1
          stale-issue-label: "stale"
          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
          close-issue-message: "This issue was closed because it has been inactive for 1 day since being marked as stale."
          stale-pr-label: "stale"
          stale-pr-message: "This pull request is stale because it has been open for 30 days with no activity."
          close-pr-message:  "This pull request was closed because it has been inactive for 1 day since being marked as stale."
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

There are [several options](https://github.com/marketplace/actions/close-stale-issues#repo-token) available. You can set different stale/close days for issues and PRs, or the same number of days (as is done above). Could whip up a PR for this if anyone has any recommendations on the messages used, or any other particular options/settings for the action.

ericsizemore:
Any thoughts? :)

innocenzi:
The action looks good to me, but it will close a lot of the backlog @brendt has built up in the issues, so we probably need to use one of the `exempt` options

Not sure if we should exempt using assignees or labels? Or maybe just clear the backlog before... it's never good to have a big backlog 😆

ericsizemore:
Another option would be to use the `start-date` setting: https://github.com/marketplace/actions/close-stale-issues#start-date .

> The action looks good to me, but it will close a lot of the backlog @brendt has built up in the issues, so we probably need to use one of the `exempt` options
> 
> Not sure if we should exempt using assignees or labels? Or maybe just clear the backlog before... it's never good to have a big backlog 😆

[...]

PR #682, opened 2024-11-06
ci: close stale issues and pull requests

ericsizemore:
Fixes [#435](https://github.com/tempestphp/tempest-framework/issues/435)

The `start-date` setting was set to November 6th @ midnight UTC to keep from potentially closing too many issues initially. This obviously can be changed, or we could decide on a different method of exempting certain issues PR's (by label, assignee, etc.).

There are [several options](https://github.com/marketplace/actions/close-stale-issues#repo-token) available. You can set different stale/close days for issues and PRs, or the same number of days (as is done in this PR). Open to recommendations on the messages used, or any other particular options/settings for the action.


ericsizemore:
@aidan-casey

ericsizemore:
> Thanks, @ericsizemore! This looks great.
> 
> I do think we should update the label to be "Stale" with a capital "S" to be consistent with our current labels.
> 
> I'm also wondering if we do want to allow for a couple of exempt labels. Let me take a look at our labels and see if there are any worth adding.

Sure, just let me know. :)

aidan-casey:
Going to merge this as is. Thank you, @ericsizemore!

PR #685, opened 2024-11-06
feat(commandbus): async commands

### LINKED ISSUE #314: Async commands
We need a way to dispatch background jobs to a queue. We could make a CLI-only interface for starters. We already have a scheduler component which might come in handy, though still tbd.

brendt:
Closes #314 

TODO

- [ ] Docs


src/Tempest/CommandBus/src/MonitorAsyncCommands.php
```diff
@@ -0,0 +1,84 @@
...
+                $this->sleep(0.5);
+
+                continue;
+            }
+
+            // Start a task
+            $uuid = $availableUuids->first();
+            $time = new DateTimeImmutable();
+            $this->writeln("<h2>{$uuid}</h2> started at {$time->format('Y-m-d H:i:s')}");
+            $process = new Process(['php', 'tempest', 'command:handle', $uuid], getcwd());
```
brendt, line 74: Make configurable

src/Tempest/CommandBus/src/AsyncCommandRepository.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\CommandBus;
+
+interface AsyncCommandRepository
```
brendt, line 7: Rename to `CommandRepository`?

src/Tempest/CommandBus/src/AsyncCommandRepositoryInitializer.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Tempest\CommandBus\AsyncCommandRepositories\FileCommandRepository;
+use Tempest\Container\Container;
+use Tempest\Container\Initializer;
+
+final readonly class AsyncCommandRepositoryInitializer implements Initializer
+{
+    public function initialize(Container $container): AsyncCommandRepository
+    {
+        // TODO: refactor to make it configurable
```
brendt, line 15: Make configurable

brendt:
@aidan-casey @innocenzi can you give a review?

src/Tempest/CommandBus/src/AsyncCommandRepositories/FileCommandRepository.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function markAsFailed(string $uuid): void
+    {
+        rename(
+            from: __DIR__ . "/../stored-commands/{$uuid}.pending.txt",
+            to: __DIR__ . "/../stored-commands/{$uuid}.failed.txt",
+        );
+    }
+
+    public function getPendingCommands(): array
```
brendt, line 48: We'll run into performance issue if we're always returning _all_ pending commands. I think it outside the scope of this PR to optimize that though. Let's first play a bit with it as-is and see what other problems surface

src/Tempest/Console/src/Input/ConsoleArgumentBag.php
```diff
@@ -132,6 +132,11 @@ public function add(ConsoleInputArgument $argument): self
         return $this;
     }
 
+    public function getBinaryPath(): string
```
brendt, line 135: I plan on refactoring this after this PR: #711

src/Tempest/CommandBus/src/CommandBusConfig.php
```diff
@@ -14,6 +15,9 @@ public function __construct(
 
         /** @var array<array-key, class-string<\Tempest\CommandBus\CommandBusMiddleware>> */
         public array $middleware = [],
+
+        /** @var class-string<\Tempest\CommandBus\CommandRepository> $commandRepositoryClass */
+        public string $commandRepositoryClass = FileCommandRepository::class,
```
innocenzi, line 20: If we want a different driver in production, we'll need to specify the FQCN of the repository in environment variables so we can do this:

```php
return new CommandBusConfig(
    commandRepositoryClass: env('COMMAND_REPOSITORY'),
);

// .env
COMMAND_REPOSITORY="Tempest\CommandBus\AsyncCommandRepositories\FileCommandRepository"
```

This is not ideal. How about abstracting this away? Might be out-of-scope for this PR, but improving on this later will be a breaking change too.

```php
use Tempest\CommandBus\Driver;

return new CommandBusConfig(
    asyncCommandDriver: Driver::from(env('COMMAND_DRIVER')),
);
```

innocenzi (commented):
Functionality-wise: seems good to me. This is something we can improve on later to provide more advanced functionality, like balancing strategies.

Code-wise: just curious about the usage of `HasConsole`. Since we inject `Console` in the constructor anyway, why not use it directly?

brendt:
> Code-wise: just curious about the usage of HasConsole. Since we inject Console in the constructor anyway, why not use it directly?

No big reason for this… just something I've gotten used to as I prefer to call `$this->` instead of `$this->console`.

src/Tempest/CommandBus/src/CommandBusConfig.php
brendt: Gave it some thought, it's not as trivial as I imagined: if we want a mapping somewhere, we need to know of all available `CommandRepository` implementations. Which means adding discovery for it, because we want to support project and third part repository implementations as well. That's a lot of overhead. 

This is actually a problem not limited to this PR, so I'm going to keep it as-is for now, and we need to think about ways to improve it.

PR #690, opened 2024-11-06
feat(container): support injecting properties using `#[Inject]`

brendt:


src/Tempest/Console/src/HasConsole.php
```diff
@@ -5,12 +5,12 @@
...
 use Closure;
+use Tempest\Container\Inject;
 
 trait HasConsole
 {
-    public function __construct(private readonly Console $console)
-    {
-    }
+    #[Inject]
+    private readonly Console $console; // @phpstan-ignore-line
```
gturpin-dev, line 13: Why do we need to ignore something from phpstan here ?
It will be always the case in all other implementations ?

brendt: The ignore-line tag actually doesn't work, since it's being used via a trait… IDK yet what the best solution is

PR #703, opened 2024-11-08
feat(console): support dynamic style injections

innocenzi:
This pull request adds support for dynamic style injections, responding to the need to write arbitrary styles. 

For context, I have this need in my upcoming "prompts" pull request—I am currently adding injections as I see the need for them, but I need too many colors (eg. `danger`, `warning`, `hint`, `dim`...) which don't really fit the semantic injection paradigm we have now.

Essentially, the following is now possible:

```php
$this->console->write('<style="fg-cyan bg-white underline">Tempest</style>');
```

<details>
<summary>Before edit</summary>

```php
// foreground color
$this->console->write('<fg=cyan>Tempest</fg>');
$this->console->write('<fg=darkcyan>Tempest</fg>');

// background color
$this->console->write('<bg=cyan>Tempest</bg>');
$this->console->write('<bg=darkcyan>Tempest</bg>');

// modifiers
$this->console->write('<mod=underline>Tempest</mod>');
$this->console->write('<mod=bold>Tempest</mod>');
```

</details>

This works by adding a new `DynamicInjection` injection, which will parse the tag and styles and will look for a corresponding values in the `TerminalStyle` enum from `tempest/highlight`.

See also: https://github.com/tempestphp/highlight/pull/162

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -117,6 +117,7 @@
             "Tempest\\Cache\\Tests\\": "src/Tempest/Cache/tests",
             "Tempest\\Clock\\Tests\\": "src/Tempest/Clock/tests",
             "Tempest\\CommandBus\\Tests\\": "src/Tempest/CommandBus/tests",
+            "Tempest\\Console\\Tests\\": "src/Tempest/Console/tests",
             "Tempest\\Container\\Tests\\": "src/Tempest/Container/tests",
             "Tempest\\Core\\Tests\\": "src/Tempest/Core/tests",
             "Tempest\\Database\\Tests\\": "src/Tempest/Database/tests",
--- a/src/Tempest/Console/composer.json
+++ b/src/Tempest/Console/composer.json
@@ -21,6 +21,11 @@
             "Tempest\\Console\\": "src"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "Tempest\\Console\\Tests\\": "tests"
+        }
+    },
     "bin": [
         "bin/tempest"
     ]
--- a/src/Tempest/Console/src/Highlight/DynamicTokenType.php
+++ b/src/Tempest/Console/src/Highlight/DynamicTokenType.php
@@ -11,30 +11,22 @@ use function Tempest\Support\str;
 final readonly class DynamicTokenType implements TokenType
 {
     public function __construct(
-        private string $tag,
-        private string $mod,
+        private string $style,
     ) {
     }
 
     public function getStyle(): TerminalStyle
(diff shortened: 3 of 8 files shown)


brendt:
I think it's a good feature, but I always disliked Symfony's style tags (these look very similar). Couldn't we do something like 

`<style="fg-cyan bg-red underline">…</style>`

Or something like 

`<fg-cyan bg-red underline>…</>`

IDK about the syntax, but I'd like to brainstorm some more about it first.

innocenzi:
Fair enough, here are my thoughts:

- The advantage of `<fg>`, `<bg>` and `<mod>` is that it's easy to implement with regex and concise, but you need multiple tags to apply multiple styles at once
- The advantage of `<style>` is that you can have multiple styles at once, but the tag itself is longer
- I personally don't really like `<fg-cyan bg-red>` because it would be harder to parse and it's an unusual syntax

Out of those three, I don't have a strong opinion, I like both the `<fg>` one and `<style>` one

@aidan-casey any opinion here?

brendt:
> A token-based approach, vastly different than the tag-based one. Super easy to parse, very flexible for the user, but a slightly worse readability. I think the readability tradeoff might be acceptable because it's a syntax that would not often be needed.

There might be something to this. I _do_ think with need something else than the `@` approach, because of whitespacing, that'll get confusing very quickly.

```
{bg-dark-red}{fg-white}{underline}Tempest{reset}
```

But writing it like this reminds me that it's essentially the same thing Symfony does…

```php
<info>Tempest</>
```

I was reminded of how tempest/highlight works:

```php
{:bg-dark-red:fg-white:underline:Tempest:}
```

But that's more difficult to parse again, and not really readable… 

That makes me circle back to what you originally implemented, or maybe the `style` tag. I like the idea of being able to combine multiple styles in one tag, I think that's slightly more convenient. It's definitely something I missed when using Symfony's console styling

brendt:
Cool! Is this ready for a final review then?

innocenzi:
@brendt yes it is!

brendt:
This is so great, I've got nothing to add! 

Oh maybe one more thing after merging: did you intend to write docs for this, or do you want me to tackle that?

innocenzi:
To be honest I initially thought of this as a kind of internal feature, but now that I think more about it, you're right—it could totally be documented

If you want to tackle it, feel free, otherwise I'll do it with my console components PR :)

brendt:
I've made an issue for it, we'll see who gets to it first

PR #704, opened 2024-11-08
feat(console): ensure `tempest serve` supports routes with file extension

blackshadev:
# What

- extends `tempest serve` with a router script to control what it will serve as static file and what will continue through tempest
- The old/current implementation doesn't allow routes with known file extensions (eg `.json`) to hit tempest  even if the file does not exist

# How to test

- Add an route `#[Get('/example/{file}')]` to a known controller
- run `tempest serve` and try accessing the route through `curl http://localhost:8000/example/my_file.json` 
  - On my branch this will work
  - On main this will result in a 404 (given that `example/my_file.json` doesn't exist)


brendt:
Nice, thanks! Ideally we add some tests for this, feel free to take a stab at it if you want to: https://github.com/tempestphp/tempest-framework/issues/708

I can take care of it as well btw, no problem

PR #706, opened 2024-11-09
fix(commandbus): disallow having two `#[CommandHandler]` for the same command

### LINKED ISSUE #683: Throw exception for two or more command handlers that handle the same command

Bapawe:
Fixes #683


brendt:
Thank you! Appreciate it!

Issue #547, opened 2024-10-05
[Feature Request]: Support for missing SQL data types


hanieas:
### Description

After reviewing the current table statement implementations, I noticed that some types are not yet supported, including `BOOLEAN `, `ENUM `, `SET`, and `JSON`. I would like to contribute by implementing these missing statements.

Would it be acceptable to work on these types and submit them as separate PRs?

### Benefits

/

brendt:
Definitely!

PR #709, opened 2024-11-09
feat(database): add json data type

MrYamous:
Part of #547 to add the `json` data type to the table statements

Latest versions of MySQL accept default value for `json` but this require a version check.
I didn't want to add this complexity (at least for this pr) as this was not done either for `text` type before

Do you think we should check if value provided as `default` is a valid json ? `json_validate` seems free to use as Tempest require 8.3+

### DIFF
--- a/src/Tempest/Database/src/QueryStatements/CreateTableStatement.php
+++ b/src/Tempest/Database/src/QueryStatements/CreateTableStatement.php
@@ -166,6 +166,20 @@ final class CreateTableStatement implements QueryStatement
         return $this;
     }
 
+    public function json(
+        string $name,
+        bool $nullable = false,
+        ?string $default = null
+    ): self {
+        $this->statements[] = new JsonStatement(
+            name: $name,
+            nullable: $nullable,
+            default: $default,
+        );
+
+        return $this;
+    }
+
     public function compile(DatabaseDialect $dialect): string
     {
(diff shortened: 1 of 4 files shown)


brendt:
> Do you think we should check if value provided as default is a valid json ? json_validate seems free to use as Tempest require 8.3+

Yes, that makes sense!

> Latest versions of MySQL accept default value for json but this require a version check.
> I didn't want to add this complexity (at least for this pr) as this was not done either for text type before

Yeah we can ignore this for now :)

We should also make sure that JSON columns are eventually mapped to array properties. But that's outside the scope of this PR.

src/Tempest/Database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -166,6 +166,20 @@ public function boolean(
         return $this;
     }
 
+    public function json(
+        string $name,
+        bool $nullable = false,
+        ?string $default = null
```
brendt, line 172: Let's use validate_json 👍

brendt (changes requested):

MrYamous:
`json_validate` added :) 
I'm not sure about the Exception name and the new test, let me know if this is suitable

brendt (approved):

brendt:
Thanks!

PR #712, opened 2024-11-10
refactor(http): use `symfony/process` to run `tempest serve`

blackshadev:
During a separate discussion on discord. Aiden brought to my attention env-variables shouldn't be used for this.

I drew inspiration from [laravel serve](https://github.com/laravel/framework/blob/37f96f1d6b515f55ea7dfef144fd41519cca38b6/src/Illuminate/Foundation/Console/ServeCommand.php#L103). Which doesn't use an env-var, but uses the `cwd` of the sub-process instead.

The while-loop with a `usleep` is also inspired by laravel's serve command. I couldn't find an easier way to start a subprocess (like `passtru`) for which I could alter the `cwd`

### DIFF
--- a/src/Tempest/Http/src/Commands/ServeCommand.php
+++ b/src/Tempest/Http/src/Commands/ServeCommand.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Http\Commands;
 
+use Symfony\Component\Process\Process;
 use Tempest\Console\ConsoleCommand;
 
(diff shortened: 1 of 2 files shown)


blackshadev:
#717  seems an easier fix. I would prefer that to be merged. We can close mine, and look back at it if the other fix doesn't work as expected or if we do need more control over the child process at some point.

brendt:
So I guess this one can be closed in favor of #717 ?

PR #713, opened 2024-11-10
Feat/add str start helper

pxpm:
Inline with the `finish()` string helper, this function allow the developer to specify a prefix (start) for the string.

### DIFF
--- a/src/Tempest/Support/src/StringHelper.php
+++ b/src/Tempest/Support/src/StringHelper.php
@@ -183,6 +183,16 @@ final readonly class StringHelper implements Stringable
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Tempest/Support/src/StringHelper.php
```diff
@@ -182,6 +182,16 @@ public function finish(string $cap): self
         );
     }
 
+    /**
+     * Prefix the instance with the given string.
```
innocenzi, line 186:
```suggestion
     * Prefixes the instance with the given string.
```

innocenzi (approved):

innocenzi:
Thanks!

PR #714, opened 2024-11-10
feat(routing): add regex chunking to route regex

blackshadev:
# What

Add regex chunking to route regex. When loading in massive amounts of routes a singular regex can become to large for PCRE to process. A solution for this is to split it up in multiple separate regexes, this is called chunking. With this PR is should be possible to load in as many routes as you like, as long as your memory can hold it. 

In order to facilitate this I needed to rework how we build up the regex, instead of using recursion and having the stack implicitly as a php call stack. I needed to implement it manually such that, when it is detected we require a new chunk, we could rebuild the regex prefix based on the stack and also finish it based on it.

# Outstanding debates

## RouteMatchingRegexesBuilder performance considerations

I isolated the regex buildin into `src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php` and one singular function. This is mostly because of performance reasons.  The inner part of the loop is called so many times, every function call needed to count. So there is a trade-off  between readability and performance. Currently I opted to go slightly more towards the performance side. If you think it should be implemented into multiple methods, do know it comes at a cost. (If needed, we can use the benchmarks to signify the cost of alterations).

This is also the reason for using `null` as a end marker instead of wrapping all nodes into a VO which also would translate to an end-marker. I saw a performance drop of arround 60%, I went from 120μs to 190μs

## PHPunit memory limit

I bumped the phpunit memory limit. I couldn't find a "leak" in the newly added code. And it mainly breaks on the integration part. My expectation is the integration tests are already quite hefty on memory and my changes tipped it over the limit. I used memprof to find any culprits, but I couldn't find any glaring issues, most memory consumption is not in "my" code or even near it.

### DIFF
--- a/src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
+++ b/src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+use Tempest\Http\Routing\Matching\MatchingRegexes;
+
+final readonly class RouteMatchingRegexesBuilder
+{
+    // This limit is guesstimated using a small script with an ever in pattern feed into preg_match
+    private const int PREG_REGEX_SIZE_LIMIT = 32764;
+
+    private const int REGEX_SIZE_MARGIN = 264;
+
+    private const REGEX_SIZE_LIMIT = self::PREG_REGEX_SIZE_LIMIT - self::REGEX_SIZE_MARGIN;
+
+    public function __construct(private RouteTreeNode $rootNode)
+    {
+    }
+
+    public function toRegex(): MatchingRegexes
+    {
+        // Holds all regex "chunks"
+        $regexes = [];
+
+        // Current regex chunk
+        $regex = '';
+        // Used to track how to 'end' a regex chunk partially in the building process
+        $regexBack = '';
+
+        /** @var (RouteTreeNode|null)[] $workingSet */
+        $workingSet = [$this->rootNode];
+
+        // Track how 'deep' we are in the tree to be able to rebuild the regex prefix when chunking
+        /** @var RouteTreeNode[] $stack */
+        $stack = [];
+
+        // Processes the working set until it is empty
+        while ($workingSet !== []) {
+            // Use array_pop for performance reasons, this does mean that the working set works in a fifo order
+            /** @var RouteTreeNode|null $node */
+            $node = array_pop($workingSet);
+
+            // null values are used as an end-marker, if one is found pop the stack and 'close' the regex
+            if ($node === null) {
+                array_pop($stack);
+                $regex .= $regexBack[0];
+
+                $regexBack = substr($regexBack, 1);
+
+                continue;
+            }
+
+            // Checks if the regex is getting to big, and thus if we need to chunk it.
+            if (strlen($regex) > self::REGEX_SIZE_LIMIT) {
+                $regexes[] = '#' . substr($regex, 1) . $regexBack . '#';
+                $regex = '';
+
+                // Rebuild the regex match prefix based on the current visited parent nodes, known as 'the stack'
+                foreach ($stack as $previousNode) {
+                    $regex .= '|' . self::routeNodeSegmentRegex($previousNode);
+                    $regex .= '(?';
+                }
+            }
+
+            // Add the node route segment to the current regex
+            $regex .= '|' . self::routeNodeSegmentRegex($node);
+            $targetRouteRegex = self::routeNodeTargetRegex($node);
+
+            // Check if node has children to ensure we only use branches if the node has children
+            if ($node->dynamicPaths !== [] || $node->staticPaths !== []) {
+                // The regex uses "Branch reset group" to match different available paths.
+                // two available routes /a and /b will create the regex (?|a|b)
+                $regex .= '(?';
+                $regexBack .= ')';
+                $stack[] = $node;
+
+                // Add target route regex as an alteration group
+                if ($targetRouteRegex) {
+                    $regex .= '|' . $targetRouteRegex;
+                }
+
+                // Add an end marker to the working set, this will be processed after the children has been processed
+                $workingSet[] = null;
+
+                // Add dynamic routes to the working set, these will be processed before the end marker
+                foreach ($node->dynamicPaths as $child) {
+                    $workingSet[] = $child;
+                }
+
+                // Add static routes to the working set, these will be processed first due to the array_pop
+                foreach ($node->staticPaths as $child) {
+                    $workingSet[] = $child;
+                }
+
+            } else {
+                // Add target route to main regex without any children
+                $regex .= $targetRouteRegex;
+            }
+        }
+
+        // Return all regex chunks including the current one
+        return new MatchingRegexes([
+            ...$regexes,
+            '#' . substr($regex, 1) . '#',
+        ]);
+    }
+
+    /**
+     * Create regex for the targetRoute in node with optional slash and end of line match `$`.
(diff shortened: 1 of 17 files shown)


src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
```diff
@@ -0,0 +1,136 @@
...
+                $regexBack .= ')';
+                $stack[] = $node;
+
+                // Add target route regex as an alteration group
+                if ($targetRouteRegex) {
+                    $regex .= '|' . $targetRouteRegex;
+                }
+
+                // Add an end marker to the working set, this will be processed after the children has been processed
+                $workingSet[] = null;
```
blackshadev, line 85: I really didn't like using `null` as a meaning-full value . But... When wrapping these things into the smallest possible value-object this whole section became around 60% slower.. I didn't think the readability was worth that much performance loss. (See PR description `RouteMatchingRegexesBuilder performance considerations` for more info)

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
```diff
@@ -0,0 +1,136 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+use Tempest\Http\Routing\Matching\MatchingRegexes;
+
+final readonly class RouteMatchingRegexesBuilder
+{
+    // This limit is guesstimated using a small script with an ever in pattern feed into preg_match
+    private const int PREG_REGEX_SIZE_LIMIT = 32764;
```
brendt, line 12: Wouldn't it make more sense if this was a factor of 2?

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
brendt: Agree :)

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Construction;
+
+use Tempest\Http\Routing\Matching\MatchingRegexes;
+
+final readonly class RouteMatchingRegexesBuilder
```
brendt, line 9: Would something like `RouteMatchingRegexBuilder` also be a correct name? `Regexes` feels weird to me, and I think `Regex` is already plural

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
```diff
@@ -12,12 +12,12 @@
 final class RouteTreeNode
 {
     /** @var array<string, RouteTreeNode> */
-    private array $staticPaths = [];
+    public array $staticPaths = [];
```
brendt, line 15: My guess is that these props can't be readonly?

src/Tempest/Http/src/Routing/Matching/MatchingRegexes.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function match(string $uri): RouteMatch
+    {
+        foreach ($this->patterns as $pattern) {
+            $matchResult = preg_match($pattern, $uri, $matches);
+
+            if ($matchResult === false) {
+                throw new RuntimeException("Failed to use matching regex. Got error " . preg_last_error());
+            }
```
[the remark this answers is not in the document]
brendt: If I understand you correctly, this won't happen in a real application where we should always have valid regex patterns?

src/Tempest/Http/src/Routing/Matching/RouteMatch.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Http\Routing\Matching;
+
+use Tempest\Http\Routing\Construction\MarkedRoute;
+
+final readonly class RouteMatch
+{
+    private function __construct(
+        public bool $isFound,
```
brendt, line 12: I find it a bit weird that there's a boolean here. I'd expect this object to be always valid when a route was matched. In other words: it should be `null` when there's no match. Does that make sense?

brendt:
Great PR 👍

src/Tempest/Http/src/Routing/Construction/RouteTreeNode.php
blackshadev: Correct these are mutated curing route construction. see `findOrCreateNodeForSegment` and `setTargetRoute` in this file

src/Tempest/Http/src/Routing/Matching/RouteMatch.php
blackshadev: Personally I don't like to put a meaning onto the value "null" if I can help it. I can however, make "isFound" a method which checks on the existence of `$mark` field. That way we do still have an explicit meaning of a "not found" route. But you don't have the need for this explicit boolean  field. Would that suffice?

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
blackshadev: Makes sense, will rename

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
blackshadev: I feel so stupid now. The limit IS a factor of 2. But I didn't specify that limit.

I found this value with an attached script. basically by increasing a pattern in `#^a$#` where the a is the size of the limit. To note that I have 4 extra characters in this regex. Thus my defined limit needs to be 4 bigger. which is a power of 2: `32768`.

Will fix.

blackshadev: done

src/Tempest/Http/src/Routing/Construction/RouteMatchingRegexesBuilder.php
blackshadev: done

src/Tempest/Http/src/Routing/Matching/RouteMatch.php
blackshadev: Implemented isFound as computed, would that suffice?

brendt:
Do you wanna take a look at the merge conflict? Other than that I'm fine merging this 👍

It's not a blocker for me, but I _do_ like for `null` to have meaning. So to me it makes more sense that `RouteMatch` can be nullable instead being `notFound`.

Look at it from a language perspective: "I have a route match" — that always means that there _is_ a matching route. When you say "I don't have a route match", it means there is _no_ matching route (ie. `null`).

blackshadev:
> Do you wanna take a look at the merge conflict? Other than that I'm fine merging this 👍

Done 

 
> It's not a blocker for me, but I _do_ like for `null` to have meaning. So to me it makes more sense that `RouteMatch` can be nullable instead being `notFound`.
> 
> Look at it from a language perspective: "I have a route match" — that always means that there _is_ a matching route. When you say "I don't have a route match", it means there is _no_ matching route (ie. `null`).

It is a matter of opinion I changed it to `null`. I do not like null because you end up with allot of null checks everywhere. While not all nulls are equal. So in our case the Route match was not found. Which is something different than an error (maybe a case for later) or a route that is missing (as in, not completely loaded in). By returning null you will be missing these cases. But as state, it is a matter of opinion, and I do not see any harm in going your way on this. More so because it is a bit more in the "tempest" style.

brendt:
Thanks! Great PR 💪

Issue #504, opened 2024-10-01
Support json bodies in requests


NeoIsRecursive:
### Description

When trying to send a json payload to a tempest controller, the requests body isn't set to an array (I believe this is how the  laminas-diactoros package works, or atleast according to https://stackoverflow.com/a/73671455), which causes errors when doing validation or accessing the properties on a "custom" reqeust.

When I added something like this in the router dispatch method it seems to work, but I guess its not the best solution:

```php
/** @var PsrRequest $request */
if ($request->getHeader('content-type')[0] === 'application/json') {
    $request = $request->withParsedBody(json_decode($request->getBody()->getContents()));
}
```

### Benefits

Well, the ability to build json apis

brendt:
Yeah this should definitely work. I'll look into it!

brendt:
Fix will be included in alpha.6

PR #715, opened 2024-11-11
fix(http): parse json body from request

mtk3d:
Fix for: https://github.com/tempestphp/tempest-framework/issues/504

**This PR fixes a couple of issues:**

_Raw Body Mocking and Header Resetting in HttpRouterTester:_

- JSON body data doesn’t show up in global variables but in the `php://input` stream, so to test it, I've added `$rawBody` in `HttpRouterTester` to mock that body.
- Also found a problem during testing—headers were being set in global variables and sticking around across tests until the PHP process ends. Now, headers get reset between tests to avoid that.

_JSON Body Parser for Laminas Requests:_

- Added a `JsonParserServerRequest` class to handle JSON parsing in Laminas requests. This class grabs data from the body stream, parses it, and adds it to the request as `$parsedBody`.
- Thought about other ways to do this, like using `ServerRequestFilter` from Laminas, but that seems more for filtering.
- Middleware was another option, but Tempest middlewares are executed after mapping PsrRequest to Tempest Request

### DIFF
--- a/src/Tempest/Http/src/Mappers/PsrRequestToRequestMapper.php
+++ b/src/Tempest/Http/src/Mappers/PsrRequestToRequestMapper.php
@@ -7,6 +7,7 @@ namespace Tempest\Http\Mappers;
 use Psr\Http\Message\ServerRequestInterface as PsrRequest;
 use Psr\Http\Message\UploadedFileInterface;
 use Tempest\Http\GenericRequest;
+use Tempest\Http\Json\JsonParserServerRequest;
 use Tempest\Http\Method;
 use Tempest\Http\Request;
 use Tempest\Http\Upload;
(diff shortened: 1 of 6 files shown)


src/Tempest/Http/src/Mappers/PsrRequestToRequestMapper.php
```diff
@@ -24,6 +25,8 @@ public function canMap(mixed $from, mixed $to): bool
     public function map(mixed $from, mixed $to): array|object
     {
         /** @var PsrRequest $from */
+        $from = (new JsonParserServerRequest())($from);
```
brendt, line 28: I find the naming a bit odd here. This is an action class that makes a `ServerRequestInterface`, couldn't it be called `ParseJsonServeRequest` or something alike?

Also, is there a reason you opted for making this a class instead of a private method?

brendt (changes requested):
Nice PR, thanks. I've got one question

PR #717, opened 2024-11-11
fix(http): use document root instead of env in `tempest serve`

aazsamir:
This https://github.com/tempestphp/tempest-framework/issues/704 pull request added router for `./tempest serve` to correctly handle paths with file extension.

It uses `putenv` but on default php environment `$_ENV` is not populated.
It depends on `php.ini` setting `variables_order` that often is set to `GPCS` instead of `EGPCS`.

But `php` cli allows to set `$_SERVER['DOCUMENT_ROOT']` with `-t` flag, so we can use it to pass public path to router.


brendt:
Thanks!

PR #718, opened 2024-11-12
feat(log): allow usage of multiple same log channels

aazsamir:
Right now, it isn't possible to correctly declare `AppendLogChannel` more than once in `LogConfig`. It will result in same log, being logged in same channel multiple times.

This changes this behaviour, by storing `Monolog` instance by `spl_object_id`, not class name.

### DIFF
--- a/src/Tempest/Log/src/GenericLogger.php
+++ b/src/Tempest/Log/src/GenericLogger.php
@@ -10,7 +10,7 @@ use Stringable;
 
 final class GenericLogger implements Logger
 {
-    /** @var array<class-string, Monolog> */
+    /** @var array<int, Monolog> */
     private array $drivers = [];
 
     public function __construct(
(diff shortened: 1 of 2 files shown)


src/Tempest/Log/src/GenericLogger.php
```diff
@@ -74,14 +74,14 @@ private function writeLog(Level $level, string $message, array $context): void
 
     private function resolveDriver(LogChannel $channel, Level $level): Monolog
     {
-        if (! isset($this->drivers[$channel::class])) {
-            $this->drivers[$channel::class] = new Monolog(
+        if (! isset($this->drivers[spl_object_id($channel)])) {
```
brendt, line 77: why `spl_object_id` and not simply append `$this->drivers[]` ?

brendt (changes requested):

src/Tempest/Log/src/GenericLogger.php
aazsamir: I think I don't understand? `$drivers` is a map of `LogChannel` to `Monolog` instance, but as we cannot use `LogChannel` as offset, I decided to use its id (as `::class` is not unique).

Do you mean to resolve it once, and just iterate over drivers if already resolved?

brendt: My bad, I was confused for a second 😅

brendt (approved):

brendt:
Thanks!

PR #719, opened 2024-11-12
fix(http): use default log config only if no config is provided

aazsamir:
Currently tempest always adds `AppendLogChannel` and hardcoded paths, even if user provided `LogConfig`.

I think better way would be to publish `log.config.php` with sane defaults with framework, but I feel it will bring discussion about project directory structure and probably better to just avoid it for now?

Also, as this code currently lives in `HttpApplication` boot method, I don't feel comfortable with calling this in test.


brendt:
> probably better to just avoid it for now?

Yeah the idea is that Tempest ships with reasonable defaults, but people can overwrite it when needed by making their own `.config.php` files. 

>  I don't feel comfortable with calling this in test.

Fair, we need to come up with a proper way of testing these kinds of config stuff. I don't think any of them are really tested now

Issue #543, opened 2024-10-05
[Feature Request]: Improve array helper with others methods


gturpin-dev:
### Description

Hi 👋
I've read other issues and PR and the code related to the Array Helper component.
I think the purpose of it is not to be a clone of Laravel collection but a much simpler wrapper around php's native functions. ( tell me if i'm wrong )
I guess there's some methods missing like a pluck or reduce or other ones.
If you are ok with this I can write some implementations.
But is it possible to list some of methods you want and don't want ? 

Here's the updated TODO from the original PR :
- [x] put()
- [x] combine()
- [x] keys()
- [x] dump()
- [x] diff()
- [x] diffKeys()
- [x] merge()
- [x] intersect()
- [x] intersectKeys()
- [x] unique()
- [x] flip()
- [x] pad()
- [x] add()
- [x] push()
- [x] reduce()
- [x] pluck()
- [x] chunk() ?
- [x] random()
- [x] pull() ?
- [x] findKey() ?
- [x] slice()
- [x] sort() => and probably its variations
- [x] flatten() => I must check if unwrap isn't doing the same before
- [x] wrap() ?
- [x] remove()
- [x] forget() => alias of remove()
- [x] shuffle
- [x] find => inspired by `array_find` in PHP 8.4 ( check for retro-compat )

### Benefits

This will improve the array helper component and make it more useful

shaffe-fr:
I often have the use case to remove a value from an array. For such cases, I use `filter` or `reject`.
What do you think about adding the following method?
```php
@method self without(mixed $item, bool $strict = true);
```

gturpin-dev:
I don't know.
Actually, this helper don't aim to be fully-featured as the Laravel collections or others libs.
Its main goal is to have a clean API to wrap native PHP array functions and others methods that's missing in PHP.
That said, to have a good DX I'm ok to add `reject` method as an alias of `filter` to avoid frustration for developers who's used to this syntax.
But, if the `without` method behavior can be replicated easily using `filter` I don't really see the added value of such an additional method.

As I understand this, this allow you to do something like :
```php
$array = arr([1, 2, 3])->without(2, true); // [1, 3]
```

So It can be achieve using `filter` like this :

```php
$array = arr([1, 2, 3])->filter(fn (int $value) => $value !== 2); // [1, 3]
```
This is not really verbose and a lot of developers are used to this syntax.
I wonder if a without method which does less things may confuse users.

Any thought ? @aidan-casey

[...]

PR #721, opened 2024-11-12
feat(support): add more methods to `ArrayHelper` and `StringHelper`

innocenzi:
This pull request adds many methods to `ArrayHelper` and `StringHelper`. This is stuff I had the need for when working on an upcoming console components overhaul PR.

Normally, I would send multiple PRs for each change, but there are too many. Reviewing this PR commit-by-commit might be easier.

### Changes to `ArrayHelper`

- `arr(null)` instanciates an empty instance instead of `[null]`
- Added generics to `first` and `last`
- Added `int` support to `get` and `has`
- Added `flatten` and `flatMap`

### Changes to `StringHelper`

- Instanciating `StringHelper` with `null` will no longer crash
- Added `replaceAt`, for replacing a portion of a string with another a the specified position
- Added `insert`, which inserts a string at the specified position
- Added `limit`, for truncating a string
- Added `substr`, which proxies `mb_substr`
- Added `take`, which keeps only the specified amount of characters
- Added `split`, which chunks the string using `str_split`
- Added `stripTags`, for stripping HTML tags in a string

### DIFF
--- a/src/Tempest/Support/src/ArrayHelper.php
+++ b/src/Tempest/Support/src/ArrayHelper.php
@@ -811,6 +811,37 @@ final class ArrayHelper implements Iterator, ArrayAccess, Serializable, Countabl
     }
 
     /**
+     * Flattens the instance to a single-level array, or until the specified `$depth` is reached.
+     *
+     * ### Example
+     * ```php
+     * arr(['foo', ['bar', 'baz']])->flatten(); // ['foo', 'bar', 'baz']
+     * ```
+     */
+    public function flatten(int|float $depth = INF): self
+    {
+        $result = [];
+
+        foreach ($this->array as $item) {
+            if (! is_array($item)) {
+                $result[] = $item;
+
+                continue;
+            }
+
+            $values = $depth === 1
+                ? array_values($item)
+                : arr($item)->flatten($depth - 1);
+
+            foreach ($values as $value) {
+                $result[] = $value;
+            }
+        }
+
+        return new self($result);
+    }
+
+    /**
      * Dumps the instance.
      */
     public function dump(mixed ...$dumps): self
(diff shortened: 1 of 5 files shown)


MrYamous:
Great work ! 
Some personal thoughts on names
`replaceAt` and `insert` have similarities in principle, WDYT about `replaceAt` and `insertAt` or `replace` and `insert` 
`limit` is for truncating a string, what about `truncate` function ?

brendt:
> replaceAt and insert have similarities in principle, WDYT about replaceAt and insertAt or replace and insert

Agree, my preference is `replace` and `insert`, maybe we could also rename the position parameter to `$at`? Just a thought:

```php
str()->replace(at: 5, …)
```

> limit is for truncating a string, what about truncate function ?

I think `truncate` is a better name

brendt:
But of course, great PR 👍

innocenzi:
@MrYamous I agree, yes!

@brendt while I also like `replace` and `insert`, `replace` already exists. Do you suggest updating it with a new parameter? I think it would be better to go with `replaceAt` and `insertAt` to make it clearer.

Definitely agree with `truncate` though 👍

innocenzi:
Ready for another review 👍

brendt:
> while I also like replace and insert, replace already exists. Do you suggest updating it with a new parameter? I think it would be better to go with replaceAt and insertAt to make it clearer.

Oh great point! Let's do that, no problem 👍

brendt:
Btw, feel free to merge when it's done, no need waiting for me again :)

brendt:
Ah, you already made the changes. Peeeeerfect 💪

gturpin-dev:
Just updated the TODO of the `ArrayHelper` methods issue #543 😄

Issue #699, opened 2024-11-07
[Feature Request]: Use symfony/uid instead of ramsey/uuid


jonagoldman:
### Description

Symfony UID supports both [UUIDs](https://symfony.com/doc/current/components/uid.html#uuids) and [ULIDs](https://symfony.com/doc/current/components/uid.html#ulids).

To be honest, this is mainly a personal preference, as I like to use ULIDs instead of UUIDs in my code, but I also think [Symfony's UID](https://github.com/symfony/uid) code is nicer and simpler than [Ramseys UUID](https://github.com/ramsey/uuid).

### Benefits

Support for both UUIDs and ULIDs.
Developed and supported by big respected company.
Minimum PHP version PHP 8.2 instead of 8.0.
Seems to be faster: https://github.com/laravel/framework/discussions/47631

MrYamous:
I'd be glad to give a try on this topic if nobody has already started or is planning to do :)

brendt:
I don't have a strong opinion on it. I'm open for a PR :)

PR #724, opened 2024-11-13
chore: replace `ramsey/uuid` with `symfony/uid`

jonagoldman:
Replace [ramsey/uuid](https://github.com/ramsey/uuid) with [symfony/uid](https://github.com/symfony/uid). Adds support for both [UUIDs](https://symfony.com/doc/current/components/uid.html#uuids) and [ULIDs](https://symfony.com/doc/current/components/uid.html#ulids).

See https://github.com/tempestphp/tempest-framework/issues/699


innocenzi:
Thanks!

PR #725, opened 2024-11-13
feat(database): add `set` data type

MrYamous:
Part of #547  to add the `set` data type to the table statements

To be honest I have a doubt about this implementation. 
I think it can be confusing for end users to silently ignore this data type if not supported by the database. I did so inspired by `BelongsTo` or `DropConstraint` statements but I prefer the way it's done in `ShowTables` which throw an `UnsupportedDialect` exception.
But this broke tests (or at least, I don't know how to handle multiples cases from a Dialect to an other)

Btw I think it can be a good point to try to be consistent in how handle this situation


brendt:
> I think it can be confusing for end users to silently ignore this data type if not supported by the database. 

I agree. 

> But this broke tests (or at least, I don't know how to handle multiples cases from a Dialect to an other)

I want to help figure this out with you, because indeed it's important to get this right. Could you maybe push the version with the exception, that way I can play around with it some more and see what happens in the tests

MrYamous:
> > I think it can be confusing for end users to silently ignore this data type if not supported by the database.
> 
> I agree.
> 
> > But this broke tests (or at least, I don't know how to handle multiples cases from a Dialect to an other)
> 
> I want to help figure this out with you, because indeed it's important to get this right. Could you maybe push the version with the exception, that way I can play around with it some more and see what happens in the tests

@brendt I've come up with an idea to deal with this situation, so I'm back with a new version that tests all the cases, including `UnsupportedDialect` Exception, and green CI :) but that I find very verbose

brendt:
This looks good, what did you find too verbose about it?

MrYamous:
> This looks good, what did you find too verbose about it?

Nevermind, if it's good for you it's good for me

I'll send another PR a bit later to use this Exception in other statements for consistency as suggested in my first message

brendt:
Perfect, thanks!

PR #726, opened 2024-11-13
fix(console): handle nested `style` tags

innocenzi:
This pull request is a follow-up of #703. It improves support for nested `<style>` tags and better handles ending tags (`</style>`).

&nbsp;

## Nesting <style>

The first change is the support for nested `<style>` tags. I changed the regular expression to use a negative lookahead (`(?!\<style)`) that prevents matching nested `<style>` tags. Without this, the following would be matched as one block:

```
<style="dim">[<style="fg-cyan">cyan</style>]</style>
```

The replacement would then miss styles. Now, it properly matches two distinct blocks.

&nbsp;

## Handling resets

The second change is an update to how `</style>` resets the current style. Instead of using `0m` to reset all styles, it will use more specific reset codes depending on the opening tag.

For instance:

```
<style="dim"><style="fg-gray">dim-gray</style> just-dim</style>
```

Previously, after the first `</style>`, all colors and modifiers would be reset, so `just-dim` would use the default text style.

Now, `</style>` will reset the foreground color only, keeping the previous `dim` modifier intact, so `just-dim` will still be dimmed.

&nbsp;

> [!NOTE]
> This pull request depends on https://github.com/tempestphp/highlight/pull/163 for the new reset cases on `TerminalStyle`.
> Once `tempest/highlight` is tagged a new version, I'll update this PR with the new version and refactor the part of the code where I used strings instead of the missing `TerminalStyle` cases.

### DIFF
--- a/src/Tempest/Console/src/Highlight/TempestConsoleLanguage/Injections/DynamicInjection.php
+++ b/src/Tempest/Console/src/Highlight/TempestConsoleLanguage/Injections/DynamicInjection.php
@@ -20,41 +20,33 @@ final readonly class DynamicInjection implements Injection
 
     public function parse(string $content, Highlighter $highlighter): ParsedInjection
     {
-        $pattern = '/(?<match>\<style=\"(?<styles>(?:[a-z-]+\s*)+)\"\>(.|\n)*\<\/style\>)/';
+        $pattern = '/(?<match>\<style=\"(?<styles>(?:[a-z-]+\s*)+)\"\>(?:(?!\<style).|\n)*?\<\/style\>)/';
 
-        $result = preg_replace_callback(
-            pattern: $pattern,
-            callback: function ($matches) use ($highlighter, $pattern) {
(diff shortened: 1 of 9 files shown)


src/Tempest/Console/src/Highlight/TempestConsoleLanguage/Injections/DynamicInjection.php
```diff
@@ -20,41 +20,33 @@ public function getTokenType(): ConsoleTokenType
...
-                        ->toString(),
-                    subject: $matches[0],
-                );
-
-                if (preg_match($pattern, $result)) {
-                    return $this->parse($result, $highlighter)->content;
+                        ->toString();
                 }
+            );
+        } while (preg_match($pattern, $content));
```
brendt, line 48: Oh this is a clever way to avoid the nested stuff!

brendt:
Merged!

PR #732, opened 2024-11-14
feat(framework): add `config:show` command

aazsamir:
This pull request brings `tempest config:show [format=pretty {dump|pretty|file}] [--search=false] [filter=null]` command.
This command loads config loaded by tempest, and dumps it to terminal in given format.

Supported formats are
- dump: simply dumps all configs using `lw` function
- pretty: uses reflection on configs, and shows config as valid JSON. All class entries have `@type` field to annotate its class
- file: reads content of resolved configs, and dumps it along with its filepath

Both `pretty` and `file` formats are formatted using Highlight.

`--filter` arg allows to narrow returned configs by filepath or classname.
So this is possible
```
tempest config:status --filter=database.config.php
```
and
```
tempest config:status --filter=DatabaseConfig
```

`--search` arg allows you to interactively select the configuration file you want to display with `SearchComponent`.

### DIFF
--- a/src/Tempest/Console/src/Console.php
+++ b/src/Tempest/Console/src/Console.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Tempest\Console;
 
 use Closure;
+use Tempest\Highlight\Language;
 
 interface Console
 {
@@ -17,6 +18,8 @@ interface Console
     public function write(string $contents): self;
 
     public function writeln(string $line = ''): self;
+
+    public function writeWithLanguage(string $contents, Language $language): self;
 
     /**
      * @param \Tempest\Validation\Rule[] $validation
--- a/src/Tempest/Console/src/GenericConsole.php
+++ b/src/Tempest/Console/src/GenericConsole.php
@@ -19,6 +19,7 @@ use Tempest\Console\Highlight\TempestConsoleLanguage\TempestConsoleLanguage;
 use Tempest\Console\Input\ConsoleArgumentBag;
 use Tempest\Container\Tag;
 use Tempest\Highlight\Highlighter;
+use Tempest\Highlight\Language;
 
 final class GenericConsole implements Console
 {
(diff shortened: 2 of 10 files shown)


innocenzi:
So, I have a few thoughts about this:

- I love the idea

- I think the current `JSON` format should be called `PRETTY` (or similar), so `JSON` can be a non-formatted JSON string for interop (right now, reading it from stdout prints terminal codes)

- I think it would be great to allow searching for a specific config file, through the `search` console component. If I'm looking at `config:status`, I'm probably looking for a specific config, so this would help me find it.

- I'm not actually sure about the name `config:status`. It's not showing the status of a config, but it dumps it. I have no strong opinion about it, but the name doesn't feel right here

aazsamir:
> I think the current `JSON` format should be called `PRETTY` (or similar), so `JSON` can be a non-formatted JSON string for interop (right now, reading it from stdout prints terminal codes)

Right now I don't see where non-formatted JSON would be needed (maybe in some SSH connections between different systems?), But I changed it to `pretty` to get ahead of the future.

> I think it would be great to allow searching for a specific config file, through the `search` console component. If I'm looking at `config:status`, I'm probably looking for a specific config, so this would help me find it.

I didn't know about `SearchComponent`, I added a `search` argument that allows you to select a configuration. I'm not sure about this `search` naming, if it's not confusing with the `filter` argument.

> I'm not actually sure about the name `config:status`. It's not showing the status of a config, but it dumps it. I have no strong opinion about it, but the name doesn't feel right 

I agree, I changed it to my first take on this, to `config:show` with the alias `config`.

innocenzi:
👌 

> Right now I don't see where non-formatted JSON would be needed

It's useful for interop—for instance, an IDE extension

> if it's not confusing with the filter argument.

The search component and `--filter` definitely overlap, but the later is still useful for non-pretty output

brendt:
> I'm not actually sure about the name config:status. It's not showing the status of a config, but it dumps it. I have no strong opinion about it, but the name doesn't feel right here

Agree now that I see it. Let's take `config:show`, sorry for proposing status 🫣

@innocenzi feel free to review and merge when you're comfortable with it, I really like it 👍

innocenzi:
I like this, let's merge :)

PR #735, opened 2024-11-15
refactor(console): exit code as value object

brendt:
Alternative to #741
closes #729

PR #736, opened 2024-11-15
refactor(console):  exit codes as enum

### LINKED ISSUE #729: Dynamic exit codes
We need to figure out of to tackle returning exit codes besides the most common ones from console commands

brendt:
Alternative to #735
Closes #729


brendt:
This approach is really bad :p gonna close!

PR #753, opened 2024-11-19
fix(generation): simplify traits and method parameter attributes

gturpin-dev:
I don't think I opened an issue about it.
But there was an issue where the `ClassManipulator` do not simplifies class traits when they are in another namespace of the current class.
Same goes for class method parameter attributes.

This PR make those two things to be simplified.
It cleans the current `make:model` command generated stub file


brendt:
Nice!

PR #755, opened 2024-11-20
feat(validation): enhance enum validation

hungthai1401:
This PR adds:
- Supports varidic functions for only and expects
- Chainable only or expects


brendt:
Nice, thanks!

Issue #759, opened 2024-11-20
[Feature Request]: TODO list for make commands


gturpin-dev:
### Description

Hey @brendt This is my TODO list for future make commands that I'll work on soon
Tell me if you prefer to skip issue and make it and discord or elsewhere.
Could you assign me to this ?

Maybe you can tell me if you want some stuff in priority or if some commands seems useless for you.
I'll update this list while reading the whole docs

- [ ] `make:initializer`
- [ ] `make:response`
- [ ] `make:route` ( for custom routes, not sure if it's helpful )
- [ ] `make:view` ( maybe one command with flags to make class or class component or raw view file )
- [ ] `make:middleware` ( with some flags like `--http` for middleware type )
- [ ] `make:config` ( not sure if it's helpful, but it could create a file named `xxx.config.php` with the suggested name as usual and maybe return a basic config object ? ) || Or another idea could be to have a select component where user can choose the config type to create ( e.g. `DatabaseConfig` or `EventBusConfig` other config types )
- [ ] `make:migration` => could create a migration class or with a flag like `--raw` a sql migration
- [ ] `make:command` => a generator command that make another command, the circle is now complete 😄 
- [ ] `make:generator-command` => maybe a basic stub that follow other basic make commands using `PublishesFiles` and prompts
- [ ] `make:event` or `make:event-handler` => a class with a method with the `#[EventHandler]` attribute

### Benefits

Keep an eye on progress of make commands 😄

brendt:
> `make:route` make:route ( for custom routes, not sure if it's helpful )

I don't think this is necessary atm

>  make:view ( maybe one command with flags to make class or class component or raw view file )

Do you intend this command to also make regular view files, or only components? I think this will be especially useful to make view components

>  make:middleware ( with some flags like --http for middleware type )

Hmm, a flag could work. If no flags were provided, you'd get a dropdown list. But maybe you should use an enum instead of a flag?

`make:middleware [type]`

>  make:config ( not sure if it's helpful, but it could create a file named xxx.config.php with the suggested name as usual and maybe return a basic config object ? ) || Or another idea could be to have a select component where user can choose the config type to create ( e.g. DatabaseConfig or EventBusConfig other config types )

Both are cool, I think the most useful implementation would be `make:config` without  arguments, getting a prompt with all available configs, selecting one, and then create a default file in the right place.

>  make:generator-command => maybe a basic stub that follow other basic make commands using PublishesFiles and prompts

[...]

PR #760, opened 2024-11-20
feat(console): add "make:response" command

gturpin-dev:
This PR add a basic implementation of the "make:response" command.
Not sure if we must give code example in the generated file of if we must keep it empty.

It updates the TODO on #759


brendt:
Nice!

PR #762, opened 2024-11-22
feat(database): custom model query builder

hungthai1401:
This PR introduces `QueryBuilder` attribute allows custom model query builder for resuable and complexity logic

```php
#[QueryBuilder(UserQueryBuilder::class)]
final class User implements DatabaseModel
{
    use IsDatabaseModel;
    // ...
}
```

### DIFF
--- a/src/Tempest/Database/src/Builder/ModelQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/ModelQueryBuilder.php
@@ -11,8 +11,9 @@ use function Tempest\map;
 
 /**
  * @template TModelClass of DatabaseModel
+ * @phpstan-ignore-next-line because this model query builder is extended
  */
-final class ModelQueryBuilder
+class ModelQueryBuilder
 {
     private ModelDefinition $modelDefinition;
 
--- a/src/Tempest/Database/src/IsDatabaseModel.php
+++ b/src/Tempest/Database/src/IsDatabaseModel.php
@@ -10,6 +10,7 @@ use Tempest\Database\Exceptions\MissingRelation;
 use Tempest\Database\Exceptions\MissingValue;
 use function Tempest\get;
 use function Tempest\make;
+use function Tempest\reflect;
 use Tempest\Reflection\ClassReflector;
 use Tempest\Reflection\PropertyReflector;
 
@@ -73,7 +74,11 @@ trait IsDatabaseModel
      */
     public static function query(): ModelQueryBuilder
     {
-        return new ModelQueryBuilder(self::class);
+        $queryBuilderAttribute = reflect(self::class)->getAttribute(QueryBuilder::class);
+
+        return $queryBuilderAttribute === null
+            ? new ModelQueryBuilder(self::class)
+            : new ($queryBuilderAttribute->builderClass)(self::class);
     }
 
     /** @return self[] */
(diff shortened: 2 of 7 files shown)


src/Tempest/Database/src/Builder/ModelQueryBuilder.php
```diff
@@ -11,8 +11,9 @@
 
 /**
  * @template TModelClass of DatabaseModel
+ * @phpstan-ignore-next-line because this model query builder is extended
  */
-final class ModelQueryBuilder
+class ModelQueryBuilder
```
brendt, line 16: We'll need to extract an interface + trait combo, we usually don't rely on inheritance :)

src/Tempest/Database/src/IsDatabaseModel.php
```diff
@@ -73,7 +74,11 @@ public static function new(mixed ...$params): self
      */
     public static function query(): ModelQueryBuilder
     {
-        return new ModelQueryBuilder(self::class);
+        $queryBuilderAttribute = reflect(self::class)->getAttribute(QueryBuilder::class);
```
brendt, line 77: Better to use `new ClassReflector` if we already know it's a class

src/Tempest/Database/src/QueryBuilder.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class QueryBuilder
```
brendt, line 10: I'm not quite sure if the attribute is beneficial here. It could be a method in the `IsDatabaseModel` trait that classes could override.

brendt: But I'm undecided on it. @aidan-casey @innocenzi, any feedback?

aidan-casey: I'm kind of indifferent on it. Attributes feel a little more Tempest-y, but that's pretty much all I've got. 😛

brendt: True. Let's see what Enzo has to say. I'm fine with attributes if that what we all feel best about

innocenzi: I agree with Aidan here. I like attributes better than overriding methods

brendt: Ok, attributes it is

hungthai1401: @brendt @aidan-casey @innocenzi thank for your feedback. I'll keep it

src/Tempest/Database/src/IsDatabaseModel.php
hungthai1401: Yeah, thanks for your review. I'll update asap

brendt:
FYI I'm not sure about the direction the ORM will be going in the future. So I'm hesitant going further down the current route if we decide to refactor it somewhere in the coming months 🤔

brendt:
Actually, gonna close and revisit if needed later.

hungthai1401:
@brendt thanks. I will see our roadmap.
Best regards.

PR #763, opened 2024-11-22
feat(core): partial discovery cache

brendt:
See #395


src/Tempest/Cache/src/DiscoveryCacheStrategy.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Cache;
+
+enum DiscoveryCacheStrategy: string
+{
+    case ALL = 'all';
```
brendt, line 9: Rename to FULL

src/Tempest/Cache/src/CacheDiscovery.php
```diff
@@ -4,34 +4,31 @@
...
-    public function createCachePayload(): string
+    public function apply(): void
     {
-        return serialize($this->cacheConfig->caches);
-    }
-
-    public function restoreCachePayload(Container $container, string $payload): void
-    {
-        $this->cacheConfig->caches = unserialize($payload);
+        foreach ($this->discoveryItems->flatten() as $className) {
```
brendt, line 30: Make `$this->discoveryItems` iterable

src/Tempest/Console/src/Middleware/OverviewMiddleware.php
```diff
@@ -68,5 +65,16 @@ private function renderOverview(bool $showHidden = false): void
                 (new RenderConsoleCommand($this->console))($consoleCommand);
             }
         }
+
+        $this->console
+            ->writeln()
+            ->when(
```
brendt, line 71: Don't think we need this, I added it for debugging

src/Tempest/Core/src/Commands/DiscoveryClearCommand.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Tempest\Console\Commands;
+namespace Tempest\Core\Commands;
```
brendt, line 5: I moved the discovery commands to Core, similar to how the `InstallCommand` is in core.

src/Tempest/Core/src/Commands/DiscoveryGenerateCommand.php
```diff
@@ -0,0 +1,114 @@
...
+        private Kernel $kernel,
+        private DiscoveryCache $discoveryCache,
+    ) {}
+
+    #[ConsoleCommand(
+        name: 'discovery:generate',
+        description: 'Compile and cache all discovery according to the configured discovery caching strategy',
+        aliases: ['dg'],
+    )]
+    public function __invoke(): void
```
brendt, line 35: Need to clean this up

src/Tempest/Core/src/DiscoveryCache.php
```diff
@@ -4,41 +4,48 @@
...
+        $this->pool = $pool ?? new PhpFilesAdapter(
             directory: path($this->cacheConfig->directory, 'discovery')->toString(),
         );
     }
 
-    public function get(string $key): mixed
+    public function restore(string $className): ?DiscoveryItems
     {
-        return $this->getParent(str_replace('\\', '_', $key));
+        // TODO: cache and check whether the strategy has been changed
```
brendt, line 34: Already done in Kernel

src/Tempest/Core/src/DiscoveryLocation.php
```diff
@@ -17,4 +17,26 @@ public function isVendor(): bool
         return str_contains($this->path, '/vendor/')
             || str_contains($this->path, '\\vendor\\');
     }
+
+    public function toClassName(string $path): string
```
brendt, line 21: Just some cleanup to get this out of the way

src/Tempest/Core/src/Kernel.php
```diff
@@ -76,29 +81,48 @@ private function createContainer(): Container
...
     {
         $this->container->singleton(Composer::class, new Composer($this->root));
 
         return $this;
     }
 
-    private function loadEnv(): self
+    public function loadEnv(): self
     {
+        // The discovery:generate command must always run without discovery cache enabled
```
brendt, line 93: So, this part sucks, but all alternatives are worse.

The `discovery:generate` command must always run with discovery cache disabled. If we don't do that, and switch our discovery caching from `all` to `partial` or the other way around, you end up with a corrupt discovery cache with no way to regenerate it anymore (`./tempest` won't boot).

That's why we also throw an exception a little further in this method, to force people to rerun `discovery:generate` whenever they change their caching strategy (except when you turn discovery cache off).

I considered other places to add this logic, but it _must_ happen before we try to access any dependency from the container, since at that point we need a valid discovery cache.

src/Tempest/Core/src/Kernel.php
```diff
@@ -76,29 +81,48 @@ private function createContainer(): Container
...
+        }
+
+        // Load from .env
         $dotenv = Dotenv::createUnsafeImmutable($this->root);
         $dotenv->safeLoad();
 
+        // Check whether the discovery caching strategy in .env has changed
+        $current = env('DISCOVERY_CACHE');
+
+        if ($current && $original = @file_get_contents(DiscoveryGenerateCommand::CURRENT_DISCOVERY_STRATEGY)) {
```
brendt, line 109: I don't like storing the currently generated discovery cache strategy in a file. But I have no idea how else I could handle it. 

If we don't have this check, Tempest will fail _somewhere_ whenever you change your environment variable (most likely when you first try to resolve a dependency from the container).

So I rather have a clear exception being thrown telling the user they should rerun `discovery:generate`.

On the other hand… maybe it isn't a big deal? I doubt anyone will often change their discovery cache config…

brendt:
TODO: 

- [ ] Testing, but it should be done outside our normal test suite (since clearing caches affects the whole testsuite). I think the best approach is to move discovery-related cache tests to the `tempest/app` testsuite
- [ ] Some cleanup (see review comments)
- [ ] Documentation

For reference, these are the benchmark comparisons:

| branch  | cache enabled | cache disabled | partial cache enabled | 
| ------------- | ------------- |  ------------- |  ------------- |
**framework (main)** | 0.12s | 0.16s | /
**framework (partial)** | 0.11s ✅ | 0.21s ❗️ | 0.20s ❗️
**project (main)** | 0.11s | 0.15s | /
**project (partial)** | 0.11s ✅ | 0.15s ✅ | 0.10s ✅
**big project (main)** | 0.42s | 0.74s | /
**big project (partial)** | 0.36s ✅ | 0.71s ✅ | 0.58s ✅

| branch  | test suite | 
| ------------- | ------------- |
| main | 00:16 |
| partial-discovery-cache | 00:15 ✅ |

PR #774, opened 2024-11-25
chore: cleanup after release

brendt:


innocenzi:
So, what's the reason for this? Is this for better DX (local development) and/or isolation in CI?

brendt:
@aidan-casey told me not to use self.version 🫣

aidan-casey:
Yeah, `self.version` shouldn't be used. I had an explanation for everything in a previous issue. Will see if I can dig it up or put down my thoughts later.

aidan-casey:
Okay, so long story short, there are a couple problems with `self.version`:

1. `self.version` resolves to the commit hash in CI which causes a slew of CI problems when testing Tempest as extensively as we are.
2. `self.version` is _exact_ there is no room for different package versions within the same major version. Given the monorepo, this primarily comes into play when a third-party package requires a specific version. For example: Tempest requires `self.version`, the project uses `1.1.1` therefore, that resolves to `1.1.1.` A third-party package is installed that requires `^1.1.2.` This will fail since `self.version` is used and exact.

I'm not opposed to better ways here, but I don't think `self.version` is it.

PR #779, opened 2024-11-26
refactor: move classes from Http to Router package

aidan-casey:


brendt:
You could also turn the argument around: maybe http-client shouldn't be a separate component then either.

No but, I get it. I don't oppose it. I'm just saying we shouldn't aim for tempest/router to be a truly usable standalone component anytime soon. It's ok to pull them apart already today and work on it later.

aidan-casey:
> You could also turn the argument around: maybe http-client shouldn't be a separate component then either.

You could also turn the argument around: maybe Tempest shouldn't have components and everything should be bundled. Yet here we are and separation of concerns has some value.

> I'm just saying we shouldn't aim for tempest/router to be a truly usable standalone component anytime soon. It's ok to pull them apart already today and work on it later.

That's pretty much all I'm going for here. Rip it into another standalone component and then we can continue to finesse.

brendt:
👍 Feel free to merge when you're ready :)

aidan-casey:
I'm gonna close this and re-create the PR.

PR #784, opened 2024-11-26
fix(http): remove empty directories when running `static:clean`

### LINKED ISSUE #777: [Bug]: `static:clean` does not remove empty directories
### Tempest Version

1.0-alpha4

### PHP Version

8.3

### Operating System

MacOS

### Description

Currently, the `static:clean` command does not remove empty directories after removing all the containing `.html` files. This looks like expected behaviour (see [here](https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Http/src/Static/StaticCleanCommand.php#L45-L51) and most importantly [this assertion](https://github.com/tempestphp/tempest-framework/blob/main/tests/Integration/Http/Static/StaticCleanCommandTest.php#L24)).

In the test setup, the `public/static/a/b` directory is left alone, and any requests to that route (`static/a/b`) don't fall through to PHP anymore because of the default `.htaccess` file Tempest ships with. Instead, people are greeted with PHP's 404 page. I expected the request to be handled by PHP again.

I can work on a fix, but first wanted to confirm whether this was expected behaviour.

TL;DR: should the `static:clean` command remove empty directories it leaves behind?

### Steps to Reproduce

```php
<?php

use Tempest\Http\Get;
use Tempest\Http\StaticPage;

class StaticController
{
    #[Get('/test/{param}')]
    #[DataProvider(ParamProvider::class)]
    public function __invoke(string $param)
    {
        return view('test.view.php');
    }
}
```

```php
<?php

use Generator;
use Tempest\Http\DataProvider;

class ParamProvider implements DataProvider
{
    public function provide(): Generator
    {
        foreach (['a', 'b', 'c'] as $param) { yield ['param' => $param]; }
    }
}
```

```shell
$ ./tempest static:generate
- /test/a > /var/www/public/test/a/index.html
- /test/b > /var/www/public/test/b/index.html
- /test/c > /var/www/public/test/c/index.html

$ tree public/test
public/test
├── a
│   └── index.html
├── b
│   └── index.html
└── c
    └── index.html

3 directories, 3 files

$ ./tempest static:clean
- /var/www/public/test/a/index.html removed
- /var/www/public/test/b/index.html removed
- /var/www/public/test/c/index.html removed

$ tree public/test # notice that the `test` folder also still exists. Should it?
public/test
├── a
├── b
└── c

3 directories, 0 files

$ ./tempest serve

# Go to http://localhost:8000/test/a in your browser, and notice that you get a 404 instead of hitting the controller.
```

svenluijten:
Fixes #777.

I opted to split the functionality out into private methods to keep it nice and readable. If that's unwanted, let me know and I'll change it back to all be contained within the `__invoke()` method.

---

- I also removed the unnecessary `$console` property from the command class. It's injected in the parent class and was unused anyway.
- I also fixed the namespace for the test.

### DIFF
--- a/src/Tempest/Http/src/Static/StaticCleanCommand.php
+++ b/src/Tempest/Http/src/Static/StaticCleanCommand.php
@@ -4,10 +4,10 @@ declare(strict_types=1);
 
 namespace Tempest\Http\Static;
 
+use FilesystemIterator;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use SplFileInfo;
-use Tempest\Console\Console;
 use Tempest\Console\ConsoleCommand;
 use Tempest\Console\HasConsole;
 use Tempest\Console\Middleware\CautionMiddleware;
@@ -19,7 +19,6 @@ final readonly class StaticCleanCommand
     use HasConsole;
 
     public function __construct(
-        private Console $console,
         private Kernel $kernel,
     ) {
     }
@@ -30,10 +29,20 @@ final readonly class StaticCleanCommand
     )]
     public function __invoke(): void
     {
+
+        $directoryIterator = new RecursiveDirectoryIterator($this->kernel->root.'/public');
+        $directoryIterator->setFlags(FilesystemIterator::SKIP_DOTS);
+
+        $this->removeFiles($directoryIterator);
+        $this->removeEmptyDirectories($directoryIterator);
+
+        $this->success('Done');
+    }
+
+    private function removeFiles(RecursiveDirectoryIterator $directoryIterator): void
+    {
         /** @var SplFileInfo[] $files */
         $files = [];
-
-        $directoryIterator = new RecursiveDirectoryIterator($this->kernel->root . '/public');
 
         /** @var SplFileInfo $file */
         foreach (new RecursiveIteratorIterator($directoryIterator) as $file) {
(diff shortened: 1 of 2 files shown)


aidan-casey:
Thanks for the PR, @svenluijten! @brendt how strongly do you feel about outputting the file names? We could also utilize the `deleteDirectory` method in our [`Filesystem` class](https://github.com/tempestphp/tempest-framework/blob/461193d9f939aad9a92137476099a795a6a709c5/src/Tempest/Filesystem/src/LocalFilesystem.php#L100).

brendt:
In this case, I think outputting the filenames is beneficial, since it's kind of a destructive command, and there might be cases where we could use that data for debugging later. 

@svenluijten could you run `composer qa` locally and push to fix the codestyle issues?

svenluijten:
@brendt I'd love to, but... 😬 

```shell
> vendor/bin/php-cs-fixer fix --allow-risky=yes
PHP needs to be a minimum version of PHP 7.4.0 and maximum version of PHP 8.3.*.
Current PHP version: 8.4.1.
To ignore this requirement please set `PHP_CS_FIXER_IGNORE_ENV`.
If you use PHP version higher than supported, you may experience code modified in a wrong way.
Please report such cases at https://github.com/PHP-CS-Fixer/PHP-CS-Fixer .
Script vendor/bin/php-cs-fixer fix --allow-risky=yes handling the csfixer event returned with error code 1
Script composer csfixer handling the qa event returned with error code 1
```

I've run it with `PHP_CS_FIXER_IGNORE_ENV=true composer qa` to circumvent that problem for now.

svenluijten:
Odd observation: I had to run `composer qa` twice to appease it because of how Rector runs _after_ PHP CS Fixer. Rector changed the code in a way that removed newlines, which PHP CS Fixer needs.

brendt:
Thanks!

PR #785, opened 2024-11-26
refactor(http): update Request object to use PHP 8.4 features

aidan-casey:


aidan-casey:
Okay, this is by no means finalized, but is intended more as a proof of concept. It seems to make sense to have requests and responses be objects and utilize the PHP 8.4 features we have now.

A couple things we would need to figure out:

1. How do request objects work? They cannot use the trait `IsRequest` anymore.
2. Should we move headers/parameters to collections?

I've talked to @brendt in the past about using builders (example [here](https://github.com/REST-PHP/Client/blob/main/src/Http/Request/RestRequestBuilder.php)) as _a solution_ for the request objects, but I think he is inclined to find a different route, so open to some suggestions or feedback here.

brendt:
What's the problem with traits?

aidan-casey:
@brendt we can use traits, I'm not entirely sure we should use traits.

At the end of the day, a request is a request is a request. That feels better off being a value object that can be leveraged in different ways from _other_ classes. As I was refactoring, it felt much more clear as well.

But that's just my opinion.

brendt:
> At the end of the day, a request is a request is a request. That feels better off being a value object that can be leveraged in different ways from other classes. As I was refactoring, it felt much more clear as well.

Well that's kind of the approach we settled on very early on and decided to make consistent throughout Tempest.

> Edit: I forgot! You also cannot define properties in a contract/interface which is the biggest benefit to this approach.

Why can't you? That's now possible thanks to property hooks

PR #786, opened 2024-11-27
fix: change order of scripts in composer.json

svenluijten:
As noted in #784, running Rector _after_ PHP CS Fixer can cause the code style check to fail. Switching these two around in the `composer qa` script ensures that Rector runs _first_ (potentially changing code), and PHP CS Fixer comes in _after_ to clean it up and format it.


brendt:
Nice!

PR #790, opened 2024-11-27
feat(framework): extend http testing methods

Xammie:
When writing documentation for https://github.com/tempestphp/tempest-app/pull/8 I noticed not all HTTP methods are supported by the `HttpRouterTester`. This adds all methods which were already described in `Tempest\Http\Method` enum.


innocenzi:
That's great, thanks! I think those should be tested in `HttpRouterTesterIntegrationTest`. Could you add these?

Xammie:
Tests have been added. In addition, I needed to create 3 more route attributes to use in the test controller.

brendt:
Nice! Thanks :)

PR #791, opened 2024-11-28
feat(core): add `TEMPEST_START` constant

innocenzi:
This pull request adds a `TEMPEST_START` constant which holds the system's [high-resolution time](https://www.php.net/manual/en/function.hrtime.php) for performance measuring purposes.

```php
echo (hrtime(true) - TEMPEST_START) * 1e-6; // 244.245834
```

### DIFF
--- a/src/Tempest/Core/src/Kernel.php
+++ b/src/Tempest/Core/src/Kernel.php
@@ -39,6 +39,8 @@ final class Kernel
         array $discoveryLocations = [],
         ?Container $container = null,
     ): self {
+        define('TEMPEST_START', value: hrtime(true));
+
         return (new self(
             root: $root,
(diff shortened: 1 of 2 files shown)


aidan-casey:
Nice! Not sure on the answer, b he  would it make sense to store a stopwatch instance?

innocenzi:
Not sure honestly. I think a scalar value makes more sense here, but I guess a stopwatch instance is just as lightweight. Would you just do `new GenericClock`? Because the container is not yet initialized at this point

brendt:
I think an int is good for now

innocenzi:
@aidan-casey I think only the `hrtime()` function is built in PHP, but the `HRTime` namespace is in an extension that's not installed by default

PR #793, opened 2024-11-30
refactor(http): refactor http route attribute

### LINKED ISSUE #789: Refactor Route attribute
People can use `Route` directly, as well as `Get`, `Post`, … 

In order to make custom route attributes, we allow people to extend from `Route`. This is very inconsistent with the rest of the framework. I propose to make `Route` an interface, provide an `IsRoute` trait, and to not have `Route` available as an attribute anymore.

blackshadev:
# What 

Fixes #789 

- Refactor Route attributes to derive from instance instead of common parent class
- Split different "types" of routes. 
  - All user defined (attribute) routes are normal Routes.
  - After discovery we have a DiscoveredRoute with the handler attached to it
  - During route construction we have a Marked route
  - After matching we has a MatchedRoute

# Outstanding questions

- See question to brendt in comments


brendt:
Ping me when ready for review :)

blackshadev:
> Ping me when ready for review :)

@brendt Ping 

I do got a question for you :

- Do we even want the distinction between different types of routes?  And if so, what are they?

I really struggled with this one, initially I introduced it because initially we mutated the user provided Route attribute and add the handler to it. before passing it along. To me, this seemed rather hacky. 

## Pro multiple routes

Clearly we only need to set the handler internally, so why should we expose this in an common interface to the user. It could be easily "misused" by fixating the handler and ignoring the attribute applied one, this also seems like a really easy and hard to spot security vector to exploit. 

By having multiple different type of routes we have a better way of communicating "this is  what a user should see" at this point in time. This is also shown in the benchmark code, previously I didn't need to have a `handler` set , but now I need to set it due to the non nullable typing. I like this, because the RouteConfig, and the RoutingTree should never have routes without a handler, it shouldn't be possible because what are you going to call when matching it?

## Con multiple routes

Naming is hard, and now we need to name 3 different things. Also , object construction doesn't come cheap. So doing this in a performance sensitive place in the framework can become quite a performance hit. And I doubt if it clearifies more or if this is just confusing people.

aidan-casey:
Sorry, @blackshadev. That was me. 😬

src/Tempest/Router/src/IsRoute.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Router;
+
+use Tempest\Http\Method;
+use Tempest\Reflection\MethodReflector;
+
+trait IsRoute
+{
+    private Method $method;
```
aidan-casey, line 12: I'll preface this by saying I don't have a solution, but I'm not a huge fan of this approach because in essence, the route attribute corresponds to the method.

blackshadev: Ah, shame. Maybe Brendt has an idea. This is how I interpreted his issue https://github.com/tempestphp/tempest-framework/issues/789.

aidan-casey:
More for discussion than anything... I'm not really a huge fan of eliminating the `#[Route]` attribute, even though @brendt suggested it. 🙂 

I think I'd propose an API similar to the following:

```php
final class Route
{
    // ...
}

final class GET implements RouteProvider
{
    private Route $route;

    public function __construct(string $uri)
    {
        $this->route = new Route(...);
    }

    public function getRoute(): Route
    {
        return $this->route;
    }
}
```

src/Tempest/Router/src/IsRoute.php
aidan-casey: No, I think you interpreted right. I'm just thinking out loud. 😛

blackshadev:
> More for discussion than anything... I'm not really a huge fan of eliminating the `Route` class, even though @brendt suggested it. 🙂
> 
> I think I'd propose an API similar to the following:
> 
> ```
> final class Route
> {
>     // ...
> }
> 
> interface RouteProvider
> {
>     private(set) string $route { get; }
> }
> 
> final class GET implements RouteProvider
> {
>     public function __construct(string $uri)
>     {
>         $this->route = new Route(...);
>     }
> }
> ```

Wait.. you can have an interface on property hooks o.0. That is amazing. Can't we basically remove the whole IsRoute trait and type the interface like: That way we do not need any default methods.

```
interface Route
{
    private(set) string $uri{ get; }
    private(set) Method $method{ get; }
    private(set) array $middleware{ get; }
}
```

brendt:
> The rebasing itself was the hardest rebase I did in years

I'm actually going to revert that commit for the biggest part 😬 I'll merge this one first though, so that you don't have to do double work. I'll review and fix things where needed myself. Sorry for the overhead

src/Tempest/Router/src/IsRoute.php
brendt: Keep in mind that this is only true for the default route attributes. The goal is that people make their own to model route groups, and I foresee the method not always being part of the attribute.

brendt:
I read your question. I don't have a good answer at the moment. However, I think we can merge this as is, and figure out if we need further improvements later on. As far as I can tell, these classes are internal, so even if we change them later on, users wouldn't be impacted.

brendt:
@blackshadev did you plan on changing the docs as well, or do you want me to take care of that?

blackshadev:
> @blackshadev did you plan on changing the docs as well, or do you want me to take care of that?

I will create an MR for the docs tonight.

brendt:
Thanks! The docs aren't updated yet to reflect the namespace changes though. We need to make a decision on that first, just so that you're not doing double work. I'll keep you posted

PR #795, opened 2024-12-02
feat(log): emit `MessageLogged` when logs are written

innocenzi:
This pull request does two things:
- It gives ownership over the "log level" to Tempest,
- It makes `GenericLogger` emit a `MessageLogged` event.

&nbsp;

## Ownership over the log level

Right now, `GenericLogger#log` accepts a mixed `$level` which can be either a string or a `Monolog\Level` enum case.

We now have a `Tempest\Log\LogLevel` enum, which has the same log levels as specified by PSR-3. This is needed for the second part of the pull request, since the `MessageLogged` event has a `$level` property, which should not be a `Monolog\Level` instance, but something Tempest provides.

Note: this is not a breaking change, as `GenericLogger#log` still accepts strings and `Monolog\Level` instances, in addition to the new `Tempest\Log\LogLevel` enum.

&nbsp;

## New `MessageLogged` event

Writing any log message through Tempest's `Logger` will emit a `MessageLogged` event, which contains the log level, the message and the context.

This is specifically useful for third-party packages and external services like Ray.


brendt:
Nice!

PR #796, opened 2024-12-02
feat(debug): emit `ItemsDebugged` on debug

innocenzi:
Similarly to https://github.com/tempestphp/tempest-framework/pull/795, this pull request adds a `ItemsDebugged` event that is emitted when items are debugged through `Tempest\Debug`.

A new `le` (log event) function has also been added. It doesn't write to `STDOUT` nor a log file.

Very happy to change the name of the event and the function! Please suggest.


brendt:
Let's merge and see how it works in practice :)

PR #797, opened 2024-12-02
style: add more php-cs-fixer rules

innocenzi:
This one's for you, @aidan-casey 😄 

This pull request adds more rules to the current CS Fixer config. Most of these rules are motivated by habits I have, working with a stricter CS Fixer config on my projects, and seeing some of the auto-fixes not being applied when working on Tempest.

This includes trailing comas everywhere, spaces after casts, ordered imports (eg. `use function` after normal `use`s), explicit variables in strings, and some more. 

Notes:
- This PR is better reviewed commit-by-commit, the first one being the rule changes, and the last one being the new rules applied to the codebase
- If this is going to be merged, it should be done after all current PRs to minimize the risk of conflict
- I'm obviously totally open to adding more rules or removing some that we don't think should be added


aidan-casey:
Very nice, @innocenzi! I like it.

I am wondering about the following additions while we are at it:

- `@PHP84Migration` - primarily deals with handling nullable type declarations. I don't think we a have a lot but seems we've had a few
- `ordered_attributes` - might as well order attributes
- `ordered_traits` - might as well order traits too
- `attribute_empty_parentheses` - removes parenthesis if no arguments are passed to the attribute
- `yoda_style` - I'd love to see this get added. I prefer non-yoda style, but it'd be nice to be standardized
- `class_keyword` - convert FQDNs to `::class`

That's probably enough additions. 😜

innocenzi:
I love all of this, the stricter the better. Regarding yoda style, I like both styles so I don't have a preference, as long as it's consistent. 

I'll add these tomorrow

aidan-casey:
Thanks for your work on this!

gturpin-dev:
Hey 👋 

Pretty cool rules and PR !

Just giving personal opinion on Yoda style :
While we can think of them as something more readable, in fact most of people don't write them in this order, so this may be weird to change habits of 90% people.
Also, coming from WordPress which have yoda style as standard, they were adopted a long time ago and are contested by more and more people in the community, for example Human Made ( which is a big actor in WP ecosystem ) has made their own coding standard and removed yoda styles ( e.g. https://engineering.hmn.md/standards/style/php/#yoda-conditions )
Furthermore I think there is no PHP Framework that use yoda conditions ( e.g. Larevel/Symfony ).

WDYT ?

innocenzi:
Let's use non-Yoda style 👍

aidan-casey:
Nice touch adding the parallel runner, by the way. 🙂

innocenzi:
Any idea why CI fails but local is fine?

aidan-casey:
@innocenzi - I'm guessing the GitHub action is using an older version of PHP so the linter is failing. We'd get more detail if the arguments had `-vvv` instead of just `-v`.

My suggestion would be to avoid using the GitHub action for cs-fixer and simply build up PHP + Dependencies like we are for PHPStan and run CS-Fixer from the vendor directory. We don't want to run `composer csfixer` because it should be a dry run only.

innocenzi:
Well, I didn't expect this to be harder to debug than the console PR. I got confused with the multiple workflow files, too.

@aidan-casey if this look good to you, we can merge!

aidan-casey:
Sorry, I really should document those. 😛 

Looks good to me. Let's get this merged!

PR #803, opened 2024-12-03
fix (view): check the existing of the `$tagName` property

yassiNebeL:


yassiNebeL:
Test added. In the meanwhile, I use Linux so, I'm unable to fix the windows issue!

brendt:
Thanks for the issue and the PR. However, I don't think this is what we want to do: comments need to be kept in the rendered view. I'll go ahead and merge your PR though, and improve on it some more.

PR #809, opened 2024-12-04
feat(console): improve rescuing enum console parameters

### LINKED ISSUE #807: [Feature Request]: add better fallback to enum command parameter
### Description

Hey 👋 

I wanted to improve the console command fallback system.
I mean, currently in case we have a command argument that's an enum, if we don't provide it, Tempest will ask us for the value as a plain string without giving more indications.

Capture d'écran 2024-12-04 162618

Then, I wonder if we could provide a better DX on it.
So my thought is that we can give the beautiful Tempest select component with all Enum cases to the user.

WDYT ?

If you are ok, I can give it a look

### Benefits

Provide a better DX for console commands

innocenzi:
This pull request improves the rescuing of backed enum console parameters for console commands. Previously, they were treated as text input. After this PR, they are treated as choices.

https://github.com/user-attachments/assets/936e75d9-bd09-43f9-b8cf-f334c88a96be

Closes #807

### DIFF
--- a/src/Tempest/Console/src/Middleware/InvalidCommandMiddleware.php
+++ b/src/Tempest/Console/src/Middleware/InvalidCommandMiddleware.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Console\Middleware;
 
+use BackedEnum;
 use Tempest\Console\Actions\ExecuteConsoleCommand;
 use Tempest\Console\Console;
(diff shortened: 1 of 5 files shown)


innocenzi:
cc @gturpin-dev, can you confirm this is what you wanted?

gturpin-dev:
@innocenzi Yes, that was that ! 🎉

PR #810, opened 2024-12-04
refactor: move http router to separate package

aidan-casey:


aidan-casey:
This is PR 1 of a few, that will refactor the router to a separate package and remove it as a dependency of `http-client`. We will need to refactor a little bit how requests work, which I'd like to do anyway now that we are on 8.4.

brendt:
Gonna share here what I wrote on Discord as well:

Now that the Router refactor has been merged, I've played around with it a bit. I really don't like it 😅

Here's my problem:

`Tempest\Http\Request` became `Tempest\Router\Request`
`Tempest\Http\Get` became `Tempest\Router\Request`
`Tempest\Http\HttpMiddleware` became `Tempest\Router\HttpMiddleware`

etc.

OTOH

`Tempest\Http\Method` and `Tempest\Http\Status` stayed the same.

The thing is, concepts like `Request`, `Get`, and `HttpMiddleware` don't belong to the router.  (ok we could argue about the middleware one, but not the others) They belong to HTTP. We even say this in natural language: it's an HTTP method, it's an HTTP request, an HTTP response.

I feel like with this refactor, we've grouped code based on technical requirements, instead of their real world meaning. And that's not a Tempest thing to do.

That being said, I understand the need for this refactor from a component point of view, but if that means bending the code to mak it fit within a frame defined by technical requirements, I don't think it's worth it.

My proposal is that the Router can still be a subcomponent, but it should only contain router specific code, and indeed have a dependency on `tempest/http`.

brendt:
Or… giving it some more thought: I'm also circling back to my very first idea: let's just merge router, http, and http client into one component. 

I think we're really overcomplicating stuff for something the framework doesn't benefit from

PR #811, opened 2024-12-05
feat: add mailer

### LINKED ISSUE #149: Mail support

aidan-casey:
Closes #149

### DIFF
--- a/src/Tempest/Mailer/src/Components/Address/ImmutableAddressCollection.php
+++ b/src/Tempest/Mailer/src/Components/Address/ImmutableAddressCollection.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer\Components\Address;
+
+class ImmutableAddressCollection
+{
+    use ReadsAddresses;
+}
--- a/src/Tempest/Mailer/src/Email.php
+++ b/src/Tempest/Mailer/src/Email.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer;
+
+use Tempest\Mailer\Components\Address\ImmutableAddressCollection;
+
+final readonly class Email
+{
+    public function __construct(
+        public ImmutableAddressCollection $recipients,
+        public ImmutableAddressCollection $cc,
+        public ImmutableAddressCollection $bcc,
+        public string $subject,
+    ) {
+    }
+}
--- a/src/Tempest/Mailer/src/EmailBuilder.php
+++ b/src/Tempest/Mailer/src/EmailBuilder.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer;
+
+use Tempest\Mailer\Components\Address\Address;
+use Tempest\Mailer\Components\Address\AddressCollection;
+
+final class EmailBuilder
+{
+    private(set) public AddressCollection $to;
+
+    private(set) public AddressCollection $cc;
+
+    private(set) public AddressCollection $bcc;
+
+    private(set) public string $subject;
+
+    public function __construct()
+    {
+        $this->to = new AddressCollection();
+        $this->cc = new AddressCollection();
+        $this->bcc = new AddressCollection();
+    }
+
+    public function to(Address|string $to): self
+    {
+        $this->to->add($to);
+
+        return $this;
+    }
+
+    public function cc(Address|string $cc): self
+    {
+        $this->cc->add($cc);
+
+        return $this;
+    }
+
+    public function bcc(Address|string $bcc): self
+    {
+        $this->bcc->add($bcc);
+
+        return $this;
+    }
+
+    public function subject(string $subject): self
+    {
+        $this->subject = $subject;
+
+        return $this;
+    }
+
+    public function make(): Email
+    {
+        return new Email(
+            recipients: $this->to->toImmutable(),
+            cc: $this->cc->toImmutable(),
+            bcc: $this->bcc->toImmutable(),
+            subject: $this->subject,
+        );
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -69,6 +69,7 @@
         "tempest/http": "self.version",
         "tempest/http-client": "self.version",
         "tempest/log": "self.version",
+        "tempest/mailer": "self.version",
         "tempest/mapper": "self.version",
         "tempest/reflection": "self.version",
         "tempest/router": "self.version",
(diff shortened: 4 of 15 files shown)


aidan-casey:
@brendt / @innocenzi before I go too far down this path, can I get some eyes on this API?

innocenzi:
Regarding the implementation details, my first thoughts are that I like the collection classes and traits, they feel clean. I don't like the constructors in the traits, though.

Regarding the API, I have mixed feelings. I am not a fan of having a `MailBuilder` in addition to the `Email` value object... maybe it's just a naming thing, having the pattern name in the class name?

I'm also thinking the `EmailBuilder` should accept constructor parameters:

```php
$mail = new EmailBuilder(
    to: 'person@example.com',
    subject: '...',
    html: '...',
);
```

Under the hood, we can instantiate the properties accordingly:

```php
public function __construct(
    null|string|array $to,
    null|Stringable|string $subject,
) {
    $this->to = new AddressCollection(
        array_filter(is_array($to) ? $to : [$to]), // or let `AddressCollection`'s constructor handle that
    );

    $this->subject = (string) $subject;
}
```

As for naming, I'm not sure honestly. If we keep the builder pattern, maybe rename `EmailBuilder` to `Envelope`?

```php
$envelope = new Envelope(  // previously `EmailBuilder`
    to: 'person@example.com',
    subject: '...',
    html: '...',
);

$email = $envelope->wrap(); // previously `make`
```

I'm wondering if we really need to have `Email` be an immutable object? While it's valuable to have that protection, it also slightly worsen the DX at the end.

aidan-casey:
Great feedback!

> I don't like the constructors in the traits, though.

That's fair. It's something I've been doing recently to keep the trait requirements internal. That let's you have a read-only collection with only the read trait and you could even, in theory, have a write-only collection.

> I'm also thinking the EmailBuilder should accept constructor parameters

That's easy enough to do.

> If we keep the builder pattern, maybe rename EmailBuilder to Envelope?

I think those represent different things. We could probably rename the builder or eliminate it if we think we should.

> I'm wondering if we really need to have Email be an immutable object?

Yeah, I get that. From my perspective, the moment email hits a transport (e.g., `SmtpTransport` or `OAuthTransport`) it shouldn't be modifiable, so there should be some immutability there. That's just my own preference though.

src/Tempest/Mailer/src/Components/Address/ImmutableAddressCollection.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer\Components\Address;
+
+class ImmutableAddressCollection
```
brendt, line 7: Why do we need a mutable and immutable variant? Only immutable is fine I think

src/Tempest/Mailer/src/Email.php
```diff
@@ -0,0 +1,18 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer;
+
+use Tempest\Mailer\Components\Address\ImmutableAddressCollection;
+
+final readonly class Email
+{
+    public function __construct(
```
brendt, line 11: Shouldn't an email have a body?

brendt: As well as optional attachements?

src/Tempest/Mailer/src/EmailBuilder.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mailer;
+
+use Tempest\Mailer\Components\Address\Address;
+use Tempest\Mailer\Components\Address\AddressCollection;
+
+final class EmailBuilder
```
brendt, line 10: IDK… we talked about this for requests as well. I understand that you like the pattern, but I don't think it's all that useful for most cases. 

- We could make `Email` itself immutable, add these methods on it, and always return a clone of the object. 
- Or we could not have these methods at all, why would you want to change something after the first time constructing an email/builder anyway?

My preference is to add immutable builder methods on `Email` directly.

src/Tempest/Mailer/src/Email.php
brendt: Ah it's probably just WIP :)

brendt (changes requested):
What's missing here to get a proper idea of where this component is heading is _how_ we'll send mails. 

- Transports should be configurable
- I'd like some kind of `Mailer` interface with a simple `send(Email $email)` method
- People should be able to make their own `Email` implementations, so `Email` should be an interface with trait.
- Emails should be able to be sent asynchronously, we should use async commands for that behind the scenes. Perhaps allow custom `Email` classes with the `#[Async]` attribute?

aidan-casey:
> What's missing here to get a proper idea of where this component is heading is how we'll send mails.

@brendt I have ideas here, but I wanted to make sure we had the email object/builder down first before starting to work on them (since it will necessarily impact those parts).

This is basically the answer to all your review comments as well. These commits are just a POC to ensure we agree on the email format and then I'll flesh it out.

> People should be able to make their own Email implementations, so Email should be an interface with trait.

I kind of heavily disagree with this. An email is made up of a certain set of fields defined by an RFC.What makes up an email doesn't change.

src/Tempest/Mailer/src/Email.php
aidan-casey: Just POC to ensure we agree on the format and then I'll fill it in.

src/Tempest/Mailer/src/Components/Address/ImmutableAddressCollection.php
aidan-casey: Those addresses should be modifiable during the build phase but not after making the email object. We can't have just an immutable because then it wouldn't be modifiable at all.

src/Tempest/Mailer/src/EmailBuilder.php
aidan-casey: > We could make Email itself immutable, add these methods on it, and always return a clone of the object.

I'm going to shock you real quick... the way PSR handles immutable objects was informed by limitations in PHP at the time and there was much debate in the drafting of it. IMHO, it's just wrong because it's not truly immutable.

If a transport gets passed an email, it could modify the contents of the email. That shouldn't be possible.

brendt:
> I kind of heavily disagree with this. An email is made up of a certain set of fields defined by an RFC.What makes up an email doesn't change.

Well the `Email` interface enforces the right structure (I'd prefer public properties over methods):

```php
class UserSubscribedEmail implements Email
{
    use IsEmail;
    
    public function __construct(User $user, Subscribable $subscribable) {
        $this->to = $user->email;
        $this->bcc[] = $subscribable->email;
        $this->subject = "You have been subscribed to " . $subscribable->email;
        
        // …
    }
}
```

Of course we'd also have a `GenericEmail` or something alike as a default implementation.

aidan-casey:
@brendt so what benefits are we gaining with this?

brendt:
Gonna close this one, I don't think it's the direction I want to go in

PR #814, opened 2024-12-05
feat(console): provide command suggestions when using `:` shorthands

innocenzi:
This pull request adds support for using the `:` shorthand when running console commands.

For instance, typing `tempest ca:cl` will now suggest `cache:clear`. This is a functionality that the Symfony console component has (so does Composer and Laravel).

I use this shorthand all the time, and I thought it was missing in Tempest, especially seeing all these commands with hardcoded shorthands (eg. `cache:clear` -> `cc`).

## Some questions

- With this implementation, Tempest will still ask for confirmation before running a matched command. Should we run the command if there is a sole result?
- Should we remove the hardcoded aliases if so?

## How it looks

https://github.com/user-attachments/assets/158d28c8-7617-43e7-9c35-b90d5253f130

In comparison, here's how it look in Laravel:

https://github.com/user-attachments/assets/1c6f31ed-9730-4389-8991-9f43e10bb522

### DIFF
--- a/src/Tempest/Cache/src/CacheClearCommand.php
+++ b/src/Tempest/Cache/src/CacheClearCommand.php
@@ -18,7 +18,7 @@ final readonly class CacheClearCommand
     ) {
     }
 
-    #[ConsoleCommand(name: 'cache:clear', description: 'Clears all or specified caches', aliases: ['cc'])]
+    #[ConsoleCommand(name: 'cache:clear', description: 'Clears all or specified caches')]
     public function __invoke(bool $all = false): void
     {
         $caches = $this->cacheConfig->caches;
--- a/src/Tempest/Cache/src/CacheStatusCommand.php
+++ b/src/Tempest/Cache/src/CacheStatusCommand.php
@@ -20,7 +20,7 @@ final readonly class CacheStatusCommand
     ) {
     }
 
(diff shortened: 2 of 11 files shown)


aidan-casey:
> Tempest will still ask for confirmation before running a matched command

I like this approach better, actually. Take for example, a command like `destroy:my-app`. I don't want that run automatically on accident. 😛

innocenzi:
Right, that's what I thought too. 

What about the aliases? `ca:cl` is longer than `cc`, but `cc` could be ambiguous in the future. I think we should get rid of them but that'll probably mess up @brendt's muscle memory. 😁

aidan-casey:
I'm kinda indifferent on the alias front. Symfony has them and Laravel has them. I can see their use and I can see the argument against them.

innocenzi:
To be clear, I specifically mean the really short aliases we currently have. I'm totally in favor of other aliases.

eg.
`cache:clear` → ❌ `cc`, ✅ `clear:cache`
`make:controller` → ✅ `controller:make`

aidan-casey:
Ooooh gotcha. Get rid of them if not for any other reason than to mess with @brendt and confuse him on the next live stream. 😉

brendt:
The problem with manual defined aliases is that they don't scale. Sooner or later we're running into name collisions, so I'm not against removing them. I'll adjust :)

This is a nice feature BTW :)

PR #815, opened 2024-12-06
perf(discovery): cleanup discovery manager and add potential caching capabilities

aidan-casey:
…ing capability

### DIFF
--- a/src/Tempest/Console/src/Discovery/ConsoleCommandDiscoveryAgent.php
+++ b/src/Tempest/Console/src/Discovery/ConsoleCommandDiscoveryAgent.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Tempest\Console\Discovery;
+
+use ReflectionClass;
+use Tempest\Console\ConsoleCommand;
+use Tempest\Console\ConsoleConfig;
+use Tempest\Container\Discovery\Agent\DiscoveryAgent;
+use Tempest\Reflection\ClassReflector;
+
+final class ConsoleCommandDiscoveryAgent implements DiscoveryAgent
+{
+    public function __construct(
+        private readonly ConsoleConfig $consoleConfig,
+    ) {
+    }
+
+    public function inspect(ReflectionClass $class): void
+    {
+        $class = new ClassReflector($class);
+
+        foreach ($class->getPublicMethods() as $method) {
+            $consoleCommand = $method->getAttribute(ConsoleCommand::class);
+
+            if (! $consoleCommand) {
+                continue;
+            }
+
+            $this->consoleConfig->addCommand($method, $consoleCommand);
+        }
+    }
+}
--- a/src/Tempest/Container/src/Discovery/Agent/DiscoveryAgent.php
+++ b/src/Tempest/Container/src/Discovery/Agent/DiscoveryAgent.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Tempest\Container\Discovery\Agent;
+
+use ReflectionClass;
+
+interface DiscoveryAgent
+{
+    public function inspect(ReflectionClass $class): void;
+}
--- a/src/Tempest/Container/src/Discovery/Agent/SampleAgent.php
+++ b/src/Tempest/Container/src/Discovery/Agent/SampleAgent.php
@@ -13,7 +13,7 @@ final class SampleAgent implements DiscoveryAgent
     {
         foreach (reflect($class)->getPublicMethods() as $method) {
             if ($method->hasAttribute(ConsoleCommand::class)) {
-                var_dump($class->getName() . '::' . $method->getName());
+//                var_dump($class->getName() . '::' . $method->getName());
             }
         }
     }
--- a/src/Tempest/Container/src/Discovery/Agent/SampleAgent.php
+++ b/src/Tempest/Container/src/Discovery/Agent/SampleAgent.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Tempest\Container\Discovery\Agent;
+
+use ReflectionClass;
+use Tempest\Console\ConsoleCommand;
+use Tempest\Container\Container;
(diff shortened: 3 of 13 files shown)


aidan-casey:
Okay, @brendt this is a mess, so pardon all of that for a second... I think I may be on to something here.

I've been finding one of the huge bottlenecks appears to be the reading from the filesystem we are currently doing. Since composer is already putting the work into mapping out namespaces, I had a couple of ideas to avoid that bottleneck by using the classmap composer generates (side benefit, we get support for more than just PSR-4).

The other thing I realized while I was refactoring this was that we were redoing a lot of the class reflection work for every discovery class. I tidied this up a little bit by first resolving the "discovery agents", then resolving the reflection classes, and finally running those same reflection class instances through the agents.

Now, keep in mind, I haven't moved over any discoverers to "agents" as this PR calls them, but the early results of this PR are seeming to be very promising. I performed the following tests:

**Current Discovery**

1. Added an early return to `DiscoveryDiscovery` so that no additional discoveries are added.
2. Added microtime benchmarks to the start and end of the `build` method on `LoadDiscoveryClasses`.
3. Ran `php tempest`.
4. Tracked times pretty consistently around 0.283984899520874. (This is on Apple M1 Max, 64 GB RAM)

**This implementation**
1. Created a test file with the following code:

```php
<?php

require 'vendor/autoload.php';

$start = microtime(true);

new DiscoveryManager()
    ->addClassLoader(
        new CacheableClassLoader(
            new VendorClassLoader(__DIR__)
        )
    )
    ->addClassLoader(new AppClassLoader(__DIR__))
    ->addAgent(new SampleAgent())
    ->run();

$end = microtime(true);

var_dump($end - $start);
```

2. Ran test file by calling `php test.php`.
3. Tracked times at `0.038229942321777344`. (This is on Apple M1 Max, 64 GB RAM)

As you can see from this, times are automatically faster, but there is also some significant room for improvement on the caching front of this PR (which we are not doing at all).

---

I have a few other thoughts too, that I'm playing around with such as:

1. What if instead of leaving it up to the `inspect` method on the agent on what to do with the class, we had predefined "rulesets" that could be returned. Then whether the ruleset matches a class (e.g., `HasAttributeRule`) could be cached as well.
2. If we embraced an async library in Tempest and had a proper event loop, could we keep discovery constantly happening in the background during development?

brendt:
Ok this is very interesting! But how can you run benchmarks when you've only changed a handful of discovery classes? Shouldn't we change all for a more accurate picture? Maybe I misunderstood though, I only scrolled through it quickly this morning.

Some general thoughts:

- This is an area of Tempest where I think performance should be an absolute priority. I'm fine with doing weird internal stuff if it improves performance.
- Ideally, I want it to be still relatively easy for people to make their own discovery classes; I don't think I fully understand yet how it works, but I'd like it when people don't have to implement multiple classes to get their own discovery working.

aidan-casey:
@brendt see the step:

> Added an early return to DiscoveryDiscovery so that no additional discoveries are added.

brendt:
> If we embraced an async library in Tempest and had a proper event loop, could we keep discovery constantly happening in the background during development?

Yes, but I'd say out of scope :)

brendt:
Do you want me to give it a proper read before you continue work on it?

aidan-casey:
@brendt at least tell me you don't hate the approach before I keep going down the path. 🙃

brendt:
If I understand it correctly, you benchmarked by disabling all other discovery classes in both implementations?

I still think we need to see performance impact on the whole before making any decision, which unfortunately means refactoring everything. I know the pain, I went through it three times when refactoring discovery last time 😅

I also don't understand what's changed now? I still see a nested loop. Is this about not creating new reflection instances for every discovery but instead reusing them? 

Also:

> Ideally, I want it to be still relatively easy for people to make their own discovery classes; I don't think I fully understand yet how it works, but I'd like it when people don't have to implement multiple classes to get their own discovery working.

Is there a technical reason all of this can't be in the same class?

brendt:
One more thought: maybe we should first PR the composer classloader part, and see how that impact performance, then move on to the next part

aidan-casey:
> If I understand it correctly, you benchmarked by disabling all other discovery classes in both implementations?

Basically I disabled auto discovery and then manually registered a console discovery in each implementation.

> Is this about not creating new reflection instances for every discovery but instead reusing them?

Yes, but also creating better cache points for each step as well.

> Is there a technical reason all of this can't be in the same class?

I'm not sure I understand. All someone has to do is implement `DiscoveryAgent`.

brendt:
> I'm not sure I understand. All someone has to do is implement DiscoveryAgent.

Right, ok. So this completely replaces the `Discovery` interface then? Why the namechange?

> Basically I disabled auto discovery and then manually registered a console discovery in each implementation.

I meant I wanna see benchmarks with all discoveries, because it'll better show the real impact of this change.

blackshadev:
@aidan-casey  you could write a small benchmark using phpbench. You can look in `tests/Benchmark/` for some example files and a readme how to run it. The benefit is you get more "stable" numbers. comparing a single run, or even the number `0.038229942321777344` is kind of hard, because you will never see small deviations caused by your changes, or by background tasks executed by your OS.  PHPBench helps you a bit, you can easily isolate the thing you want to benchmark and let it do so as much as you like using iterations, this will make the end result a bit more stable and thus easier to compare between runs. 

For the Http rework I wrote a benchmark against the old code and new code to get a good view of the improvement (although this can be quite some work). But what I found a bigger benefit, while working I could easily identify which things hit the performance hard, and which things didn't. I saw for instance that in one instance array_pop + array_reverse was faster and array_unshift was faster.

Wulfheart:
@aidan-casey Why did you close this?

aidan-casey:
@Wulfheart I think if you are going to refactor as mentioned in #854, we'll have to rework this a bit, which is fine. 🙂

PR #821, opened 2024-12-06
chore: upgrade phpstan to v2

blackshadev:
# What

- Upgrade phpstan to v2
- Upgrade required dependencies to other versions
  - rector v2.0-rc2
  - spaze/phpstan-disallowed-calls v4
- Remove phpat, it doesn't yet support phpstan, and I cannot see any progress on it. I have found [this issue](https://github.com/carlosas/phpat/issues/293) , but no branches revering it etc. 

## Why not wait?

Well , as it stands, phpstan doesn't allot for property hooks, I tried a property hook in an interface and phpstan paniced to the point it became unusable. In my opinion, I rather have php 8.4 features, than the features added by phpat, which can be re-added when phpat does provide phpstan 2.0 support.

## Why the baseline

I rather introduce phpstan 2.0 early and fix the issues from the baseline later, the alternative is my trying to clean up while others keep on piling new issues without them even knowing they did it (because phpstan 2 has become more strict)

PR #822, opened 2024-12-06
refactor(router): convert route attributes to plain objects

blackshadev:
# What

@aidan-casey made an alternative suggestion to removing the `Route` as abstract base. With this rework we also lose the `IsRoute` trait and all `Route` builder "verbs" (`Get`, `Post`, etc) become plain old php objects. No method, no logic, no magic. This will make it easier to use the `Route` interface. But it will make it "harder" to do logic in these Route builder "verbs" as you do not have any method to hook into, you should do everything in the constructor. 

Personally, I think this is as clean as it gets.

# DO NOT MERGE

Based on  #821 as PHPStan 2.0 is required, the old one will panic at the sight of property hooks in the interface


.php-cs-fixer.dist.php
```diff
@@ -9,6 +9,7 @@
     ])
     ->name('*.php')
     ->notName('*.cache.php')
+    ->notPath('Tempest/Router/src/Route.php') // phpcs doesn't yet support property hooks in interfaces
```
aidan-casey, line 12: Simply commenting to say let's keep an eye on [this PR](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/8312).

aidan-casey:
Cool PR!

src/Tempest/Router/src/Route.php
```diff
@@ -8,10 +8,9 @@
...
 {
-    public function method(): Method;
+    public Method $method { get; }
 
-    public function uri(): string;
+    public string $uri { get; }
 
-    /** @return class-string<HttpMiddleware>[] */
-    public function middleware(): array;
+    public array $middleware { get; }
```
brendt, line 15: Can't you keep the docblock?

brendt:
Nice!

src/Tempest/Router/src/Route.php
blackshadev: Yes I forgot, thanks!

blackshadev:
PHPStan seems drunk, can someone rerun it? It reports an error of unnecessary null safe, while on that line (and even in the file) no nullsafe operators are used...

aidan-casey:
@blackshadev I reran it from mobile, but same error. What file/line is it complaining about? It's not complaining locally?

blackshadev:
> @blackshadev I reran it from mobile, but same error. What file/line is it complaining about? It's not complaining locally?

It is not, but I found out why. Somebody other than me rebased this  branch from main , which contained the new error. I rebased the phpstan-v2 branch and added the error to the baseline. I will rebase this one to phpstan-v2 and it should be green.

brendt:
Very nice!

brendt:
Do we need to update the docs as well?

blackshadev:
> Do we need to update the docs as well?

A bit. I can do it tomorrow.

blackshadev:
> Do we need to update the docs as well?

updated [docs in this PR](https://github.com/tempestphp/tempest-docs/pull/48)

PR #824, opened 2024-12-07
feat(console): allow calling console commands via fqcn

innocenzi:
This pull request adds support for calling console commands with the following syntaxes:

```php
$this->console->call(MyCommand::class);
$this->console->call([MyCommand::class, 'method']);
$this->console->call(MyCommand::class, arguments: ['arg1', '--flag']);
```

Previously, only the following syntax was allowed:
```php
$this->console->call('my-command arg1 --flag');
```

An inconsistency with `ConsoleTester#call` and `GenericConsole#call` has also been fixed. The former already allowed the FQCN and array syntaxes, while the later did not. They now use the same resolving logic.

Since calling a command with parameters in a single string is still allowed to preserve existing functionalities, there are a lot more changes in this pull request that it could have if the parameters could only be added via the `$arguments` parameter.

> [!CAUTION]
> Both the current and previous implementation will not work with arguments that contain spaces, since we do not use PHP's argument parser and instead explode by space.

<details>
<summary>Examples</summary>

### PHP argument parser

CleanShot 2024-12-07 at 02 39 20@2x

### Our naive implementation

CleanShot 2024-12-07 at 02 40 27@2x

</details>

### DIFF
--- a/src/Tempest/Console/src/Actions/ExecuteConsoleCommand.php
+++ b/src/Tempest/Console/src/Actions/ExecuteConsoleCommand.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Console\Actions;
 
+use Tempest\Console\ConsoleCommand;
 use Tempest\Console\ConsoleConfig;
 use Tempest\Console\ConsoleInputBuilder;
 use Tempest\Console\ConsoleMiddlewareCallable;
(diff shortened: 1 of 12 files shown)


brendt:
Nice!

brendt:
> Both the current and previous implementation will not work with arguments that contain spaces, since we do not use PHP's argument parser and instead explode by space.

We should probably refactor that?

innocenzi:
> We should probably refactor that?

Absolutely. To be honest, I was thinking or removing the ability to pass arguments as string altogether

brendt:
I believe it's required for console command autocompletion though

innocenzi:
Oh okay! Well let's keep it this way for now

PR #825, opened 2024-12-08
chore: resolve phpstan issues

blackshadev:
# what

Fixes all phpstan errors in the baseline

There are a lot of changed files, but luckily most of them are relatively simple changes. Most errors fell into these categories:

- The biggest category: using named arguments for phpunit assert functions. Apparently phpunit has marked their functions with "no-named-arguments" disallowing you to use them as named arguments.
- nullable return types when no null is returned. Some function had a nullable return type, while never return null. All of the classes had an interface they implemented where the nullable was on the interfacing method, but apparently you can narrow your types in interfaces perfectly fine.
- Unnecessary safe path navigation (`?->`) allot of times we slap this `?->` thing in between nullable types. But you do not need it if you already got a null coercion on that same line `??`. 

I did add some `// @phpstan-ignore <identifier>` to some files. The main reason was because for these cases the typing phpstan assumed to be fact, are not. The best example is the changes in `src/Tempest/Support/src/ArrayHelper.php` , the parameter `$generator` has a phpdoc describing how it should be used. But because this function is public to our users there is no guarantee they do as we say, there is also no way to let phpstan treat phpdocs as certain while others not (at least not that I found). We can disable this entirely, but than we would lose the ability to type array-s using phpdocs. So I opted for this approach instead. 

# Outstanding question:

- HandlesDiscoveryCache is unused. Can this be removed?


brendt:
> Apparently phpunit has marked their functions with "no-named-arguments" disallowing you to use them as named arguments.

🤢 Can we override this check? We'll deal with the breaking change when it happens, but disallowing named arguments is stupid…

> HandlesDiscoveryCache is unused. Can this be removed?

Yes, that's my oversight. Let's remove.

Let me know your thoughts, happy to merge this as soon as possible 👍

blackshadev:
> 🤢 Can we override this check? We'll deal with the breaking change when it happens, but disallowing named arguments is stupid…

Technically we can, but I don't think you want to. phpstan doesn't disallow all named argument. Only the once which are marked with the `@no-named-arguments` annotation [(docs)](https://psalm.dev/docs/running_psalm/issues/NamedArgumentNotAllowed/) . PHPUnit explicitly marks it's assertion methods with this annotation. Personally I think we should listen to them ;P 

> Yes, that's my oversight. Let's remove.

:ok: will do

brendt:
> I think I can only remove any check on the @no-named-arguments annotation, not specifically the phpunit one.

Right, that's what I meant, I think it's a really stupid rule to have. I understand why it's there, but I'll happily deal with the breaking changes whenever I shoot myself in the foot because of it.

But, it's a detail :) it's not a hill worth dying on, so I'm fine either way. 

Just FYI, in case you want some more background info, I wrote about it here a while ago: https://stitcher.io/blog/named-arguments-and-variadic-functions

Ping me when I can merge :)

innocenzi:
> @blackshadev 
> PHPUnit explicitly marks it's assertion methods with this annotation. Personally I think we 
should listen to them 

They only do that because they don't want to cover backward compatibility for argument names. It's pretty unlikely these arguments change anytime soon, so if we can disable the rule for test files, I think it's pretty safe to do so

brendt:
Maybe I should make a new PHPStan rule: `@no-no-named-arguments` 😂

blackshadev:
> > @blackshadev
> > PHPUnit explicitly marks it's assertion methods with this annotation. Personally I think we
> > should listen to them
> 
> They only do that because they don't want to cover backward compatibility for argument names. It's pretty unlikely these arguments change anytime soon, so if we can disable the rule for test files, I think it's pretty safe to do so

Ok I will check and see if I can disable it specifically for phpunit. If not, I will disable the entire rule.

blackshadev:
@brendt done.

PR #827, opened 2024-12-08
chore: update issue template to use php 8.4

kevinfrom:
In the Github bug report issue template, PHP 8.3 was the only available PHP version. Since tempest requires a minimum of 8.4 now, I've updated the option.


innocenzi:
Thanks!

PR #829, opened 2024-12-09
feat: add Vite support

innocenzi:
This pull request adds support for [Vite](https://vite.dev/). 

### How it works

**In development**, Tempest will look for a "bridge file" created by `vite-plugin-tempest`. This file contains the URL of the development server, which allows Tempest to generate `script` tags that point to it.

When the development server is not running, Tempest will look for the manifest file generated by `vite build`. Thanks to the manifest, Tempest can generate `script`, `link` and preload tags.

### How to use it

As the end user, you may use Vite as follows:
- Create `vite.config.ts` as usual
- Add `vite-plugin-tempest` to the list of plugins
- Add `{!! \Tempest\vite_tags(['src/main.ts']) !!}` in the `head` tag of your template
  - Note that you may specify global entrypoints in `ViteConfig`. When specified, you may call `vite_tags()` without arguments. When there are global entrypoints _and_ arguments to `vite_tags()`, only these arguments will be used.
- You may configure a few things by creating a `vite.config.php` that exports a `ViteConfig` instance.

### Notes
- There are a few commits in this PR that I will PR to `main` directly, I'll rebase after.
- ~~Prefetching is not yet implemented. I want to implement it before merging.~~
- ~~`vite-plugin-tempest` is not written yet. We need to decide where it should go.~~
  - **I opted to include it in the repository for now, as I believe this is the best DX option.** We can still discuss this!
  - ~~It could be in this repository, eg. `src/vite-plugin-tempest`. We'd just need a simple GitHub action to automate the release when a tag is pushed, but we haven't automated the release of the PHP packages yet...~~
  - ~~In the meantime, it can go to a separate `tempestphp/vite-plugin-tempest` repository? I would like your input on that @aidan-casey @brendt~~
- Only one `vite.config.ts` file can be used, for now.
  - Laravel has support for multiple configs because it allows specifying the build directory on the `@vite` directive
  - `innocenzi/laravel-vite` supports multiple configs in `config/vite.php`
  - I was not sure how to implement this in a way that I feel is clean and maintainable, so I decided to go with supporting just one config for now.
- There are lot of potential improvements, such as a built-in CSP HTTP middleware, or discovery of entrypoints, but I want to keep things simple in this PR.

### To-dos

- [x] Prefetching
- [ ] Installer?
- [x] `vite-plugin-tempest` tests and CI
- [x] More tests
- [x] Iron out end-user testing DX
- [x] Docs

PR #830, opened 2024-12-09
feat(support): add `HtmlHelper` class

innocenzi:
This pull request introduces a `HtmlHelper` class, which purpose is to help working with HTML.

For now, its only method is `createTag`. It properly boolean handles attributes and self-closing tags.

```php
HtmlHelper::createTag('script', [
  'src' => 'https://localhost:5137/@vite/client',
]);
```

See https://github.com/tempestphp/tempest-framework/pull/829

### DIFF
--- a/src/Tempest/Support/src/HtmlHelper.php
+++ b/src/Tempest/Support/src/HtmlHelper.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Support;
+
+final class HtmlHelper
+{
+    public static function createTag(string $tag, array $attributes = [], ?string $content = null): string
+    {
+        $attributes = self::compileAttributes($attributes);
+
+        if ($content || ! self::isSelfClosing($tag)) {
+            return sprintf('<%s%s>%s</%s>', $tag, $attributes, $content ?? '', $tag);
+        }
+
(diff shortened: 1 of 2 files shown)


src/Tempest/Support/src/HtmlHelper.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Support;
+
+final class HtmlHelper
+{
+    public static function createTag(string $tag, array $attributes = [], ?string $content = null): string
```
brendt, line 9: Not really a fan of static helpers, I'd prefer to move to instantiable, immutable classes. That's why I recently refactoered `PathHelper` as well

brendt (changes requested):

src/Tempest/Support/src/HtmlHelper.php
innocenzi: While I'm not a fan of them either, sometimes they make sense I think.

In this case, none of the methods in that helper class would act on an existing string. `createTag` could be a static constructor, but then the class would basically be an empty `Stringable` instance...

Do you have suggestions?

innocenzi: Actually... https://github.com/tempestphp/tempest-framework/pull/842

PR #831, opened 2024-12-09
refactor(container): mark container exceptions with `ContainerException`

innocenzi:
This pull request introduces a marker interface, `ContainerException`. It's useful for catching only exceptions related to the container:

```php
try {
    $this->container->get(SomeDependency::class);
} catch (ContainerException) {
    // Do something
}
```

I am using this in https://github.com/tempestphp/tempest-framework/pull/829. Not merging immediately, as I know @brendt is not a fan of marker interfaces 👀


brendt:
Feel free to merge! It makes sense in case of exceptions 👍

brendt:
Oh, I already pressed merge. Well; good.

aidan-casey:
> It makes sense in case of exceptions 👍

@brendt are you says that's the... one exception? 🥁

PR #834, opened 2024-12-09
fix(support): support keys with dots in `ArrayHelper#get`

innocenzi:
This pull request fixes a bug with `ArrayHelper#get` that prevented from accessing array values which keys contained dots:

```php
$array = arr([
  'src/main.ts' => '...',
]);

$array->get('src/main.ts'); // null
```

This raises an issue with this method though: as it is, it converts any array retrieved to an `ArrayHelper` instance, which might be unexpected.

I'm wondering if we should rename it to something else, and make the behavior of `get` more predictable—or simply remove the bahavior that transforms returned arrays.

Any thoughts?

### DIFF
--- a/src/Tempest/Support/src/ArrayHelper.php
+++ b/src/Tempest/Support/src/ArrayHelper.php
@@ -675,6 +675,12 @@ final class ArrayHelper implements Iterator, ArrayAccess, Serializable, Countabl
     {
         $value = $this->array;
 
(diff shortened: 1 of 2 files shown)


innocenzi:
I have no clue how I managed to open both this one and https://github.com/tempestphp/tempest-framework/pull/832. I think I'm tired.

PR #839, opened 2024-12-10
feat(core): optionally run `composer up` after installers

### LINKED ISSUE #835: `install framework` should run `composer up` after it's done

brendt:
closes #835


brendt:
Had to move the executor interface from console to core so that I could properly mock it for testing.

PR #840, opened 2024-12-10
feat(container): add `has` and `unregister`

innocenzi:
This pull request adds the ability to unregister bindings, as well as checking for the existence of a binding.

```php
$this->container->register(SomeInterface::class, fn () => new SomeImplementation());
$this->container->has(SomeInterface::class); // true

$this->container->unregister(SomeInterface::class);
$this->container->has(SomeInterface::class); // false
```

Both `has` and `unregister` act on normal bindings and singletons. I think it would be pretty uncommon to have both a classic binding and a singleton with the same name, so this avoids having two extra methods for checking the type of the binding.

> [!NOTE]
> The `has` method only checks for the existence of a binding. `has` may return `false` even though `get` would succeed because of autowiring.

### DIFF
--- a/src/Tempest/Container/src/Container.php
+++ b/src/Tempest/Container/src/Container.php
@@ -12,6 +12,8 @@ interface Container
 {
     public function register(string $className, callable $definition): self;
 
+    public function unregister(string $className): self;
+
     public function singleton(string $className, mixed $definition, ?string $tag = null): self;
 
     public function config(object $config): self;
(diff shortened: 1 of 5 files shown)


brendt:
These look good. One thing I'm worried about is the `Container` interface growing so big. I've thought about it before whether it's worth having an interface here or should just have a class: we're never really switching the container implementation for eg. testing, and we don't support PSR. Anyway, it's a sidenote, and I'm not convinced by my own arguments. So let's merge this :)

PR #842, opened 2024-12-11
feat(support): add `HtmlString` class

innocenzi:
Second take on https://github.com/tempestphp/tempest-framework/pull/830

---

This pull request introduces a `HtmlString` class. This class can be passed to Tempest's view renderer and its content won't be escaped as for other strings.

It has the ability to be converted to `StringHelper`, and `StringHelper` can also be converted to `HtmlString`:

```php
$html = new HtmlString('<h1></h1>');
$string = $html->toStringHelper();
$html = $string->toHtmlString();
```

It also has a `createTag` static constructor, which creates an HTML tag with properly-parsed attributes:

```php
HtmlString::createTag('script', [
  'src' => 'https://localhost:5137/@vite/client',
]);
```

### DIFF
--- a/src/Tempest/Support/src/HtmlString.php
+++ b/src/Tempest/Support/src/HtmlString.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Support;
+
+use Stringable;
+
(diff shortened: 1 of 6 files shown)


brendt:
Let's merge, and see how we can improve it over time

PR #849, opened 2024-12-12
Refactor rename database migration interface

### LINKED ISSUE #788: Name collision between Migration
There's both the `Migration` interface and the `Migration` model. We should rename the `Migration` model so that there's no naming collision

kevinfrom:
Closes #788


brendt:
Hey! So I already fixed this myself earlier today when I was doing the PHP 8.4 refactor (#848 )

Sorry for the inconvenience! I still appreciate it though :)

PR #857, opened 2024-12-13
feat(console): add `task` component

innocenzi:
This pull request adds a`task` component that shows a spinner and prints the time it took to execute a handler.

**EDIT**: removed the key/value component, I'm adding it in another PR.

https://github.com/user-attachments/assets/a478467f-97ea-4cfd-b007-ebb539a831e6

> [!NOTE]
> The task component requires `pcntl`.
> I didn't implement a static component alternative, but someone can definitely PR one.

~~I don't like that the design is very much the same as Laravel. I was out of inspiration. If you have any ideas, feel free to step in. I'm not even sure it fits with the rest of the current design system.~~ (see next comment)


innocenzi:
Design update:

https://github.com/user-attachments/assets/357d1536-d41e-46f7-ac18-ba835121c933

innocenzi:
Ready for review. CI failures seem unrelated

src/Tempest/Console/src/Components/Interactive/TaskComponent.php
```diff
@@ -0,0 +1,183 @@
...
+        if ($this->handler === null) {
+            $this->state = ComponentState::SUBMITTED;
+
+            yield $this->renderTask($terminal);
+
+            return true;
+        }
+
+        $this->sockets = stream_socket_pair(domain: STREAM_PF_UNIX, type: STREAM_SOCK_STREAM, protocol: STREAM_IPPROTO_IP);
+        $this->processId = pcntl_fork();
```
brendt, line 62: Couldn't we use `symfony/process` instead? It's already a dependency because we use for async commands, and it has cross-platform support.

brendt (changes requested):

src/Tempest/Console/src/Components/Interactive/TaskComponent.php
innocenzi: The process component can't execute closures, can it?

brendt: Aah, that's true! Could fibers be an alternative? Or do you want to run tasks in parallel?

innocenzi: I tried fibers because I really didn't want to use `pcntl`, but it doesn't work because yeah, the task needs to run in parallel for the spinner to spin 😔

innocenzi:
So, the CI fails on MySQL because "the server has gone away". After investigation, what I think happens is that the connection is duplicated because of `pcntl_fork`, and it's then being mixed up with the main process.

What we would ideally need to do is close the database connection before forking, but we can't do that yet (see #872). Instead, I'm opting out of database-specific tests for this, since it's unrelated and likely won't affect actual users as it is development commands, not anything that would be used in production.

src/Tempest/Console/src/Components/InteractiveComponentRenderer.php
```diff
@@ -153,6 +154,8 @@ private function applyKey(InteractiveConsoleComponent $component, Console $conso
                 continue;
             }
 
+            Fiber::suspend();
```
brendt, line 157: What's this one about?

innocenzi: This is further improved and commented in another upcoming PR because this class is complex, but it's needed to re-render components before they return a value.

src/Tempest/Console/src/Components/InteractiveComponentRenderer.php
```diff
@@ -231,6 +234,19 @@ private function validate(mixed $value, array $validation): ?Rule
         return null;
     }
 
+    public function isComponentSupported(Console $console, InteractiveConsoleComponent $component): bool
+    {
+        if (! arr($component->extensions ?? [])->every(fn (string $ext) => extension_loaded($ext))) {
+            return false;
+        }
+
+        if (! new Terminal($console)->supportsTty()) {
```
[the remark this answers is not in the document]
innocenzi: Making `supportsTty` static is annoying because we'd need to have 2 methods instead of one, to still support the `disableTty` method.

I quite liked that `Console` was not dependent on `Terminal`, even though I sometimes wished it did. Maybe that's the best bet, but I have another alternative ; making the instantiation of `Terminal` less heavy.

Generally-speaking, I'm not a fan of having non-lightweight logic in constructors. At the same time, we really need an up-to-date terminal size when we create the instance. So I suggest just moving away `switchToInteractiveMode` from the instantiation.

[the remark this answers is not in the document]
brendt: You know, I've thought about it some more. It's not something we need to solve in this PR, because it'll lead us too far down a rabbit hole. Let's keep it as-is for now, and worry about fixing it later.

FWIW

> making the instantiation of Terminal less heavy.

I think this is the best approach, I've made an issue for it so that we don't have to worry about it in this PR: https://github.com/tempestphp/tempest-framework/issues/880

brendt:
Oh you already did the refactor, fair enough 😂

innocenzi:
Haha yes! It was easy enough to do, and maybe can be further improved later

brendt:
Ok, let's see if PHPStan is fixed after rebase, feel free to merge afterwards if I'm not around to do so. Great feature!

brendt (approved):

PR #860, opened 2024-12-15
ci: clean up workflow files

innocenzi:
This pull request cleans up workflow files:
- I made the task names consistent in casing
- I removed two seemingly unused workflows
- I fixed `coding-conventions.yml`, which was corrupted in a rebase
- I removed the updating of `CHANGELOG.md` in the GitHub release workflow, as that doesn't work with branch protections
- I took the opportunity to update `package.json` scripts


.github/workflows/coding-conventions.yml
```diff
@@ -35,8 +35,8 @@ jobs:
       - name: Run eslint
         run: bun run lint
 
-  phpstan:
-    name: "Run Static Analysis: PHPStan"
+ phpstan:
+    name: "Run static analysis: phpstan"
```
aidan-casey, line 39: It looks to me like you are adopting the brand casing for GitHub, Bun, etc. in this PR. We should probably do the same for __PHPStan__.

.github/workflows/coding-conventions.yml
```diff
@@ -35,8 +35,8 @@ jobs:
       - name: Run eslint
```
aidan-casey, line 35: It looks to me like you are adopting the brand casing for GitHub, Bun, etc. in this PR. We should probably do the same for __ESLint__.

.github/workflows/coding-conventions.yml
innocenzi: Absolutely. Good catch!

aidan-casey:
@innocenzi looks like there are still a few cases of the brand names being off (another phpstan, etc.). Once that's cleaned up, I like it and let's get this merged.

Side note, this is one of those things we should probably document in contributing.

innocenzi:
@aidan-casey Thanks! Here I was, thinking I was clever with my perfectionism, and I missed one 😔

PR #804, opened 2024-12-04
feat(console): add `make:middleware` command

gturpin-dev:
This PR check a TODO on #759
It add the `make:middleware` command which have different available types :
- http
- console
- event-bus
- command-bus

Please tell me if I forgot one ?

Also, could you review the `Stubs` that I provide, not sure if they fit properly to the common use cases.
I started from the docs examples.

PR #863, opened 2024-12-16
feat(console): add `make:config` command

gturpin-dev:
This PR add the `make:config` command
This close a TODO on #759 

I have created a config type enum to have a list of supported config type, similar to #804 
So it supports all of these config types : 
- database
- twig
- blade
- view
- event-bus
- command-bus
- log
- cache
- console

I'm not really familiar with all components so please give me feedbacks on related stub files, I have stick as much as possible to the doc 😄 

Notice that I didn't add a support for `AppConfig` it is because the doc seems to be outdated as params are no longer the same as the current codebase and I'm not sure that we can create a custom `AppConfig` or if this is only internal ( have seen a default one which take env variables )
Also checked the `FrameworkInstaller` which don't seems to create a default `AppConfig` too.

Tell me if I forgot any Config type 😄

### DIFF
--- a/src/Tempest/Console/src/Commands/MakeConfigCommand.php
+++ b/src/Tempest/Console/src/Commands/MakeConfigCommand.php
@@ -65,6 +65,7 @@ final class MakeConfigCommand
             $stubPath = dirname( __DIR__ ) . '/Stubs';
             
             return match ($configType) {
+                ConfigType::VIEW => StubFile::from( $stubPath . '/view.config.stub.php'),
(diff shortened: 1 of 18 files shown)


brendt:
This looks good! We might tweak it over time, but I'm good to merge.

To answer your question about `AppConfig`: indeed, it's never available as a `.config.php` file. That's because AppConfig is passed manually into `Kernel`, before all other config files are loaded. It contains the bare minimum to be able to boot Tempest, and can't be discovered. Maybe we should rename it to `KernelConfig` over time to make it more clear, but that's just me thinking out loud. 

Thanks for the new PR!

PR #864, opened 2024-12-17
feat(console): add `make:view` command

gturpin-dev:
This PR adds the `make:view` command
This PR check a TODO on #759 

I'm not familiar with the view system in Tempest so I need detailed review on stubs and supported generated classes.
That being said, I currently support the creation of a basic view file like this : 
`make:view home` => this will ask for `home.view.php` path and then creates an empty php file
`make:view HomeView class` => this second enum parameter will specify that we want to create the view class documented here https://tempestphp.com/docs/framework/views/#view-objects

I don't know if more view class are needed for this command, but I think this is something that can be added when needed 😄


src/Tempest/View/src/Commands/MakeViewCommand.php
```diff
@@ -0,0 +1,86 @@
...
+    use PublishesFiles;
+
+    #[ConsoleCommand(
+        name: 'make:view',
+        description: 'Creates a view file',
+        aliases: ['view:make', 'view:create', 'create:view'],
+    )]
+    public function __invoke(
+        #[ConsoleArgument(
+            help: 'The file name of the view',
```
innocenzi, line 30:
```suggestion
            description: 'The file name of the view',
```

src/Tempest/View/src/Commands/MakeViewCommand.php
```diff
@@ -0,0 +1,86 @@
...
+        )]
+        string $fileName,
+        #[ConsoleArgument(
+            name: 'type',
+            help: 'The type of the view to create',
```
innocenzi, line 35:
```suggestion
            description: 'The type of the view to create',
```

src/Tempest/View/src/Commands/MakeViewCommand.php
```diff
@@ -0,0 +1,86 @@
...
+
+            $this->success(sprintf('View successfully created at "%s".', $targetPath));
+        } catch (FileGenerationAbortedException|FileGenerationFailedException|InvalidArgumentException $e) {
+            $this->error($e->getMessage());
+        }
+    }
+
+    private function getStubFileFromViewType(ViewType $viewType): StubFile
+    {
+        try {
```
innocenzi, line 76: I don't think it's necessary to throw an exception type, and re-throw another type after. Can you just return the `match` result?

I'm also curious about that "guardrail", wouldn't phpstan always catch it if we added a new `ViewType`? We don't need to throw here, right?

innocenzi (dismissed):

src/Tempest/View/src/Commands/MakeViewCommand.php
gturpin-dev: This is because `StubFile` can also throw an `InvalidArgumentException` so we catch this to throw a more specific exception.

For the "guardrail", you're right, phpstan will catch it until we change the config ( I don't think this should happen )
But, it's more a personal preference to have a more specific error message instead of a generic phpstan error.
I can remove it on make commands where I've implemented that if you prefer ?

brendt: I think a more specific exception contained within this command makes sense 👍

brendt:
Awesome!

Issue #867, opened 2024-12-20
`nette/php-generator` do not take care of replacements before converting FQCNs


gturpin-dev:
### Tempest Version

dev-main

### PHP Version

8.4

### Operating System

Windows

### Description

Working with make commands or in general with the `nette/php-generator` wrapper from Tempest ( in `ClassManipulator` for example ).
I wanted to make replacements in the stub file to replace a string by a FQCN.
I therefore expected that the replacements will transform the string to my FQCN and then the `ClassManipulator` will convert the FQCN in method body to class short name and use statement. ( located in `SimplifiesClassNames::extractFqcnFromBody()`

It seems like the replacements are made after the `extractFqcnFromBody`.
I suggest to dig it and check if we can just reverse the order or if this require further tweaks.

I also don't know if reversing the order could break anything or not.

### Steps to Reproduce

Have a stub file like this : 
```php
final class MigrationModelStub implements DatabaseMigration
{
    public function up(): ?QueryStatement
    {
        return CreateTableStatement::forModel('DummyModel')
            ->primary()
            ->text('name')
            ->datetime('created_at')
            ->datetime('updated_at');
    }
}
```

If we add replacements via the `StubFileGenerator`  like this :
```php
$this->stubFileGenerator->generateClassFile(
    stubFile: $stubFile,
    targetPath: $targetPath,
    shouldOverride: $shouldOverride,
    replacements: $replacements = [
        "'DummyModel'" => sprintf('%s::class', '\Tempest\Database\SomethingThatImplementsDatabaseModel'),
    ]
);
```
Then the FQCN stay there without being processed to be shortened

The code is run through abstractions but the main instructions are those ones :
```php
$classManipulator = (new ClassManipulator($stubFile->filePath))
    ->setNamespace($namespace)
    ->setClassName($classname);
    
foreach ($replacements as $placeholder => $replacement) {
    if (! is_string($replacement)) {
        continue;
    }

    $classManipulator->manipulate(fn (StringHelper $code) => $code->replace($placeholder, $replacement));
}

// Run all manipulations
$classManipulator = array_reduce(
    array: $manipulations,
    callback: fn (ClassManipulator $manipulator, Closure $manipulation) => $manipulation($manipulator),
    initial: $classManipulator,
);

$classManipulator->save($targetPath);
```

gturpin-dev:
@brendt As the PR is merged I think this issue can be safely closed

Shouldn't it be done automatically when the PR is merged ?

[...]

Issue #866, opened 2024-12-18
`nette/php-generator` support of property hooks is missing in our implementation


gturpin-dev:
### Tempest Version

dev-main

### PHP Version

8.4

### Operating System

Windows

### Description

Working with make commands or in general with the `nette/php-generator` wrapper from Tempest ( in `ClassManipulator` for example ).
And notice that it does not transform properly the property hooks.

Just looking this blog post to check that they support them https://blog.nette.org/en/php-generator-brings-the-power-of-php-8-4
So I think we need to update the `SimplifiesClassNames::simplifyClassNames` or something like that to determine that a given class has property hooks and then don't try to alter them 🤷 

### Steps to Reproduce

Hard to say, but in my case having a class like this 
```php
final class MigrationStub implements DatabaseMigration
{
    public string $name {
        get => 'dummy-date_dummy-table-name';
    }
}
```

Gives this code 
```php
public string $name {
    get {
    }
}
```

The code is run through abstractions but the main instructions are those ones :
```php
$classManipulator = (new ClassManipulator($stubFile->filePath))
    ->setNamespace($namespace)
    ->setClassName($classname);
$classManipulator->save($targetPath);
```

gturpin-dev:
I have dig into it, I have no solutions for now, but here's my founds.

If I install `nette/php-generator` and `nikic/php-parser` on a fresh PHP install and try this code :

```php
$class = new ClassType('Foo');
$class->addProperty('bar')
    ->addHook('get', '\'test string\'');

$second = ClassType::from(Foo::class);
    
$third = ClassType::fromCode(file_get_contents(__DIR__ . '/app/Foo.php'));
    
$printer = new Printer();
dd(
    $printer->printClass($class),
    $printer->printClass($second),
    $printer->printClass($third),
);
```

With this `Foo` class
```php
namespace App;

class Foo {
    public $bar {
        get => 'test string';
    }
}
```

Then the generated code works for the manual way ( which we cannot use ) so it validate that the package handle property hooks.  
It doesn't works with the second way ( which is used in Tempest because we scan `PHPFile` )
It works with the third way, so from raw code taken from the class. This is pretty weird and I don't know if I'm missing something about the package or if it's an issue.

![screenshot_2025-01-03_130358](https://github.com/user-attachments/assets/9ca6f683-9bc0-43e3-87f5-11fa80ea0767)

brendt:
[...]

PR #871, opened 2024-12-23
feat(console): add `make:migration` command

gturpin-dev:
This PR adds the `make:migration` command
This checks a TODO on #759 

The command allows you to create a raw migration file in SQL with the following command :  
`tempest make:migration create-book-table raw`

You can create a classic migration file in PHP using :  
`tempest make:migration CreateBookTable`

Or you can even create a migration file for a specific existing model with :  
`tempest make:migration CreateBookTable model`
Tempest will scan the main namespace and let you chose in all classes that implements `DatabaseModel`  the Model you want to create the migration for  
The final generated file and statements are then slightly differents


gturpin-dev:
Note that this PR should not be merged without fixing #866  and #867 as those issues have been discovered while working on this PR.

brendt:
@gturpin-dev do I remember correct that this PR is blocked because of missing property hooks support in nette/generator? I haven't looked into the detail yet, but I assume that's because of `$name` property? You can actually write it without property hooks: `$name = 'dummy-name'`. Would be solve the issue?

gturpin-dev:
@brendt 
> do I remember correct that this PR is blocked because of missing property hooks support in nette/generator?

Absolutely

> I haven't looked into the detail yet, but I assume that's because of $name property? You can actually write it without an explicit property hook: public $name = 'dummy-name'. Would be solve the issue?

I don't remember which thing in particular but it's actually a property hook in the stub  
So yes, if you prefer we can move this to an old declaration for now

brendt:
Let's do the old declaration and get this merged 😁

gturpin-dev:
@brendt Some tests aren't related to my changes

But some tests seems to fail because the raw sql migration stub are discovered I guess
Do you know how to avoid it ?

innocenzi:
@gturpin-dev you might want to update discovery to ignore stub files: https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Database/src/MigrationDiscovery.php#L38

gturpin-dev:
@innocenzi I think this is good for me  
Is the CI broken on #main ?

brendt:
@innocenzi is this because of the pathhelper changes?

tests/Integration/Database/Commands/MakeMigrationCommandTest.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Tempest\Integration\Database\Commands;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use Tempest\Core\ComposerNamespace;
```
innocenzi, line 9: This has changed to `Tempest\Support\Namespace\Psr4Namespace`

innocenzi (changes requested):

innocenzi:
> @innocenzi is this because of the pathhelper changes?

Yes it is!

brendt:
Hey @gturpin-dev just checking, would you like some help to finish this PR? I'm happy to make the changes but if you still planned to do them that's fine as well :)

brendt:
Let's merge! Thanks @gturpin-dev

gturpin-dev:
Thanks for help 🙏

PR #873, opened 2024-12-27
chore: add back php architecture tests

blackshadev:
# Why
phpat supports phpstan 2 

# What

- Re-add phpat based on the deletion in https://github.com/tempestphp/tempest-framework/pull/821
- I needed to change one extra file: src/Tempest/Generation/src/Exceptions/FileGenerationFailedException.php, it was newly introduced after the removal of phpat and it did not adhere to our final rules. I checked the catch statements where the extended clauses where used and this change shouldn't affect it.


brendt:
Nice! Appreciate it :)

Issue #756, opened 2024-11-20
Model Relation attributes


brendt:
Currently we have a `HasOne` attribute that allows you to define and optionally override the inverse relation of `BelongsTo`. 

We need attributes for other types of relations as well though, to make sure that users have to ability to override Tempest's default relation inference.

- [ ] `#[BelongsTo('inverseName')]`
- [ ] `#[HasMany('inverseName')]`

Optionally, we may allow something like this:

```php
class Book
{
    #[BelongsTo('books.author_id', 'authors.id')]
    public Author $author;

    // …
}
```

This would be in line with how migrations handle relations.

These attributes shouldn't be used by default, but should be available to solve the edge cases.

Docs:

- [ ] Document these attributes
- [ ] Document relation loading via the query builder

blackshadev:
Bad bot? can this be reopened?

PR #874, opened 2024-12-29
feat(database): add explicit relation attributes

blackshadev:
# What

Fixes https://github.com/tempestphp/tempest-framework/issues/756

- Adds explicit relation attributes for HasMany and BelongsTo
- Add tests for both explicit and implicit relation definitions

# Outstanding questions

- I implemented it by "database" reference. So you supply the inverse and local database field names. Is this what we want? Or is it preferred to supply the inverse property name and infer the sql name from that?

# After merge

- If / when approved I will create an PR for the docs repo.

### DIFF
--- a/src/Tempest/Database/src/BelongsTo.php
+++ b/src/Tempest/Database/src/BelongsTo.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final readonly class BelongsTo
+{
+    public function __construct(public string $localPropertyName, public string $inversePropertyName = 'id')
+    {
+    }
+}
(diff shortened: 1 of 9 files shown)


brendt:
> I implemented it by "database" reference. So you supply the inverse and local database field names. Is this what we want? Or is it preferred to supply the inverse property name and infer the sql name from that?

This is a good question that we haven't properly thought through yet. I'm actually inclined to say "database", since Tempest's ORM is meant to be very lightweight and close to the database. 

That being said, I think it'll be likely we'll have some kind of `#[MapFrom('other_name')]` attribute in the future that will link a property to another column name. But, we'll tackle that issue separately if we decide we need it.

I really appreciate this PR, btw, I was kind of procrastinating doing it myself 😂

blackshadev:
> I really appreciate this PR, btw, I was kind of procrastinating doing it myself 😂

I don't mind cleaning up your breadcrumbs, you leave enough for me to follow. So most of the time it is a nice and gentle introduction into a new area of the framework.

Issue #872, opened 2024-12-24
[Bug]: Connections to the database cannot be closed


innocenzi:
In PHP, we cannot manually close a PDO connection. All references to the PDO instance must be garbage-collected in order to close the connection.

Internally, Tempest passes the PDO object around in multiple classes (`Database`, `TransactionManager` and `PdoInitializer`). We would need to wrap it so we can close connections manually.

blackshadev:
I am unsure if the label "bug" is applicable. But I created a PR which should help your efforts in #857 .

PR #875, opened 2024-12-29
feat(database): add closable connection wrapper for PDO connection

blackshadev:
# What

Closes https://github.com/tempestphp/tempest-framework/issues/872

- add closable connection wrapper for PDO connection

# Why

As mentioned in https://github.com/tempestphp/tempest-framework/pull/857 it is sometimes needed to manually close the PDO connection. With this PR you can manually close (and re-connect) the PDO connection.


src/Tempest/Database/src/ConnectionInitializer.php
```diff
@@ -0,0 +1,32 @@
...
+use Tempest\Container\Singleton;
+
+final class ConnectionInitializer implements Initializer
+{
+    private static Connection|null $instance = null;
+
+    #[Singleton]
+    public function initialize(Container $container): Connection
+    {
+        // Reuse same connection instance in unit tests
```
innocenzi, line 18: If this is only meant to be used in tests, can we ensure we are actually running in tests? We can check the current environment from the Kernel I believe

innocenzi (changes requested):
That looks great! Can you think of a way to test that the connection is actually closed? 🤔

src/Tempest/Database/src/ConnectionInitializer.php
blackshadev: This was a cary over from the PDOInitializer which had the same. But yeah, good suggestion. I will make it such that it only "caches" when in testing. This is more of a test speed+memory improvement than anything else.

blackshadev:
> That looks great! Can you think of a way to test that the connection is actually closed? 🤔

It depends on your meaning of "test".. I can Add a test for the close functionality: yes. It will just check if the PDO connection has been set to null. I cannot test if the PDO driver really closes the connection, PDO doesn't expose anything for this. 

Do you want me to add a unit test for the connect + close functionality?

innocenzi:
I think what would be ideal is that the `PdoConnection` class throws a `ConnectionClosed` exception when the internal `$pdo` variable is accessed if it's `null`. Right now, I think if we close the connection and try a query, we will have a `cannot call method on null` type exception.

Then, we can write a test that first ensure the connection works by issuing a query, then closing the connection, and finally expecting the exception when issuing another query

src/Tempest/Database/src/ConnectionInitializer.php
blackshadev: Implemented the check.

blackshadev:
> I think what would be ideal is that the `PdoConnection` class throws a `ConnectionClosed` exception when the internal `$pdo` variable is accessed if it's `null`. Right now, I think if we close the connection and try a query, we will have a `cannot call method on null` type exception.
> 
> Then, we can write a test that first ensure the connection works by issuing a query, then closing the connection, and finally expecting the exception when issuing another query

Implemented the exception and the test using sqllite.

src/Tempest/Database/src/ConnectionInitializer.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Tempest\Database;
+
+use Tempest\Container\Container;
+use Tempest\Container\Initializer;
+use Tempest\Container\Singleton;
+use Tempest\Core\AppConfig;
+
+final class ConnectionInitializer implements Initializer
+{
+    private static Connection|null $instance = null;
```
brendt, line 14: Doesn't `#[Singleton]` solve this?

src/Tempest/Database/src/Connection.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database;
+
+use PDOStatement;
+
+interface Connection
```
brendt, line 9: I'm a little afraid the name will cause confusion:

We'd now have `DatabaseConnection` and `Connection`. `DatabaseConnection` is implemented by `MySqlConnection`, `SQLiteConnection`, …

`DatabaseConnection` was originally called `DatabaseDriver`, but I renamed it because `DatabaseConnection` seemed a better name from the user config point of view.

Could we come up with a better name for this new `Connection` interface? Or should we revert `DatabaseConnection` back to `DatabaseDriver`? 

@innocenzi @aidan-casey would appreciate your input as well

brendt (dismissed):

src/Tempest/Database/src/Connection.php
innocenzi: I honestly can't think of a better name for `Connection` right now. `DatabaseDriver` seems perfectly fine instead of `DatabaseConnection`, but I get why you made the change. I think most people would understand the terminology and would have no problem with it in the config though.

src/Tempest/Database/src/ConnectionInitializer.php
brendt: Gotcha. It's unfortunate that we need to include test-specific code into the source though. Maybe we could register a separate initializer instead of this one while running tests? Something like `TestConnectionInitializer`?

blackshadev: That seems like an incredible good idea. Lets do that.

blackshadev: Implemented it using the `CachedConnectionInitializer` and adding the initializer in the `FrameworkIntegrationTestCase` . It was incredably straight forward to do. Kudos to you all, amazingly designed.

blackshadev:
@brendt I do not know why phpstan fails. It occurred because I rebased my branch on the upstream main branch.  The errors are not even the files I changed in this PR, or due to it? Any ideas?

src/Tempest/Database/src/ConnectionInitializer.php
brendt: Aaah nice! 😁

src/Tempest/Database/src/Connection.php
brendt: Oooh I have an even better idea: we can have an interface `DatabaseConfig`, and have `PostgresConfig`, etc implement it. Then we don't even need the environment one. 

This is definitely outside the scope of this PR. I'll create a new issue for it :)

innocenzi (approved):

brendt:
Gonna merge this, will still have to refactor with #881 as a followup

PR #876, opened 2025-01-02
fix(generation): `ClassManipulator` now make replacements before simplifies classnames

gturpin-dev:
This PR fix the #867 

It update `ClassManipulator` to make replacements before simplifies classnames which could cause issues.
The code is then printed two times, one time to get the code, then we make replacements and simplifies classnames ( from the updated code ) and finally we print the updated file

I haven't found a better way. I don't think this is a big deal as those operations are generally made during console commands


innocenzi:
I think the class name simplification should be done both before and after, because the manipulators expect the final code

gturpin-dev:
> I think the class name simplification should be done both before and after, because the manipulators expect the final code

Not sure about the point.
The simplification is done after replacements and then the manipulators gets the simplified code, or I have missed something ?
Maybe you could give a failing test according to it ?

innocenzi:
I mean that `simplifyClassNames` is called after the `$this->manipulations` callbacks are executed, which I'm not a fan of. I think the class names should be simplified _before_.

gturpin-dev:
I can hear that, I don't know if there is a perfect solution in doing manipulations before or after `simplifyClassNames`.
Currently, this PR change the order and doesn't break any tests, and solve the issue #867  
Do you have a better solution to solve this without changing the order ?

innocenzi:
Yes, I suggested doing it both before and after

gturpin-dev:
> Yes, I suggested doing it both before and after

You mean doing manipulations once, then simplify classnames then doing manipulations once again and finally simplify classnames again ?
I think this can cause issues because manipulations aren't only "replacements"

gturpin-dev:
> There's no concept of "replacements" in this code, right? It's somewhere else?

True, this is only my concrete example of what can be the problem with manipulations order because in my case the replacements are made through the manipulations 😄 

> I feel like I'm missing something or we're not discussing the same thing

So I think it's ok, I understand your thoughts.

> No, I mean simplifying class names before and after manipulations

I think this could be okay, I need to test this

gturpin-dev:
@innocenzi Could you check the last commit, it looks good to me and for tests. But I'm not confident that it was your thoughts

innocenzi:
Yes, it's what I was thinking. I don't have time to test this right now, but if tests pass, it's good enough for me 👍

gturpin-dev:
Is the test suite broken ?
It seems like the failing test aren't related to my changes

brendt:
Tests are now fine :)

brendt:
Thanks!

PR #878, opened 2025-01-09
feat(support): add enums support

gturpin-dev:
This PR follows the discussion on Discord https://discord.com/channels/1236153076688359495/1323681600768315472
It add `enums` support using various traits to enable useful features for dealing with `enums`


src/Tempest/Support/src/Enums/InvokableCases.php
```diff
@@ -0,0 +1,45 @@
...
+
+use BackedEnum;
+use InvalidArgumentException;
+
+use function Tempest\Support\arr;
+
+/**
+ * This trait provides the ability to call an enum case as a function to get its value
+ */
+trait InvokableCases
```
innocenzi, line 15: I don't see the value in this. Do you have usage examples?

src/Tempest/Support/src/Enums/Comparable.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Check if the current enum case is in the given list of enums
+     *
+     * @param \Traversable<\UnitEnum>|array<\UnitEnum> $enums The list of enums to check
+     *
+     * @return boolean True if the current enum is in the list, false otherwise
+     */
+    public function in( Traversable|array $enums ): bool {
```
innocenzi, line 43: Nitpick: bracket and space styling issue here, as well as all other methods.

src/Tempest/Support/src/Enums/Comparable.php
```diff
@@ -0,0 +1,122 @@
...
+trait Comparable
+{
+    /**
+     * Check if the current enum case is equal to the given enum
+     *
+     * @param \UnitEnum $enum The enum to compare
+     *
+     * @return boolean True if the enums are equal, false otherwise
+     */
+    public function is(\UnitEnum $enum): bool {
```
innocenzi, line 21: I'm not sure if this is really useful. It's syntactic sugar for `===` and `!==`... I don't mind it _that much_ but it opens the door to many trivial methods, then.

src/Tempest/Support/src/Enums/Comparable.php
```diff
@@ -0,0 +1,122 @@
...
+     * @param \Traversable<\UnitEnum>|array<\UnitEnum> $enums The list of enums to check
+     *
+     * @return boolean True if the current enum is in the list, false otherwise
+     */
+    public function in( Traversable|array $enums ): bool {
+        $iterator = match (true) {
+            is_array($enums) => new \ArrayIterator($enums),
+            $enums instanceof \Iterator => $enums,
+            $enums instanceof \IteratorAggregate => $enums->getIterator(),
+            default => throw new \InvalidArgumentException(sprintf('The given value must be an iterable value, "%s" given', get_debug_type($enums))),
```
innocenzi, line 48: We could accept a non-array value and wrap it as well.

src/Tempest/Support/src/Enums/Enumerates.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Tempest\Support\Enums\InvokableCases;
+use Tempest\Support\Enums\HelperMethods;
+use Tempest\Support\Enums\Comparable;
+
+/**
+ * Use this trait to supercharge your enums with all enums features at once
+ * You can still use the traits individually if you want to have more fine-grained control
+ */
+trait Enumerates
```
innocenzi, line 15: I don't feel like this trait name conveys what it actually does. I'm also not personally a fan of nested traits, but that's just me.

src/Tempest/Support/src/Enums/HelperMethods.php
```diff
@@ -0,0 +1,58 @@
...
+            ? array_column(static::cases(), 'value')
+            : array_column(static::cases(), 'name');
+    }
+
+    /**
+     * Wrap all enum cases in an array
+     *
+     * @return ArrayHelper<static>
+     */
+    public static function collect(): ArrayHelper {
```
innocenzi, line 43: Why is this one the only one that returns `ArrayHelper` instead of `array`?

innocenzi (changes requested):
I allowed myself to review before it's ready because I like the idea of this PR.

I love some enum helper methods, but I think they are too many. Specifically, I don't think `InvokableCases` is useful, and I don't feel like `Comparable::is` and `isNot` are.

brendt:
I like the idea, and many of the methods, but I think it should be one trait: `IsEnum`, or `IsImprovedEnum` or `HasEnumMethods` or something alike. I don't think splitting it into separate traits adds value.

gturpin-dev:
> but I think they are too many. Specifically, I don't think InvokableCases is useful, and I don't feel like Comparable::is and isNot are.

My intend is to create a set of enums support that's firstly useful internally, but for end users, so I think we must provide some useful methods or DX even if we don't use them internally.
But this idea can be challenged, tell me :) 


> I like the idea, and many of the methods, but I think it should be one trait: IsEnum, or IsImprovedEnum or HasEnumMethods or something alike. I don't think splitting it into separate traits adds value.

I agree, but to support as many people as possible, I have split things in multiple traits if people don't want to import everything.
And for the most case if we want to import everything we have the `Enumerates` trait which import every traits, did you missed it ?

src/Tempest/Support/src/Enums/InvokableCases.php
gturpin-dev: Tests are basic but you can see what can be done with it.
But this allows you to do that : `Enum::CASE()` which will give you the value instead of doing this `Enum::CASE->value` everytime.
This is especially useful when you deal with multiple cases value, for example when doing array or match on multiple cases values.
Consider status enum where you want to define an array of transitions between them, the `->value` visually pollute the code :
```php
private array $allowedTransitions = [
    OrderStatus::DRAFT->value => [
        OrderStatus::PENDING->value,
        OrderStatus::CANCELLED->value
    ],
    OrderStatus::PENDING->value => [
        OrderStatus::PROCESSING->value,
        OrderStatus::CANCELLED->value
    ],
    OrderStatus::PROCESSING->value => [
        OrderStatus::SHIPPED->value,
        OrderStatus::CANCELLED->value
    ],
    OrderStatus::SHIPPED->value => [
        OrderStatus::DELIVERED->value,
        OrderStatus::CANCELLED->value
    ]
];
```

```php
private array $allowedTransitions = [
    OrderStatus::DRAFT() => [
        OrderStatus::PENDING(),
        OrderStatus::CANCELLED()
    ],
    OrderStatus::PENDING() => [
        OrderStatus::PROCESSING(),
        OrderStatus::CANCELLED()
    ],
    OrderStatus::PROCESSING() => [
        OrderStatus::SHIPPED(),
        OrderStatus::CANCELLED()
    ],
    OrderStatus::SHIPPED() => [
        OrderStatus::DELIVERED(),
        OrderStatus::CANCELLED()
    ]
];
```

gturpin-dev: The idea is to focus on business case and not on technical aspect

src/Tempest/Support/src/Enums/Comparable.php
gturpin-dev: Yes, sorry you had to see that, it's actually a draft work and I haven't run the QA yet. 😄

src/Tempest/Support/src/Enums/Comparable.php
gturpin-dev: Yes, I think this is more to open range of possibilities to do basic things.

src/Tempest/Support/src/Enums/Comparable.php
gturpin-dev: I'm not a fan of this as it complicate the typing of the input, but why not, it's useful when you don't want to change the method.
I've marked this in my TODO

src/Tempest/Support/src/Enums/Enumerates.php
gturpin-dev: I don't have a better name yet, we can change that.
But for the goal of this, refer to this comment 😄 https://github.com/tempestphp/tempest-framework/pull/878#issuecomment-2582017766

src/Tempest/Support/src/Enums/HelperMethods.php
gturpin-dev: Because in my mind if we want to collect items this is to make transformation on them, so it's a shortcut to `arr(Enum::cases())`.
So based on this, if we want a basic array with enum cases, we have the default `Enum::cases()`.

Do you think I'm wrong on this ?

brendt:
> if people don't want to import everything.

I think it's an edge case we don't need to support :) It's better to be in line with `StringHelper` and `ArrayHelper`: one class/trait with lots of methods

gturpin-dev:
> I think it's an edge case we don't need to support :) It's better to be in line with StringHelper and ArrayHelper: one class/trait with lots of methods

I'm ok to be straightforward with this, so should we stay with the name `Enumerates` ? Or do you prefer things you mentionned ?
Consider that `StringHelper` and `ArrayHelper` names are currently being challenged in https://discord.com/channels/1236153076688359495/1318231202166997062

src/Tempest/Support/src/Enums/InvokableCases.php
innocenzi: Yeah, I get what it does, but I don't see the value. It's non-typed syntactic sugar to save a property call and the syntax isn't even pretty

I definitely don't think this is useful

src/Tempest/Support/src/Enums/HelperMethods.php
innocenzi: I get the thinking behind this, but I don't think we should have the method, then. Other methods return plain arrays, and this one feel out-of-place. A bit too Laravelish.

src/Tempest/Support/src/Enums/InvokableCases.php
gturpin-dev: I can hear that.
As it could be a personal preference, should we remove this ?
Any thoughts @brendt ?

src/Tempest/Support/src/Enums/HelperMethods.php
gturpin-dev: Yeah probably, should we remove this then ?

Anyu thoughts @brendt ?

brendt:
> I'm ok to be straightforward with this, so should we stay with the name Enumerates ? Or do you prefer things you mentionned ?

Good question. Our current naming strategy for traits involves prefixing with `Is` or `Has`. But I can't really think of a good name atm…

- `IsEnum`
- `IsImprovedEnum`
- `HasImprovedEnumMethods`

IDK

gturpin-dev:
Same here, so I'll start by replacing `Enumerates` with `IsEnum` for now

brendt:
Also: tests now seem to be fine after merging main :)

innocenzi:
I asked Claude because I'm not good at naming, and here's what he suggests:
- `Accessible` -> `HasCaseAccessUtils` (I'd use just `HasAccessUtils`)
- `Comparable` -> `HasComparisonUtils`
- `HelperMethods` -> `HasEnumerationUtils`
- ~~`InvokableCases`~~ (sorry 😉)
- `IsEnum` -> `HasEnumUtils`, `IsExtendedEnum`, `HasEnumExtensions`

I do prefer the word "util" over "helper", but maybe for consistency we should go with what's already used in Tempest, so "helper".

gturpin-dev:
To be sure of what to do and what to decide @brendt @innocenzi 

1. Traits combination decision:
Should we combine separate traits into one? ( both are ok for me )
   - Pro: Keeps simplicity
   - Con: Developers can't use individual enum features

2. Naming conventions:
The `IsEnum` name fits well with the rest of the codebase  
We currently have more `Helpers` than `Utils` even if I prefer `Utils` too  
But what if we rename `ArrayHelper` and `StringHelper` ( which is in discussion ) ?  

3. Apart from this, should I remove the `InvokableCases` part ?

innocenzi:
> Traits combination decision

Let's let @brendt decide. imo both are fine but I understand the will to have just one trait.

> The `IsEnum` name fits well with the rest of the codebase

It does but the name is too generic, it should be more specific

> But what if we rename `ArrayHelper` and `StringHelper` (which is in discussion)?

Let's go with "helper" for consistency, if we rename that later we'll do it in bulk

> Apart from this, should I remove the `InvokableCases` part?

I personally definitely think so!

brendt:
> Traits combination decision

Let's have one trait for now 👍

> It does but the name is too generic, it should be more specific

`IsEnumHelper`? If possible, I would really like to keep our naming convention of `Is` or `Has` for traits. I think `Helper` makes sense, since we have it in other places as well. We could debate whether `Helper` should be renamed, but I think that's outside the scope of this PR. I'd say, let's go with `IsEnumHelper`, and have another debate after this PR on what we want our helpers to be named

> Apart from this, should I remove the InvokableCases part?

Yes, let's remove

brendt:
Ping me when ready btw :)

gturpin-dev:
@brendt then it's ready for me👌
Does @innocenzi wants to review something else ?

innocenzi:
LGTM 👍

brendt:
Thanks for all the effort you put into this @gturpin-dev ! Let's merge :)

Would you be able to also send a PR to the docs about it? It could go on https://tempestphp.com/docs/framework/primitive-helpers/

gturpin-dev:
@brendt Sure, I'll check to make input there 🎉 

Thanks for pointing out the locations 👌

brendt:
I've added a docs entry for this :)

PR #887, opened 2025-01-24
feat(mapper): allow mapping value to object via cast method

Jeroen-G:
This PR allows the following code:

```php
$data = ['name' => 'Jeroen', 'updated' => new DateTime('now')];
$user = map($data)->to(User::class);
```

```php
final class User
{
    public function __construct(
        public Name $name,
        public DateTime $updated,
    ){}
}
```

```php
final class Name
{
    private function __construct(
        private string $name,
    )
    {}

    public static function cast(mixed $value): self
    {
        return new self((string) $value);
    }
}
```

So what is now possible is the mapping of any data, also nested within arrays, to be mapped to objects without having to create a CastWith caster for each of them. For me this removes the papercut that for every (small) value object I had to create a corresponding caster.


Jeroen-G:
I don't see how all those tests fail due to these changes 🤨

brendt:
So, I'm on the fence about this. I understand the reasoning, but it goes against my vision that we have one way of doing something, instead of two ways.

On top of that, this "cast" method is "just there", it feels very magic-like. In Tempest we try to have everything defined either by attributes or interfaces. 

If I may suggest an alternative? The attribute approach is more consistent within Tempest's codebase. But there's nothing stopping us from doing something like this: 

```php
#[CastWith(ObjectCaster::class)]
final class Name
{
    private function __construct(
        private string $name,
    )
    {}
}
```

`ObjectCaster` in this case would be able to cast any object. We could even make it easier by having a dedicated attribute:

```php
#[CastAsObject]
final class Name
{
    private function __construct(
        private string $name,
    )
    {}
}
```

But, writing all of that down leads me to another conclusion: the mapper should be able to map `['name' => 'Brent']` to `new Name('Brent')` without any casters defined. In other words: your example should work without any configuration needed. Like you say: it's a simple value object, our mapper should know how to handle it. Come to think of it, I actually thought it already did, so if it doesn't, I'd say that's a bug. I created an issue for it: https://github.com/tempestphp/tempest-framework/issues/906

PR #902, opened 2025-02-23
feat(database): refactor DatabaseConfig interface

### LINKED ISSUE #881: DatabaseConfig interface
Instead of a `DatabaseConfig` with one variable, we could instead have an interface `DatabaseConfig` and have a dedicated `PostgresConfig`, `MysqlConfig`, …

blackshadev:
Fixes #881

Simplifies the configuration of databases. Instead of wrapping the actual connection config in DatabaseConfig, we can just return the actual database config. This meant I also needed to find an alternative place for the migrations.


src/Tempest/Container/src/GenericContainer.php
```diff
@@ -105,6 +106,9 @@ public function singleton(string $className, mixed $definition, ?string $tag = n
 
     public function config(object $config): self
     {
+        if ($config instanceof DatabaseConfig) {
+            $this->singleton(DatabaseConfig::class, $config);
+        }
```
blackshadev, line 111: Unsure, will think of a more generalized approach

brendt: The approach I was thinking about was something like this (dummy code):

```php
    public function config(object $config): self
    {
        $this->singleton($config::class, $config);
        
        foreach (reflect($config)->getInterfaces() as $interface)
        {
            $this->singleton($interface, $config);
        }

        return $this;
    }
```

src/Tempest/Core/src/functions.php
```diff
@@ -76,6 +76,6 @@ function env(string $key, mixed $default = null): mixed
      */
     function defer(Closure $closure): void
     {
-        get(DeferredTasks::class)->add($closure);
+    get(DeferredTasks::class)->add($closure);
```
brendt, line 79: What's up with all these code style changes?

src/Tempest/Database/src/Config/DatabaseConfig.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Tempest\Database\Tables\NamingStrategy;
+
+interface DatabaseConfig
+{
+    public string $dsn { get; }
+
+    public NamingStrategy $namingStrategy { get; }
+
+    public DatabaseDialect $dialect { get; }
```
brendt, line 15: Do we still need the dialect? My understanding is that it will be a one-to-one mapping between the config instance and the dialect. Maybe we should simply have a mapping somewhere?

brendt:
> Fix migrations

What exactly is the problem with migrations?

src/Tempest/Core/src/functions.php
blackshadev: Correct, I did not touch it deliberately.

blackshadev:
> > Fix migrations
> 
> What exactly is the problem with migrations?

Migrations previously were added on the DatabaseConfig, which I removed. Now they need to live somewhere else. I do not want to mix user defined config with application state. I do think there already is a logical  place in the code somewhere, I just need to add it to it.

src/Tempest/Database/src/Config/DatabaseConfig.php
blackshadev: We do need the DatabaseDialect of some sorts somewhere, as it will change how queries are build. We can move it outside the DatabaseConfig if you want. But honestly, I don't think it will be "better". Currently if you want to change the databaseConfig (I don't know why), you can, if we hard map the DatabaseDialect to our known DatabaseConfig children one cannot create their own config without modifying the dialect mapper as well. I seems a bit weird to me to introduce this lock-in.

src/Tempest/Container/src/GenericContainer.php
blackshadev: Thanks, was thinking about something like this as well. Nice

src/Tempest/Database/src/Config/DatabaseConfig.php
brendt: > I don't know why

That was my reasoning: if we'll never change it, why keep it here?

OTOH, as you say: 

> I seems a bit weird to me to introduce this lock-in.

That's also a good point. It's not really in the way, so I'm fine keeping it as is.

brendt:
> Migrations previously were added on the DatabaseConfig, which I removed. Now they need to live somewhere else. I do not want to mix user defined config with application state. I do think there already is a logical place in the code somewhere, I just need to add it to it.

Oh, yeah, absolutely, good point. The fact that they were on the config class to begin with was just laziness from my side, and I knew I should refactor it _someday_ 😇

src/Tempest/Database/src/Connection/ConnectionInitializer.php
```diff
@@ -15,7 +17,11 @@ public function initialize(Container $container): Connection
     {
         $databaseConfig = $container->get(DatabaseConfig::class);
 
-        $connection = new PDOConnection($databaseConfig->connection());
+        if ($databaseConfig === null) {
```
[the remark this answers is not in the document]
blackshadev: Isn't it possible for a user not to define a database config? When that happens it would be nice to have a proper error response. Or is that not at all possible?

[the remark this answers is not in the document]
blackshadev: You are right, on a fresh install the databaseConfig would still default to the default SQLiteConfig. today I learned.

blackshadev:
> I love how this PR simplifies so many small things across the framework. Awesome!
> 
> As discussed on Discord: I agree that there's room for more unit tests in general, but I think it's out of scope for this PR. I don't want it to end up in limbo because of pending unit tests. What do you think?
> 
> Anything besides the unit tests question and my two small remarks that prevents this from being merged

No nothing is preventing this from going in, more (and or different) test coverage can be done afterwards. The integration tests we have do cover most of the made changes anyway.

blackshadev:
And the docs update as well: https://github.com/tempestphp/tempest-docs/pull/54

brendt:
Awesome, and thanks for the docs as well!

PR #909, opened 2025-02-24
feat(generation): support property hooks in `ClassManipulator`

gturpin-dev:
This provides tests and try fixing #866

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "guzzlehttp/guzzle": "^7.8",
         "laminas/laminas-diactoros": "^3.3",
         "monolog/monolog": "^3.7.0",
-        "nette/php-generator": "^4.1.6",
+        "nette/php-generator": "dev-master",
         "nikic/php-parser": "^5.3",
         "php": "^8.4",
         "psr-discovery/http-client-implementations": "1.x-dev",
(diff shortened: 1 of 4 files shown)


gturpin-dev:
I only printing it right now  
I wanted to use snapshots, but don't know how to make a "wanted" version of it while this is not working as expected.

brendt:
Feel free to update to `dev-master` and push to this branch. We'll just have to wait until it's tagged before merging. When we do that, we'll revert our composer requirement back to normal

brendt:
Hm, yes. But don't we want to know whether the changes on `nette/php-generator` actually work?

gturpin-dev:
Ok, I see your point.  
I actually tested this in local and make the test snapshot with the working version in `dev-master`  
That's why I fallback to the original version ( snapshot not updated to make a comparison between now and the future release of `nette/php-generator`  
That allow us to have red tests until a new release is deployed, then our CI will use it when building composer  and the tests should be green, anyway that's was my strategy.  

If you prefer, we can come back to dev-master to make the tests green ( which will 100% pass because the snapshot was made using this version )

> Let's convert to draft until it's ready?
Yes, a draft is a good choice I guess

brendt:
Let's push the working version, and then keep the PR as draft, and hope nette tags its release soon

gturpin-dev:
Is the CLI uses the composer.json of the branch or of the main ?  
Because looking at issues in tests here we have the generated code like the issue wasn't fixed although in local everything is ok

brendt:
Gonna close, but feel free to reopen

PR #919, opened 2025-02-26
feat(mapper): two-way casters

### LINKED ISSUE #843: Two-way casters
Currently, a caster only casts values one way. Especially for the ORM, it would be useful to cast both ways, since data is constantly stored and loaded.

brendt:
Closes #843


brendt:
Closed in favor of #920

PR #940, opened 2025-03-02
refactor(support)!: improve architecture of support utilities

innocenzi:
This pull request introduces several improvements to the support package, most notably in its architecture and the flexibility with which its utilities can be used.

## Namespacing

All support utilities are now located in a namespace that represents their scalar equivalent. This includes the following namespaces:
- `Arr`
- `Html`
- `Language`
- `Namespace`
- `Random`
- `Regex`
- `Str`

## Most helpers now have functions

This pull request changes how "helpers" are structured in Tempest. Instead of only having "helper classes", we now have access to single-purpose, namespaced utility functions.

This provides a major benefit: no longer needing to instantiate a class just to access a single utility method and then converting it back to the original scalar value.

```php
// Before
str($matches['name'] ?? null)->kebab()->toString();

// After
use Tempest\Support\Str;

Str\to_kebab_case($matches['name'] ?? null);
```

```php
// Before
arr($this->repository->getPendingCommands())->first();
arr($from)->unwrap()->toArray();

// After
use Tempest\Support\Arr;

Arr\first($this->repository->getPendingCommands());
Arr\unwrap($from);
```

Even though they are namespaced, most function names should not conflict with other global and namespaced functions. This is in order to reduce the cognitive overhead of finding the right function in the suggested imports.

## Arrays and string utilities can be mutable and immutable

Even though we all prefer immutability, there are occasions where it's more of a hindrance than a benefit.

For this reason, array and string helper classes now have a mutable equivalent:

| Primitive | Immutable | Previous class | New class         |
| --------- | --------- | -------------- | ----------------- |
| array     | yes       | `ArrayHelper`  | `ImmutableArray`  |
| array     | no        | N/A            | `MutableArray`    |
| string    | yes       | `StringHelper` | `ImmutableString` |
| string    | no        | N/A            | `MutableString`   |

These classes are fully-typed, use PHPDoc generics where relevant, and most importantly share the same logic as functions.

For instance, `ImmutableArray` and `MutableArray` use the `ManipulatesArray` trait under the hood, which proxies method calls to equivalent function. The trait has dedicated methods and does not use magic methods.

This paradigm does not increase the maintenance and allows for more flexibility regarding which type of helper is needed.

The `arr` and `str` helper use the immutable classes, and methods in array or string classes that return another primitive always return the corresponding immutable class. So by default, everything is immutable, unless the developer manually reaches for a mutable class.

## Other changes

- There are new regex-specific functions in `Tempest\Support\Regex`
- There are new HTML-specific functions in `Tempest\Support\Html`
- There are new random-specific functions in `Tempest\Support\Random`

## Breaking changes

- The `Tempest\Support\StringHelper` and `Tempest\Support\ArrayHelper` classes are now `Tempest\Support\Str\ImmutableString` and `Tempest\Support\Arr\ImmutableArray`, respectively. You only need to update the imports and class names, the methods are the same.
- `Tempest\Support\LanguageHelper` no longer exist. Use equivalent functions in the `Tempest\Support\Language` namespace.
- `Tempest\Support\HtmlString` moved to `Tempest\Support\Html\HtmlString`


brendt:
I like it! I'll need to go through it in depth, will try so today

brendt (approved):
I'm fine merging this! @innocenzi are you up for updating with main first? There might be some things that need fixing though 😬

Let's get it in before alpha.6 though (as in: alpha.6 has to wait until this is done), just so that we have enough time to test

brendt:
Let's go!!

@innocenzi if you have the time, maybe also update the docs: https://tempestphp.com/docs/framework/primitive-helpers/
I'll make an issue for it though, I might pick it up myself when I have the time

innocenzi:
Awesome!

I want to work on the docs website, and when I do I'll probably write and update a bunch of docs as well. But feel free to update docs in the meantime if you feel like it!

PR #945, opened 2025-03-02
feat(vite): add `<x-vite-tags />` component

### LINKED ISSUE #939: Add `<x-vite-tags />`
### Description

Would be great to have a more Tempest-like syntax to inject Vite tags:

```php
<x-vite-tags :entrypoints="['main.css', 'main.ts']" />
```

### Benefits

d

innocenzi:
## Main feature

This pull request adds a new `<x-vite-tags />` component:

```html
<x-vite-tags :entrypoints="['src/main.css', 'src/main.ts']" />
<x-vite-tags :entrypoint="src/main.css" />
```

Closes #939

## Other changes

Additionally, the pull request makes the resolution of entrypoints stricter: previously, in development, any entrypoint could be used, even if it didn't exist on the filesystem. This would result in broken `link` and `script` tags.

Now, using an entrypoint that doesn't exist on the filesystem will throw a `FileSystemEntrypointNotFoundException`.

Lastly, I updated the Vite installer to specify the CSS file as an entrypoint instead of importing it from `main.css`. This is in order to avoid FOUCs in development—something that I thought was fixed in a previous Vite version, but apparently not.

## Failing test

Unfortunately, it seems that https://github.com/tempestphp/tempest-framework/commit/0fe0df917a251294682c3a094929c6ee05f91350 did not actually fix the parsing issue. I added a failing test that demonstrates how custom components are still rendered outside of `<head />`:

https://github.com/innocenzi/tempest-framework/blob/91f2416b4feedd6984694bfebf2910921d191ca1/tests/Integration/View/ViewComponentTest.php#L507


brendt:
I fixed `<head>` parsing. Not sure how I feel about it though: https://github.com/tempestphp/tempest-framework/pull/945/commits/304b1092235df036282df711b52ce2b4388c1d9f

I think if we don't tell @aidan-casey , we should be fine 🙈

I also think the failing tests are unrelated, right @innocenzi ?

innocenzi:
So, that works (except with body attributes, which I fixed), but it does feel off... I'm not completely against keeping it (the most important is that it works), but we should probably explore other workarounds less likely to conflict with actual user code

I'll fix the remaining Windows-related issues and we can merge if that works for you

innocenzi:
I fixed the Windows test, however during my testing I found out another issue @brendt... if you add any HTML after a custom component, it isn't rendered in the final template. 

I added a failing test

brendt:
I fixed the failing test :)

PR #947, opened 2025-03-02
feat(mapper): `ObjectToArrayMapper` use `Caster::serialize` to serialize the property value

### LINKED ISSUE #942: `ObjectToArrayMapper` does not use `Caster::serialize`
### Description

I think the inverse is ok because `Caster::cast` was the original one way of mapping

But now we have serialize, I think we should use them to serialize an object to an array

For instance we'd like DateTime to be serialized properly when mapping an object to an array

### Benefits

Better two-way mapping

gturpin-dev:
Closes #942


brendt:
Awesome!

Issue #963, opened 2025-03-07
Migrations with string-backed enum columns with FQCN's values don't set up the DB column with backslashes


mattdinthehouse:
### Tempest version

1.0-alpha.5

### PHP version

8.4

### Operating system

macOS

### Description

TIL you can create a string-backed enum using FQCN's which is handy for identifying the concrete class for a polymorphic model

(I'm experimenting with Tempest rn and my actual use-case for this might be shit but the bug with enum columns still exists so stay with me)

In my case I have an `abstract class Source` that I'm going to build implementations of like `final class WordPressSource extends Source` and I can work out the concrete instance using an enum like
```php
enum Type: string
{
	case WordPress = WordPressSource::class;
}
```

I create a migration like
```php
public function up(): QueryStatement|null
{
    return CreateTableStatement::forModel(Source::class)
        ->primary()
        ->enum('type', Type::class);
}
```

And it sets up the DB table like
```sql
CREATE TABLE `sources` (
  `id` int NOT NULL AUTO_INCREMENT,
  `type` enum('AppSourcesIntegrationsWordPressSource') NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

Which you can see is missing the backslashes from the `WordPressSource::class` FQCN so when I call `$model->save()` it throws a `SQLSTATE[01000]: Warning: 1265 Data truncated for column 'type' at row 1 INSERT INTO `sources` (type) VALUES ():type); bindings: { "type": "App\\Sources\\Integrations\\WordPressSource" }`

### Steps to reproduce

as above

brendt:
Gonna move this to post-1.0, but I'm always open for a PR (with tests) that fixes it

PR #964, opened 2025-03-08
feat(database): add support for Single-Table Inheritance

mattdinthehouse:
Related to #963 where I'm trying to have an `abstract class Source` database model and I want queries to instantiate one of the concrete classes like `final class WordPressSource extends Source` based on the value in a column.

While writing this PR I realise I could've put the `getModelInstanceClass()` method into `DatabaseModel` and `IsDatabaseModel` but at the time I think it's better of being an opt-in feature.

Other frameworks like Rails just have a class property denoting the column that the class name comes from, and polymorphic relations typically have a `{relation}_type` column, but I've had scenarios before where I want to do some pre-setup or whatever so a static method gives more flexibility imo.

---

Based on my use-case above the code would look like this:

```php
<?php

namespace App\Sources;

use Tempest\Database\DatabaseModel;
use Tempest\Database\AbstractDatabaseModel;

abstract class Source implements DatabaseModel, AbstractDatabaseModel
{
	use IsDatabaseModel;

	public function __construct(
		public string $name,
		public Type $type,
	)
	{ }

	public static function getModelInstanceClass(mixed ...$params): string
	{
		return $params['sources.type'];
	}

	abstract public function fetch(): array;
}
```

```php
<?php

namespace App\Sources;

use App\Sources\Integrations\WordPressSource;

enum Type: string
{
	case WordPress = WordPressSource::class;
}
```

```php
<?php

namespace App\Sources\Integrations;

use App\Sources\Source;

final class WordPressSource extends Source
{
	public function fetch(): array
	{
		return [
			'whatever'
		];
	}
}
```

```php
$source = Source::query()->whereField('name', '...')->first();

print $source::class; // App\Sources\Integrations\WordPressSource

var_dump($source->fetch()); // [ 0 => 'whatever' ]
```


brendt:
Interesting! May I suggest not having two interfaces? Instead of `AbstractDatabaseModel`, you can add the method in `DatabaseModel`, and have a default implementation for it in the `IsDatabaseModel` trait:

```php
public static function getModelInstanceClass(mixed ...$params): string
{
    return static::class;
}
```

mattdinthehouse:
Done but need to add tests.

Originally I wanted it in a separate interface so it was opt-in to avoid burdening someone writing a custom implementation of `DatabaseModel`

(sidenote, what's the background or vision for `DatabaseModel` + `IsDatabaseModel` instead of extending an `abstract class DatabaseModel`?)

src/Tempest/Database/src/IsDatabaseModel.php
```diff
@@ -76,6 +76,11 @@ public static function query(): ModelQueryBuilder
         return new ModelQueryBuilder(self::class);
     }
 
+    public static function getModelInstanceClass(mixed ...$params): string
```
brendt, line 79: Perhaps in this case it makes more sense to pass in an array instead of spreading the arguments. Tempest uses argument spreading when the user needs control over the names of the variables, which in this situation isn't really the case.

src/Tempest/Database/src/IsDatabaseModel.php
```diff
@@ -76,6 +76,11 @@ public static function query(): ModelQueryBuilder
...
+    {
+        return self::class;
```
brendt, line 81: I'd go with `static::class`. That way it will still work if people make an abstract model that uses the trait, but make the query from the child model (when they are not overriding `getModelInstanceClass`)

src/Tempest/Database/src/IsDatabaseModel.php
brendt, line 79: Also: maybe let's rename to `getModelClass`?

brendt (changes requested):

src/Tempest/Database/src/IsDatabaseModel.php
mattdinthehouse: I agree in principle but I feel to make that work properly we need an ability to define scopes or default model conditions

For my use-case if I ran `WordPressSource::find($id)` and that ID wasn't for a WordPressSource model I would want it to fail and not find the model rather than proceed anyway with the wrong concrete class and cause unexpected behaviour

src/Tempest/Database/src/IsDatabaseModel.php
mattdinthehouse: I wanted to distinguish the "model definition" from the "model implementation" and imo `getModelClass` is ambigious, what do you think?

mattdinthehouse: Done in latest commit

src/Tempest/Database/src/IsDatabaseModel.php
brendt: Aaah yeah, that makes sense! OTOH, in that case you would have overwritten `getModelClass` anyway, wouldn't you? So there would be no reference of `static` or `self` anymore, but rather a reference to the type.

src/Tempest/Database/src/IsDatabaseModel.php
brendt: I'm not sure I understand. There's currently no existing reference to `modelClass` in the query to model mapper or the database model trait, so I'm not sure what ambiguity you're referring to?

mattdinthehouse: `IsDatabaseModel::query()` passes `self::class` to `ModelQueryBuilder` which builds a query based on that "model definition" and stores it in `$this->modelClass`

Later on it passes that to the mapper which `QueryToModelMapper` uses as the class definition to work out which properties to set, but `$class` is not necessarily the class we need it to instantiate and return for STI (the "model implementation") so if it were called `getModelClass` I can foresee confusion with why `ModelQueryBuilder` holds a `$this->modelClass` and would effectively end up with a call to `$this->modelClass::getModelClass()` as `$to::getModelClass()` (where `$to` already refers a model class) in the mapper to get a different kind of model class?

I hope that makes sense... 😅 there's just too many uses of "class" or "model" or "model class" in the query pipeline that all make sense when it's 1:1 but that won't be the case with STI potentially returning an instance of a different class to the one the query was built on

src/Tempest/Database/src/IsDatabaseModel.php
mattdinthehouse: Oh I think I get what you're saying - there are 3 scenarios:
1. you have a `final class MyModel` where `static::class` and `self::class` are the same so no biggie
2. my scenario where you customise the class using `getModelInstanceClass()` based on some data condition
3. the calling code says what model class it wants so `static::class` is different to `self::class`

So for scenario 3 you'd have a "base" `MyModel` that may be extended into a "child" `MyModelButDifferent` in certain scenarios and the calling code specifies which one it wants with `MyModel::find()` or `MyModelButDifferent::find()` and gets the one it calls on?

brendt: Correct! If we don't use `static`, scenario 3 wouldn't work

src/Tempest/Database/src/IsDatabaseModel.php
brendt: Very good point, thanks for helping me understand :) I'd rather rename the internal variable then and have `getModelClass` as the public method, as framework users are more likely to run into that one. 

I have no idea what to call the internal parameter though 😅 I've use `modelClassName` before I think. 

What do you think?

mattdinthehouse: lol i'd call it `modelDefinitionClass` to give it purpose? but that's just swapping around the issue with "model instance class" so I guess it's up to you which one is going to pop up more?

brendt: The public method is definitely gonna pop up the most, so let's make that one the easiest :)

brendt:
Gonna convert to draft until we've got some tests. I'm happy to add them if you're low on time though! I think three things still need doing

- [ ] Tests
- [ ] Rename the method
- [ ] Double-check the default implementation with `static`

PR #965, opened 2025-03-08
feat(database): add a `findBy` method to models

mattdinthehouse:
This is inspired by Rails a bit where I can fetch models with simple equality conditions like
```ruby
Book.where(out_of_print: true)
```

and using PHP named arguments and this PR I can now do the same like this:
```php
Book::findBy(out_of_print: true)->all();
```

as opposed to
```php
Book::query()->where('out_of_print = 1')->all();
```

In my use case I'll probably do a name or slug lookup for a single record with a unique constraint like:
```php
Source::findBy(name: '...')->first();
```

The `findBy` method returns a `ModelQueryBuilder` so loading relations and doing anything more complicated than simple equality conditions still works:
```php
Book::findBy(out_of_print: true)->where('year > :year', year: 2000)->with('author')->all();
```

### DIFF
--- a/src/Tempest/Database/src/DatabaseModel.php
+++ b/src/Tempest/Database/src/DatabaseModel.php
@@ -28,6 +28,8 @@ interface DatabaseModel
 
     public static function find(Id $id, array $relations = []): ?self;
 
+    public static function findBy(mixed ...$conditions): ModelQueryBuilder;
+
     public function save(): self;
 
     public function getId(): Id;
--- a/src/Tempest/Database/src/IsDatabaseModel.php
+++ b/src/Tempest/Database/src/IsDatabaseModel.php
@@ -91,6 +91,15 @@ trait IsDatabaseModel
             ->find($id);
     }
 
+    public static function findBy(mixed ...$conditions): ModelQueryBuilder
+    {
+        $query = self::query();
+        
+        array_walk($conditions, fn($value, $column) => $query->whereField($column, $value));
+
+        return $query;
+    }
+
     public function load(string ...$relations): self
     {
         $new = self::find($this->getId(), $relations);


brendt:
Good addition! However, there's a naming problem: there's already a `find` method that doesn't return the query builder but rather a model instance. If `findBy` works differently by returning the query, it might lead to confusion. How about something like `Model::whereFields(name: '…', slug: '…')->first()`?

Are you able to add some tests? I can add them but I wanted to check with you first if you had plans to do so

mattdinthehouse:
I can definitely add tests but can't commit to time, so feel free to jump in if I'm taking too long 😅

Do you feel that changing it to `Model::whereFields()` causes a different naming problem because it takes a different argument format to `Model::query()->whereField('column', $value)`?  I also feel like `whereFields()` would be more appropriate in `ModelQueryBuilder` which kinda defeats the purpose of the shortcut.

Is there another naming convention out there for doing a `find($id)` but using parameters other than the ID? Doctrine and Phalcon use `findBy()` or variations of it.

On other projects I've used `get($id)` instead of `find($id)` and if Tempest were to do that then it'd free up `find($params)` for what I'm doing here?

brendt:
> On other projects I've used get($id) instead of find($id) and if Tempest were to do that then it'd free up find(column: $value, ...) for what I'm doing here?

This might actually be a good idea! I went with `find` because that's how Laravel does it, but I'm open to changing it.

mattdinthehouse:
Done:
- changed existing `find($id)` to `get($id)`
- changed new method `findBy(column: $value, ...)` to `find`
- added test for new `find` method

brendt:
Awesome! I've added an issue for the docs, but I can take care of it if you don't have time for, we'll see who gets there first! 

https://github.com/tempestphp/tempest-framework/issues/987

(I'll merge this PR when CI is done)

PR #966, opened 2025-03-08
feat(database): add a `Virtual` attribute to exclude model properties from query builder

### LINKED ISSUE #962: Hooked property support in models
When a model has a hooked property, it gets included in the query, but that property might not be available in the database. We need a way to prevent properties from being included in the query. Maybe all hooked properties, but that feels rather aggressive as well. Maybe some kind of attribute?

mattdinthehouse:
Fixes #962

Example use-case where I have a JSON `config` column and I just want a handy readonly accessor for values in it:
```php
<?php

namespace App\Models;

use Tempest\Database\Virtual;
use Tempest\Database\DatabaseModel;

final class MyModel implements DatabaseModel
{
	use IsDatabaseModel;

	#[Virtual]
	public string $url
	{
		get => $this->config['url'] ?? '';
	}

	public function __construct(
		public string $name,
		public ?array $config,
	)
	{ }
}
```

^ in this case I mark the `$url` property with `Virtual` so that `ModelQueryBuilder` doesn't include it in the SELECT clause of queries against this model

### DIFF
--- a/src/Tempest/Database/src/Builder/ModelQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/ModelQueryBuilder.php
@@ -5,10 +5,12 @@ declare(strict_types=1);
 namespace Tempest\Database\Builder;
 
 use Closure;
-use Tempest\Database\DatabaseModel;
 use Tempest\Database\Id;
-use Tempest\Database\Query;
 use function Tempest\map;
+use Tempest\Database\Query;
+use function Tempest\reflect;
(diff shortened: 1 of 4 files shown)


brendt:
That was fast :) Could you add some tests? I can add them if you can't

mattdinthehouse:
Done :)

brendt:
Cool, thanks! If you want to you may add a section to the docs: #968, but I'm fine doing it myself as well, probably later today :)

PR #970, opened 2025-03-08
fix(database): backslashes getting stripped from enum column values

### LINKED ISSUE #963: Migrations with string-backed enum columns with FQCN's values don't set up the DB column with backslashes

mattdinthehouse:
Fixes #963

### DIFF
--- a/src/Tempest/Database/src/QueryStatements/EnumStatement.php
+++ b/src/Tempest/Database/src/QueryStatements/EnumStatement.php
@@ -25,6 +25,7 @@ final readonly class EnumStatement implements QueryStatement
     {
         $cases = arr($this->enumClass::cases())
             ->map(fn (UnitEnum|BackedEnum $case) => $case instanceof BackedEnum ? $case->value : $case->name)
+            ->map(fn (string $value) => str_replace('\\', '\\\\', $value))
             ->map(fn (string $value) => "'{$value}'");
 
         if ($this->default !== null) {


brendt:
Nice, could we add a test? Something along the lines of what you described in #963 should do :)

brendt:
Also, we'd need it in `\Tempest\Database\QueryStatements\CreateEnumTypeStatement` (which is only used in postgres)

Issue #969, opened 2025-03-08
Promoted properties with default value aren't mapped


brendt:
```php
final class BlogPost
{
    public function __construct(
        public bool $published = true
    ) {}
}
```

Since objects are created without calling the constructor, promoted properties with a default value aren't set. If you write the property as a normal one, it does get the default value:

```php
final class BlogPost
{
    public bool $published = true;
}
```

The fix here is to check whether a property is promoted, and if so manually set the default value if no value is provided.

mattdinthehouse:
Why not just invoke the constructor?

brendt:
Because we allow partial object construction to lazily load relations.

mattdinthehouse:
But by having Lazy relations as promoted properties (or any relation as a promoted property) it means users can't call `new MyModel` without providing them otherwise PHP will throw a missing argument error,

and since Tempest's mapper doesn't call the constructor it means users can't put any logic into the constructor (eg for value sanitising or runtime validation etc) because it doesn't get run all the time.

So tbh I don't quite understand why Tempest uses promoted properties for models in the first place when the constructor isn't even called/callable?

Why not just un-promote those properties and define them at the class level like normal, let people define a constructor if they need to, and make the mapper call it? For the nice `new MyModel(column: $value, ...)` syntax there's already a `MyModel::new(column: $value, ...)` method

brendt:
That's a good point. I'll need to play around with it a bit more. IIRC, I also wanted `new MyModel(…)` to be possible, but it simply might not be doable. 

I'll look into it!

PR #971, opened 2025-03-08
fix(database): handle virtual properties when generating a query from a model

mattdinthehouse:
My bad - in #966 I only did the "read" part, this PR does the "write" part that actually threw the original error in #962


brendt:
I think this isn't the right approach, since readonly properties should also be writeable. The thing with the mapper is that it allows to create partial objects (which is why the constructor is skipped, you asked about it in https://github.com/tempestphp/tempest-framework/issues/969#issuecomment-2708143152 as wel). We allow partial objects because we support lazy relations.

Anyway, long story just to say that, with Tempest's mappers, there could be scenarios where a readonly property isn't written to when the object was initially constructed, but it did load afterwards. So we cannot simply expand the readonly check. Shouldn't it work without those readonly property changes? Also we'd need a test :)

mattdinthehouse:
Sorted - I've separated out the "is writable" check into it's own method.

I realise that part's semi-unrelated to the `Virtual` attribute, it's to avoid trying to set the value on a hooked property with no `set` handler on it when reloading the model and copying the properties over.

Actually just thinking about it now, because there's likely to be logic on hooked properties should we even be trying to copy them over like this in the `load()` method?

brendt:
> Actually just thinking about it now, because there's likely to be logic on hooked properties should we even be trying to copy them over like this in the load() method?

Isn't that what we're preventing with the added isWritable check? Maybe I misunderstand what you mean though

mattdinthehouse:
> Isn't that what we're preventing with the added isWritable check?

The `isWritable` method will return true if there's a "set" hook available, so it would run whatever logic is in there

So both of these cases:
```php
public string $foo;

public string $bar = {
    set(string $value) {
        $this->bar = strtoupper($value);
    }
}
```

which might lead to unexpected behaviour if the set hook does something less trivial - I wouldn't want someone to call `$myModel->load('relation')` and get a corrupted model.

I don't know the background about why `IsDatabaseModel::load()` recreates the model and copies over the values so I'll need your guidance on when/if/how to skip hooked properties regardless of whether they're marked `Virtual` or not

brendt:
Gotcha, thanks for explaining!

> I don't know the background about why IsDatabaseModel::load() recreates the model and copies over the values 

Convenience at the time 😇

> I'll need your guidance on when/if/how to skip hooked properties regardless of whether they're marked Virtual or not

I don't think we need to explicitly do anything with them for now. If people decide to add a set hook for a relation, I'd say they are on their own 😅

Would you be able to add a test for this one?

aidan-casey:
@mattdinthehouse - Would you be willing to make the updates above or should we have someone else take a look at this PR?

mattdinthehouse:
@aidan-casey feel free to pull someone else onto this i'm far too unreliable

PR #972, opened 2025-03-08
fix(view): ensure body tag isn't removed from view output

### LINKED ISSUE #967: Views - situation where <body> tag is stripped
### Tempest version

1.0-alpha.5 - specifically 92321b3f145f9207ed8593b7067dc3f7483e02f5

### PHP version

8.4

### Operating system

Linux

### Description

The \<body> tag is stripped from view output when you have a parent "base" view component,  but put the \<html> tag inside the child view.

This is most likely related to #886 where the DOM parser's behaviour needed to be taken into account differently.

### Steps to reproduce

In this example,  I've used a "base" component,  but kept the \<body> tag inside my child view:

// base.view.php
```php
<x-component name="x-base">

<!doctype html>
<html lang="en">

  <x-slot />

</html>

</x-component>
```

// child.view.php
```php
<x-base>

  <body>
    Body tag is missing
  </body>

</x-base>
```

(output)
```html
<!DOCTYPE html>
<html lang="en">Body tag is missing
  





</html>
```

**Use-case:**  I do this so as to have several different *html layouts* (which would be child components not shown in my example),  but have the \<head> tags managed by the single base component.   Each of the child components would have their own \<body class="xxx"> custom styling.

I could get around this by keeping the \<body> tag in the base view,  and pass the css classes as attributes into the base view.

However it would make the layout of my use-case clearer,  and I wouldn't have expected the current behaviour.

What do you think?

code-distortion:
Added a test for issue #967 - Test that the \<body> tag isn't removed when it's included in a child view component

closes #967

### DIFF
--- a/tests/Integration/View/ViewComponentTest.php
+++ b/tests/Integration/View/ViewComponentTest.php
@@ -528,8 +528,6 @@ final class ViewComponentTest extends FrameworkIntegrationTestCase
 
     public function test_body_tag_is_not_removed_from_child_view(): void
     {
-        $this->container->get(AppConfig::class)->environment = Environment::PRODUCTION;
-
         $this->registerViewComponent('x-layout', <<<'HTML'
         <html lang="en"><x-slot /></html>
         HTML);
--- a/tests/Integration/View/ViewComponentTest.php
+++ b/tests/Integration/View/ViewComponentTest.php
@@ -526,6 +526,23 @@ final class ViewComponentTest extends FrameworkIntegrationTestCase
         HTML, $html);
     }
 
+    public function test_body_tag_is_not_removed_from_child_view(): void
+    {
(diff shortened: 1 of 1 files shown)


code-distortion:
I think I've followed all the conventions but am happy to receive any feedback

brendt:
Perfect. I don't think the environment setup is necessary in this test, but that's ok, I'll work on it! Appreciate the broken test :)

brendt:
I looked into it, unfortunately `<x-layout><body>Content</body></x-layout>` cannot be parsed by PHP's HTML parser. It will switch the elements like so: `<body><x-layout>Content</x-layout></body>`. That's because `<body>` can only go directly into `<html>`. 

So unfortunately this is something we can't work around. I've tried switching to PHP's XML parser, but that one is missing way too many HTML-specific features that we rely on. So I'm afraid this is one of the very few edge cases we can't really work around at the moment.

PR #981, opened 2025-03-08
fix(view): allow head and body tags to appear in child views

code-distortion:
I had a go and worked out a way that I think solves the problem of `<head>` and `<body>` tags not working when appearing in child views #967.

This PR is a proof of concept as I think there are some things to get your input on (if these changes are actually desired in the first place).

It's possible that I've overlooked some obvious concept which nullifies this approach.  However I've tried to treat this carefully as I know the view code is important.

Sorry for the long description.  You don't need to look at this right away!

----

The main work is done in `Tempest\View\Renderers\TempestViewCompiler::parseDom()`, where it uses `\Dom\HTMLDocument` to find `<html>`, `<head>` and `<body>` tags and switch them with placeholders.

It identifies them by splitting the template using a regex initially,  but then testing the relevant parts with `\Dom\HTMLDocument` to see if they really are `<html>`, `<head>` and `<body>` tags.  This avoids issues with regexes getting confused when tags appear in other places, like comments and attributes.

It only splits based on those tag names,  it's not splitting it into every html element or anything.

I think these changes mean that we can remove the code in `TempestViewCompiler::parseDom()` that strips the `<head>` tags from the template and replaces them afterwards.

----

There are a few drawbacks, and questions I'd like your thoughts on:

It is a bunch of code to fix a particular problem.  I don't think it's over-the-top,  but you might not want to include it because of the complexity.

It runs the templates through `\Dom\HTMLDocument` an extra time for each `<html>`, `<head>` and `<body>` tag it finds.  Realistically that should only be up to three extra times plus one extra,  however I removed two from the original code.

Another class, `Tempest\View\Elements\ElementFactory` resolves and includes parent templates.  But it *also* resolves all the other PHP code that gets rendered in there at the same time.  I'd liked to have swapped the placeholder tags back to their original `<html>`, `<head>` and `<body>` state in `TempestViewCompiler`,  but I can't do it after the PHP code has been rendered in.

So I put the swapping code into `ElementFactory` for now,  and the tag swapping happens at the same time.  `ElementFactory` might need to be looked some more because I don't think it really should know about the placeholder tags.

This has also introduced another small problem for the same reason as above;  there isn't a good place to check the html structure with `Dom\HTMLDocument` *after* piecing the templates together,  but *before* rendering in the PHP code.  So I can't so easily enforce the fact that the html must start with `<html>`, and only have `<head>` / `<body>` children.  The check that was being enforced zealously before by `Dom\HTMLDocument` is now being side-stepped essentially.

Perhaps this is a feature and developers can choose what to start their HTML with themselves… ?

I created test `test_invalid_html_can_be_produced` because that's the new current behaviour,  but this can be reviewed of course.

----

I added some extra tests to cover the new situations.

The other changes are just tweaks to tests.  Because the code that strips the `<head>` tags from the template and places them back in has been removed,  a superfluous new line character has been removed from the output.


brendt:
It's interesting, I experimented with this approach as well, but I ran into the problem that you solved by moving the "unescaping" into `ElementFactory`. Let's take a look at your questions:

> It is a bunch of code to fix a particular problem. I don't think it's over-the-top, but you might not want to include it because of the complexity.

Yeah, I'm unsure whether it's worth the complexity, yet. But I'm honestly undecided about it atm.

> It runs the templates through \Dom\HTMLDocument an extra time for each <html>, <head> and <body> tag it finds. Realistically that should only be up to three extra times plus one extra, however I removed two from the original code.

Everything that happens in the compiler is cached, so while we shouldn't make it super slow, we also don't have to worry about a handful of extra operations.

> Another class, Tempest\View\Elements\ElementFactory resolves and includes parent templates. But it also resolves all the other PHP code that gets rendered in there at the same time. I'd liked to have swapped the placeholder tags back to their original <html>, <head> and <body> state in TempestViewCompiler, but I can't do it after the PHP code has been rendered in.

So this is what I ran into as well. This is the complexity part I'm most scared of atm. Not sure whether it's worth it

> This has also introduced another small problem for the same reason as above; there isn't a good place to check the html structure with Dom\HTMLDocument after piecing the templates together, but before rendering in the PHP code. So I can't so easily enforce the fact that the html must start with <html>, and only have <head> / <body> children. The check that was being enforced zealously before by Dom\HTMLDocument is now being side-stepped essentially.
>
>  Perhaps this is a feature and developers can choose what to start their HTML with themselves… ?

Yeah, I wouldn't worry about it too much. If people want to mess up their HTML, so be it.

---

So, in summary: I'm not sure. I feel like, for your specific case, it might not be worth the complexity, but I also feel like there might be more edge cases popping up in the future. I wouldn't be surprised if at one point there will be enough edge cases to warrant this approach.

The best approach would of course be a DOM parser that's a bit more lenient 😅

brendt:
Awesome, thanks for understanding! I'll close this PR but I've labeled it in case I need again in the future :)

Issue #925, opened 2025-02-27
Explore Mago for linting and formatting


innocenzi:
[Mago](https://github.com/carthage-software/mago) is a relatively new project which goal is to be a unified toolchain for PHP.

With php-cs-fixer not being compatible with PHP 8.4 weeks after its release, and code-quality tooling giving us trouble in CI recently, I think now is the time to explore Mago.

I actually have tried Mago's formatter already, but it was too opinionated for me, compared to php-cs-fixer. However, azjezz, the maintainer, listens to feedback and is quite fast at fixing issues—I've yet to try the fixes he landed recently.

Mago's linter, on the other hand, was really good right from the start. There is a lot of rules already, the linting is fast, and I've only had one issue which was more of an edge case and was fixed fast too.

Note that Mago has the ability to let us write specific plugins for Tempest as well, which is a big plus.

innocenzi:
Update on this: Mago has gone a long way since I created the issue already. A few more fixes to the formatter and I'll PR it as a replacement for php-cs-fixer.

PR #986, opened 2025-03-10
style: format and lint using mago

innocenzi:
This pull request introduces [Mago](https://github.com/carthage-software/mago), a stellar toolchain for PHP that currently includes a linter and a formatter.

Mago is in its early days, but is already packed with features. Its linter has dozens of really interesting rules, and its formatted allows us to completely replace php-cs-fixer. While less configurable, it's pretty solid and conforms to our current coding style.

Note that if we encounter issues with Mago, its maintainer, azjezz, is very reliable. He has helped me with a lot of reports and is usually super fast to respond.

Closes https://github.com/tempestphp/tempest-framework/issues/925

## Changes
- I removed `php-cs-fixer` completely.
- I added `mago lint` and `mago fmt` to the CI.
- I updated `composer qa` to use Mago.
- I configured `mago.toml` with a formatting style corresponding to the codebase, and configured rules that were relevant to us.

## Potential linting rules
Some lint rules are interesting, but I don't know if I should enable them:

- `analysis/override-attribute` 
  This one enforces the usage of [`#[Override]`](https://www.php.net/manual/en/class.override.php). It's an attribute I use in app code, but I don't know how relevant it would be in the framework code. I am not convinced we should use it.

- `comment/no-untagged-todo`
  This one requires `TODO` comment to have an issue or someone assigned to it, eg. `TODO(#123)`. I think it's a very good practice that we should adopt, but we have 30 TODOs currently that need to be tracked. 😅


rector.php
```diff
@@ -40,6 +41,7 @@
         ExplicitNullableParamTypeRector::class,
     ])
     ->withSkip([
+        NewlineAfterStatementRector::class, // TODO(https://github.com/carthage-software/mago/issues/143)
```
azjezz, line 44: this is fixed. can be removed from rector.

innocenzi: Good catch!

brendt:
> This one enforces the usage of [#[Override]](https://www.php.net/manual/en/class.override.php). It's an attribute I use in app code, but I don't know how relevant it would be in the framework code. I am not convinced we should use it.

No, I don't think we should add #[Override]: https://stitcher.io/blog/override-in-php-83#some-thoughts

> This one requires TODO comment to have an issue or someone assigned to it, eg. TODO(#123). I think it's a very good practice that we should adopt, but we have 30 TODOs currently that need to be tracked. 😅

I don't mind it!

brendt:
I suggest we go ahead and merge. We can always finetune later, but I want to avoid too many merge conflicts with main

brendt:
I merged :p

PR #1020, opened 2025-03-14
feat(view): support merging class attributes

innocenzi:
This pull request adds failing test for a new feature.

We should be able to support the following:

```html
<div class="foo" :class="$someStringOrArray"></div>
```

If `$someStringOrArray` is `['bar', 'baz']`, the resulting class should be `foo bar baz`. I already implemented support for passing arrays as attributes.

However, duplicate attributes other than `class` should raise an exception, because their merging behavior would be undefined.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,6 +14,9 @@ DISCOVERY_CACHE=false
 # Enable or disable config cache
 CONFIG_CACHE=false
 
+# Enable or disable icon cache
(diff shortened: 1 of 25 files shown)


brendt:
I'm making progress, but still some work to be done :)

brendt:
I suggest when handle the attribute precedence in another PR: #1026

PR #1024, opened 2025-03-16
feat(http): fix http header casing on retrieval

### LINKED ISSUE #1006: case sensitive request header
`$request->headers['Referer']` should be the same as `$request->headers['referer']`. Either by making all headers lowercase, or by implementing a custom collection object that will change the keys on the fly.

Also check cookies etc.

blackshadev:
# What

Fixes #1006 .

- Ensures all headers keys are lowercased when constructing the request and when retrieving the values from the headers

# How to test

1. Have a controller with this method
```
    #[Get('/headers')]
    public function headers(Request $request): Response {
        return new Ok([
            'host' => $request->headers['host'] ?? null,
            'Host' => $request->headers['Host'] ?? null,
        ]);
    }
```
2. Call the method and see as a result `{"host":"localhost:8000","Host":"localhost:8000"}`. 

# Notes

1. I did not yet do any work on the cookie retrieval. I will do that in a later PR. The reason being, the cookies in a request are accessed through the container instead of passing them through (like any other array). I think I might rework that in one go.

### DIFF
--- a/src/Tempest/Router/src/RequestHeaders.php
+++ b/src/Tempest/Router/src/RequestHeaders.php
@@ -18,7 +18,7 @@ final readonly class RequestHeaders implements ArrayAccess, IteratorAggregate
     public static function normalizeFromArray(array $headers): self
     {
         $normalized = array_combine(
-            array_map('strtolower', array_keys($headers)),
+            array_map(strtolower(...), array_keys($headers)),
             array_values($headers),
         );
         return new self($normalized);
--- a/src/Tempest/Router/src/RequestHeaders.php
+++ b/src/Tempest/Router/src/RequestHeaders.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Router;
+
+use ArrayAccess;
+use ArrayIterator;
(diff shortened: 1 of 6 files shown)


src/Tempest/Router/src/RequestHeaders.php
```diff
@@ -0,0 +1,74 @@
...
+
+final readonly class RequestHeaders implements ArrayAccess, IteratorAggregate
+{
+    /**
+     * @param array<string, string> $headers
+     */
+    public static function normalizeFromArray(array $headers): self
+    {
+        $normalized = array_combine(
+            array_map('strtolower', array_keys($headers)),
```
brendt, line 21: I prefer to use `strtolower(...)`

brendt (dismissed):
Perfect! One minor nitpick but this is exactly what I had in mind as wel :)

src/Tempest/Router/src/RequestHeaders.php
blackshadev: Done, I always forget about the syntax you mentioned, but very nice suggestion.

brendt:
Thanks!

PR #1028, opened 2025-03-16
feat(console): add `editor:metadata` command

nhedger:
### Summary

This PR adds a new _hidden_ `editor:metadata` command to the framework.

### Description

The goal of the `editor:metadata` command is to expose metadata about the framework and the host application to text editors such as VS Code, allowing for a richer development experience.

In this initial PR, we break the ground by providing a simple `framework` field that exposes the name of the framework, and its version. In the future, the metadata will contain information such as configuration values, discovery locations, component names, etc.

### Exchange format

Metadata is exchanged in the form of a JSON payload directly emitted on stdout.

```shell
~/Code/tempest/framework php tempest editor:metadata
{
    "framework": {
        "name": "Tempest",
        "version": "1.0.0-alpha.5"
    }
}%
```

### DIFF
--- a/src/Tempest/Console/src/Commands/EditorMetadataCommand.php
+++ b/src/Tempest/Console/src/Commands/EditorMetadataCommand.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console\Commands;
+
+use Tempest\Console\Console;
+use Tempest\Console\ConsoleCommand;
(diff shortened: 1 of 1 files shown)


innocenzi:
I think that before merging, we should get a PoC of the vscode extension working with that first. Once this is done, we can iterate if needed, then merge. Would that work for you?

nhedger:
Sure. I actually already have one ready. I'll publish it and link it here afterwards

PR #1035, opened 2025-03-17
refactor(view): use `is_self_closing_tag` where relevant

innocenzi:


brendt:
The correct term is actually "void" element: https://developer.mozilla.org/en-US/docs/Glossary/Void_element

PR #1037, opened 2025-03-18
refactor(view): is_void_tag

brendt:


innocenzi:
Thanks!

Please remember that PR titles are changelog entries, so before merging they should make sense without context 🥹

PR #1038, opened 2025-03-19
feat(support): add `mapFirstTo` and `mapLastTo` to array utils

innocenzi:
This pull request adds support for using the mapper on the first or last element of an array:

```php
arr(glob(__DIR__ . "/content/{$version->value}/{$category}/*.md"))
  ->map(fn (string $path) => before_first(basename($path), '.'))
  ->sort()
  ->mapFirstTo(ImmutableString::class)
  ->title();
```

### DIFF
--- a/src/Tempest/Mapper/src/Mappers/StringToStringableMapper.php
+++ b/src/Tempest/Mapper/src/Mappers/StringToStringableMapper.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Mapper\Mappers;
+
+use Stringable;
+use Tempest\Mapper\Mapper;
+
+final readonly class StringToStringableMapper implements Mapper
+{
+    public function canMap(mixed $from, mixed $to): bool
+    {
+        if (! is_a($to, Stringable::class, allow_string: true)) {
+            return false;
+        }
+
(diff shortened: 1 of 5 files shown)


brendt:
Wouldn't `->limit(1)->mapTo()` be more flexible?

innocenzi:
We would then have to do:

```php
arr(glob(__DIR__ . "/content/{$version->value}/{$category}/*.md"))
  ->map(fn (string $path) => before_first(basename($path), '.'))
  ->sort()
  ->limit(1)
  ->mapTo(ImmutableString::class)
  ->first();
```

It achieves similarly but it's quite longer. That being said, I can implement `limit`/`nth` in this PR and implement `mapFirstTo` and `mapLastTo` on top of it?

innocenzi:
Well... I added `at` (modeled after the ECMScript implementation, supporting negative values), but realized it's not particularly useful to build `mapFirstTo` and `mapLastTo` on top of it...

I also can't find a name for something that would map the specified index using the mapper (`mapAtTo`?)

I suggest keeping the `map` methods this way, in addition to the new `at`, who's also very useful

brendt:
I'm ok with that :)

brendt:
Feel free to merge btw when you're happy

PR #1042, opened 2025-03-19
feat(support): add `words` and `sentence` methods to string utils

innocenzi:
This pull request adds a `words` and `sentence` method to the string classes.

Both of these are quite hard to get right.
- `words` tries to be accurate here, by counting actual words—tests could be improved, but I think it's a fine start.
- `sentence` is pretty naive and built on top of `words`. It uppercases the first word and lowercase the rest. Actual sentence case is super hard to achieve, and I expect this method to be improved over time, so it's a fine start as well.


brendt:
Cool!

PR #1043, opened 2025-03-19
fix(validation): prevent type errors in rules using `preg_match`

jonkerw85:
**Summary:**
This PR adds a check to ensure the value passed to `preg_match` is a string in various validation rules, preventing a `TypeError` when non-string values (e.g., arrays, booleans, or objects) are provided.

**Changes:**
- Added type checks for string values in the **Alpha**, **AlphaNumeric**, **Numeric**, **RegEx**, **Time**, **Ulid**, and **Uuid** validation rules.
- Updated tests to ensure invalid types return `false` instead of causing a `TypeError`.

**Purpose:**
To prevent errors when non-string values are passed to `preg_match` and improve the stability of the validation rules.

### DIFF
--- a/src/Tempest/Validation/src/Rules/Alpha.php
+++ b/src/Tempest/Validation/src/Rules/Alpha.php
@@ -12,6 +12,10 @@ final readonly class Alpha implements Rule
 {
     public function isValid(mixed $value): bool
     {
+        if (! is_string($value)) {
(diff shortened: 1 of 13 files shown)


innocenzi (approved):

innocenzi:
Thanks!

PR #1046, opened 2025-03-20
chore(linter): exclude cache directories

NeoIsRecursive:
When running things (`composer qa`) locally I got failures in the lint due to cache files. Dunno if this is the best way to do this but it seems to work.


innocenzi:
Thanks! Yeah, I think that should be ignored indeed.

Issue #1049, opened 2025-03-20
Add Back response


thoresuenert:
### Description

For a convenient way to redirect a user back after a form request we need a Back Response class.

### Benefits

We can use the `Back::class` to redirect the user back:
```php
public function store(): Response
{
    return new Back();
}
```

PR #1050, opened 2025-03-20
feat(router): add redirect back response

thoresuenert:
Fixes: #1049


src/Tempest/Router/tests/Responses/BackResponseTest.php
```diff
@@ -0,0 +1,59 @@
...
+use Tempest\Router\GenericRequest;
+use Tempest\Router\Header;
+use Tempest\Router\Request;
+use Tempest\Router\Responses\Back;
+use Tests\Tempest\Integration\FrameworkIntegrationTestCase;
+
+/**
+ * @internal
+ */
+final class BackResponseTest extends FrameworkIntegrationTestCase
```
brendt, line 18: You should move this test to the integration test folder instead: https://github.com/tempestphp/tempest-framework/tree/main/tests/Integration

brendt:
The failing tests aren't related to this PR, I'll fix them before merging.

One thing I was wondering about: how to set the `referer` header with eg. a normal GET request? I believe Laravel stores the previous URL in a session to cover more cases. I could be wrong though. Not saying we need it, just wanted to check

thoresuenert:
I will look into it. it will take some time because its the first time doing something with temptest.
Should i convert the pr to draft until i am finished?

brendt:
Currently global HTTP middleware is added here: https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Router/src/RouterInitializer.php#L19

This will be refactored soon, but you can add it over there for now.

brendt:
Middleware has now been refactored. The difference is that middleware is discovered, so you won't need to add it anywhere anymore. Just make sure to add a priority like with this one: https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Router/src/Cookie/SetCookieMiddleware.php#L13

thoresuenert:
I introduced a new middleware, only for the purpose of storing the previous url in the session.
Finally i found the tests for the other responses in `Tests\Tempest\Integration\Http\Responses` i was looking for `Router\Responses` something.

I borrowed two functions from laravel `getPreviousUrl` and `setPreviousUrl` and added these to the Session Class.
Later on i figured out that the `Request` has a function `getSessionValue`.

It is possible to remove the new functions from the session and use the new const `PREVIOUS_URL` in combination with the set and get functions.

Let me know if i should refactor `getPreviousUrl` and `setPreviousUrl`. 

PS: I did my best to figure out the right way to test. It is way harder then expected to contribute to the core of a framework ;) Thanks for your patience and help.

brendt:
Looking great! Thanks for sticking with it and figuring it out, I actually have nothing to add :) 

I'll let CI run and then merge it!

Are you up for PRing the docs as well? A simple entry in this list should be fine: https://github.com/tempestphp/tempest-docs/blob/main/src/Web/Documentation/content/main/1-essentials/02-controllers.md?plain=1#L363

thoresuenert:
@brendt of couse. https://github.com/tempestphp/tempest-docs/pull/71

PR #1053, opened 2025-03-20
fix(discovery): pass real paths to discovery classes

innocenzi:
Apparently, `RecursiveDirectoryIterator` does not yield canonical paths. This pull request ensures that `discoverPath` gets the actual canonical path to the file.

This shouldn't break anything, but I would like you to double-check @brendt before merging


brendt:
Totally fine with it. Is there a benefit with using realpath?

innocenzi:
What prompted me to use it is the new Vite Discovery, the generated asset links are not canonical

PR #1054, opened 2025-03-20
feat(database): Add migration hash checking

erikaraujo:
This PR adds Hash-checking to database migrations - preventing migration files from being tempered with.

WIP - will update PR description later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,11 @@ final class MigrateUpCommand
         description: 'Run all new migrations',
         middleware: [ForceMiddleware::class, CautionMiddleware::class],
     )]
-    public function __invoke(): void
-    {
-        $this->migrationManager->up();
+    public function __invoke(
+        #[ConsoleArgument(description: '')]
+        bool $allowChanges = false,
+    ): void {
+        $this->migrationManager->up($allowChanges);
 
         $this->console->success("Everything migrated");
     }
--- a/src/Tempest/Database/src/Migrations/MigrationException.php
+++ b/src/Tempest/Database/src/Migrations/MigrationException.php
@@ -12,4 +12,9 @@ final class MigrationException extends Exception
     {
         return new self('Migrations table does not exist. Nothing to roll back.');
     }
+
+    public static function hashMismatch(): self
+    {
+        return new self('Migration hash mismatch.');
+    }
 }
--- a/src/Tempest/Database/src/Migrations/MigrationManager.php
+++ b/src/Tempest/Database/src/Migrations/MigrationManager.php
@@ -10,8 +10,10 @@ use Tempest\Database\Config\DatabaseConfig;
 use Tempest\Database\Config\DatabaseDialect;
 use Tempest\Database\Database;
 use Tempest\Database\DatabaseMigration as MigrationInterface;
+use Tempest\Database\DatabaseMigration;
 use Tempest\Database\Exceptions\QueryException;
 use Tempest\Database\Query;
+use Tempest\Database\QueryStatement;
 use Tempest\Database\QueryStatements\DropTableStatement;
 use Tempest\Database\QueryStatements\SetForeignKeyChecksStatement;
 use Tempest\Database\QueryStatements\ShowTablesStatement;
@@ -29,7 +31,7 @@ final readonly class MigrationManager
     ) {
     }
 
-    public function up(): void
+    public function up(bool $allowChanges): void
     {
         try {
             $existingMigrations = Migration::all();
@@ -39,13 +41,17 @@ final readonly class MigrationManager
             $existingMigrations = Migration::all();
         }
 
-        $existingMigrations = array_map(
-            static fn (Migration $migration) => $migration->name,
-            $existingMigrations,
-        );
-
         foreach ($this->migrations as $migration) {
-            if (in_array($migration->name, $existingMigrations, strict: true)) {
+            $existingMigration = array_find(
+                $existingMigrations,
+                static fn (Migration $existingMigration) => $existingMigration->name === $migration->name,
+            );
+
+            if ($existingMigration !== null) {
+                if (! $allowChanges) {
+                    $this->verifyMigrationHash($migration, $existingMigration);
+                }
+
                 continue;
             }
 
@@ -126,6 +132,7 @@ final readonly class MigrationManager
 
             Migration::create(
                 name: $migration->name,
+                hash: $this->getMigrationHash($migration),
             );
         } catch (PDOException $pdoException) {
             event(new MigrationFailed($migration->name, $pdoException));
@@ -198,5 +205,40 @@ final readonly class MigrationManager
             },
             new ShowTablesStatement()->fetch($dialect),
         );
+    }
+
+    private function getMigrationHash(DatabaseMigration $migration): string
+    {
+        $minifiedDownSql = $this->getMinifiedSqlFromStatement($migration->down());
+        $minifiedUpSql = $this->getMinifiedSqlFromStatement($migration->up());
+
+        return hash('sha256', $minifiedDownSql . $minifiedUpSql);
+    }
+
+    private function getMinifiedSqlFromStatement(?QueryStatement $statement): string
+    {
+        if ($statement === null) {
+            return '';
+        }
+
+        $query = new Query($statement->compile($this->databaseConfig->dialect));
+
+        // Remove comments
+        $sql = preg_replace('/--.*$/m', '', $query->getSql()); // Remove SQL single-line comments
+        $sql = preg_replace('/\/\*[\s\S]*?\*\//', '', $sql); // Remove block comments
+
+        // Remove blank lines and excessive spaces
+        $sql = preg_replace('/\s+/', ' ', trim($sql));
+
+        return $sql;
+    }
+
+    private function verifyMigrationHash(DatabaseMigration $migration, Migration $existingMigration): void
+    {
+        $hash = $this->getMigrationHash($migration);
+
+        if ($hash !== $existingMigration->hash) {
+            event(new MigrationFailed($migration->name, MigrationException::hashMismatch()));
+        }
     }
 }
--- a/src/Tempest/Framework/Commands/MigrateUpCommand.php
+++ b/src/Tempest/Framework/Commands/MigrateUpCommand.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Tempest\Framework\Commands;
 
 use Tempest\Console\Console;
+use Tempest\Console\ConsoleArgument;
 use Tempest\Console\ConsoleCommand;
 use Tempest\Console\Middleware\CautionMiddleware;
 use Tempest\Console\Middleware\ForceMiddleware;
@@ -29,9 +30,11 @@ final class MigrateUpCommand
         description: 'Runs all new migrations',
         middleware: [ForceMiddleware::class, CautionMiddleware::class],
     )]
-    public function __invoke(): void
-    {
-        $this->migrationManager->up();
+    public function __invoke(
+        #[ConsoleArgument(description: '')]
+        bool $allowChanges = false,
+    ): void {
+        $this->migrationManager->up($allowChanges);
 
         $this->console
             ->success(sprintf('Migrated %s migrations', $this->count));
--- a/src/Tempest/Database/src/Migrations/CreateMigrationsTable.php
+++ b/src/Tempest/Database/src/Migrations/CreateMigrationsTable.php
@@ -18,7 +18,8 @@ final class CreateMigrationsTable implements DatabaseMigration
     {
         return new CreateTableStatement(Model::table()->tableName)
             ->primary()
-            ->text('name');
+            ->text('name')
+            ->varchar('hash', 64);
     }
 
     public function down(): QueryStatement
--- a/src/Tempest/Database/src/Migrations/Migration.php
+++ b/src/Tempest/Database/src/Migrations/Migration.php
@@ -12,4 +12,6 @@ final class Migration implements DatabaseModel
     use IsDatabaseModel;
 
     public string $name;
+
+    public string $hash;
 }
--- a/tests/Integration/Database/MigrationManagerTest.php
+++ b/tests/Integration/Database/MigrationManagerTest.php
@@ -17,13 +17,13 @@ final class MigrationManagerTest extends FrameworkIntegrationTestCase
     {
         $migrationManager = $this->container->get(MigrationManager::class);
 
-        $migrationManager->up();
+        $migrationManager->up(allowChanges: false);
 
         $migrations = Migration::all();
         $this->assertNotEmpty($migrations);
         $oldCount = count($migrations);
 
-        $migrationManager->up();
+        $migrationManager->up(allowChanges: false);
 
         $migrations = Migration::all();
         $this->assertNotEmpty($migrations);


README.md
```diff
@@ -56,9 +56,11 @@ final class MigrateUpCommand
         description: 'Run all new migrations',
         middleware: [ForceMiddleware::class, CautionMiddleware::class],
     )]
-    public function __invoke(): void
-    {
-        $this->migrationManager->up();
+    public function __invoke(
+        #[ConsoleArgument(description: '')]
+        bool $allowChanges = false,
```
brendt, line 61: I think we need to come up with a more descriptive name, but not sure what

src/Tempest/Database/src/Migrations/MigrationManager.php
```diff
@@ -199,4 +206,39 @@ private function getTableDefinitions(DatabaseDialect $dialect): array
             new ShowTablesStatement()->fetch($dialect),
         );
     }
+
+    private function getMigrationHash(DatabaseMigration $migration): string
+    {
+        $minifiedDownSql = $this->getMinifiedSqlFromStatement($migration->down());
+        $minifiedUpSql = $this->getMinifiedSqlFromStatement($migration->up());
+
+        return hash('sha256', $minifiedDownSql . $minifiedUpSql);
```
brendt, line 215: Can we use another (less complex) hashing algo? This doesn't have to be secure, we care more about performance here. OTOH, it would probably make no difference in practice

src/Tempest/Database/src/Migrations/MigrationManager.php
```diff
@@ -199,4 +206,39 @@ private function getTableDefinitions(DatabaseDialect $dialect): array
...
+    }
+
+    private function getMinifiedSqlFromStatement(?QueryStatement $statement): string
```
brendt, line 218: Why do we need the minified version? Can't we hash the non-minified sql?

src/Tempest/Database/src/Migrations/MigrationManager.php
```diff
@@ -199,4 +206,39 @@ private function getTableDefinitions(DatabaseDialect $dialect): array
...
+
+        return $sql;
+    }
+
+    private function verifyMigrationHash(DatabaseMigration $migration, Migration $existingMigration): void
+    {
+        $hash = $this->getMigrationHash($migration);
+
+        if ($hash !== $existingMigration->hash) {
+            event(new MigrationFailed($migration->name, MigrationException::hashMismatch()));
```
brendt, line 241: Probably add a new event `MigrationHashMismatch` or something

brendt:
So what exactly is the flow here? Let's say I have a migration file that has changed. Then what? We run `migrate:up` locally and get an error? I'm not sure I fully understand the flow yet.

README.md
erikaraujo: `--allow-migration-files-tempering`? `--skip-tempering-check`?

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: I don't mind changing it to `xxh128` or some other algorithm.

While I do agree with you that in practice this will likely not make a huge difference, I do think we should pick the faster algorithm.

We _could_ have the algo being an Config - but that may be overkill.

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: This is more for styling purposes - specially on raw sql migration files. Adding/removing an indentation shouldn't really prevent you from running migrations.

I don't really feel too strongly about this, but I think it makes sense.

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: Will do!

erikaraujo:
> So what exactly is the flow here? Let's say I have a migration file that has changed. Then what? We run `migrate:up` locally and get an error? I'm not sure I fully understand the flow yet.

Yup, that's it. Migrations will not run if existing migrations have been tempered with.

src/Tempest/Database/src/Migrations/MigrationManager.php
brendt: We're only using it to generate the hash, is that correct? I would get rid of it

brendt:
I'm fine with it being the default mode. Let's not hook into `--force` though, but keep it as a separate argument. I'm still struggling to understand how this works in practice.

Let's say you make a change during development and run `migrate:fresh`, no error will be thrown since we start from scratch. Next we deploy and run `migrate:up`. This time the migration fails because an old migration has errors. What's the next step then? How do you get out of this "error state"?

Possibly by running with `--allow-changes`. Does that mean that the old migration hash gets updated to the new one? Is there another way to fix the error state without running with `--allow-changes`? 

To me it feels like, by the time this error is triggered, it's already pretty (too) late: we're deploying to production. That's why I suggested adding a warning of some sorts (maybe an error) that's shown when running `migrate:fresh`, I think that makes more sense since that way we'll get feedback during development instead of deployment. We can still keep the check in `migrate:up` as a last-resort, but ideally it should never trigger.

The end goal here is that people don't update existing migrations, but instead make new ones that apply on top of the original one.

I hope I'm making sense 😅

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: Yes - but if I get rid of it, the hash will be different if indentation changes on the migration file, which is not what we want.

erikaraujo:
> I'm fine with it being the default mode. Let's not hook into `--force` though, but keep it as a separate argument. I'm still struggling to understand how this works in practice.
> 
> Let's say you make a change to an existing migration during development and run `migrate:fresh`, no error will be thrown since we start from scratch. Next we deploy and run `migrate:up`. This time the migration fails because the existing migration's hash has a mismatch. What's the next step then? How do you get out of this "error state"?
> 
> Possibly by running with `--allow-changes`. Does that mean that the old migration hash gets updated to the new one? Is there another way to fix the error state without running with `--allow-changes`?
> 
> To me it feels like, by the time this error is triggered, it's already pretty (too) late: we're deploying to production. That's why I suggested adding a warning of some sorts (maybe an error) that's shown when running `migrate:fresh`, I think that makes more sense since that way we'll get feedback during development instead of deployment. We can still keep the check in `migrate:up` as a last-resort, but ideally it should never trigger.
> 
> The end goal here is that people don't update existing migrations, but instead make new ones that apply on top of the original one.
> 
> I hope I'm making sense 😅

I understand your concern. One of the most popular migration libraries (Redgate Flyway) doesn't have a `:fresh` command per se, but you can do a `clean` followed by a `migrate`, which is the same thing in the end. When you do such a thing, there's no warning or anything - it's like you're recreating it all from scratch.

And I don't know how we'd accomplish a warning or something on Tempest's `:fresh` command either without compromising the performance of it. Like, we'd need to make all checks before the `->dropAll` call.

src/Tempest/Database/src/Migrations/MigrationManager.php
brendt: Ah, didn't we generate the hash based on the generate SQL? I figured there wouldn't be any indentation changes when we use the builder. But in case of SQL migrations there might be…

I'm ok keeping it in! Can I suggest one refactor then? Create a method `hashQuery` or something alike that will both minify and hash the query, so that it's one self-contained part.

erikaraujo: >Ah, didn't we generate the hash based on the generate SQL? I figured there wouldn't be any indentation changes when we use the builder. But in case of SQL migrations there might be…

Exactly - for raw sql migrations we need to remove comments and minify to guarantee that style changes have no impact on the generated hash.

> I'm ok keeping it in! Can I suggest one refactor then? Create a method hashQuery or something alike that will both minify and hash the query, so that it's one self-contained part.

You mean create a trait `hasQuery` and use it inside the MigrationManager?

README.md
erikaraujo: @brendt @innocenzi which one should we go with?

- `--allow-changes`
- `--allow-tempering`
- `--allow-migration-files-tempering`
- `--skip-tempering-check`
- `--skip-validation` (because we'll also have a `migrate:validate` command)

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: @brendt How about `MigrationValidationFailed`?

README.md
erikaraujo: Or - because we shouldn't be shipping breaking changes anymore, should we have an opt-in `--validate` flag instead?

src/Tempest/Database/src/Migrations/MigrationManager.php
erikaraujo: >I'm ok keeping it in! Can I suggest one refactor then? Create a method hashQuery or something alike that will both minify and hash the query, so that it's one self-contained part.

Actually, I don't think I get it. I have a `getMigrationHash` method already, that takes a `DatabaseMigration` and generates the hash by calling `getMinifiedSqlFromStatement` twice (one for the `up` `QueryStatement` and one for the `down` `QueryStatement`.

Do you just want me to rename `getMigrationHash` to `hashQuery` instead?

erikaraujo: I can definitely rename `getMigrationHash` to `hashMigration` and `getMinifiedSqlFromStatement` to `minifyQueryStatement` - if that makes it more inline with other methods in the framework.

brendt: I was talking about merging them together, but I misremembered that you made two hashes. Ignore all of what I said before then 😁 Sorry for the noise!

src/Tempest/Database/src/Migrations/MigrationManager.php
brendt: Perfect!

brendt:
Ping me when ready :)

erikaraujo:
> Ping me when ready :)

@brendt I believe the code itself is ready. I'm just having a hard time figuring out how to properly add Unit Tests for when a migration file has been tampered with 😅

erikaraujo:
> Ping me when ready :)

@brendt Ready! :)

src/Tempest/Framework/Commands/MigrateUpCommand.php
```diff
@@ -28,12 +30,24 @@ public function __construct(
...
         middleware: [ForceMiddleware::class, CautionMiddleware::class],
     )]
-    public function __invoke(): void
-    {
+    public function __invoke(
+        #[ConsoleArgument(description: 'Validates the integrity of existing migration files by checking if they have been tampered with.')]
+        bool $validate = false,
+    ): ExitCode {
+        if ($validate) {
+            $validationSuccess = $this->console->call('migrate:validate');
```
innocenzi, line 38:
```suggestion
            $validationSuccess = $this->console->call(MigrateValidateCommand::class);
```

Nitpick: prefer the FQCN here (previous code doesn't use that yet because support for FQCN was recently introduced)

innocenzi (commented):
Super cool PR! Is there a reason why validation is opt-in though? I'd prefer having to specify `--no-validate` to opt-out

src/Tempest/Framework/Commands/MigrateUpCommand.php
erikaraujo: Oh, I didn't know this was added. I **much** prefer it like this also. I'll make sure to update it.

erikaraujo:
> Super cool PR! Is there a reason why validation is opt-in though? I'd prefer having to specify `--no-validate` to opt-out

@innocenzi I thought we were already avoiding breaking changes (I was wrong 😅 ).
I can definitely change this. Should I go ahead and do it, or is it better to wait for @brendt 's feedback?

PS: I hate that github doesn't allow replying to such comments

src/Tempest/Framework/Commands/MigrateUpCommand.php
erikaraujo: Pushed the changes.

innocenzi:
> I can definitely change this. Should I go ahead and do it, or is it better to wait for @brendt 's feedback?

I think Brent would agree with me, but you can wait to be sure to avoid having to revert the change


> What would a good name for the argument be? I'm not a huge fan of `--no-validate` because I usually prefer having "positive" options. Like `--no-validate=false` is confusing.

[We have support for negating arguments](https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Console/src/Input/ConsoleInputArgument.php#L63-L64), so it's fine: you can set the current `validate` argument to `true` by default, and passing `--no-validate` will set it to `false`. So no issue with confusing booleans

> PS: I hate that github doesn't allow replying to such comments

I'm with you! 😅

erikaraujo:
> [We have support for negating arguments](https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Console/src/Input/ConsoleInputArgument.php#L63-L64), so it's fine: you can set the current `validate` argument to `true` by default, and passing `--no-validate` will set it to `false`. So no issue with confusing booleans

🤯 
Neat! I actually have all the changes done locally already - but I'll wait on the feedback to push them.

README.md
```diff
@@ -44,29 +44,46 @@ final class BookController
 ```
 
 ```php
+#[Singleton]
```
brendt, line 47: Let's not make changes to the README in this PR, we can do that separately

src/Tempest/Database/src/Migrations/MigrationException.php
```diff
@@ -12,4 +12,14 @@ public static function noTable(): self
     {
         return new self('Migrations table does not exist. Nothing to roll back.');
     }
+
+    public static function hashMismatch(): self
```
brendt, line 16: I prefer separate exception classes. Maybe add an interface `MigrationException` on top of all concrete migration classes.

src/Tempest/Database/src/Migrations/MigrationManager.php
```diff
@@ -108,6 +110,60 @@ public function dropAll(): void
...
+    public function rehashAll(): void
+    {
+        try {
+            $existingMigrations = Migration::all();
+        } catch (PDOException) {
+            return;
+        }
+
+        foreach ($existingMigrations as $existingMigration) {
+            $databaseMigration = array_find(
```
brendt, line 122: What exactly are we doing here? Shouldn't we just loop over the existing migrations and update them?

src/Tempest/Framework/Commands/MigrateFreshCommand.php
```diff
@@ -30,8 +31,18 @@ public function __construct(
...
-    public function __invoke(): ExitCode
-    {
+    public function __invoke(
+        #[ConsoleArgument(description: 'Validates the integrity of existing migration files by checking if they have been tampered with.')]
+        bool $validate = false,
+    ): ExitCode {
+        if ($validate) {
+            $validationSuccess = $this->console->call(MigrateValidateCommand::class);
+
+            if ($validationSuccess !== 0 && $validationSuccess !== ExitCode::SUCCESS) {
```
brendt, line 41: What's the `!== 0` for?

brendt (changes requested):
Looks good, some minor questions, but that's ok. 

FYI I'm merging #1076 today, I'll check this PR afterwards to see if anything breaks because of it. If it does, I'll fix it on your branch.

brendt:
> I think Brent would agree with me, but you can wait to be sure to avoid having to revert the change

Yeah it's fine, make it true by default :)

src/Tempest/Framework/Commands/MigrateFreshCommand.php
erikaraujo: Because `->call` can return `ExitCode` and `int` I thought I'd handle both cases. I can gladly remove it, though.

src/Tempest/Database/src/Migrations/MigrationManager.php
brendt: Aaah that makes sense! Could you add that explanation as a comment in the code? Just so that we remember two months from now :)

src/Tempest/Framework/Commands/MigrateFreshCommand.php
brendt: Aah true 🤔 Let's keep it in

README.md
erikaraujo: Undid the changes.

brendt:
Thanks!

PR #1059, opened 2025-03-22
feat(router): allow checking an action against the current route

innocenzi:
This pull request adds the ability to check whether the current route matches the specified controller action:

```php
// /user/1

Tempest\is_uri(UserController::class); // true
Tempest\is_uri(UserController::class, id: 1); // true
Tempest\is_uri(UserController::class, id: 2); // false

$router->isCurrentUri(UserController::class); // true
$router->isCurrentUri(UserController::class, id: 1); // true
$router->isCurrentUri(UserController::class, id: 2); // false
```

Note that this doesn't currently apply the regular expression constraint for a given parameter, allowing to match an invalid parameter against the current route (which should never happen, since the route would not have matched in the first place)


src/Tempest/Router/src/functions.php
```diff
@@ -25,4 +25,24 @@ function uri(array|string|MethodReflector $action, mixed ...$params): string
             ...$params,
         );
     }
+
+    /**
+     * Checks whether the given controller action matches the current URI.
+     */
+    function is_uri(array|string|MethodReflector $action, mixed ...$params): bool
```
brendt, line 32: Shouldn't this be called `is_current_uri` then?

innocenzi: My thoughts was that `is_uri` as a function makes sense and is clear, while I'd like a more explicit method on the actual router object, hence the longer name. Both have comments anyway.

If the inconsistency is bothering here, I'd rather change `isCurrentUri` on the router to `isUri`, but if it's not an issue, I'd prefer keeping it that way

brendt: But isn't there a difference? When I read `is_uri`, I expect this function to return true for any string that is a URI. Which isn't the case, it only returns true when it's the _current_ uri>

CoolGoose: random guy dropping in 🧇 
But could that be named `is_matching_uri`  ? :P or `matches_uri` ? :P

innocenzi: > But isn't there a difference? When I read `is_uri`, I expect this function to return true for any string that is a URI. Which isn't the case, it only returns true when it's the _current_ uri?

Very good point actually, I don't know how I missed that. I like `matches_uri` as well but let's keep `is_current_uri` for consistency

brendt: Thank you random guy @CoolGoose for solving our naming issue :D

PR #1060, opened 2025-03-23
feat(database): allow overriding table names through model class attributes

innocenzi:
This pull request offers the ability to specify table names through attributes:

```php
#[TableName('aircraft')]
final class Aircraft implements DatabaseModel
{
    use IsDatabaseModel;
}
```

This was already possible through overriding the static `table` method, but this way is less verbose and feels more like Tempest.


innocenzi:
Yes!

PR #1062, opened 2025-03-23
chore: update mago to latest version

azjezz:


src/Tempest/Auth/src/Install/UserPermission.php
```diff
@@ -14,6 +14,5 @@ final class UserPermission implements DatabaseModel
     public function __construct(
         public User $user,
         public Permission $permission,
-    ) {
-    }
```
azjezz, line 18: Most of the diff is regarding this change, we can disalbe this using the new `inline_empty_constructor_braces` option.

src/Tempest/Log/tests/LogLevelTest.php
```diff
@@ -24,13 +24,13 @@ public static function levelsProvider(): array
...
-            [Level::Info, LogLevel::INFO],
-            [Level::Debug, LogLevel::DEBUG],
+            [Level::Alert,     LogLevel::ALERT],
+            [Level::Critical,  LogLevel::CRITICAL],
+            [Level::Error,     LogLevel::ERROR],
+            [Level::Warning,   LogLevel::WARNING],
+            [Level::Notice,    LogLevel::NOTICE],
+            [Level::Info,      LogLevel::INFO],
+            [Level::Debug,     LogLevel::DEBUG],
         ];
```
azjezz, line 34: This feature of aligning arrays as tables is enabled by default, not sure if we should disable it here or not ( controlled with `array_table_style_alignment` ).

src/Tempest/View/src/Stubs/view.stub.php
```diff
@@ -1 +1,3 @@
+<?php
 
+declare(strict_types=1);
```
azjezz, line 3: broke phpstan somehow? 😅 should this be reverted?

azjezz:
cc @innocenzi

src/Tempest/Auth/src/Install/UserPermission.php
innocenzi: This is actually a great change 👀

src/Tempest/Log/tests/LogLevelTest.php
innocenzi: I don't mind that. Let's keep it and see if we get issues over time

src/Tempest/View/src/Stubs/view.stub.php
innocenzi: Yeah, let's ignore stub files. I'll do the change

innocenzi:
This is awesome, thanks a lot @azjezz. I've temporarily disabled `strictness/require-strict-types` though, waiting for https://github.com/carthage-software/mago/issues/146.

PR #1071, opened 2025-03-24
feat(support): rename `map_array` to `map_iterable`

innocenzi:
`Arr\map` was changed to `Arr\map_array` in https://github.com/tempestphp/tempest-framework/pull/1045 in order to not conflict with `Tempest\map`, which was understandably frustrating when auto-importing with IDEs.

However, the new name feels like a cheap version of the existing `array_map`. Having `array_map` and `map_array` available feels quite awkward.

`map_iterable` as a name makes a clearer distinction with `array_map`.


brendt:
Great!

PR #1075, opened 2025-03-25
feat(database): store default sqlite database in internal storage

innocenzi:
This pull request updates the location of the default SQLite database to put it in the internal storage, instead of being nested into the vendor directory.


brendt:
Should be fine, we're still before beta.1 :)

innocenzi:
Let's go, then 👍

PR #1080, opened 2025-03-26
refactor(support): improve path utilities

innocenzi:
This pull request refactors our path utilities so they can be consistent in what they return. It also adds a few more path utilities in the support package.

## Modified

- `root_path` now returns an absolute path scoped to the root of the project, even if the parts in it are also an absolute path
- `src_path` now returns an absolute path scoped to the main project directory instead of a relative one
- `Path` now uses `ManipulatesStrings` and is refactored to use underlying functions

## Added

- `registered_namespace` allows creating a namespace from a file path, scoped in the existing project namespaces (the main one first, the others if not). If the given file path doesn't correspond to any of the current namespaces' paths, it will throw
- `is_relative_path` determines if the given path is a relative path
- `is_absolute_path` does the opposite
- `to_absolute_path` converts the given path to an absolute path, provided a current working directory to resolve relative paths from
- `to_relative_path` converts the given path to a path relative to the given directory
- `normalize` converts slashes to forward slashes, alongside other normalizations (no canonicalization). It's the equivalent of the `Path` object's contructor

## Cheatsheet

```php
// root is `/user/you/code/tempest-app`

root_path('file.txt'); // /user/you/code/tempest-app/file.txt
src_path('file.txt'); // /user/you/code/tempest-app/src/file.txt

to_relative_path(root_path(), 'file.txt'); // file.txt
to_relative_path(root_path(), src_path('file.txt')); // src/file.txt
to_relative_path(src_path(), root_path('file.txt')); // ../file.txt

to_absolute_path(root_path(), 'file.txt'); // /user/you/code/tempest-app/file.txt
to_absolute_path(src_path(), 'file.txt'); // /user/you/code/tempest-app/src/file.txt
```


innocenzi:
I realized I renamed `src_namespace` to `main_namespace`, because when doing so I thought it should be consistent with the "main namespace" concept in our Composer class. But now, it's inconsistent with `src_path`.

Any naming ideas here?

brendt:
> I realized I renamed src_namespace to main_namespace, because when doing so I thought it should be consistent with the "main namespace" concept in our Composer class. But now, it's inconsistent with src_path.

I think that's fine, keep it `src_namespace`

brendt:
I'm fine with these changes 👍 Feel free to merge. Could you check whether nothing breaks on tempest-docs though afterwards?

PR #1083, opened 2025-03-26
feat(router): support returning `string` and `array` from controllers

innocenzi:
This pull request allows returning string and arrays directly from controllers. Both the `ResponseSender` and `Response` classes already supported it already, but not the router's `createResponse`, so this PR changes that.

```php
#[Get('/')]
public function __invoke(): string
{
    return 'hello world'; // no longer throws!
}
```


brendt:
Makes sense!

Issue #701, opened 2024-11-08
Custom 404 and 500 pages


brendt:

innocenzi:
I just saw that this is in post-1.0. @brendt I'm thinking this is pretty essential, especially when we think of 404s alongside fallback routes. Can we try getting this before 1.0?

PR #1084, opened 2025-03-26
feat(router): introduce response processors

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/701

This pull request introduces the concept of "response processors", which is the same as view processors but for responses.

This allows intercepting responses right before they are sent. In contrast to HTTP middleware, they will be called if there was an error somewhere in the HTTP layer, since they are last in the router pipeline.

This notably allows for implementing custom error pages in a very simple way:

```php
use function Tempest\view;

final class ErrorStatusResponseProcessor implements ResponseProcessor
{
    public function process(Response $response): Response
    {
        if (! $response->status->isSuccessful()) {
            return $response->setBody(view('./error.view.php', status: $response->status));
        }

        return $response;
    }
}
```

The pull request now includes a built-in error page displayed for HTTP 500, 404, 403, and 401. Other error statuses will keep sending a blank page for now.

CleanShot 2025-03-26 at 23 25 53

## Questions

- I think we should not send the built-in error pages if the request doesn't expect JSON responses, but I don't believe we currently have utilities to detect that
- Should we display the built-in error page for more/less statuses?
- I think this makes the current `HttpProductionErrorHandler`'s `showErrorPage` useless?

I'd appreciate criticism about the approach taken here.

### DIFF
--- a/src/Tempest/Router/src/ErrorResponses/ErrorResponseProcessor.php
+++ b/src/Tempest/Router/src/ErrorResponses/ErrorResponseProcessor.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Tempest\Router\ErrorResponses;
+
+use Tempest\Http\Status;
+use Tempest\Router\Response;
+use Tempest\Router\ResponseProcessor;
+use Tempest\View\GenericView;
+use Tempest\View\View;
+
+final class ErrorResponseProcessor implements ResponseProcessor
+{
+    public function process(Response $response): Response
+    {
+        return match ($response->status) {
+            Status::INTERNAL_SERVER_ERROR, Status::NOT_FOUND, Status::FORBIDDEN, Status::UNAUTHORIZED => $response->setBody($this->renderErrorView($response->status)),
+            default => $response,
+        };
+    }
+
+    private function renderErrorView(Status $status): View
+    {
+        return new GenericView(__DIR__ . '/error.view.php', [
+            'css' => $this->getCss(),
+            'status' => $status->value,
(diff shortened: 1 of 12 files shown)


brendt:
> I think we should not send the built-in error pages if the request doesn't expect JSON responses, but I don't believe we currently have utilities to detect that

True, you can manually check the request's headers. This will need improvement

> Should we display the built-in error page for more/less statuses?

Which ones were you thinking of?

> I think this makes the current HttpProductionErrorHandler's showErrorPage useless?

True, feel free to remove!

brendt:
So I'm not against the feature, although I wonder _why_ we couldn't do it with middleware? We take the same approach with the console already: https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Console/src/Middleware/ResolveOrRescueMiddleware.php

innocenzi:
This is because we need to be able to return any response from `Router#dispatch` and still be able to catch them, even if the router is replaced. Currently, we also have specific error handling in the dispatch method, that wouldn't catch cases like no route matching, returning a `new NotFound` response.

Having a different step last in the pipeline ensures that _any_ error happening in the HTTP stack can get caught, even weird server errors in the framework's own code

brendt:
Edit, or maybe: have a generic error page for all codes > 400, but have special pages for the most common ones?

innocenzi:
> Edit, or maybe: have a generic error page for all codes > 400, but have special pages for the most common ones?

That sounds great, let's try that. I'll follow up with a `needsJson` PR later

PR #1093, opened 2025-03-28
feat(router): support getting raw body from requests

innocenzi:
This pull request adds support for getting the raw body from request classes:

```php
public function __invoke(Request $request)
{
    var_dump($request->raw);
    var_dump($request->body);
}
```

The `body` property is still the parsed body as before, and will still have a key/value pair if the raw body was a string. This is because of [this part](https://github.com/innocenzi/tempest-framework/blob/feat/raw-body/src/Tempest/Router/src/Input/StdinInputStream.php#L19), I'm not sure why we want to parse the body as an url-encoded body by default, it will confuse people down the road for sure 😁 cc @brendt


brendt:
> This is because of [this part](https://github.com/innocenzi/tempest-framework/blob/feat/raw-body/src/Tempest/Router/src/Input/StdinInputStream.php#L19), I'm not sure why we want to parse the body as an url-encoded body by default, it will confuse people down the road for sure

This whole part is because of a stupid thing in PHP: dots in body field names are automatically converted to underscored. That's because PHP used to automatically extract $_POST variables into real variables, in which case `.` wasn't an allowed character. So a form field called `author.name` would be converted to `author_name`, and be available via `$_POST['author_name']` and `$author_name`.

The problem is that for request validation, we need the dot notation. The easy was would be to convert underscores back to dots, but PHP actually converts not just dots but also a handful of other characters. On top of that, people might use actual underscores in their field names as well, so you cannot convert the underscored notation back to normal. There was an RFC to remove this behaviour, but it never went to vote: https://wiki.php.net/rfc/on_demand_name_mangling

What alternative would you propose? The most important part to me is that form data and JSON data can be automatically mapped and validated to objects as described in here: https://tempestphp.com/blog/request-objects-in-tempest

innocenzi:
> What alternative would you propose?

For now we can keep it this way honestly. But we might want to update this later to check if we _should_ parse the input stream before actually doing it. Not sure the heuristics would be though

src/Tempest/Router/src/Request.php
```diff
@@ -17,6 +17,10 @@ interface Request
         get;
     }
 
+    public ?string $raw {
```
brendt, line 20: Should we also allow streams for large blobs of content?

innocenzi: I don't think so, but if the need comes, we can improve later

brendt: True-ish. We'd need to change the interface, which would be a breaking change. But I'm fine merging this in for now.

PR #1098, opened 2025-03-29
feat(database): `Column` attribute for specifying the column name

erikaraujo:
Adds `Column` attribute to specify the column name (if different from the property, for some reason).

Usage:

```php
class Book
{
    #[Column('uuid')]
    public string $id;
}
```

### DIFF
--- a/src/Tempest/Database/src/Builder/FieldDefinition.php
+++ b/src/Tempest/Database/src/Builder/FieldDefinition.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace Tempest\Database\Builder;
 
 use Stringable;
+use Tempest\Database\Column;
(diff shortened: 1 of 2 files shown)


erikaraujo:
I originally went with `ColumnName` to be consistent with the existing `TableName` one - however, I think calling it just `Column` gives more flexibility in case we want to add more things to it in the future.

I even think we should rename `TableName` to just `Table` for the same reason. For example, maybe we want to add a `Schema` attribute in the future, and it would be weird to have `TableName` and `SchemaName. Or even if that particular Model is connected to a separate DB... ¯\\_(ツ)_/¯

EranNL:
I was wondering why this PR is closed. This would be a very nice addition

PR #1099, opened 2025-03-29
refactor(database): rename `TableName` to `Table`

erikaraujo:
The idea behind this is to give the `Table` attribute more flexibility - in case more things need to be added to it in the future.

A good example I can think of right now is `schema` - a concept very commonly used in SQL Server, for example.

Having one attribute per table thing can become terrible real quick.
For example, this:
```php
#[Table(name: 'tweets', schema: 'blogging')]
class Post
{
}
```

is a lot better than this:
```php
#[TableName('tweets')]
#[TableSchema('blogging')]
class Post
{
}
```


brendt:
Thanks! Could you update the docs as well? https://github.com/tempestphp/tempest-docs/blob/main/src/Web/Documentation/content/main/1-essentials/05-models.md#table-naming

PR #1100, opened 2025-03-29
refactor(router): improve static page generation console command output

innocenzi:
This pull request improves the output of `static:generate` and `static:clean` to be more consistent with the recent changes of Tempest's console aesthetic.

It also introduces the `StaticPageGenerated`, `StaticPageGenerationFailed` and `StaticPageRemoved` events.

CleanShot 2025-03-29 at 18 20 11

CleanShot 2025-03-29 at 18 19 42

CleanShot 2025-03-29 at 18 23 37

PR #1102, opened 2025-03-30
refactor(console): improve console output consistency

innocenzi:
As a follow-up to https://github.com/tempestphp/tempest-framework/pull/1100, this pull request is another pass at console output consistency. It consists of minor fixes, as well as the following changes:

## Fixed a bug with `$console->call`

Tempest has a singleton `ConsoleArgumentBag` that initially stores the `argv` arguments. When the CLI is booted, this is what determines which command to run. However, this argument bag is still used for subsequent `$console->call` runs, resulting in arguments being mixed up.

This is fixed by adding a `withoutArgumentBag` to `ExecuteConsoleCommand`.

```php
public function call(string|array $command, string|array $arguments = []): ExitCode|int
{
    return $this->executeConsoleCommand->withoutArgumentBag()($command, $arguments);
}
```

&nbsp;

## Migration output refresh

The migrations console message didn't use a header or key/value lines, so I added them.

Additionally, I aliased `migrate:up` to `migrate` for those like me who have some Laravel muscle memory. Similarly, I added the `--fresh` flag to `migrate:up`, so we can run `migrate --fresh` as an alias of `migrate:fresh`.

**Before**

CleanShot 2025-03-30 at 13 42 30

CleanShot 2025-03-30 at 13 42 51

**After**

CleanShot 2025-03-30 at 13 43 46

CleanShot 2025-03-30 at 13 44 05

&nbsp;

## Clean up of console errors

I slightly updated the rendering of console errors. The error line is easier to spot thanks to a red line number as well as three `<<<` instead of one `<`.

The exact file and line where the error appeared is the first thing displayed below the actual error message. Lastly, the file and lines are colored to highlight the important parts.

**Before**

CleanShot 2025-03-30 at 13 40 14

**After**

CleanShot 2025-03-30 at 13 40 28

&nbsp;

## Removal of top margin in console messages

In the last console refresh, I added a top margin above `error`, `success` and `info` console messages. This was nice to automatically separate them from the input components, but it also had a major drawback—that it was not possible to have two lines without spacing.

I removed this margin, helping fix the output of console errors above, as well as other parts of the console. The drawback is that some console commands, like the `make` ones, have to manually draw a new line before their output.

**Before**

CleanShot 2025-03-30 at 13 54 09

**After**

CleanShot 2025-03-30 at 13 54 43

### DIFF
--- a/src/Tempest/Console/src/Actions/ExecuteConsoleCommand.php
+++ b/src/Tempest/Console/src/Actions/ExecuteConsoleCommand.php
@@ -41,6 +41,16 @@ final readonly class ExecuteConsoleCommand
         ));
     }
 
+    public function withoutArgumentBag(): self
+    {
+        $bag = new ConsoleArgumentBag([
+            $this->argumentBag->getBinaryPath(),
+            $this->argumentBag->getCommandName(),
+        ]);
+
+        return new self($this->container, $this->consoleConfig, $bag, $this->resolveConsoleCommand);
+    }
+
     private function getCallable(array $commandMiddleware): ConsoleMiddlewareCallable
     {
         $callable = new ConsoleMiddlewareCallable(function (Invocation $invocation) {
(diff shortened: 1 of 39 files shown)


innocenzi:
The CI failure is coveralls being down until 19:00 CET.

PR #1103, opened 2025-03-30
feat(event-bus): add event bus testing utilities

innocenzi:
This pull request adds a new `EventBusTester` class, with an associated property in `IntegrationTest`. It allows faking the event bus, preventing handlers from being called, and enabling assertions on dispatches and listeners.

```php
$this->eventBus->fake(); // register `FakeEventBus` as a singleton for `EventBus`

$this->eventBus->assertDispatched('my-event');
$this->eventBus->assertDispatched(MyEvent::class);
$this->eventBus->assertDispatched(new MyEvent());

$this->eventBus->assertNotDispatched('my-event');
$this->eventBus->assertListeningTo('my-event');
```

`assertDispatched` has optional `count` and `callback` arguments, the former asserting a specific number of dispatches, and the later accepting a callback that should return `false` if some conditions don't match.

I'll add documentation once this is merged

### DIFF
--- a/src/Tempest/EventBus/src/Testing/EventBusTester.php
+++ b/src/Tempest/EventBus/src/Testing/EventBusTester.php
@@ -5,6 +5,8 @@ namespace Tempest\EventBus\Testing;
 use BackedEnum;
 use Closure;
 use PHPUnit\Framework\Assert;
+use PHPUnit\Framework\ExpectationFailedException;
+use PHPUnit\Framework\GeneratorNotSupportedException;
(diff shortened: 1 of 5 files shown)


innocenzi:
I'm thinking `preventEventHandling` could be a better name than `fake`. It would align with the Vite testing utils as well, but it might be less discoverable than "fake", which is wildly used

innocenzi:
Was just waiting for your opinion! Let's merge then

Issue #339, opened 2024-08-16
Multiple database connections


brendt:
How about we change our database config like so?

```php
return new DatabaseConfig(
    connections: [
        'main' => new SQLiteConnection(
            path: 'database.sqlite'
        ),
        'backup' => new MysqlConnection(
            // …
        ),
        // …
    ]
);
```

We'd change the name of the interface `Driver` to `Connection`, I think that makes more sense anyway.

By default, the first registered connection is used. However, users can manually switch their connection:

```php
(new Query($sql))->connection('backup')->execute();
```

We'd need to consider whether we want the `DatabaseModel` trait to allow for connection switching on per-model basis. That's up for debate, although I lean more towards no.

brendt:
@aidan-casey feel free to share your thoughts

brendt:
Fixed with #1120!

PR #1105, opened 2025-04-01
feat(core): introduce tagged configurations

innocenzi:
## Tagged configurations

This pull request adds support for tagged configurations.

This is a new concept that attempts to solve the "multiple X" issue at the foundation level. This could be used, for instance, to support multiple databases by simply defining tagged configurations, and injecting a tagged instance of `Database`:

```php
// database.backup.config.php
return new SQLiteConfig(
    path: __DIR__ . '/backup.sqlite',
    tag: 'backup',
);

// database.config.php
return new SQLiteConfig(
    path: __DIR__ . '/main.sqlite',
);

// some class
public function __construct(
    private readonly Database $mainDatabase,
    #[Tag('backup')]
    private readonly Database $backupDatabase,
) {}
```

&nbsp;

## Further considerations

Right now, this can only work because `DynamicInitializer`s have been modified to accept a `$tag`. An initializer can determine if it will initialize a class as usual, but it can also resolve dependencies using the given tag, allowing to pass it down to the chain of dependencies.

This is quite verbose, since it would require an initializer for every class or interface for a feature that would implement this concept.

To avoid that, I added `#[AllowDynamicTags]` and `#[ForwardTag]`. The former prevents a class from not being resolved when it has a tag but no initializer and is not registered yet. The latter resolves the marked parameter using the same tag that was used for the parent class.

This is the least verbose pattern I could come up with.

I'm wondering though, if normal `Initializer` should also receive the tag. This would allow not using a `DynamicInitializer` for the `Database`, for instance. However, the API would become more confusing, with `#[Singleton]` accepting a tag.

&nbsp;

## Tags as enums

I have taken the opportunity to support enums in the `#[Tag]` attribute and all container methods accepting a `$tag` parameter, as this is usually better than specifying plain strings, which are prone to typos and not easily refactored.

```php
$this->container->get(Service::class, tag: ExampleEnum::FOO);
```

&nbsp;

## Proof of concept

I included a proof of concept of the internal usage of tagged configurations by implementing it on Vite.

Currently, it's only possible to have one Vite configuration in a Tempest project—this means a you can't build multiple manifests or have multiple development servers while using the back-end integrations.

This is solved by specifying the `tag` property in `ViteConfig`, as well as replacing the default `bridgeConfigFile`, `buildDirectory` and `manifest`:

```php
// vite.backoffice.config.php
return new ViteConfig(
    tag: 'backoffice',
    bridgeFileName: 'vite-tempest-backoffice',
    manifest: 'backoffice/manifest.json',
    buildDirectory: 'backoffice/build',
    entrypoints: [
        'src/BackOffice/main.ts',
    ],
);

// in a view
<x-vite-tags config="backoffice" />
```

Related to https://github.com/tempestphp/tempest-framework/issues/339


brendt:
I really like the idea, but would like to brainstorm some more. Here are my thoughts:

1. Tagging a config file shouldn't be done with a magic variable on the config itself. What about some kind of helper function that wraps the config in a proxy?

```php
// sqlite.config.php

return tag(new SqliteConfig(), 'sqlite');
```

Since config objects are registered in the container via the `->config` method, we could easily check if we're dealing with eg. `TaggedObject`, and unwrap it there.

2. If we have this `tag` wrapper, I'm not sure if we'd still need the `DynamicInitializer` changes? Maybe I'm missing something
3. Tags as enums: makes sense 👍

innocenzi:
I agree with the variable having to be specified in the config not being ideal. However, it's not that bad: for configurations like the database config, we could have a different name than `tag` to be more specific about what it is.

To be fair, I don't like the wrapper function. It's not convenient to type nor pretty:

```php
return tag(new SQLiteConfig(
    path: __DIR__ . '/backup.sqlite',
), 'backup');
```

Also, no, this unfortunately not fix the `DynamicInitializer` changes. We definitely need the `$tag` to be accessible in `initialize` to instantiate the object's dependencies that rely on the tag:

```php
final class DatabaseInitializer implements Initializer
{
    // ...

    #[Singleton]
    public function initialize(ClassReflector $class, Container $container, ?string $tag = null): Database
    {
        return new GenericDatabase(
            $container->get(Connection::class, $tag),
            $container->get(TransactionManager::class, $tag),
        );
    }
}
```

Even with `#[AllowDynamicTags]` and `#[ForwardTag]`, there are cases like the above, where an initializer is needed and it needs to forward the tag it is being initialized with

innocenzi:
@brendt the last commit is a proof of concept of the database working with multiple tagged configs. I feel like the API is still confusing but at least it's not verbose anymore because dynamic initializers are no longer needed: the current tag is injected into a property of the initializer, thanks to the `#[CurrentTag]` attribute. For this to work the initializer also has to have `#[AllowDyanmicTags]` though...

Overall I think the pattern (tagged configs and passthrough tags) is good but the API/code should be cleaned. Would love your help in this regard

brendt:
Ah, we could add an interface? `Taggable` with a `public string|array|UnitEnum $tags { get; }` property? Then anything can be made taggable, but it's more explicit than relying on a random property with attribute?

I don't oppose the changes to `DynamicInitializer`, I just want to make sure I understand the need for them. That's what confuses me: I can't find the added `$tag` parameter being used in any of the dynamic initializers in your PR? You mention `DatabaseInitializer`, but that's not a dynamic initializer? Or is that just WIP?

innocenzi:
Hey @brendt, this is ready for review. If possible I'd like you to play with the feature, see how it stands conceptually, criticize the API—because this would be a fairly important part of future Tempest APIs.

The important parts:
- You can implement `TaggedConfig` on a configuration class, effectively allowing multiple instances of this configuration to coexist.
- You can add `#[AllowDynamicTags]` to any class and their initializer (if any). In this situation, a class can be resolved with any tag.
- You can add `#[ForwardTag]` to a constructor parameter to resolve this parameter using the same tag that was used to resolve the owner class.
- You can add `#[CurrentTag]` to a class property to inject it with the tag that was used to resolve the owner class. This is what allows us to keep using normal initializers to initialize dependencies with `#[AllowDynamicTags]`.

brendt:
Ok but I still don't understand this part:

```
You can add #[AllowDynamicTags] to any class and their initializer (if any). In this situation, a class can be resolved with any tag.
You can add #[ForwardTag] to a constructor parameter to resolve this parameter using the same tag that was used to resolve the owner class.
You can add #[CurrentTag] to a class property to inject it with the tag that was used to resolve the owner class. This is what allows us to keep using normal initializers to initialize dependencies with #[AllowDynamicTags].
```

Let me write up another PR to illustrate what I mean though, because I feel like we're going in circles.

brendt:
Here's my PR: https://github.com/tempestphp/tempest-framework/pull/1120/files

What does your implementation do with the attributes that can't be solved with the interface?

innocenzi:
> What does your implementation do with the attributes that can't be solved with the interface?

It's completely unrelated 😅 Your `HasTag` interface is just a different abstraction to the `TaggedConfig` interface.

The rest of the implementation is less about tagged configurations and more about giving the framework the ability to initialize objects _and their own dependencies_ with the same tag.

I recommend you try this PR out locally: create multiple database configs and inject them in a class with their own tag

innocenzi:
Closing this in favor of https://github.com/tempestphp/tempest-framework/pull/1120

PR #1113, opened 2025-04-03
feat(view): improve boolean attributes

brendt:


brendt:
Merged this in #1115 and will continue work there

PR #1115, opened 2025-04-04
refactor(view): implement custom html parser

brendt:
TODO

- [x] Clean up TOKEN_MAPPING
- [ ] Mark the view engine as experimental
- [x] Check todos
- [ ] ~Cleanup `RawConditionalAttribute`: should there be an interface? Should we refactor more to raw attributes?~
- [x] Move lexer, parser, and compiler into the same namespace
- [ ] ~Maybe refactor `Token` to an interface and have multiple implementations? (will make it easier to merge elements and tokens in the future as well)~ should be done later
- [x] Profile the lexer and parser to identify performance improvements


brendt:
Some preliminary benchmarks on the homepage of tempestphp.com during local development (view cache disabled):

```
// PHP DOM
Time per request:       205.783 [ms] (mean)
Time per request:       205.783 [ms] (mean, across all concurrent requests)

// Tempest's parser
Time per request:       222.260 [ms] (mean)
Time per request:       222.260 [ms] (mean, across all concurrent requests)
```

It's a bit slower, though I'm optimistic: right now we simply replaced the DOM parsing part, but there's a lot of potential to optimize the whole engine now that we control the lexer and parser as well.

brendt:
These were done on another laptop, so don't mind the absolute values. 

```
// DOM
Time per request:       335.760 [ms] (mean)
Time per request:       335.760 [ms] (mean, across all concurrent requests)

// Tempest before optimizations
Time per request:       382.743 [ms] (mean)
Time per request:       382.743 [ms] (mean, across all concurrent requests)

// Tempest with optimizations
Time per request:       373.134 [ms] (mean)
Time per request:       373.134 [ms] (mean, across all concurrent requests)
```

PR #1127, opened 2025-04-08
fix(validation): more lenient scalar validation

### LINKED ISSUE #1110: Allow Request objects to be properly hydrated from request query parameters
### Description

Right now the Request objects work as intended when the values are hydrated from a request body, because they can have types, but if we wanted to create a Request object to be hydrated from the request query parameters we couldn't do it because the validation would fail, as currently the query parameters in GenericRequest objects are always strings.

This is because the way the GenericRequest object is mapped from PSR requests, in **PsrRequestToGenericRequestMapper**, line **46**:

```php
parse_str($from->getUri()->getQuery(), $query);
```

### Plan to solve it

There are two major ways to solve this:

#### Cast the query parameters after the router starts dispatching the request
We would need to change the previous line to this:

```php
$query = $from->getQueryParams();
```

And then on the **GenericRouter**, on line **61**:
```php
$request = $this->queryParametersCaster->cast($request, $matchedRoute);
```

Then create a service like this:

```php
final readonly class QueryParametersCaster
{
    public function cast(ServerRequestInterface $request, MatchedRoute $matchedRoute): Request
    {
        $castQueryParameters = $this->castQueryParameters(
            $request->getQueryParams(),
            $this->getTargetRequestClass($matchedRoute)
        );

        return new ServerRequest(
            uploadedFiles: $request->getUploadedFiles(),
            uri: $request->getUri(),
            method: $request->getMethod(),
            headers: $request->getHeaders(),
            cookieParams: $request->getCookieParams(),
            queryParams: $castQueryParameters,
            parsedBody: $request->getParsedBody(),
        );
    }

    /**
     * @return class-string<Request>
     */
    private function getTargetRequestClass(MatchedRoute $matchedRoute): string
    {
        foreach ($matchedRoute->route->handler->getParameters() as $parameter) {
            if ($parameter->getType()->matches(Request::class)) {
                return $parameter->getType()->getName();
            }
        }

        return GenericRequest::class;
    }

    private function castQueryParameters(array $queryParameters, string $targetRequestClass): array
    {
        $queryParameters = arr($queryParameters);
        $classReflector = new ClassReflector($targetRequestClass);

        foreach ($classReflector->getPublicProperties() as $property) {
            if ($property->hasAttribute(SkipValidation::class)) {
                continue;
            }

            if (!$queryParameters->has($property->getName()) && $property->hasDefaultValue()) {
                continue;
            }

            $value = $queryParameters->get($property->getName());

            $queryParameters = $queryParameters->put(
                $property->getName(),
                $this->getPropertyValue($property, $value)
            );
        }

        return $queryParameters->toArray();
    }

    private function getPropertyValue(PropertyReflector $property, mixed $value): mixed
    {
        if (is_null($value)) {
            return null;
        }

        return match ($property->getType()->getName()) {
            'bool' => filter_var($value, filter: FILTER_VALIDATE_BOOLEAN, options: FILTER_NULL_ON_FAILURE) ?? $value,
            'int' => filter_var($value, filter: FILTER_VALIDATE_INT, options: FILTER_NULL_ON_FAILURE) ?? $value,
            'float' => filter_var($value, filter: FILTER_VALIDATE_FLOAT, options: FILTER_NULL_ON_FAILURE) ?? $value,
            default => $value,
        };
    }
}
```

#### Cast the query parameters on the last step, before instantiating the custom request object

We would need to change **RequestToObjectMapper** in line **30** to do something like this:

```php
$castQueryParameters = $this->queryParametersCaster->cast($from->query, $to);
```

And on line **34**:

```php
$data = [
    ...[
        'method' => $from->method,
        'uri' => $from->uri,
        'body' => $from->body,
        'headers' => $from->headers,
        'path' => $from->path,
        'query' => $castQueryParameters,
        'files' => $from->files,
        'cookies' => $from->cookies,
    ],
    ...$from->files,
    ...$castQueryParameters,
    ...$data,
];
```

The caster service would be slightly different in this case because we already know the target Request class beforehand:

```php
final readonly class QueryParametersCaster
{
    /*
     * @param class-string<Request>
     */
    public function cast(ServerRequestInterface $request, string $targetRequestClass): Request
    {
        $castQueryParameters = $this->castQueryParameters(
            $request->getQueryParams(),
            $targetRequestClass
        );

        return new ServerRequest(
            uploadedFiles: $request->getUploadedFiles(),
            uri: $request->getUri(),
            method: $request->getMethod(),
            headers: $request->getHeaders(),
            cookieParams: $request->getCookieParams(),
            queryParams: $castQueryParameters,
            parsedBody: $request->getParsedBody(),
        );
    }

    private function castQueryParameters(array $queryParameters, string $targetRequestClass): array
    {
        $queryParameters = arr($queryParameters);
        $classReflector = new ClassReflector($targetRequestClass);

        foreach ($classReflector->getPublicProperties() as $property) {
            if ($property->hasAttribute(SkipValidation::class)) {
                continue;
            }

            if (!$queryParameters->has($property->getName()) && $property->hasDefaultValue()) {
                continue;
            }

            $value = $queryParameters->get($property->getName());

            $queryParameters = $queryParameters->put(
                $property->getName(),
                $this->getPropertyValue($property, $value)
            );
        }

        return $queryParameters->toArray();
    }

    private function getPropertyValue(PropertyReflector $property, mixed $value): mixed
    {
        if (is_null($value)) {
            return null;
        }

        return match ($property->getType()->getName()) {
            'bool' => filter_var($value, filter: FILTER_VALIDATE_BOOLEAN, options: FILTER_NULL_ON_FAILURE) ?? $value,
            'int' => filter_var($value, filter: FILTER_VALIDATE_INT, options: FILTER_NULL_ON_FAILURE) ?? $value,
            'float' => filter_var($value, filter: FILTER_VALIDATE_FLOAT, options: FILTER_NULL_ON_FAILURE) ?? $value,
            default => $value,
        };
    }
}
```

I think the first one is a cleaner solution because it fixes the `PsrRequestToGenericRequestMapper` to create a `GenericRequest` from the PSR request query parameters instead of parsing them again from the uri, but involves touching the `GenericRouter`. The second one, on the other hand, it would only alter the `RequestToObjectMapper`.

### Benefits

We could do this:

```php
final readonly class Controller
{
    #[Get('/test')]
    public function __invoke(MyRequest $request): Ok
    {
        return new Ok();
    }
}

final readonly class MyRequest
{
    #[Get('/test')]
    public function __construct(
        #[Between(18, 100)]
        public int $age;
    ) {}
}
```

Right now, a call to `/test?age=20` would fail, because the age parameter inside the query array in the `GenericRequest` is a string. When the `RequestToObjectMapper` tries to map it into a MyRequest instance, it throws a `ValidationError`.

brendt:
Closes #1110

### DIFF
--- a/src/Tempest/Validation/src/Rules/IsFloat.php
+++ b/src/Tempest/Validation/src/Rules/IsFloat.php
@@ -20,15 +20,12 @@ final readonly class IsFloat implements Rule
             return true;
         }
 
-        if ($value === null || $value === false || $value === '') {
+        if ($value === null || $value === false || $value === '' || $value === []) {
             return false;
         }
 
-        if (floatval($value) == $value) {
-            return true;
-        }
-
-        return is_float($value);
+        // @mago-expect strictness/require-identity-comparison
+        return is_float($value) || floatval($value) == $value;
     }
 
     public function message(): string
--- a/src/Tempest/Validation/src/Rules/IsInteger.php
+++ b/src/Tempest/Validation/src/Rules/IsInteger.php
@@ -20,15 +20,12 @@ final readonly class IsInteger implements Rule
             return true;
         }
 
-        if ($value === null || $value === false || $value === '') {
+        if ($value === null || $value === false || $value === '' || $value === []) {
             return false;
         }
 
-        if (intval($value) == $value) {
-            return true;
-        }
-
-        return is_int($value);
+        // @mago-expect strictness/require-identity-comparison
+        return is_int($value) || intval($value) == $value;
     }
 
     public function message(): string
--- a/src/Tempest/Validation/tests/Rules/IsFloatTest.php
+++ b/src/Tempest/Validation/tests/Rules/IsFloatTest.php
@@ -23,5 +23,6 @@ final class IsFloatTest extends TestCase
         $this->assertFalse($rule->isValid(''));
         $this->assertFalse($rule->isValid(null));
         $this->assertFalse($rule->isValid(false));
+        $this->assertFalse($rule->isValid([]));
     }
 }
--- a/src/Tempest/Validation/tests/Rules/IsIntegerTest.php
+++ b/src/Tempest/Validation/tests/Rules/IsIntegerTest.php
@@ -22,5 +22,6 @@ final class IsIntegerTest extends TestCase
         $this->assertFalse($rule->isValid(''));
         $this->assertFalse($rule->isValid(null));
         $this->assertFalse($rule->isValid(false));
+        $this->assertFalse($rule->isValid([]));
     }
 }


brendt:
Failing tests aren't related to this PR. Discussing it on Discord, but I'm going ahead and merging

PR #1135, opened 2025-04-09
Add missing DoNotDiscover attribute to Console stubs

### LINKED ISSUE #1134: Stub classes are included in discovery locations
### Tempest version

1.0-alpha.6

### PHP version

8.4

### Operating system

macOS

### Description

`*Stub` classes are being included in Discovery locations despite the fact that they are not "real" classes.

I noticed this when running a console command, but this likely applies to other discovery locations as well.

Image

I thought of a couple options to resolve this:

1. Add the `DoNotDiscover` attribute to the class stubs, and then use `ClassManipulator::removeClassAttribute()` during the generation process. This could potentially be an enhancement where an option is added to the `make:` command to maintain the class as `DoNotDiscover`.
2. Don't add the `ConsoleCommand` attribute until the generation process. Similar to the point above, this can be done with `ClassManipulator::addMethodAttribute()`. This is more specific to the Console stubs, but it could be extrapolated elsewhere as needed.
3. Modify the `LoadDiscoveryClasses::shouldSkipDiscoveryForClass()` method to check whether the found class is a stub. This could be done by looking at the namespace, or possibly by the class name itself, since most (all?) have `*Stub` as the class name.

### Steps to reproduce

1. Create a new Tempest application
2. Run `./tempest`
3. See `dummy-command-slug` as one of the available commands.

JPry:
Fixes #1134.

1. Adds the `DoNotDiscover` attribute to Console stubs
2. Updates the Console `make:` commands to remove the attribute when generating the class
3. Updates the integration tests to reflect the expected behavior

### DIFF
--- a/src/Tempest/Console/src/Commands/MakeCommandCommand.php
+++ b/src/Tempest/Console/src/Commands/MakeCommandCommand.php
@@ -8,6 +8,7 @@ use Tempest\Console\ConsoleArgument;
 use Tempest\Console\ConsoleCommand;
 use Tempest\Console\Stubs\CommandStub;
 use Tempest\Core\PublishesFiles;
+use Tempest\Generation\ClassManipulator;
 use Tempest\Generation\DataObjects\StubFile;
 
 use function Tempest\Support\str;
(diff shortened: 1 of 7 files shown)


brendt:
@gturpin-dev I don't really understand what's happening here. Shouldn't `DoNotDiscover` be removed automatically when installed?

brendt:
Did some digging, apparently we don't remove `DoNotDiscover`. I'm not sure if that's wanted behavior, but I'll follow up with @gturpin-dev on this.

I'm closing this one in favor of #1136

PR #1139, opened 2025-04-10
fix(console): select default option in ask component

### LINKED ISSUE #897: alt+enter in console doesn't work on Mac
Followup after #888: alt+enter to select the default option in a choice component doesn't work on Mac, instead it's registered as a simple enter.

brendt:
Closes #897

TODO:

- [ ] ~Alt+enter in multiline. Should probably be replaced with pressing `esc` to "exit" edit mode and be able to press enter to submit~ #1140

### DIFF
--- a/src/Tempest/Console/src/Components/Concerns/OpensInEditor.php
+++ b/src/Tempest/Console/src/Components/Concerns/OpensInEditor.php
@@ -22,9 +22,9 @@ trait OpensInEditor
         return is_subclass_of(static::class, CanOpenInEditor::class) && ((bool) $this->getEditorCommand());
     }
 
-    private function getEditorCommand(): string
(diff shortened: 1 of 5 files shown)


brendt:
@innocenzi I did find one more reference to alt_enter in `HasTextInputRenderer`, but when I try to run a multiline ask component, I apparently need an `EDITOR` env variable. How do I force it to open inline?

innocenzi:
`alt+enter` in multiline mode confirms the output, so you can use `enter` to go to a new line. Does that not work for you as well?

I'm not sure how `EDITOR` is related, it's another shortcut entirely

https://github.com/user-attachments/assets/9673517f-91e7-428a-b637-351e7890bfed

PR #1149, opened 2025-04-12
feat(storage): add storage component

innocenzi:
This pull request adds a new storage component, as a wrapper around [Flysystem](https://flysystem.thephpleague.com/docs/). Most adapters are included, excluding path-prefixing, AsyncAws S3, MongoDB GridFS, SFTP (V2) and WebDAV. These can be added later.

Currently, a single storage at a time is supported. Multiple storages depend on https://github.com/tempestphp/tempest-framework/pull/1120 or https://github.com/tempestphp/tempest-framework/pull/1105.

The implementation works by providing a `Tempest\Storage\Storage` interface, implemented by `Tempest\Storage\GenericStorage`. It takes a `StorageConfig` as a parameter, from which a `FlysystemAdapter` is created.

## Usage

By default, the framework configures the local storage to be in `vendor/.tempest/storage`. Users should override this by creating a `storage.config.php` file:

```php
// storage.config.php
return new LocalStorageConfig(
  path: __DIR__ . '/storage',
);
```

The `Storage` interface can then be injected and used:

```php
$storage = $this->container->get(Storage::class);
$storage->write('file.txt', $content);
```

## Testing

The `IntegrationTestCase` has a `storage` property that provides testing utilities.

```php
$this->storage->fake();

// Call some app code...

$this->storage->assertFileExists('file.txt');
```

When tagged configurations are implemented, a tag can be passed to the `fake` method to fake the specified storage.


aidan-casey:
@innocenzi - Sweet! Should we not have the config accept an array of drivers (named, maybe?) that can be used?

aidan-casey:
I was thinking something like this:

```php
return new StorageConfig(
    locations: [
        'default' => new LocalStorageConfig(...),
        'aws' => new S3StorageConfig(...),
    ]
);
```

Then you are able to utilize multiple?

innocenzi:
That would work, but I'm looking into a more Tempest-y way to achieve this, take a look: https://github.com/tempestphp/tempest-framework/pull/1105

Basically, we'd use tags instead of arrays:

```php
// no tag—default
return new LocalStorageConfig(
  path: __DIR__ . '/storage',
);
```

```php
return new S3StorageConfig(
  tag: Disk::USER_AVATARS, // <-- tag!
  bucket: env('S3_BUCKET'),
  region: env('S3_REGION'),
  accessKeyId: env('S3_ACCESS_KEY_ID'),
  secretAccessKey: env('S3_SECRET_ACESS_KEY'),
);
```

```php
public function __construct(
  private readonly Storage $main, // no tag, main storage
  #[Tag(Disk::USER_AVATARS)]
  private readonly Storage $userAvatarStorage, // tag, use named storage
) {}
```

brendt:
This is great, I've got nothing to add honestly. I'll merge! @innocenzi would also write the docs for this?

innocenzi:
I will!

PR #1150, opened 2025-04-14
refactor(filesystem): move from package to functions

innocenzi:
This pull request retires the `tempest/filesystem` package in favor of namespaced functions in `tempest/support`. There are multiple (minor) reasons why I think this is a nice change:

- It's slightly more convenient to use one-off functions than to instantiate a class to access one method:
```php
// Before
new LocalFilesystem()->ensureDirectoryExists();

// After
Filesystem\ensure_directory_exists();
```

- Most Tempest packages already depend on `tempest/support`. When I tried using `LocalFilesystem` in the new `tempest/storage` package, I didn't have issues locally, but CI made me realize I had a new dependency on `tempest/filesystem`.

- This is more maintainable than to update the `Filesystem` interface. We can simply add functions in the namespace, and that's it.

- It is a bit more discoverable for userland. Typing `file` or `dir` will make an IDE bring up related functions. It's also less confusing than to have both a `Storage` package and a `Filesystem` one.

Overall, though, this pull request is more about convenience and consistency with other utilities than these specific reasons.

Note that a lot of the functions implemented here come from the PSL.


aidan-casey:
I'm not sure I love the seemingly procedural shift in this PR, but that's just my own preference. 😛

innocenzi:
It's not procedural, it's functional :p 

And no worries, I don't plan on making _everything_ functional. Just what makes sense. What prompted me to make this PR is actually me using this component multiple times in tests, and wishing it was just a function instead. 😅 

That being said, if you could tell your Windows CI to please cooperate... 😡

src/Tempest/Support/tests/Filesystem/UnixFunctionsTest.php
```diff
@@ -11,17 +11,17 @@
...
-    private string $fixtures;
+    private string $fixtures = __DIR__ . '/Fixtures';
 
     protected function setUp(): void
     {
         parent::setUp();
 
-        $this->fixtures = normalize(__DIR__, '/Fixtures/', uniqid('tempest', more_entropy: true));
+        if (PHP_OS_FAMILY === 'Windows') {
+            $this->markTestSkipped('Irrelevant on Windows.');
```
erikaraujo, line 23: Why is the test irrelevant on windows?
I mean, I get the name of the test, but... shouldn't this work on Windows as well?

innocenzi:
@erikaraujo You'd use the [Storage](https://github.com/tempestphp/tempest-framework/pull/1149) component instead

This PR is purely related to the local filesystem, primarily for within the framework itself, but available as a polished functional API for Tempest developers as well

aidan-casey:
Ack, beat by @innocenzi again.

innocenzi:
They don't always have an object wrapper (regex utils for instance don't). Even if they did, it's a matter of whether it makes sense or not. Adding an object wrapper just for the sake of it is not necessarily useful. I'd be happy to add a `Filesystem` class if it had another purpose than to proxy the functional API (eg. by scoping the path).

I get that the functional approach is not something we're use to too much in PHP, but it just works great. Why instantiate a whole class when you need to call just one function? It's also clean to use. I personally am quite a fan of this syntax:

```php
Filesystem\ensure_directory_exists();
```

brendt:
Looking at the current `Filesystem` interface, I think Enzo does make a good argument here. Nowhere in the interface are there chainable methods, meaning the "interface" is merely a collection of functions. 

This functional approach is also something I'm not super used to, but it also feels weird having to construct or inject a class that doesn't really do anything but serve as a gateway to access those functions.

aidan-casey:
No one ever made an argument for non-functional because of method chaining or better syntax. It'd be one thing if this was just syntax sugar over an object like Laravel's helpers, but this completely eliminates any ability to swap out implementations.

By way of example, one thing we looked at (and deferred) was a test filesystem implementation that wrapped in vfsStream under the hood. This would be immensely useful to testing filesystem operations in unit tests. You cannot easily implement that because you are totally tied to the functions present here.

Now, personally, I hate working with functions for this stuff anyway, the string and array helpers drive me nuts. The total inability to swap out implementations is a bigger deal though.

innocenzi:
Alright, this PR was made with the assumption that we are not ever going to change the underlying implementation of file operations.

How big of a deal is it actually, though? I don't know much about `vfsStream`, a quick search tells me that it's a library that implements a virtual filesystem, so we don't have to touch the actual one in tests.

Can we actually use such a thing? What do we do when we need to deal with the actual filesystem because the underlying code don't use the `Filesystem` implementation that uses this library?

Either way, if this ever becomes a thing, we can add back the `Filesystem` package, and re-implement the `LocalFilesystem` using `Tempest\Support\Filesystem` functions under the hood. 

By the way, we weren't using `LocalFilesystem` through the container, and I think in some places we didn't have the container available to swap the implementation too.

aidan-casey:
> I don't know much about vfsStream, a quick search tells me that it's a library that implements a virtual filesystem, so we don't have to touch the actual one in tests.

That is correct and it's super convenient because it doesn't make a mess of your filesystem when unit testing file-related operations. Some of that stuff can get pretty hairy especially around file permissions. We [use it](https://github.com/tempestphp/tempest-framework/blob/main/src/Tempest/Filesystem/tests/LocalFilesystemTest.php) in our filesystem tests.

> What do we do when we need to deal with the actual filesystem because the underlying code don't use the Filesystem implementation that uses this library?

Can you think of an example of that? I can only think of testing an external third-party package and that feels outside of the scope of our test cases.

> By the way, we weren't using LocalFilesystem through the container, and I think in some places we didn't have the container available to swap the implementation too.

I think there were a lot of places where we weren't using a filesystem object at all because we didn't have one for the longest time. Personally, I'd rather see everything resolve from the container, whether through a helper function or not.

innocenzi:
> Can you think of an example of that?

For instance, the `DependencyInstaller` class calls package managers under the hood, which will use the real filesystem. That might be the only one though.

> Personally, I'd rather see everything resolve from the container, whether through a helper function or not.

What about places where we don't use `FrameworkIntegrationTestCase`?

So yeah, if we ever go this route, potentially we can add back the `Filesystem` package, and the `LocalFilesystem` can be implemented on top of the primitive functions in `Tempest\Support\Filesystem`, which can be very useful to userland too.

aidan-casey:
> For instance, the DependencyInstaller class calls package managers under the hood, which will use the real filesystem. That might be the only one though.

A number of ways to slice that. I'd probably abstract the package manager to its own interface, but it's more of an integration test anyway.

> What about places where we don't use FrameworkIntegrationTestCase?

It doesn't matter because we utilize dependency injection?

brendt:
Hmm I can follow both sides of the argument 🤔

The way we currently use FS throughout the framework definitely isn't the best. For example: we _should_ inject FS everywhere if we decide to keep the class implementation. It's just one of these things that we haven't gotten to yet (and IMO don't need for a 1.0 release, this can wait until 2.0 if it requires breaking changes).

That being said: we just added a storage component with swappable drivers. Why do we then need another FS implementation that achieves the same? Following that thought, there's something to say for a "primitive helper" implementation on top of PHP's one, as an enhancement. It's what we do with string and array and other helpers, and they are really useful, even if they can't be replaced. It's not like we can replace or mock PHP's built-in functions either, so it also shouldn't be a design goal for our helper functions.

So, the main question: what Aidan's talking about is useful, but I think we have the storage component for that — is that correct? My vision for this FS helper is that is a wrapper around primitive PHP functionality. If those assumptions are correct, then it means we can go with the functional approach (and add a OO wrapper later if needed, just like our array and string implementations)

innocenzi:
@brendt I see it that way too. However, we can't really use the storage component internally, or it should be tagged with a tag that can't conflict with one in userland

aidan-casey:
@brendt we talked about separating storage all the way back in that original filesystem PR. The distinction we made was:

__Filesystem__: interaction directly with the local filesystem. Aware of how data is organized, structured, and permissioned.

__Storage__: a location unaware way to store application files such as uploads. Unaware of how data is organized, structured, and permissioned.

Storage would not allow for swapping out the local filesystem in a test case if that is what you are implying.

brendt:
> The distinction we made was

Yeah I agree, and it also aligns with how I envisioned the FS helpers: a wrapper around PHP's built-in functions, just like our other primitive helpers. 

For me, the "switching during test" happens on another layer: by eg. mocking config with other file paths. That's already what's happening today, so I feel like this PR is in line with how Tempest is handling it. 

_That being said_ (again): I wouldn't oppose another approach where we inject `Filesystem` everywhere, with a mockable implementation for testing, … But that would imply a very big refactor within the whole of Tempest. We don't have that now. Does that mean we should remove `Filesystem` and add it back later (if we ever do), or are we going to commit doing that refactor? 

To me, that's the deciding factor for this PR: are we going to refactor all FS interactions within Tempest? Are we going to commit to that? If the answer is yes, then my suggestion would be to still refactor to functions, but keep the object implementation as well (using the functions internally). If no one is willing to make a commitment here, then I say we go with the simpler function-based approach until someone's able to pick it up again.

innocenzi:
> Does that mean we should remove Filesystem and add it back later (if we ever do), or are we going to commit doing that refactor?

> If the answer is yes, then my suggestion would be to still refactor to functions, but keep the object implementation as well (using the functions internally).

It seems that in both case, you're also leaning towards the primitive helpers, which can then be wrapped in the `Filesystem` package, which we can really easily add back later.

brendt:
@innocenzi true, the only question I have is whether we should remove the FS interface and potentially add it back later, or just keep it as is.

PR #1153, opened 2025-04-18
feat(view): attribute precedence

### LINKED ISSUE #1157: foreach on a tag doesn't render `href`
This doesn't work:

```html
<a :foreach="$stories as $story" :href="$story->uri">{{ $story->date->format('d/m/Y') }} — {{ $story->title }}</a>
```

But this does:

```html
    <x-template :foreach="$stories as $story">
        <a :href="$story->uri">{{ $story->date->format('d/m/Y') }} — {{ $story->title }}</a>
    </x-template>
```

erikaraujo:
When looping in a tag, like `<div :foreach="$iterable as $value"`, `$value` cannot be used within the same tag in an attribute.

Example:

```html
<a :foreach="$links as $link" :href="$link->uri">
    {{ $link->label }}
</a>
```

In the example above, the link tag will be rendered with the label, but the `href` attribute will be missing.

~Alternative solution: allow for using [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/template) for things such as `:foreach` and `:if`, etc. It's better than wrapping things in `div` because `template` doesn't get rendered as a tag itself in the page.~

~Currently, in Tempest, `<template>` and everything in it is not rendered at all.~

_Never mind, Tempest has `<x-template>` 😅_ 

Fixes #1157

### DIFF
--- a/src/Tempest/View/src/Elements/IsElement.php
+++ b/src/Tempest/View/src/Elements/IsElement.php
@@ -32,7 +32,23 @@ trait IsElement
             $wrappingAttributes = [];
         }
 
-        return [...$this->attributes, ...$wrappingAttributes];
+        $attributes = [...$this->attributes, ...$wrappingAttributes];
+
+        // Some attributes should always come after others,
+        // so that these expressions have access to the data attributes
+        $attributePrecedence = [
+            ':foreach' => 1,
+            ':if' => 2,
+        ];
+
+        uksort($attributes, function (string $a, string $b) use ($attributePrecedence) {
+            $precedenceA = $attributePrecedence[$a] ?? 0;
+            $precedenceB = $attributePrecedence[$b] ?? 0;
+
+            return $precedenceA <=> $precedenceB;
+        });
+
+        return $attributes;
     }
 
     public function hasAttribute(string $name): bool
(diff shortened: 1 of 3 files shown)


brendt:
The problem here was that some attributes need precedence over others. I've added a fix, but we might need to further improve it down the line. It's good for now though.

wryk:
Are there any shortcoming to use the order of directives instead of a fixed precedence ? It would be useful to be able to choose their execution order.

```html
<div :if="$condition" :foreach="$items as $item">{{ $item->label }}</div>

<div :foreach="$items as $item" :if="$item->condition">{{ $item->label }}</div>
```
as

```php
if ($condition) foreach ($items as $item) $item->label;


foreach ($items as $item) if ($item->condition) $item->label;
```

The current fixed precedence only properly allow the second example, it can also act like the first exemple but only in a unoptimized way.

I'm not sure about regular attributes. I only see downsides of them being able to have a higher precedence than directives.

innocenzi:
> Are there any shortcoming to use the order of directives instead of a fixed precedence ?
It would be quite confusing, this is not something HTML itself does either. Instead, you should wrap your `if` in `<x-template>` or the other way around:

```php
<x-template :if="$condition">
  <div :foreach="$items as $item">{{ $item->label }}</div>
</x-template>
```

```php
<x-template :foreach="$items as $item">
  <div :if="$condition">{{ $item->label }}</div>
</x-template>
```

innocenzi:
> If order importance of directives is too much of a stretch for an HTML looking language maybe this should be a disallowed behavior by throwing an error at compilation.

I actually agree! Feel free to open an issue about that

PR #1154, opened 2025-04-18
feat(database): add `HasConditions` to query builders

erikaraujo:
This PR allows for dynamically building queries, like:

```php
query('my_table')
    ->select('id', 'name')
    ->when(
        $request->query['sort'],
        fn (SelectQueryBuilder $query) => $query->orderBy($request->query['sort']),
    )
    ->all();
```

```php
query('my_table')
    ->update(name: 'Some Name')
    ->when(
        $myCondition,
        fn (UpdateQueryBuilder $query) => $query->orderBy($request->query['sort']),
    )
    ->execute();
```

It adds the `HasConditions` trait to the following Query Builders: `SelectQueryBuilder`, `UpdateQueryBuilder` and `DeleteQueryBuilder`.

Looking at the methods in the `InsertQueryBuilder` I thought it didn't make a ton of sense to add it there.

I've also considered having an abstract `QueryBuilder` class (with a different name) that would serve as a base for all query builders and include the trait in there - but I wasn't so sure about it.


brendt:
Nice!

brendt:
FYI I can't merge currently because GitHub is having some API issue

PR #1158, opened 2025-04-19
feat(datetime): add datetime component

### LINKED ISSUE #985: Improve `tempest/clock`
Improve the implementation of `tempest/clock` to provide utilities regarding manipulating the clock.

Potential inspirations:
- Carbon
- PSL's [`DateTime`](https://github.com/azjezz/psl/blob/next/src/Psl/DateTime/DateTime.php)

innocenzi:
This pull request adds a new `tempest/datetime` package, bringing a better interface for working with dates in Tempest. The implementation is directly ported from the [PSL](https://github.com/azjezz/psl), and slightly adapted to our needs.

The `tempest/clock` package has been updated to use the new `DateTimeInterface`, and some parts of the framework have been refactored to no longer use PHP's own `DateTimeImmutable` class.

The `tempest/support` package now has a new `Json` and `Math` namespaces with new utilities that were needed for implementing `tempest/datetime`.

Closes #985

---

**Why use a custom implementation and not PHP's native `DateTimeInterface`?**
> PHP's native datetime implementation lacks a lot of convenience and flexibility. Adding or substracting durations is super cumbersome, and not everything is typed. Tempest's DateTime aims to provide more quality of life and flexibility in this regard.

**Why not use Carbon?**
> While Carbon solves the "convenience" part, it also has its own flaws. Most notably, it is mutable by default. By using Carbon directly, we also lose some flexibiltiy that we gain with our own implementation. The downside is that Carbon currently has more functionalities (mostly regarding formatting), but the gap can be closed other time.

---

**To-do**:
- Update Tempest's validation and casters to use `tempest/datetime`


innocenzi:
I'm marking this as ready for review, if we go forward with it I'll make another PR for the validation/casters

brendt:
I'm adding this to the beta.1 milestone. Not because I'm certain it will be merged, but to make sure we make a decision on this before tagging beta.1

brendt:
Ok there's no point in waiting, I guess. We've discussed this on Discord and I decided to merge it, especially since it's only 3k lines of actual code, and most of those are tests. 

We'll mark the component as experimental for the time being though

PR #1163, opened 2025-04-22
fix(view): allow specifying dynamic icon names

brendt:
Closes #1146
Closes #1144

Issue #1146, opened 2025-04-11
`<x-icon>` does not accept a variable as the icon name


erikaraujo:
### Tempest version

dev-main

### PHP version

8.4

### Operating system

macOS

### Description

Cannot use variable in the `name` attribute of the Icon component - nothing gets rendered.

May be related #1144 

### Steps to reproduce

Use the `<x-icon` with a variable in its name inside of a view:

```html
<x-icon :name="$myIconName" />
```

Issue #1144, opened 2025-04-11
`ViewComponentElement` `getAttribute` not working properly for expression attributes


erikaraujo:
### Tempest version

dev-main

### PHP version

8.4

### Operating system

macOS

### Description

Using `ViewComponentElement`'s `->getAttribute('')` or `->getAttributes()` does not return the actual attribute values for expression attributes.

### Steps to reproduce

Given the following component usage:

```html
<div>
    <x-my-custom-view-component :some-variable="$myVariable" />
</div>
```

When you have this component class:
```php
final class MyCustomViewComponent implements ViewComponent
{
    public static function getName(): string
    {
        return 'x-my-custom-view-component';
    }

    public function compile(ViewComponentElement $element): string
    {
        dd(
            $element->getAttribute('some-variable')
        );
    }
}
```

The output of the `dd` is the following string:
```
"<?= Tempest\View\Attributes\ExpressionAttribute::resolveValue($myVariable) ?>"
```

PR #1165, opened 2025-04-23
fix(view): switch to runtime icon view component

brendt:
Closes https://github.com/tempestphp/tempest-framework/issues/1146
Closes https://github.com/tempestphp/tempest-framework/issues/1144
Closes #1163


erikaraujo:
What if we add the `Component` interface back and use Discovery to automap component classes to their "tags"?

That way, we'd do `<x-icon name="lucide:cloud" />` and the renderer would replace it with `new Icon(name: 'lucide:cloud')->render()` or `get(Icon::class)->render(name: 'lucide:cloud')`...

brendt:
The problem there is that we don't have access to the local variables. You can't access `$name` or `$class` within that view component class, since the actual variable is only present in the view. You could write `global $name` within the view component class, but that's really ugly.

PR #1173, opened 2025-04-24
fix(view): wrong matched imports in view component slots

erikaraujo:


brendt:
Just FYI, I spent two hours debugging this, eventually this was the fix:

```php
            // The import contains escaped slashes, meaning it's a var_exported string; we can ignore those
            if (str_contains($matches[0], '\\\\')) {
                return $matches[0];
            }
```

All the rest is just some cleanup I did while I was working there…

PR #1174, opened 2025-04-24
feat(database): add `Count` query builder and statement

### LINKED ISSUE #1164: Get count of records from the DB
### Description

I'm not sure if this is a bug or a feature request!

I've tried the following and neither worked to retrieve the total records from a table in the DB:

```php
query(User::class)
    ->select('count(*)')
    ->all();
```

```php
query('users')
    ->select('count(*)')
    ->build()
    ->execute();
```

In both scenarios I got the following:
> SQLSTATE[HY000]: General error: 1 no such column: count(*)

So I'm not sure if we should have a ->count() method as a feature or what 😅

### Benefits

Be able to retrieve the count of records from a table in the DB

erikaraujo:
This pull request introduces a new `CountQueryBuilder` to the Tempest database library, enabling efficient SQL `COUNT` queries with fluent condition chaining. It includes the implementation of the builder, integration with existing query builders, and comprehensive unit and integration tests.

### New Feature: `CountQueryBuilder`

* Added the `CountQueryBuilder` class to support building SQL `COUNT` queries with methods for chaining conditions (`where`, `andWhere`, `orWhere`, etc.) and specifying columns. It integrates with `ModelDefinition` and `TableDefinition` for resolving table and field details. (`src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php`)
* Introduced a `count()` method in the `QueryBuilder` class to create instances of `CountQueryBuilder`, allowing seamless integration with existing query workflows. (`src/Tempest/Database/src/Builder/QueryBuilders/QueryBuilder.php`)

### Supporting Query Statement

* Implemented the `CountStatement` class, which compiles the SQL for `COUNT` queries, including support for optional `WHERE` clauses and column specification. (`src/Tempest/Database/src/QueryStatements/CountStatement.php`)

### TODOs:

- [ ] ~Add support for `GROUP BY` and `HAVING`~ _(As brought up on discord, this makes more sense on the SELECT query builder only)_
- [x] ~Add support for aliasing the `COUNT`~ _(There's no point aliasing the count if the only thing returned is an `int`)_
- [x] Add support for `DISTINCT`
- [ ] Include `COUNT` in a `->select`

### Caveats (?)

The problem with Tempest's approach of defining the "query type" (`select`/`update`/`delete`) before building the rest of the query can make it difficult to reuse the query conditions.

For example - say you need to implement pagination from a filterable query - you'll need to repeat yourself in order to get the results and also a total count:

```php
$selectQuery = query('books')->select();
$countQuery = query('books')->count();

if ($request->search) { 
    $selectQuery = $selectQuery->where('title LIKE ?', "%{$request->search}%");
    $countQuery = $countQuery->where('title LIKE ?', "%{$request->search}%");
}

//...

$count = $countQuery->execut();

$selectQuery = $selectQuery->limit(10)->offset(10);
$results = $selectQuery->all();

```

Closes #1164

### DIFF
--- a/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
@@ -41,14 +41,6 @@ final class CountQueryBuilder
     }
 
     /** @return self<TModelClass> */
-    public function as(string $alias): self
-    {
-        $this->count->alias = $alias;
-
-        return $this;
-    }
-
-    /** @return self<TModelClass> */
     public function distinct(): self
     {
         if ($this->count->column === null || $this->count->column === '*') {
--- a/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
@@ -36,9 +36,15 @@ final class CountQueryBuilder
 
     public function execute(mixed ...$bindings): int
     {
-        $key = "COUNT({$this->count->countArgument})";
+        return $this->build()->fetchFirst(...$bindings)[$this->count->getKey()];
+    }
 
-        return $this->build()->fetchFirst(...$bindings)[$key];
+    /** @return self<TModelClass> */
+    public function as(string $alias): self
+    {
+        $this->count->alias = $alias;
+
+        return $this;
     }
 
     /** @return self<TModelClass> */
--- a/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
@@ -6,6 +6,7 @@ namespace Tempest\Database\Builder\QueryBuilders;
 
 use Tempest\Database\Builder\ModelDefinition;
 use Tempest\Database\Builder\TableDefinition;
+use Tempest\Database\Exceptions\CannotCountDistinctWithoutSpecifyingAColumn;
 use Tempest\Database\Query;
 use Tempest\Database\QueryStatements\CountStatement;
 use Tempest\Database\QueryStatements\WhereStatement;
@@ -43,6 +44,18 @@ final class CountQueryBuilder
     public function as(string $alias): self
     {
         $this->count->alias = $alias;
+
+        return $this;
+    }
+
+    /** @return self<TModelClass> */
+    public function distinct(): self
+    {
+        if ($this->count->column === null || $this->count->column === '*') {
+            throw new CannotCountDistinctWithoutSpecifyingAColumn();
+        }
+
+        $this->count->distinct = true;
 
         return $this;
     }
--- a/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
+++ b/src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Builder\QueryBuilders;
+
+use Tempest\Database\Builder\ModelDefinition;
+use Tempest\Database\Builder\TableDefinition;
+use Tempest\Database\Query;
+use Tempest\Database\QueryStatements\CountStatement;
+use Tempest\Database\QueryStatements\WhereStatement;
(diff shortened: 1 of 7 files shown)


erikaraujo:
@brendt Not sure if the query builder is marked as experimental or not, but IF we can have breaking changes on this `count` thing, this PR can be merged as is and I can work on the TODOs on follow-up ones.

aidan-casey:
Nice PR! As I mentioned on the Discord, I think I'd keep this under the select query builder. I'd envision this being something like:

```php
public function count(string $column = '*'): int
{
    // Do the work, accept where statements.
}
```

I'd add to the query builder a count helper that could be something like:

```php
public function count(string $column = '*'): int
{
    return $this->select()->count($column);
}
```

Columns previously specified can simply be ignored, I'd think.

That's just IMO.

brendt:
Oh, interesting approach, I admit I also thought it would have be part of the `SelectQueryBuilder`, but having seen this solution, I like it more, I think. That's because our `count` queries should only do single counts. Ie. they should return an integer, that's it. On the SQL level, you could also use `count` on select queries, you could even have more than one count per query; but we don't want to support any of that with our querybuilders. If people have more complex cases, they should build the query themselves. 

That's why I like the separation, as it clearly indicates that `query()->count()` is something on its own, apart from normal select queries.

> Not sure if the query builder is marked as experimental or not, but IF we can have breaking changes on this count thing, this PR can be merged as is and I can work on the TODOs on follow-up ones.

The whole ORM is experimental, so we're fine 👍

> The problem with Tempest's approach of defining the "query type" (select/update/delete) before building the rest of the query can make it difficult to reuse the query conditions.

This is something we can solve with traits in the long run. I prefer to keep the implementations apart for a while though, to make sure we don't over-abstract up front. Especially in an experimental component like the ORM.

src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $this->count = new CountStatement(
+            table: $this->resolveTable($model),
+            column: $column,
+        );
+    }
+
+    public function execute(mixed ...$bindings): int
+    {
+        $key = "COUNT({$this->count->countArgument})";
```
brendt, line 39: Could we move retrieving the key to the statement? Something like `$this->count->getKey()`?

brendt (changes requested):

src/Tempest/Database/src/Builder/QueryBuilders/CountQueryBuilder.php
erikaraujo: We could and we should - because of aliasing the count. Let me work on that first.

erikaraujo:
> > The problem with Tempest's approach of defining the "query type" (select/update/delete) before building the rest of the query can make it difficult to reuse the query conditions.
> 
> This is something we can solve with traits in the long run. I prefer to keep the implementations apart for a while though, to make sure we don't over-abstract up front. Especially in an experimental component like the ORM.


I think you misunderstood me, @brendt, because I don't think trait would solve the problem I described 🤔.

What I mentioned is not the fact that `public function where` is duplicated on all builders - it's that the fact that you must select a "query builder type" upfront by immediately doing `->select()` or `->update()` or `->count()` makes it difficult to reuse where clauses between multiple queries.

brendt:
Gotcha, yeah I misunderstood entirely 😁 

It's an interesting idea, but I suggest we discuss it in another issue, it's out of scope for this PR :)

erikaraujo:
@brendt Added support for distinct count and aliasing. This PR is now ready for review.

Here's how to do it:

```php
query('books')
    ->count('title')
    ->distinct()
    ->as('book_title')
    ->execute();
```

erikaraujo:
@brendt Added support for distinct count and removed the useless alias.
This PR is now (really) ready for review.

Here's how to do it:

```php
query('books')
    ->count('title')
    ->distinct()
    ->execute();
```

If you do the following, an exception will be thrown, since you must specify a column for distinct:

```php
query('books')
    ->count() // or ->count('*')
    ->distinct()
    ->execute();
```

brendt:
Awesome! Are you up for adding the docs for this as well?

erikaraujo:
> Awesome! Are you up for adding the docs for this as well?

Sure thing, but I may not be able to get around to it until after may 1st.

PR #1176, opened 2025-04-25
build: cleans up coverage analysis

aidan-casey:
Full disclosure, @brendt is really going to hate this one... 😜


brendt:
"Hate" is such a strong word. 

But ok, my feelings about it are strong, so yes, "hate" is correct 😂

So, what have we achieved by doing this? We're being forced to refactor hundreds of tests to run in isolation, or we simply add more coverage attributes and basically achieve nothing. We have a codebase that's 2 years old by now. How long will it take to refactor everything? Who's going to spend time doing that? I know I won't because:

1. I have other Tempest issues I think are more important to solve
2. I have other things to work on during my daytime job
3. I have other things to do with family and friends
4. I'm with you that better coverage and cleaner tests will have a positive impact, just not enough impact that will justify the time needed to get it right

I simply don't have and am not willing to put in the time needed to get this right.

aidan-casey:
@brendt the attributes here are largely experimental on my part as I'm looking at this and may not make the final PR.

The bigger issue, imo, is how drastically we are skewing our coverage which I know we've discussed before.

We also have a very large portion of tests that are not PSR compliant, etc.

brendt:
@aidan-casey that's fair, I'm with you that there's room for improvement. I'm just saying that I won't be able to invest the time needed to make that improvement.

ericsizemore:
I notice quite a bit of "This test executed code that is not listed as code to be covered or used" in the failing checks. It looks like a lot of work, but having had to do this before it's not... too... bad. Looks like it needs more ```#[CoversClass(...)]``` entries and the addition of ```#[CoversFunction(...)]``` ([CoversFunction](https://docs.phpunit.de/en/12.0/attributes.html#coversfunction))? :)

aidan-casey:
@ericsizemore yeah, you're spot on. Combo of Covers{Class/Function} and Uses{Class/Function}. Primarily do not want to put the work in if there is zero chance of this getting merged. We had a similar PR earlier on that got closed.

PR #1181, opened 2025-04-28
feat(database): add `count()` helper to `IsDatabaseModel` trait

erikaraujo:
This pull request introduces enhancements to the `IsDatabaseModel` class in the `Tempest\Database` namespace by adding new functionality for counting records. The most notable changes include the addition of two new methods, `count` and `countAll`, and the inclusion of a new import to support these methods.

### Enhancements to `IsDatabaseModel`:

* **Added `count` method**: Introduced a static method `count` that returns a `CountQueryBuilder` instance, enabling users to construct count queries for the model. (`[src/Tempest/Database/src/IsDatabaseModel.phpR64-R73](diffhunk://#diff-71139a1d93468f4b2a4d83c6b70374139c48a5ffb42252794913fedb7ebdae39R64-R73)`)
* **Added `countAll` method**: Introduced a static method `countAll` that executes a count query and returns the total number of records for the model as an integer. (`[src/Tempest/Database/src/IsDatabaseModel.phpR64-R73](diffhunk://#diff-71139a1d93468f4b2a4d83c6b70374139c48a5ffb42252794913fedb7ebdae39R64-R73)`)

### Supporting changes:

* **New import for `CountQueryBuilder`**: Added an import for `Tempest\Database\Builder\QueryBuilders\CountQueryBuilder` to support the new counting functionality. (`[src/Tempest/Database/src/IsDatabaseModel.phpR7](diffhunk://#diff-71139a1d93468f4b2a4d83c6b70374139c48a5ffb42252794913fedb7ebdae39R7)`)


src/Tempest/Database/src/IsDatabaseModel.php
```diff
@@ -60,6 +61,16 @@ public static function find(mixed ...$conditions): SelectQueryBuilder
         return $query;
     }
 
+    public static function count(): CountQueryBuilder
+    {
+        return query(self::class)->count();
+    }
+
+    public static function countAll(): int
```
erikaraujo, line 69: I'm not so sure about this `countAll` method...

brendt: Yeah, I'm not a fan of it either. I think it's ok for `count` to directly return `int` on this level. If people want more control they should fall back to `Model::query()` or `query()`

erikaraujo: Doesn't it feel weird to have the "same method" behave differently, though?

erikaraujo: Yea, we can do that. I'll update the PR later today or tomorrow.

erikaraujo: Done!

PR #1186, opened 2025-04-29
refactor: move integration tests to package directories

aidan-casey:


brendt:
We'll have to retry this PR when there's more time :) We were now blocked by it and Aidan didn't seem to have the time to jump on it, so I decided to close it.

PR #1187, opened 2025-04-29
feat(storage): support multiple storage configurations

innocenzi:
This pull request adds support for multiple storage by adding the `HasTag` interface to the `StorageConfig` and by adapting the testing helpers.

The latter required the addition of `removeInitializer` to the container, as well as updating `unregister` to support unregistering the dependency along with all its tagged instances.

**Note**: this should be merged after the refactor in https://github.com/tempestphp/tempest-framework/pull/1186, I'll solve the conflicts myself.

### DIFF
--- a/packages/container/src/Container.php
+++ b/packages/container/src/Container.php
@@ -12,7 +12,7 @@ interface Container
 {
     public function register(string $className, callable $definition): self;
 
-    public function unregister(string $className): self;
+    public function unregister(string $className, bool $tagged = false): self;
 
     public function singleton(string $className, mixed $definition, ?string $tag = null): self;
 
--- a/packages/container/src/GenericContainer.php
+++ b/packages/container/src/GenericContainer.php
@@ -16,6 +16,7 @@ use Tempest\Reflection\FunctionReflector;
 use Tempest\Reflection\MethodReflector;
 use Tempest\Reflection\ParameterReflector;
 use Tempest\Reflection\TypeReflector;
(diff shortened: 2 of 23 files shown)


brendt:
One remark here: I would expect `unregister` to work slightly different. I'd expect it to look like this: `unregister(string $className, null|string|UnitEnum $tag = null)`. If `$tag` is null, we'll unregister all singletons for that class, otherwise only the tagged one.

WDYT?

innocenzi:
> One remark here: I would expect `unregister` to work slightly different. I'd expect it to look like this: `unregister(string $className, null|string|UnitEnum $tag = null)`. If `$tag` is null, we'll unregister all singletons for that class, otherwise only the tagged one.
> 
> WDYT?

It'd be nice to allow specifying a tag, but in this case we need to unregister all bindings for all tags. Doing that by default of no tag is specified would be confusing, right?

innocenzi:
> One remark here: I would expect `unregister` to work slightly different. I'd expect it to look like this: `unregister(string $className, null|string|UnitEnum $tag = null)`. If `$tag` is null, we'll unregister all singletons for that class, otherwise only the tagged one.
> 
> WDYT?

It'd be nice to allow specifying a tag, but in this case we need to unregister all bindings for all tags. Doing that by default if no tag is specified would be confusing, right?

brendt:
> Doing that by default if no tag is specified would be confusing, right?

Hmm yeah that's a good point… Let's keep it as-is for now

PR #1198, opened 2025-05-03
feat(core): add discovery config

### LINKED ISSUE #1197: Add discovery config
Add discovery config which allows users to define classes that should not be discovered (useful for packages or non-PHP files where you can't control the `SkipDiscovery` attribute)

brendt:
Closes #1197


brendt:
@innocenzi @aidan-casey I know we discussed this a long time before but I can't remember the details. Do you think it's worth having this feature though?

aidan-casey:
I'm a big fan of attributes only being syntax candy that exists over top of traditional features. I definitely think it makes sense to have a minimal config, but the developer shouldn't be required to touch it for things to work.

innocenzi:
Totally in favor of this

Issue #1202, opened 2025-05-05
Production exception handler


innocenzi:
By default, we don't have a production exception handler. We have Whoops as the local exception handler, which we should keep as the "emergency" exception handler since it's the first thing registered when booting the framework. We also have the console exception handler, registered last.

We need a built-in exception handler, configurable in userland, with a better API than the current `AppConfig`'s `errorHandlers` array, which doesn't have a discovery. I also think tha response processors would overlap with a better exception handler, and that we might need to sunset them.

I'm thinking the `ErrorHandler` interface needs to be cleaned to just a `handle(Throwable)` method, and that when registering it, we'd convert PHP's error to an `ErrorException`.

I would also love to have some kind of `HasContext` interface that exception classes could implement, so they kind provide more details down the pipeline (eg. for a Sentry integration).

@brendt would love your opinion here.

Issue #1189, opened 2025-04-30
`ErrorResponseProcessor` overwrites defined response bodies


wryk:
### Tempest version

1.0-alpha.6

### PHP version

8.4

### Operating system

Linux

### Description

I was trying to render a simple form page with validations errors and I can't specify a client error code or `ErrorResponseProcessor` will overwrite my body. As a default behavior I think it's getting in the way of a common normal usage of a response.

I suggest the processor only updating the response body if not defined, it could still do its job of providing a default error page without being annoying when I need control. I'll wait for #1186 before submitting a PR.

### Steps to reproduce

Dispatch a response with a body and a 4xx or 5xx status. 

```php
public function __invoke(): Response
	return new GenericResponse(
		status: Status::UNPROCESSABLE_CONTENT,
		body: 'My body'
	);
}
```

PR #1203, opened 2025-05-05
feat(core): improve exception handling

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1202 and https://github.com/tempestphp/tempest-framework/issues/1189

The goal of this PR is to have centralized exception handling.
- I have kept the existing emergency and console handler, and updated the `ErrorHandler` interface to a simpler `ExceptionHandler` one.
- I added a `ExceptionProcessor` interface and discovery, which is ran in the console exception handler and the new HTTP exception handler. This is what logs exceptions now.
- The response processor that was returning error pages now throws an `HttpException`.
- In production, error pages are rendered by the new HTTP exception handler, which converts `HttpException` to a proper error response page.

I'll add tests after https://github.com/tempestphp/tempest-framework/pull/1186.


packages/router/src/Exceptions/HttpErrorResponseProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+        private AppConfig $appConfig,
+    ) {}
+
+    public function process(Response $response): Response
+    {
+        // Throwing an HttpException during tests would make testing more
+        // complex, and is not strictly needed. During development, we
+        // don't need exceptions either, since the exception handler
+        // is different. For this reason, we skip processing here.
+        if (! $this->appConfig->environment->isProduction()) {
```
brendt, line 22: Shouldn't we make a difference here between exceptions that were thrown on the framework level, and user-based exceptions? There are a lot of cases where I want tests to simply throw the exception instead of hiding it behind a response. 

It could also be a toggle. This is what we used to do in tests: `$this->http->throwExceptions()->get(…)`

WDYT?

packages/router/src/Exceptions/SendsResponse.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Tempest\Router\Exceptions;
+
+use Tempest\Http\Response;
+
+/**
+ * Marks this exception class as one that can be converted to a response.
+ *
+ * @phpstan-require-extends \Throwable
+ */
+interface SendsResponse
```
brendt, line 12: Nitpick: it's not this class that _sends_ the response, this class can be converted _to_ a response. Maybe we need a slightly better name, although I'm not sure what that should be.

- `ConvertsToResponse`
- Laravel has `Responsable`, which is a really bad name
- Something else?

brendt (changes requested):

packages/router/src/Exceptions/HttpErrorResponseProcessor.php
innocenzi: Conceptually I agree yes. Not sure how to implement though

brendt: This actually still works, so it's fine for tests. I don't think we need to followup for now. Let's wait and see if people require more control based on them playing around with Tempest

PR #1205, opened 2025-05-06
fix(router): require hard-coded uris to start with a slash in `Router::toUri`

### LINKED ISSUE #1182: `GenericRouter::toUri` ignore invalid class `ReflectionError`
### Tempest version

1.0-alpha.6

### PHP version

8.4

### Operating system

Linux

### Description

Calling `GenericRouter::toUri` with an invalid class-string (can happen with typos and missing imports) doesn't fail and generates an incorrect uri.

I think this issue was introduced in #490.

Because a string and an invalid class-string are indistinguishable, a `ReflectionException` cannot be safely used as an heuristic to discriminate between controller class-strings and hard-coded uris.

The current best fix I can think of without removing hard-coded uris support is requiring them to starts with a slash. Everything else would be considered as a class-string and would throw if invalid.

I'm interested in submitting a PR implementing this fix (or a better one, of course)

### Steps to reproduce

Call `GenericRouter::toUri` with an invalid class string.

wryk:
Closes #1182.

Hard-coded uris must now starts with a slash.

I'll add tests after #1186.

This function signature is barebone, do the framework have a stance on external documentation vs docblocks ? I can address this issue in a later PR.

### DIFF
--- a/packages/router/src/GenericRouter.php
+++ b/packages/router/src/GenericRouter.php
@@ -131,27 +131,20 @@ final class GenericRouter implements Router
 
     public function toUri(array|string $action, ...$params): string
     {
-        try {
-            if (is_array($action)) {
-                $controllerClass = $action[0];
-                $reflection = new ClassReflector($controllerClass);
-                $controllerMethod = $reflection->getMethod($action[1]);
-            } else {
-                $controllerClass = $action;
-                $reflection = new ClassReflector($controllerClass);
-                $controllerMethod = $reflection->getMethod('__invoke');
-            }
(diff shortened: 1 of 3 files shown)


innocenzi:
> do the framework have a stance on external documentation vs docblocks ?

Detailed docblocks are awesome DX, if you can do that it'd be great 🙏 I'm personally trying to add them as much as I can as well.

brendt:
Feel free to add tests now, we're going to postpone #1186

brendt:
Great PR, thanks!

PR #1206, opened 2025-05-06
fix(filesystem): add ability to delete invalid symlinks

aazsamir:
Attempting to remove a symbolic link pointing to an invalid (non-existing) file caused an error because `file_exists($x)` returns `false` in such a case.

I noticed this because without this change `phpunit` crashes with errors on my OS, my guess is that this is related to the order in which files are removed from the directory during tests.


innocenzi:
Thank you!

PR #1214, opened 2025-05-07
feat(core): add debug exception processor

aazsamir:
I would expect the framework to log exceptions using the logger I set up in the config.

I always log everything to `php://stdout` and I don't want to keep track of some log path during development. This is especially useful in docker.

I decided to go with using `get(Logger::class)` instead of injecting it, as I thought that this should be *a bit* safer if framework fails at some really early stage? I'm not sure, I would be happy to be corrected if that's not the case.

### DIFF
--- a/packages/core/src/DebugExceptionProcessor.php
+++ b/packages/core/src/DebugExceptionProcessor.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Tempest\Core;
+
+use Tempest\Debug\Debug;
+use Throwable;
+
+final class DebugExceptionProcessor implements ExceptionProcessor
+{
+    public function process(Throwable $throwable): Throwable
+    {
+        $items = [
+            'exception' => $throwable->getMessage(),
+            'context' => ($throwable instanceof HasContext)
+                ? $throwable->context()
+                : [],
+        ];
+
+        Debug::resolve()->log($items, writeToOut: false);
+
+        return $throwable;
+    }
+}
--- a/packages/core/src/LogExceptionProcessor.php
+++ b/packages/core/src/LogExceptionProcessor.php
@@ -2,8 +2,10 @@
 
 namespace Tempest\Core;
 
-use Tempest\Debug\Debug;
+use Tempest\Log\Logger;
 use Throwable;
+
+use function Tempest\get;
 
 /**
  * An exception processor that logs exceptions.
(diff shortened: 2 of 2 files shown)


brendt (changes requested):
Hmm, we actually need to make this more configurable, because outside of Docker, the current approach is super useful. (it writes debug-styled log entries into a file, which can be [tailed with `tempest tail`](https://tempestphp.com/main/tempest-in-depth/logging))

Of course, your approach is also useful in other cases. @innocenzi how do you envision we make this part configurable? We could have two implementations of log processors, but then what? How do we pick the right one?

aazsamir:
> Hmm, we actually need to make this more configurable, because outside of Docker, the current approach is super useful. (it writes debug-styled log entries into a file, which can be [tailed with `tempest tail`](https://tempestphp.com/main/tempest-in-depth/logging))
> 
> Of course, your approach is also useful in other cases. @innocenzi how do you envision we make this part configurable? We could have two implementations of log processors, but then what? How do we pick the right one?

Maybe these should be two different processors: `LogExceptionProcessor` and `DebugExceptionProcessor`.

brendt:
Agree. I'm only not sure how we decide between the two. Enzo is working on exception handling, so I want to check whether he already had a plan for this. Probably by adding entires in AppConfig, but I wanna make sure :)

innocenzi:
Having two exception processors by default seems fine to me. FYI, I plan on improving the logging as well, I'd like to implement a web-based Ray-like UI in addition to a console server for debugging, @brendt, so the tail-based implementation we have might be less useful in the future.

In the meantime, I agree with having two different processors. I'm not a fan of having a specific configuration for them though, and I'd recommend using the kernel boot event to configure them as needed:

```php
use Tempest\Core\AppConfig;
use Tempest\Core\KernelEvent;
use Tempest\Core\LogExceptionProcessor;
use Tempest\EventBus\EventHandler;
use Tempest\Support\Arr;

final class RemoveErrorResponseProcessor
{
    public function __construct(
        private AppConfig $appConfig,
    ) {
    }

    #[EventHandler(KernelEvent::BOOTED)]
    public function __invoke()
    {
        Arr\forget_values($this->appConfig->exceptionProcessors, LogExceptionProcessor::class);
    }
}
```

brendt:
@aazsamir do you want to implement in this PR or do you want me to do it?

aazsamir:
I reverted previous processor, just named it differently. Now exception is both logged by configured logger and put in debug log.

I think I'm not sure about this configuring processors by config or another service? So I'm leaving it as is.

innocenzi:
Should one of them be off by default?

By the way, it's sage to inject the logger. The constructor is called right before the `process` method anyway.

innocenzi:
I'm working on an improved debug experience in Tempest, so this will no longer be necessary

PR #1226, opened 2025-05-09
feat(core): add `about` command

happytodev:
This pull request add the possibility 

- to get basic environment informations about your app
- to check quickly the current version of Tempest directly in the terminal.

# About command

```bash
php tempest about
```

or the following alias : 

```bash
php tempest a
```

It returns : 

image

You can also have a json version of output with this command : 

```bash
php tempest about --json
```

or the following alias : 

```bash
php tempest about -j
```

it returns : 

image

and of course you can have an help for this command : 

```bash
php tempest about --help
```

it returns

image

# Version command

Available command are : 

```bash
php tempest version
```

or the following alias : 

```bash
php tempest v
```

returns

image

The current version is retrieved from `Kernel::VERSION`.


innocenzi:
Maybe we could add the PHP version as well?

happytodev:
> **EDIT**: actually, I was thinking of a `tempest about` command, very similar to what Laravel has. I think we should go toward that, `tempest version` might offer too few information compared to what we could display

Totally agree. Do you want I work on it ?

happytodev:
Add about command to the PR.

innocenzi:
Nice addition! I'll need to clean this up a bit before merging, so I'll mark as draft for now and will try to find some time for it this week

happytodev:
Not a problem at all. 
Let me know if I need to fix something ;-)

brendt:
@innocenzi did you still plan on finetuning stuff here?

innocenzi:
Yes—sorry, I haven't taken the time yet, but I plan to

innocenzi:
Hey @happytodev, I updated how the command works as follows:
- I moved it to the core package
- I added the concept of "insight provider", which are discovered
- I added corresponding insight providers in relevant packages (for now the core package and the database one)

This way, external packages can also add data to `tempest about`. Note that I removed the cache insights for now as I'm working on this component: https://github.com/tempestphp/tempest-framework/pull/1245

PR #1227, opened 2025-05-09
feat(mailer): introduce mailer component

### LINKED ISSUE #149: Mail support

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/149

This pull request adds a mailer component, as a wrapper around Symfony Mailer. The API is not polished, but most of the features are already there, testing included.

```php
return new SmtpMailerConfig(
    scheme: Scheme::SMTP,
    host: env('MAILER_SMTP_HOST', default: '127.0.0.0'),
    port: env('MAILER_SMTP_PORT', default: 2525),
    username: env('MAILER_SMTP_USERNAME', default: ''),
    password: env('MAILER_SMTP_PASSWORD', default: ''),
);
```

```php
$this->mailer->send(new ResetPasswordEmail($user, $resetUrl));

// or

$this->mailer->send(new GenericEmail(
    subject: 'Reset your password',
    to: $user->emailAddress,  [
    html: view('./reset-password.view.php', /* ... */),
));

// or

$this->mailer->send(
    GenericEmail::build()
        ->subject('Reset your password')
        ->to($this->emailAddress)
        ->html(view('./reset-password.view.php')), /* ... */
);
```

```php
final class ResetPasswordEmail implements Email
{
    public Envelope $envelope {
        get => new Envelope(
            subject: "Reset your password",
            to: $this->user->emailAddress,
        );
    }

    public Content $content {
        get => new Content(
            html: view('./reset-password.view.php', user: $this->user, url: $this->resetUrl),
        );
    }

    public function __construct(
        private readonly User $user,
        private readonly string $resetUrl,
    ) {}
}
```

By default, sending emails is disabled during tests. There's no risk of forgetting to fake the mailer and sending actual emails.

```php
$sent = $mailer->send(new ResetPasswordEmail($user, $resetUrl));

$sent->assertSee($user->fullName);
$sent->assertSee($resetUrl);
$sent->assertSentTo($user->emailAddress);
$sent->assertFrom('no-reply@tempestphp.com');
```

I only included SMTP and Postmark for now, but we can easily add more adapters.


brendt:
I'll review next week, it's a busy weekend for me :)

innocenzi:
I'm essentially done, but I still need opinions on the `Email` interfac, `GenericEmail` name and `EmailBuilder` names, which I'm not convinced about.

packages/mailer/src/Attachments/StorageAttachment.php
```diff
@@ -0,0 +1,67 @@
...
+use Tempest\Container\GenericContainer;
+use Tempest\Mail\Exceptions\FileAttachmentWasNotFound;
+use Tempest\Storage\Storage;
+use Tempest\Support\Path;
+use UnitEnum;
+
+/**
+ * Represents an attachment that lives in the {@see Tempest\Storage\Storage}.
+ */
+final class StorageAttachment implements Attachment
```
brendt: Not sure how I feel about this. I'd rather have a simple `Attachment` class, and resolve the path from the outside. So if you want to attach something from storage, you first resolve it from eg. the controller and only then create the attachement.

packages/mailer/src/Attachments/StorageAttachment.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    public Closure $resolve {
+        get => fn () => $this->storage->readStream($this->path);
+    }
+
+    private function __construct(
+        private readonly string $path,
+        public readonly ?string $name,
+        public readonly ?string $contentType,
+        private readonly Storage $storage,
```
brendt: Why does the attachment need access to `Storage`?

packages/mailer/src/Builder/Email.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        $this->bcc = $bcc;
+
+        return $this;
+    }
+
+    /**
+     * Defines the subject of the email.
+     */
+    public function withSubject(string|Stringable $subject): self
```
brendt: There's a naming discrepancy here:

`with` methods usually indicate a clone of the original object is made. Furthermore, methods like `from` and `to` don't have a `with` prefix. 

I think these chainable builder methods are nice, but I'd drop the `with` prefix

packages/mailer/src/Builder/Email.php
```diff
@@ -0,0 +1,209 @@
...
+
+        $this->attachments = $attachments;
+
+        return $this;
+    }
+
+    /**
+     * Adds an attachment to the email.
+     */
+    public function withAttachment(Attachment $attachment): self
```
brendt: I'd call this `attach`, and have it accept `Attachment ...$attachments`, instead of having `withAttachments` and `withAttachment`

packages/mailer/src/Builder/Email.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        $this->attachments[] = $attachment;
+
+        return $this;
+    }
+
+    /**
+     * Adds an attachment from the filesystem.
+     */
+    public function withFileAttachment(string $path, ?string $name = null, ?string $contentType = null): self
```
brendt: As mentioned earlier, I'm not convinced we need to differentiate attachment types. An attachment should be a simple data object, and where exactly its file reference comes from is not its concern.

packages/mailer/src/Builder/Email.php
```diff
@@ -0,0 +1,209 @@
...
+use Tempest\Mail\Email as EmailInterface;
+use Tempest\Mail\Envelope;
+use Tempest\Mail\GenericEmail;
+use Tempest\Mail\Priority;
+use Tempest\Support\Arr;
+use Tempest\Support\Arr\ArrayInterface;
+use Tempest\View\View;
+use UnitEnum;
+
+final class Email
```
brendt: I don't mind this builder, but I'm not convinced about the name (see the naming collision with the `Email` interface).

An `Email` is a data object representing an email. This class is a builder for that email, and should probably be called `EmailBuilder`

packages/mailer/src/Content.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Tempest\Mail;
+
+use Tempest\Mail\Attachments\Attachment;
+use Tempest\Mail\Exceptions\MailContentWasMissing;
+use Tempest\View\View;
+
+/**
+ * Represents the content of an {@see \Tempest\Mailer\Envelope}.
+ */
+final class Content
```
brendt: Why do we need another object in between (both `Envelope` and `Content`). Why not have these properties directly on the `Email`?

packages/mailer/src/Email.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tempest\Mail;
+
+interface Email
+{
+    /**
+     * The envelope of the email.
+     */
+    public Envelope $envelope {
```
brendt: Referring back to my previous question: I'd rather have the properties encapsulated by `Envelope` and `Content` directly on the email instead

packages/mailer/src/EmailToSymfonyEmailMapper.php
```diff
@@ -0,0 +1,106 @@
...
+final readonly class EmailToSymfonyEmailMapper implements Mapper
+{
+    public function __construct(
+        private readonly MailerConfig $mailerConfig,
+        private readonly ViewRenderer $viewRenderer,
+    ) {}
+
+    public function canMap(mixed $from, mixed $to): bool
+    {
+        if ($from instanceof Email) {
```
brendt: You probably also want to make sure `$to` is of the correct type — if possible? In my experience, automapping must be as specific as possible to avoid collisions with other mappers

packages/mailer/src/Exceptions/ExpeditorWasMissing.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Tempest\Mail\Exceptions;
+
+use Exception;
+
+final class ExpeditorWasMissing extends Exception implements MailerException
+{
+    public function __construct()
+    {
+        parent::__construct('No expeditor address provided.');
```
brendt: I don't know what expeditor means? Maybe the exception could elaborate? Maybe I'm overthinking though

packages/mailer/src/GenericEmail.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Tempest\Mail;
+
+/**
+ * Represents a generic email.
+ */
+final class GenericEmail implements Email
```
brendt: If the plan is still to encourage people to make their own Email implementations (I think they should), then we should provide a trait `IsEmail` as well for convenience. Certainly if/when `Envelope` and `Content` get refactored out.

packages/mailer/src/GenericMailer.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->transport ??= $this->createTransport();
+    }
+
+    public function send(Email $email): SentEmail
+    {
+        $symfonyEmail = map($email)
+            ->with(fn (Email $from) => new EmailToSymfonyEmailMapper($this->mailerConfig, $this->viewRenderer)->map($from, null))
+            ->do();
+
+        return new SentGenericEmail(
```
brendt: Wouldn't it make more sense to dispatch an `EmailSent` event instead of returning a new value object? That way we can hook into the event from any place.

packages/mailer/src/MailerConfig.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Tempest\Mail;
+
+use Symfony\Component\Mailer\Transport\TransportInterface;
+
+interface MailerConfig
+{
+    /**
+     * The transport class.
+     */
+    public string $transport {
```
brendt: Add `@param class-string<Transport>` as the type

packages/mailer/src/MailerConfig.php
```diff
@@ -0,0 +1,27 @@
...
+        get;
+    }
+
+    /**
+     * The default address from which emails will be sent.
+     */
+    public null|string|Address $from {
```
brendt: Maybe rename to `defaultFrom`?

packages/mailer/src/Priority.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Tempest\Mail;
+
+enum Priority: int
```
brendt: Rename to `MailPriority`

packages/mailer/src/Priority.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * The highest priority.
```
brendt: We don't really need these docblocks imo

brendt (changes requested):

packages/mailer/src/Transports/Postmark/Scheme.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tempest\Mail\Transports\Postmark;
+
+enum Scheme: string
```
brendt: Rename to `PostmarkScheme` to avoid collisions

packages/mailer/src/Transports/Ses/Scheme.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tempest\Mail\Transports\Ses;
+
+enum Scheme: string
```
brendt: Rename to `SesScheme` to avoid collisions

packages/mailer/src/Transports/Smtp/Scheme.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Tempest\Mail\Transports\Smtp;
+
+enum Scheme: string
```
brendt: Rename to `SmtpScheme` to avoid collisions

brendt (changes requested):

packages/mailer/src/Attachments/StorageAttachment.php
innocenzi: Because its `$resolve` method uses it under the hood to lazily fetch the content from the storage

packages/mailer/src/Builder/Email.php
innocenzi: Absolutely, good call

packages/mailer/src/Builder/Email.php
innocenzi: While I conceptually agree, I personally prefer this:

```php
$email = new Email()
    ->to('michael.scott@dundermifflin.com')
    ->subject('Important: Please come to my office right away')
    ->text('...')
    ->make();
```

To this:

```php
$email = new EmailBuilder()
    ->to('michael.scott@dundermifflin.com')
    ->subject('Important: Please come to my office right away')
    ->text('...')
    ->make();
```

I initially called that class `EmailBuilder` actually, but I really dislike this kind of suffix. It feels redundant.

I know you really don't like class name conflicts, but it's sometimes necessary. In this case, not only we already have a conflict with validation rules, but in addition, since I'm instantiating the class, the interface will not even be suggeted:

https://github.com/user-attachments/assets/cc8aa7b3-09fd-4871-b5c6-812d3fd7f6b2

And when injecting it, docblocks help disambiguate classes (if the namespace is not enough).

CleanShot 2025-07-02 at 23 08 06

packages/mailer/src/Priority.php
innocenzi: Good catch

packages/mailer/src/Priority.php
innocenzi: Yeah, I got lazy. Oops. I updated them to actually be useful.

packages/mailer/src/MailerConfig.php
innocenzi: Renamed to `defaultExpeditor` to be even clearer

packages/mailer/src/Exceptions/ExpeditorWasMissing.php
innocenzi: Expeditor is the "from" email address. Is that not the usual term in english? Should that be "sender" maybe?

packages/mailer/src/Content.php
innocenzi: I felt like the interface (`Email`) was super heavy with all those properties. It's a lot.

Though for content specifically, it could be in the interface directly. The envelope could also be shared between emails, so a dedicated object still makes sense. What do you think?

packages/mailer/src/Attachments/StorageAttachment.php
innocenzi: Unless I'm misunderstanding something, the developer experience is way better with dedicated objects and methods:

```php
$email = new Email()
    ->to('michael.scott@dundermifflin.com')
    // this is nice
    ->attachFromFileystem(__DIR__ . '/paperstock.csv')
    // this is super cumbersome
    ->attach(new Attachment(
        resolve: fn () => Filesystem\read_file(__DIR__ . '/paperstock.csv')
    ))
    ->text('Gotcha!')
    ->make();
```

```php
final class SendWelcomeEmail implements Email
{
    public Envelope $envelope {
        get => new Envelope(
            subject: "Welcome {$this->fullName}",
            to: $this->address,
            from: 'no-reply@tempestphp.com',
        );
    }

    public Content $content {
        get => new Content(
            html: view(__DIR__ . '/welcome.view.php', fullName: $this->fullName),
            // this is cool
            attachments: StorageAttachment::fromPath($this->welcomeCardName, tag: 'welcome-cards')
            // this is not cool
            attachments: new Attachment(
                resolve: fn () => get(Storage::class, 'welcome-cards')->readStream($this->welcomeCardName)
            )
        );
    }

    public function __construct(
        private readonly string $address,
        private readonly string $fullName,
        private readonly string $welcomeCardName,
    ) {}
}
```

innocenzi: The alternative is an attachment builder class.

```php
new AttachmentBuilder()
    ->fromStorage($path, $tag)
    ->make();
```

packages/mailer/src/Builder/Email.php
innocenzi: See conversation in first review

packages/mailer/src/GenericMailer.php
innocenzi: This might be a good idea, but that changes the testing implementation a bit... I'll explore that

packages/mailer/src/GenericEmail.php
innocenzi: Maybe we shouldn't even have that `GenericEmail`, and encourage either creating email classes or using the email builder

In both cases, an `IsEmail` trait might indeed come in handy

packages/mailer/src/Builder/Email.php
brendt: A couple of alternatives:

- Add the builder methods to `GenericEmail` directly. 
- Have something like `Email::build()`, which returns an instance of `EmailBuilder` (similar to what we do with `QueryBuilder::select()` etc.), but this is not possible because `Email` is an interface

While I don't mind having a builder, I'm also not super convinced that it's needed. I know we've been going back and forth on this, but I have a hard time if we then also have a naming conflict.

And I know I'm a pain with naming conflicts, sorry for that :(

packages/mailer/src/Content.php
brendt: I like that! Keep the envelope as is, but move the content itself to the email :)

packages/mailer/src/Exceptions/ExpeditorWasMissing.php
brendt: I think `sender` is more clear, yeah

packages/mailer/src/GenericEmail.php
brendt: > Maybe we shouldn't even have that GenericEmail, and encourage either creating email classes or using the email builder

I like that :) Let's do that!

packages/mailer/src/Builder/Email.php
brendt: For clarification, the main problem I have with it is when you need both `Email` classes in the same context. You actually already run into the issue within the builder class itself where you need to import `Email as EmailInterface`. 

This happened way too much in my Laravel days and I cannot stand it 😅

innocenzi: Alright, you convinced me. :p

packages/mailer/src/GenericMailer.php
innocenzi: I explored it and that makes testing too hard. But I did add event support 👍

innocenzi:
Okay, after your review, are are the changes:

- `Envelope` and `Content` are still used because they are useful to avoid the use of a trait. However, the `GenericEmail` class can have parameters in its constructor, which avoids property-juggling and allows for a clean user-facing API:

```php
$this->mailer->send(new GenericEmail(
    subject: 'Hello',
    to: 'jon@doe.co',
    from: 'no-reply@tempestphp.com',
    text: 'Hello Jon',
    attachments: [
        Attachment::fromClosure(fn () => 'hello!'),
    ],
));
```

- I added a `build` method to `GenericEmail` and renamed the `Builer\Email` class to `EmailBuilder`. The intended use is now to call `GenericEmail::build()`, which conceptually makes a ton of sense.

- I removed the dedicated `DataAttachment`, `StorageAttachment` and `FileAttachment` classes in favor of a `Attachment` value object. This object has static constructors for attachments coming from the storage or the filesystem, for convenience:

```php
Attachment::fromStorage($storage, 'file.txt');
Attachment::fromFilesystem('./file.txt');
```

This is a good compromise; the `Attachment` class itself knows nothing about the storage, and there is no longer any service location, the storage has to be provided.

packages/mailer/src/EmailBuilder.php
```diff
@@ -0,0 +1,192 @@
...
+        return $this;
+    }
+
+    /**
+     * Adds attachments to the email.
+     */
+    public function attach(Attachment ...$attachments): self
+    {
+        foreach ($attachments as $attachment) {
+            if (! ($attachment instanceof Attachment)) {
```
brendt: This has been checked by PHP's type system: `Attachment ...$attachments`

packages/mailer/src/Content.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    /**
+     * @param Attachment[] $attachments
+     */
+    public function __construct(
```
brendt: Still not convinced we need `Content`. I'm ok with Envelope, but I think we need three parameters on the interface:

```php
public Envelope $envelope;
public string|View $content;
public array $attachments;
```

We loose the ability to differentiate between a view path and raw content (what used to be `$html` or `$text`). My proposal is that `$content` always assumes there's a view. Either it's a view object or a path to a view file. 

If we _really_ want to, we could have some kind of `raw` option in the builder which wraps whatever raw content in an instance of `GenericView`. Personally, I don't think we need `raw` atm, and I would just skip it for now, as it can be added later without needing a breaking change

packages/mailer/src/GenericEmail.php
```diff
@@ -0,0 +1,54 @@
...
+        get => new Content(
+            html: $this->html,
+            text: $this->text,
+            attachments: $this->attachments,
+        );
+    }
+
+    public function __construct(
+        public ?string $subject,
+        public null|string|array|Address $to,
```
brendt: I'm not a fan of the constructor on `GenericEmail` now accepting standalone parameters and then wrapping them in `Envelope`. I'd say just have people provide `Envelope`. (Assuming `Content` goes away)

packages/mailer/src/Testing/TestingMailer.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function send(Email $email): SentTestingEmail
+    {
+        $this->sent[] = $email;
+
+        $symfonyEmail = map($email)
+            ->with(fn (Email $from) => new EmailToSymfonyEmailMapper($this->mailerConfig, $this->viewRenderer)->map($from, null))
+            ->do();
+
+        $this->eventBus?->dispatch(new EmailSent($email));
```
brendt: I'm going to find it confusing that we have `EmailSent` and `SentEmail`, and both meaning something different. Can't we merge both of them in the `EmailSent` event? I noticed there's a testing implementation for `SentEmail`, but I figure those assertion methods can be moved to somewhere else.

brendt (dismissed):

packages/mailer/src/Content.php
brendt: Followup: we don't even need raw, as we can check `is_file` if `$contents` is a string

brendt:
Gonna work on this a bit myself

brendt:
Decided to not go with the view version for now. Let's first see how this goes

Issue #826, opened 2024-12-08
Tempest scripts has Windows-style line endings, which breaks the `env` call in the shebang


kevinfrom:
### Tempest Version

1.0-alpha4

### PHP Version

8.3

### Operating System

Windows, Other

### Description

Operation System: Windows/WSL2.
PHP version: 8.4.1

Some scripts, e.g. the `tempest` script has issues with Windows-style line endings (`\r\n`). By replacing all the Windows-style line endings with Unix-style line endings makes the scripts run as expected.

![billede](https://github.com/user-attachments/assets/e48d5ea2-cb53-47c1-ac65-16ba8265605b)

I checked my global Git config and I use `core.autocrlf`, meaning Git will replace all line endings with Windows-style line endings. Maybe we should set this value to false in the tempest repositories to avoid this issue?

### Steps to Reproduce

Run any tempest command, e.g. `tempest serve` in a WSL2 terminal.

ericsizemore:
I tend to use the following in my `.gitattributes` file:
```text
* text eol=lf
*.php text diff=php
*.md text diff=markdown
```

For me, personally, it doesn't seem to cause any issues. I am also on Windows with `core.autocrlf` set to `true`. 

> Maybe we should set this value to false in the tempest repositories to avoid this issue?

As far as I am aware, `core.autocrlf` is not something that can be set on a per repository level, unless you're speaking of having contributors changing this setting on their system. It is not recommended to set this to false on a Windows system. From what I remember, it's always recommended to create a `.gitattributes` file instead; so maybe could give `* text eol=lf` a go?


** My apologies if I misunderstood anything, it's very early in the morning and I'm running on no sleep :D lol

brendt:
Do you mean there are `\r\n` line endings in `tempest` which causes trouble?

kevinfrom:
@brendt yes exactly! @ericsizemore 's comment about gitattributes might be worth checking out. I can test this out when I'm home later.

brendt:
> I can test this out when I'm home later.

Thanks, and yes please do! Feel free to submit a PR to our `.gitattributes` :)

aidan-casey:
For what it's worth, a lot of this stuff is what is happening in Laravel as well, so might be a place to look for reference.

kevinfrom:
> For what it's worth, a lot of this stuff is what is happening in Laravel as well, so might be a place to look for reference.

Oh absolutely! I'll check out what Laravel does. Thanks!

PR #1229, opened 2025-05-10
chore(gitattributes): fix end-of-line handling

kapersoft:
Inspired from https://github.com/laravel/laravel/blob/master/.gitattributes

Fixes https://github.com/tempestphp/tempest-framework/issues/826


kapersoft:
It seems that the `* text=auto eol=lf`-change breaks tests on Windows. I'll investigate how to fix this.

kapersoft:
Fixed it by only applying the line endings for the cli php scripts.

@kevinfrom Can you confirm if this fixes your issue?

brendt:
I think this is ok to merge? If it breaks anything, I'm sure we'll find out soon :) Thanks for the PR!

PR #1245, opened 2025-05-14
feat(cache): separate internal and user caches

innocenzi:
This pull request is a complete overhaul of the cache component. The main purpose behind it is to have a similar implementation to the database and storage components in regards to the configuration. The internal caches (config, view, icon) are now completely separate caches with their own implementations.

### Summary of changes

- Refactored internal caches out of `tempest/cache`
- Abstracted Symfony Cache away with custom config objects
- Added several convenience methods
- Added cache locks
- Added `tempest about` insights
- Added testing utilities

### Comparison between previous and new API

Previously, configuring the user cache required instantiating one of Symfony's adapter in the `CacheConfig` object:

```php
return new CacheConfig(
    pool: new FilesystemAdapter(
        directory: __DIR__ . '/../../../../.cache',
    ),
);
```

Now, you may use one of the provided config objects:

```php
// filesystem
return new FilesystemCacheConfig(
    directory: internal_storage_path('/caches/project'),
);

// in-memory
return new InMemoryCacheConfig();

// redis (not implemented yet, requires us to add a Redis package)
return new RedisCacheConfig(
    prefix: env('REDIS_CACHE_PREFIX'),
    host: env('REDIS_HOST', default: '127.0.0.1'),
    username: env('REDIS_USERNAME'),
    password: env('REDIS_PASSWORD'),
);
```

And of course, it's now possible to add multiple caches by adding a tag to one of the config objects:

```php
// cache.config.php
return new FilesystemCacheConfig(
    directory: internal_storage_path('/caches/project'),
);

// another-cache.config.php
return new FilesystemCacheConfig(
    directory: internal_storage_path('/caches/another-cache'),
    tag: 'another-cache',
);

// or add one on-the-fly
$this->container->config(new FilesystemCacheConfig(
    directory: internal_storage_path('/caches/on-the-fly-cache'),
    tag: 'on-the-fly-cache',
));
```

The `cache:clear` and `cache:status` commands now separate user and internal caches. By default, `cache:clear` clears the default user cache. If more caches were created, you may select which ones to clear.

## Environment variables

Last but not least, the way caches are enabled or disabled have been simplified. Instead of having to have a `CACHE`, `PROJECT_CACHE`, `ICON_CACHE`, `VIEW_CACHE`, `CONFIG_CACHE`, and `DISCOVERY_CACHE` in `.env`, caches are disabled by default in a local environment, and enabled in production (except for the icon cache, which is always enabled by default).

Of course, they can still be disabled with those environment variables, but they are optional. However, the `CACHE` environment variable has been renamed to `INTERNAL_CACHES` for clarity.

In short, here are the new environment variables (all optional):

```.env
# Setting to `false` force-disables all internal caches (icon, view and config)
INTERNAL_CACHES=true

# Enable or disable discovery cache. Can be `true`, `partial` or `false`.
# `false` by default in dev, `true` by default in production.
DISCOVERY_CACHE=

# Enable or disable the icon cache. True by default all the time.
ICON_CACHE=

# Enable or disable the view cache.
# `false` by default in dev, `true` by default in production.
VIEW_CACHE=

# Enable or disable the config cache.
# `false` by default in dev, `true` by default in production.
CONFIG_CACHE=
```


brendt:
This all makes sense. I think it's better to merge before 1.0 though, because I believe it's technically a breaking change?

One remark

> except for the icon cache, which is always enabled by default

Shouldn't view cache be enabled by default as well? 

I still need to go through the code, but let's first hear your thoughts on merging it early?

innocenzi:
> Shouldn't view cache be enabled by default as well?

I'm not sure, I think it was not enabled by default before. I haven't really taken a look at how it works exactly, but if you think it should be, we can change that

> but let's first hear your thoughts on merging it early?

Totally fine by me (I would even prefer that)—but I don't mind if you prefer merging later

brendt:
> I'm not sure, I think it was not enabled by default before. I haven't really taken a look at how it works exactly, but if you think it should be, we can change that

I'm not sure either anymore, let's make a note to double check before merging

> Totally fine by me (I would even prefer that)—but I don't mind if you prefer merging later

Let's merge before beta (but after the ORM refactor so that I can review the code as well)

innocenzi:
> I'm not sure either anymore, let's make a note to double check before merging

I'm confident it wasn't enabled before! I just don't know if that was intended or not.

packages/cache/src/Cache.php
```diff
@@ -6,20 +6,88 @@
...
+
+    /**
+     * Determines whether the cache contains the specified key.
+     */
+    public function has(Stringable|string $key): bool;
+
+    /**
+     * Increments the value associated with the specified key by the specified amount. If the key does not exist, it is created with the specified amount.
+     */
+    public function increment(Stringable|string $key, int $by = 1): int;
```
brendt, line 72: Same for these methods. I'm not sure if they are really needed…

But please tell me I'm wrong :)

packages/cache/src/Config/CacheConfig.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Tempest\Cache\Config;
+
+use Psr\Cache\CacheItemPoolInterface;
+use Tempest\Container\HasTag;
+
+interface CacheConfig extends HasTag
```
brendt, line 8: I love it!

packages/cache/src/GenericCache.php
```diff
@@ -0,0 +1,169 @@
...
+{
+    public function __construct(
+        private(set) CacheConfig $cacheConfig,
+        public bool $enabled = true,
+        private ?CacheItemPoolInterface $adapter = null,
+    ) {
+        $this->adapter ??= $this->cacheConfig->createAdapter();
+    }
+
+    public function lock(Stringable|string $key, null|Duration|DateTimeInterface $expiration = null, null|Stringable|string $owner = null): Lock
```
brendt, line 26: I envisioned our cache interface to be pretty lightweight. Do we really need locks?

packages/cache/src/Testing/TestingCache.php
```diff
@@ -0,0 +1,209 @@
...
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Tempest\Cache\Cache;
+use Tempest\Cache\Config\CustomCacheConfig;
+use Tempest\Cache\GenericCache;
+use Tempest\Cache\GenericLock;
+use Tempest\DateTime\DateTimeInterface;
+use Tempest\DateTime\Duration;
+use Tempest\Support\Random;
+
+final class TestingCache implements Cache
```
brendt, line 19: `TestingCache` or `TestCache`?

brendt (commented):
Most of my remarks aren't real blockers, I think I rather just wanted to better understand your reasoning behind some features. 

It's a great PR, looking forward to hearing your thoughts on those comments, and then merging before beta.2.

Apart from the cache interface changes, what kind of things will break for users because of this? For example: we merge this, what should be changed on tempest-docs production to make sure caching still works the way it's supposed to work?

packages/cache/src/Cache.php
```diff
@@ -6,20 +6,88 @@
...
+    /**
+     * Gets the values associated with the specified keys from the cache. If a key does not exist, null is returned for that key.
+     *
+     * @template TKey of array-key
+     * @template TValue
+     *
+     * @param iterable<TKey,TValue> $array
+     * @return array<TValue,mixed>
+     */
+    public function getMany(iterable $key): array;
```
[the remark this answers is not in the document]
innocenzi: Admittedly, I've never used these myself. I added them because I expected them to be needed at some point and the implementation was simple. We can remove them, but expect them to be PR'd back by someone at some point as Tempest grows.

My opinion is that we should keep them

packages/cache/src/Cache.php
innocenzi: Those however, I have used them! It's so much nicer than having to fetch the value, increment it yourself and update it back.

packages/cache/src/Testing/TestingCache.php
innocenzi: Anything as long as it's consistent with the other components (eg. `TestingStorage`). I prefer `TestingCache` as in "the cache for testing".

packages/cache/src/GenericCache.php
innocenzi: Huh, absolutely! Locking is a very important mechanic to have. If you mean "do we need them in the interface", technically no, however it makes things quite harder. In the cache implementation, we have access to everything we need.

May I ask why do you want the interface to be lightweight? With this PR, this interface should not be expected to be implemented in userland anymore, as opposed to the previous design

innocenzi:
> Apart from the cache interface changes, what kind of things will break for users because of this? For example: we merge this, what should be changed on tempest-docs production to make sure caching still works the way it's supposed to work?

Only the `cache:clear` command in the deploy script I believe. We need to add the `--internal` flag.

packages/cache/src/Testing/TestingCache.php
brendt: Fair enough!

packages/cache/src/Cache.php
brendt: Yeah I agree that I got pretty excited reading about these :p Let's keep them in

packages/cache/src/GenericCache.php
brendt: It's mainly that there are some interfaces in Tempest that feel like they are trying to do A LOT. The container is another example. I don't think it's worth refactoring the container at this point, but that's a case where I mixed up interface and concrete implementation responsibilities. But don't tell Aidan, because he's going to say I should have gone with PSR 😅

I'm mainly wondering whether locking is a method that belongs on the base `Cache` interface, or whether it should go on a separate interface that caches can implement, when they need it. Like eg. `LockableCache`

innocenzi: Oh, I see. I personally think it's perfectly fine here, but if you want to move it, it's fine as well as long as the DX stays the same

packages/cache/src/Cache.php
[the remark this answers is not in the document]
brendt: Let's keep 👍

packages/cache/src/GenericCache.php
brendt: Let's keep it as-is for now then :)

brendt:
@innocenzi I'm fine merging this, just wanted to double-check with you if that's ok?

innocenzi:
@brendt Yes, let's go! I'm fixing the failures that are due to the merge of `main`, then I'll merge and update `tempestphp/tempest-docs`

Issue #928, opened 2025-02-28
Improve API development


jseparovic1:
In recent days it is more and more common to use PHP as a API layer with separate frontend. In this setup it is crucial to document API and allow better understanding between the teams.

While it is possible to achieve this using various packages (e.g. https://zircote.github.io/swagger-php/, https://github.com/thephpleague/openapi-psr7-validator), having official (and automated) integration would allow more flexibility.   

Core features:
- detect routes
- detect route attributes
- detect request properties (based on controller parameters)
- detect response properties (via https://tempestphp.com/docs/framework/controllers/#response-objects?)
- validate request according to schema `before` deserialization to object (long standing problem in api platform)
- generate api docs 

Inspiration and references
1. https://fastapi.tiangolo.com/#example-upgrade FastAPI does this best in my opinion. 
2. https://api-platform.com/  provides much more features but in reality it might be to complex

brendt:
I'm very much open to improve Tempest in this regard. I think I'm with you on the whole list, but we'd need to flesh it out a lot more and split in standalone tasks to get started on it.

jseparovic1:
@brendt I agree, this is just the introduction of idea or tip of the iceberg. There are whole frameworks built around this concept. I posted it to spark some discussions and see what could be potential steps to start with this.

brendt:
I've assigned it to the post-1.0 milestone, but we might pick it up earlier depending on the time available

gturpin-dev:
How can we organize this to determine some limits, what we want to support, what DX we want to achieve and so on ?

innocenzi:
> How can we organize this to determine some limits, what we want to support, what DX we want to achieve and so on ?

For such a feature, I'd like people interested in tackling this to open a discussion first, discuss the features they want to implement and how, before doing any work—because there is a lot to do in this space

brendt:
On Discord, preferably :)

gturpin-dev:
My bad we already have opened the discussion here https://discord.com/channels/1236153076688359495/1345315356696121344

innocenzi:
I'm closing this in favor of the discussion on Discord, in order to clean up the GitHub issues

PR #1246, opened 2025-05-14
[WIP] feat(api): add API "builder"

erikaraujo:
>[!WARNING]
>This is VERY MUCH a Working in Progress.

Discord discussion [here](https://discord.com/channels/1236153076688359495/1345315356696121344).

This is just to show-case an idea/PoC for a an API Development solution for Tempest.

Here's how I envision a developer would use such a feature:

## Sample File structure

Just like the rest of Tempest, the file structure DOES NOT matter - however, just for the sake of having an example, here it is.
```
.
└── app/
    ├── ApiResources/
    │   └── UserResource.php
    └── Models/
        ├── Post.php
        └── User.php
```

## Using in a dedicated resource class

Here's how a `UserResource` class could look like.

```php
#[ResourceRecord(User::class)]
final class UserResource
{
    use IsApiResource;
}
```

## Using directly in a model (🤮)

```php
final class Post
{
    use IsApiResource;

    public Id $id;

    public string $title;

    //...
}
```

## Using Only certain CRUD endpoints

```php
#[ResourceRecord(User::class)]
final class UserResource
{
    use HasApiIndexEndpoint;
    use HasApiShowEndpoint;
}
```

## Overwriting defaults

In any class that uses `IsApiResource`, defaults can be overwritten:

```php
#[ResourceRecord(User::class)]
final class UserResource
{
    use IsApiResource;
    
    public static function getResourceApiVersion(): ?string
    {
        return 'v2';
    }

    public static function getResourcePagination(): ?Pagination
    {
        return new OffsetPagination();
    }

    public static function getResourceSearchableColumns(): ?array
    {
        return ['name', 'email'];
    }

    public static function getResourceMiddlewares(): array
    {
       return [
            LogRequestMiddleware::class,
        ];
    }
}
```


gturpin-dev:
The related issue here #928

PR #1252, opened 2025-05-18
feat(kv-store): introduce key-value store component

innocenzi:
This pull request adds a new `tempest/kv-store` package, with initial support for Redis through the `redis` PHP extension or `predis`. Tempest will try to use the PHP extension if it's there, and fall back to predis otherwise. If both are not available, an exception is thrown.

```php
// redis.config.php
new RedisConfig(
    prefix: 'my_app:',
    host: env('REDIS_HOST'),
    port: env('REDIS_PORT'),
    username: env('REDIS_USERNAME'),
    password: env('REDIS_PASSWORD'),
);

// interact with Redis
use Tempest\KeyValue\Redis\Redis;

$redis = $this->container->get(Redis::class);
$redis->set('key', 'value');
$redis->command('ping');
```

The `Redis` interface is kept very simple for now—more specific methods may be added later. Similarly, there's no cluster support in this pull request, but it should be trivial to add through a new config.


innocenzi:
Marking this as draft to have a chance to test it with the cache implementation once it's merged

innocenzi:
I don't get why the connection exception is not caught...

brendt:
> Marking this as draft to have a chance to test it with the cache implementation once it's merged

Did you get a chance of doing that?

packages/kv-store/composer.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "tempest/kv-store",
```
brendt, line 2: I'd call it `key-value-store`

packages/kv-store/src/Redis/Config/RedisConfig.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace Tempest\KeyValue\Redis\Config;
+
+use UnitEnum;
+
+final class RedisConfig
```
brendt, line 7: Shouldn't this implement `HasTag`?

packages/kv-store/src/Redis/Config/Scheme.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Tempest\KeyValue\Redis\Config;
+
+/**
+ * Represents the connection scheme used to connect to Redis (predis only).
+ */
+enum Scheme: string
```
brendt: What I don't like is that now there's a naming collision between Smtp Schemes and Redis Schemes. Let's rename to `RedisScheme`

packages/kv-store/src/Redis/MissingRedisException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tempest\KeyValue\Redis;
+
+use Exception;
+use Predis;
+
+final class MissingRedisException extends Exception implements RedisException
```
brendt: Rename to `ReddisExtensionWasMissing`

packages/kv-store/src/Redis/Redis.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Tempest\KeyValue\Redis;
+
+use Stringable;
+use Tempest\DateTime\DateTimeInterface;
+use Tempest\DateTime\Duration;
+
+interface Redis
```
brendt, line 9: Do we need an interface? If so, I feel like `Redis` is not an accurate enough name. Is this a client that interacts with a Redis server? Is this a gateway that knows how to talk to it? IDK

brendt: Edit, having seen the implementations, I think `RedisClient` is more accurate

packages/kv-store/src/Redis/PredisClient.php
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        if (is_array($value)) {
+            return json_encode($value);
+        }
+
+        return $value;
+    }
+
+    public function getClient(): Client
```
brendt, line 102: I never really like `$client->getClient()`

1. Do we need to expose the client we're wrapping
2. Can't it be a `private(set)` or `readonly` property?

packages/kv-store/src/Redis/RedisInitializer.php
```diff
@@ -0,0 +1,67 @@
...
+use Predis;
+use Tempest\Container\Container;
+use Tempest\Container\DynamicInitializer;
+use Tempest\Container\Singleton;
+use Tempest\EventBus\EventBus;
+use Tempest\KeyValue\Redis\Config\RedisConfig;
+use Tempest\Reflection\ClassReflector;
+use UnitEnum;
+
+final class RedisInitializer implements DynamicInitializer
```
brendt: `RedisClientInitializer`?

brendt (changes requested):

packages/kv-store/src/Redis/Config/RedisConfig.php
innocenzi: I actually don't think tag support for this feature is necessary, like emails. I'll remove it.

packages/kv-store/src/Redis/Redis.php
innocenzi: I initially chose `Redis` as a name because I how DI looks:

```php
public function __construct(
    public readonly Redis $redis,
) {}
```

> Is this a client that interacts with a Redis server? Is this a gateway that knows how to talk to it?

Both sound the same to me. Can we keep `Redis`? 😁

packages/kv-store/src/Redis/PredisClient.php
innocenzi: 1. Yes we do, because we need that escape hatch currently
2. I considered it, but that would kind of conflict with the injected `$client` in the implementation classes:

```php
public function __construct(
    private \Redis $client,
    private RedisConfig $config,
    private ?EventBus $eventBus = null,
) {}
```

I would need to disable property promotion for the constructor's `$client` and assign it in the constructor... yuck. Is `getClient()` that bad?

packages/kv-store/src/Redis/RedisInitializer.php
innocenzi: See previous response, will rename if you really don't want a beautiful, nice, clean `Redis` interface 😌

packages/kv-store/composer.json
innocenzi: "KV" is the standard term for those (eg. [Cloudflare's](https://developers.cloudflare.com/kv/)), but if you want, we can change it. Just confirm it to me

innocenzi:
> Did you get a chance of doing that?

No actually! Will try to find some time to do so

packages/kv-store/composer.json
brendt: oh ok. It's not a big issue, I'm fine with `kv` as well

packages/kv-store/src/Redis/Redis.php
brendt: Come to think of it, we also call the interface that connects with the database simply `Database`, so I reckon this is fine :)

packages/kv-store/src/Redis/PredisClient.php
brendt: It's not super bad, it's just that I'd prefer to avoid a getter. 

We do sometimes indeed remove a promoted property and manually assign it. Personally I prefer it over a getter, but it's not a deal breaker for me

packages/kv-store/src/Redis/RedisInitializer.php
brendt: Yep, you convinced me about Redis :)

packages/kv-store/src/Redis/PredisClient.php
innocenzi: I suggest keeping the getter, as this makes both implementation slightly harder to read 🙏🏻

brendt (approved):

PR #1257, opened 2025-05-19
feat(datetime): support tempest datetime in validator and mapper

innocenzi:
I thought it was important before 1.0, so here's support for `tempest/datetime` in `tempest/mapper` and `tempest/validator`, as well as some clean up and minor improvements.

- I updated the `BeforeDate`, `AfterDate` and `DateTimeFormat` rules to support both `tempest/datetime` instances and native `\DateTimeInterface` ones. I also added a new `BetweenDates` rule, because why not.

- I updated `tempest/mapper`'s caster and serializer factories to support both `tempest/datetime` instances and native `\DateTimeInterface` ones.

- Additionally, I added more convenience methods to `DateTime`—we can now create an instance from a timestamp (we could before, but less easily), add/substract single units, and set the instance to the start/end of the current day, month, or year:

```diff
- DateTime::fromTimestamp(Timestamp::fromParts($timestamp));
+ DateTime::parse($timestamp);

- DateTime::fromTimestamp(Timestamp::fromParts(new \DateTime()->getTimestamp()));
+ DateTime::parse(new \DateTime());

- $now->plusDays(1);
+ $now->plusDay();
```


innocenzi:
All of this is now documented: https://tempestphp.com/main/features/datetime

PR #1258, opened 2025-05-19
fix(database): properly display mysql and postgresql versions in `about` command

Bapawe:
This PR fixes an issue where the regex patterns used for MySQL and PostgreSQL version insights were swapped. Additionally, it updates the handling of MySQL's VERSION() output, which may include a suffix with server build or configuration details.


innocenzi:
Very nice, thank you!

PR #1259, opened 2025-05-20
fix(database): postgres support

### LINKED ISSUE #1255: PostgreSQL driver uses backticks
MySQL allows the usage of backticks for identifiers, but PostgreSQL does not—normal quotes are used instead. Here's a simple reproduction (must have a pgsql database running):

```php
get(Container::class)->config(new PostgresConfig(
    username: 'postgres',
    database: 'testing',
));

$db = get(Database::class);

// new pgsql test table
// $db->execute(new Query(<<<SQL
//     CREATE TABLE users (
//         id SERIAL PRIMARY KEY,
//         name TEXT NOT NULL,
//         email TEXT NOT NULL
//     );
// SQL));

// insert user
$db->execute(new Query(<<<SQL
    INSERT INTO users (name, email) VALUES ('Jon', 'jon@doe.co')
SQL));

// query
$query = query('users')
    ->select('id', 'name', 'email')
    ->whereField('id', 1);

// dd($query->toSql());
dd($query->first());
```

Note that backticks are not just used in `FieldStatement`, but in many statements of our current implementation

brendt:
Closes #1255


brendt:
TODO: 

- [ ] Fix windows/postgres action

PR #1260, opened 2025-05-20
feat(router): support server-sent events

innocenzi:
This pull request adds basic support for server-sent events:

```php
#[Get('/sse')]
public function __invoke(): EventStream
{
    $stream = /** ... */;

    return new EventStream(function () use ($stream) {
        foreach ($stream as $message) {
            yield $message;
        }
    });
}
```


brendt:
Interesting! I don't have much to add here, feel free to merge

PR #1262, opened 2025-05-21
fix(database): support semicolons in queries

innocenzi:
https://github.com/tempestphp/tempest-framework/pull/1259 introduces support for multiple queries in a single `execute` call, but does that without accounting for the `;` character possibly being present anywhere in a valid query.

This PR is a failing test that shows how inserting a column with a `;` breaks the `execute` method


innocenzi:
@brendt personally, I don't think we can support multiple queries in a single `execute` statement without a parser, and that we should drop the feature entirely

brendt:
The only reason we need it is because Postgres doesn't allow multiple queries in one prepared statement, and we need multiple queries only during migrations in order to add indices. But I've now solved that on the migration manager level and without exploding on `;`, so if I manage to get the tests working, we're good :)

Issue #274, opened 2024-05-27
Environment-specific logger channels


brendt:
We need a way to define environment-specific logger channels. 

My idea was to add an `environments` array to all channels, an add `getEnvironments(): ?array` on the interface. 

```php
return new LoggerConfig(
    channels: [
        new DailyLogChannel(
            path: 'path/to/log',
            environments: [Environment::PRODUCTION],
        ),
        new AppendLogChannel(
            path: 'path/to/log',
            environments: [Environment::DEV],
        ),
    ],
)
```

PR #1266, opened 2025-05-22
feat(core): load `local` and `production` configurations last

innocenzi:
This pull request changes the order of loaded config to put `.local.config.php` and `.production.config.php` last, so they can override their "default" counterpart.

This allows having different configurations in different environments. For instance, I may have a `storage.config.php` using S3 by default, but a `storage.local.config.php` using a filesystem for local development.

Related: https://github.com/tempestphp/tempest-framework/issues/274

### DIFF
--- a/packages/core/src/Kernel/LoadConfig.php
+++ b/packages/core/src/Kernel/LoadConfig.php
@@ -8,9 +8,12 @@ use FilesystemIterator;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use SplFileInfo;
+use Tempest\Core\AppConfig;
 use Tempest\Core\ConfigCache;
 use Tempest\Core\Kernel;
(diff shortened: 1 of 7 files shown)


innocenzi:
Why not support both? People preferring conditionals in a single file can do this, and those preferring adding a `storage.dev.config.php` file to `.gitignore` can do that as well

That being said, Tempest should have a "recommendation" regarding the best way to handle this, because people are going to look for the idiomatic, Tempest way

brendt:
Ok two thoughts:

- I believe this PR prioritizes one over the other, but still loads all? If we allow this syntax, it shouldn't load non-matching environment config files at all. 
- I think the recommendation should be: 

1. Use `.config.php` by default
2. Use `env()` when possible
3. Use separate, environment specific config files when needed
4. If you need even more flexibility, do whatever you want by making your config files conditional

Issue #1281, opened 2025-05-26
Internal cache path


brendt:
Someone mentioned that, when using apache, our internal cache fails because apache is run as a separate user, without write permissions on `vendor/`. I'm not sure if this is the default behavior or not. A couple of solutions:

- Make internal cache easier to configure, or at least document how
- Write internal cache to somewhere else by default
- Add a proper exception that explains what's wrong when internal cache isn't writeable

brendt:
Input from Aidan: 

> Generally speaking you don’t want your web server user to have write access to the app directory. That’s why Laravel has a separate storage/cache directory that the web server is given explicit write permissions to.

This makes sense, we should move our internal cache directory somewhere else

innocenzi:
We can have a top-level `.tempest` directory instead

PR #1285, opened 2025-05-27
refactor(core): move internal storage to project root

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1281

Before merging, is `.tempest` the ideal name? Could also be `storage`, `.internal`, or something else. I like a dot directory because it's naturally ignored when scanning the hierarchy, but the name "tempest" is not generic and generally I don't like that

Also, we need to add it to `.gitignore` in `tempestphp/tempest-app`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /package-lock.json
 
 # Top-level+Sub directories
+.tempest/
 .cache/
 .idea/
 build/
--- a/packages/core/src/FrameworkKernel.php
+++ b/packages/core/src/FrameworkKernel.php
@@ -166,7 +166,7 @@ final class FrameworkKernel implements Kernel
 
     public function registerInternalStorage(): self
     {
-        $path = $this->root . '/vendor/.tempest';
+        $path = $this->root . '/.tempest';
 
         if (! is_dir($path)) {
             mkdir($path, recursive: true);


innocenzi:
@erikaraujo That doesn't solve the issue I have with `.tempest`, and `.tempest` would be empty besides that "internal" directory 😅

innocenzi:
There's already a few subdirectories for each concern, this example is missing local file storage and I believe there may be a few others

CleanShot 2025-05-28 at 15 29 32

I think `.tempest` might be the most explicit name unfortunately

aidan-casey:
@innocenzi - I think the question I would ask is whether there are going to be other items we need to store (e.g., `uploads`). If so, are we then creating multiple directories?

IMO, I think Laravel gets this fairly spot on in their app structure.

innocenzi:
I'm not sure I understand what you're asking. Local storage would be in `.tempest/storage` by default, in tests it would be `.tempest/tests/storage/{tag}`. Users may create additional storages wherever they want, if they chose to use `internal_storage_path('/some/path')` that would be relative to `.tempest`.

aidan-casey:
I'm of the mind `storage` is more clear to user since it will include uploads, etc.

While I can appreciate the concept of a hidden directory, I can just imagine the number of people trying to figure out where the files are because they didn't list hidden directories, etc.

innocenzi:
As far as I know, PhpStorm, WebStorm and VSC all show dot files and directories. A `storage` directory wouldn't feel like an "internal storage" though, and we'd have to move stuff like `current_discovery_strategy` somewhere else, right?

Issue #1287, opened 2025-05-31
`InvalidClosingTag` throwing on commented-out code


erikaraujo:
### Tempest version

main-dev

### PHP version

8.4

### Operating system

macOS

### Description

`InvalidClosingTag` exception is being thrown by this code:

```html
<h2 class="text-lg font-bold text-gray-900 dark:text-white flex items-center gap-2">
    <!-- <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="w-3.5 h-3.5 text-zinc-900 dark:text-zinc-50">
        <path d="M16 21v-2a4 4 0 0 0-4-4H6a4 4 0 0 0-4 4v2"></path>
        <circle cx="9" cy="7" r="4"></circle>
        <path d="M22 21v-2a4 4 0 0 0-3-3.87"></path>
        <path d="M16 3.13a4 4 0 0 1 0 7.75"></path>
    </svg> -->

    <x-slot name="header" />
</h2>
```

It's not *only* because the code is commented out - it also has to do with the `/svg` in the `xmlns` attribute.

### Steps to reproduce

Add the following snippet to any view:

```html
<h2>
    <!-- <svg xmlns="http://www.w3.org/2000/svg">
    </svg> -->
</h2>
```

PR #1288, opened 2025-05-31
fix(view): `InvalidClosingTag` should ignore commented out code attributes

erikaraujo:
This PR introduces a failing test for #1287.

The issue seems to be a combination of the code being commented out and the `/svg` part of the `xmlns` attribute of the `svg` tag in the following snippet:

```html
<h1>
    <!-- <svg xmlns="http://www.w3.org/2000/svg">
    </svg> -->
</h1>
```


brendt:
Indeed, this was fixed. Gonna merge the PR though because an additional test is always good

Issue #1114, opened 2025-04-04
Internationalization support


innocenzi:
At some point, we'll need to support i18n. I'd prefer an implementation using the ICU format (v2) syntax like Symfony (v1) rather than a placeholder-based one like Laravel.

https://unicode-org.github.io/icu/userguide/format_parse/messages/mf2.html

PR #1294, opened 2025-06-21
feat(intl): add localization support

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1114

This pull request's main goal is to add a new `tempest/intl` package with a translator and a [MessageFormat 2.0](https://messageformat.unicode.org/docs/quick-start/) implementation (parser and formatter included).

I also moved `Tempest\Support\Number`, `Tempest\Support\Language`, and `Tempest\Support\Locale` to this package, because the scope makes sense.
<br />

### MessageFormat

> [!NOTE]
> I chose to go ahead with MessageFormat 2.0 instead of 1.0 (which is built in PHP's `Intl` extension) because this specification is more flexible and future-proof. For instance, it supports [markup](https://messageformat.unicode.org/docs/reference/markup/) and [custom functions](https://messageformat.unicode.org/docs/reference/functions/).

The `Tempest\Intl\MessageFormat\Formatter\MessageFormatter` implementation provides a lower-level way to format MessageFormat 2.0 messages.

This would generally not be used in userland, but it's available (and used internally).

Note that the pluralization rules are generated thanks to publicly-available [CLDR](https://github.com/unicode-org/cldr-json) plural rules from the unicode project. To re-generate the class, we can just run the `bin/plural-rules` script.
<br />

### Translator

The `Tempest\Intl\Translator` interface has a `translate` and `translateForLocale` method, which are both straightforward. They accept a translation key and arguments, and return the translated message (or the translation key on failure).

```php
$translator->translate('hello', name: 'Jon Doe'); // Hello, Jon Doe!
```

When a translation is missing or has failed, the `TranslationMiss` or `TranslationFailed` events will be dispatched, respectively.
<br />

### Messages

Messages are stored as YAML or JSON files. YAML is recommended because MF2 messages are often multiline, which is hard to work with in JSON. We could add support for PHP files as well, but the DX is better in YAML.

By default, Tempest will discover files like `messages.fr.yaml`, where the locale part is a valid, known locale (eg. `messages.abcd.yaml` will not be discovered, but `locales.en_UK.yaml` will).

Discovery will add the file paths to the `IntlConfig`, which will be used by the `Tempest\Intl\Catalog` implementation to load translation messages.
<br />

### Configuration

From now on, the current locale is stored in `Tempest\Intl\IntlConfig` in the `currentLocale` property.

Future features that depend on the current locale should be built upon this. By default, the current locale will be `Locale::default()`, which uses [`Locale::getDefault()`](https://www.php.net/manual/en/locale.getdefault.php).

### DIFF
--- a/packages/vite/src/Installer/ViteInstaller.php
+++ b/packages/vite/src/Installer/ViteInstaller.php
@@ -97,7 +97,7 @@ final class ViteInstaller implements Installer
             ignoreNonExisting: true,
         );
 
-        $packageManager = PackageManager::detect(root_path());
+        $packageManager = PackageManager::detect(root_path()) ?? PackageManager::NPM;
 
         $this->console->instructions([
             $shouldInstallTailwind


innocenzi:
Documentation PR: https://github.com/tempestphp/tempest-docs/pull/86

aidan-casey:
Nice, PR @innocenzi! I haven't reviewed all the code, but I like the API.

Quick question, why was PHPUnit removed as a dev dependency from other sub-packages?

innocenzi:
> Quick question, why was PHPUnit removed as a dev dependency from other sub-packages?

I removed it as an attempt to fix that unknown CI error, since it was happening only for those packages. I'm not sure why none of the other packages depend on PHPUnit either

packages/intl/bin/plural-rules.php
```diff
@@ -0,0 +1,369 @@
...
+require_once getcwd() . '/vendor/autoload.php';
+
+use Tempest\Console\Console;
+use Tempest\Console\ConsoleApplication;
+use Tempest\Support\Filesystem;
+use Tempest\Support\Json;
+
+use function Tempest\get;
+
+final class PluralRulesMatcherGenerator
```
brendt, line 18: Just for my own future-sanity, could you add a doc comment here explaining what this file is used for?

packages/intl/src/Catalog/GenericCatalog.php
```diff
@@ -0,0 +1,41 @@
...
+            array: $this->catalog,
+            key: "{$locale->value}.{$key}",
+            default: Arr\get_by_key(
+                array: $this->catalog,
+                key: "{$locale->getLanguage()}.{$key}",
+            ),
+        );
+    }
+
+    public function add(Locale $locale, string $key, string $message): self
```
brendt, line 35: I need to better understand: `add` will add an entry into the in-memory catalog, but won't change any source files this catalog was loaded from. 

My assumption was that this was needed to initialize the catalog, but looking at its initializer, we never call that function there. 

That leaves the question: do we actually need this `add` function (maybe we do and I just need to keep on reading :) )

packages/intl/src/GenericTranslator.php
```diff
@@ -0,0 +1,52 @@
...
+        private(set) IntlConfig $config,
+        private(set) Catalog $catalog,
+        private MessageFormatter $formatter,
+        private ?EventBus $eventBus = null,
+    ) {}
+
+    public function translateForLocale(Locale $locale, string $key, mixed ...$arguments): string
+    {
+        if (! $this->catalog->has($locale, $key)) {
+            $this->eventBus?->dispatch(new TranslationMiss(
```
brendt, line 23: Out of curiosity: on translation miss, we dispatch an event, and the resulting string will simply be the untranslated string with its placeholders. I assume we go the event route instead of throwing an exception because we don't want to hard-error whenever a translation is missing.

I don't think we do anything with the event yet, did you plan on adding some kind of debugging tool/log later that collects missing translations?

packages/intl/src/MessageFormat/Parser/Parser.php
```diff
@@ -0,0 +1,627 @@
...
+use Tempest\Intl\MessageFormat\Parser\Node\MessageNode;
+use Tempest\Intl\MessageFormat\Parser\Node\ParsingException;
+use Tempest\Intl\MessageFormat\Parser\Node\Pattern\Pattern;
+use Tempest\Intl\MessageFormat\Parser\Node\Pattern\Placeholder;
+use Tempest\Intl\MessageFormat\Parser\Node\Pattern\QuotedPattern;
+use Tempest\Intl\MessageFormat\Parser\Node\Pattern\Text;
+use Tempest\Intl\MessageFormat\Parser\Node\SimpleMessage;
+use Tempest\Intl\MessageFormat\Parser\Node\Variable;
+
+final class Parser
```
brendt: Could we rename this to MessageFormatParser or something alike? I like class names that are identifiable regardless of the namespace, especially class names that are as generic as "parser"

packages/intl/src/MessageFormat/Parser/Parser.php
```diff
@@ -0,0 +1,627 @@
...
+        return $char;
+    }
+
+    private function peek(int $length = 1): string
+    {
+        if ($this->isEof()) {
+            return '';
+        }
+
+        return mb_substr($this->input, $this->pos, $length, 'UTF-8');
```
brendt: Just FYI, I did a lot of performance testing with Tempest's view lexer, and stuff like `$this->isEof()` instead of directly writing `$this->pos >= $this->len` does have an impact when you're in a parsing context.

I don't think it would be too noticeable when parsing messages, since they are pretty small compared to view files. So I'm totally fine keeping it as is for now, but we might need to reconsider if this turns out to be a performance bottleneck after all.

brendt (dismissed):
Great PR! I had a couple of questions, but I think this is almost ready to merge

packages/intl/bin/plural-rules.php
innocenzi: Hey, sure, good idea. TL;DR, it compiles CLDR rules to code so we don't have to parse the file every time.

packages/intl/src/Catalog/GenericCatalog.php
innocenzi: This is there so users can add translations on-the-fly, which is sometimes needed (eg. translation comes from elsewhere than the files on the disk)

I actually use this in tests as well

packages/intl/src/GenericTranslator.php
innocenzi: You're spot on!

Translation misses will simply return the translation key.

Events are there so user can monitor their translations if they need to. I can potentially add this to the debugging tool I'm working on as well.

Generally speaking, I think it's great when lots of events are dispatched throughout the apps. It's useful for monitoring and doing things

packages/intl/src/MessageFormat/Parser/Parser.php
innocenzi: Right, this makes sense. I think this parser has some room for optimization in general, so if I need to debug performance, I'll look into that first 👍

packages/intl/src/Catalog/GenericCatalog.php
brendt: I saw the tests later during my review :D  

and indeed, it makes sense to have a dynamic option when needed!

packages/intl/src/GenericTranslator.php
brendt: Absolutely! Great to have it in place already. Puuuurfect

brendt:
I'm good to merge!

PR #1307, opened 2025-06-26
refactor(router): refactor router logic to middleware

### LINKED ISSUE #1306: Handle route exceptions via middleware
1. There's a discrepancy between how console commands are handled vs route requests. For console commands, all exception handling is done via a middleware: \Tempest\Console\Middleware\ResolveOrRescueMiddleware. When a command is not found, the rescue part will trigger. With HTTP routing, that "not found" logic is handled directly in the router, without the ability to hook into the flow with middleware
2. GenericRouter actually has two ways of generating a 404: when there's no matched route, OR when a NotFoundException is thrown. This means that you could manipulate 404's when they originate from a NotFoundException, but not when there's no matched route.

brendt:
Closes #1306

### DIFF
--- a/packages/router/src/Exceptions/NoMatchedRoute.php
+++ b/packages/router/src/Exceptions/NoMatchedRoute.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Tempest\Router\Exceptions;
+
+use Exception;
+
+final class NoMatchedRoute extends Exception
+{
+    public function __construct()
+    {
+        parent::__construct('No matched route was registered in the container. Did you remove `\Tempest\Router\MatchRouteMiddleware` from the middleware stack by any chance?');
+    }
+}
--- a/packages/router/src/GenericRouter.php
+++ b/packages/router/src/GenericRouter.php
@@ -8,23 +8,16 @@ use BackedEnum;
 use Psr\Http\Message\ServerRequestInterface as PsrRequest;
 use Tempest\Container\Container;
 use Tempest\Core\AppConfig;
-use Tempest\Http\GenericRequest;
 use Tempest\Http\Mappers\PsrRequestToGenericRequestMapper;
-use Tempest\Http\Mappers\RequestToObjectMapper;
-use Tempest\Http\Mappers\RequestToPsrRequestMapper;
 use Tempest\Http\Request;
 use Tempest\Http\Response;
-use Tempest\Http\Responses\Invalid;
-use Tempest\Http\Responses\NotFound;
 use Tempest\Http\Responses\Ok;
-use Tempest\Mapper\ObjectFactory;
 use Tempest\Reflection\ClassReflector;
 use Tempest\Router\Exceptions\ControllerActionHasNoReturn;
 use Tempest\Router\Exceptions\InvalidRouteException;
-use Tempest\Router\Exceptions\NotFoundException;
+use Tempest\Router\Exceptions\NoMatchedRoute;
 use Tempest\Router\Routing\Construction\DiscoveredRoute;
 use Tempest\Router\Routing\Matching\RouteMatcher;
-use Tempest\Validation\Exceptions\ValidationException;
 use Tempest\View\View;
 
 use function Tempest\map;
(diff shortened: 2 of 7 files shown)


packages/router/src/Exceptions/HttpErrorResponseProcessor.php
```diff
@@ -1,43 +0,0 @@
...
-
-namespace Tempest\Router\Exceptions;
-
-use Tempest\Core\AppConfig;
-use Tempest\Http\HttpException;
-use Tempest\Http\Response;
-use Tempest\Router\ResponseProcessor;
-use Tempest\Router\RouteConfig;
-
-final readonly class HttpErrorResponseProcessor implements ResponseProcessor
```
brendt, line 11: Removed this class in favor of `HandleRouteExceptionMiddleware`

packages/router/src/Exceptions/NoMatchedRoute.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Tempest\Router\Exceptions;
+
+use Exception;
+
+final class NoMatchedRoute extends Exception
```
brendt, line 7: This exception should never happen, unless someone messed with the default middleware stack.

packages/router/src/GenericRouter.php
```diff
@@ -42,43 +35,28 @@ public function __construct(
...
-        if (! ($request instanceof PsrRequest)) {
-            $request = map($request)->with(RequestToPsrRequestMapper::class)->do();
-        }
-
-        $matchedRoute = $this->routeMatcher->match($request);
-
-        if ($matchedRoute === null) {
-            return new NotFound();
+        if (! ($request instanceof Request)) {
+            $request = map($request)->with(PsrRequestToGenericRequestMapper::class)->do();
```
brendt, line 39: Mapping to Tempest request, from this point on we don't have to worry about PSR anymore

packages/router/src/GenericRouter.php
```diff
@@ -42,43 +35,28 @@ public function __construct(
...
     public function dispatch(Request|PsrRequest $request): Response
     {
-        return $this->processResponse(
-            $this->processRequest($request),
-        );
-    }
-
-    private function processRequest(Request|PsrRequest $request): Response
-    {
-        if (! ($request instanceof PsrRequest)) {
```
brendt, line 52: Moved all this logic to various middeleware classes:

- `MatchRouteMiddleware`
- `HandleRouteExceptionMiddleware`

packages/router/src/GenericRouter.php
```diff
@@ -93,10 +71,7 @@ private function getCallable(MatchedRoute $matchedRoute): HttpMiddlewareCallable
 
         $callable = new HttpMiddlewareCallable(fn (Request $request) => $this->createResponse($callControllerAction($request)));
 
-        $middlewareStack = $this->routeConfig
```
brendt, line 96: We don't have a matched route at this point anymore, because that logic is moved to a middleware. That's why we provide a new `HandleRouteSpecificMiddleware`, which will run at the very end of the stack. It will gather all route-specific middleware and run those as well.

packages/router/src/GenericRouter.php
```diff
@@ -194,38 +169,4 @@ private function processResponse(Response $response): Response
 
         return $response;
     }
-
-    // TODO: could in theory be moved to a dynamic initializer
-    private function resolveRequest(PsrRequest|ObjectFactory $psrRequest, MatchedRoute $matchedRoute): Request
```
brendt, line 199: Moved and cleaned up to `MatchRouteMiddleware`

packages/router/src/HandleRouteExceptionMiddleware.php
```diff
@@ -0,0 +1,44 @@
...
+    public function __construct(
+        private RouteConfig $routeConfig,
+    ) {}
+
+    public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
+    {
+        if ($this->routeConfig->throwHttpExceptions === true) {
+            $response = $next($request);
+
+            if ($response->status->isServerError() || $response->status->isClientError()) {
```
brendt, line 26: This part was taken from `HttpRepsonseProcessor`.

packages/router/src/HandleRouteExceptionMiddleware.php
```diff
@@ -0,0 +1,44 @@
...
+                throw new HttpException(
+                    status: $response->status,
+                    cause: $response,
+                );
+            }
+
+            return $response;
+        }
+
+        try {
```
brendt, line 36: This part was taken from `GenericRouter`

packages/router/src/HandleRouteSpecificMiddleware.php
```diff
@@ -0,0 +1,36 @@
...
+
+#[Priority(Priority::LOWEST)]
+final readonly class HandleRouteSpecificMiddleware implements HttpMiddleware
+{
+    public function __construct(
+        private MatchedRoute $matchedRoute,
+        private Container $container,
+    ) {}
+
+    public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
```
brendt, line 19: Handle route-specific middleware, taken from `GenericRouter`

packages/router/src/MatchRouteMiddleware.php
```diff
@@ -0,0 +1,70 @@
...
+use Tempest\Http\Mappers\RequestToObjectMapper;
+use Tempest\Http\Request;
+use Tempest\Http\Response;
+use Tempest\Http\Responses\NotFound;
+use Tempest\Router\Routing\Matching\RouteMatcher;
+
+use function Tempest\map;
+
+#[Priority(Priority::FRAMEWORK - 9)]
+final readonly class MatchRouteMiddleware implements HttpMiddleware
```
brendt, line 17: Taken from `GenericRouter`

packages/router/src/RouteConfig.php
```diff
@@ -23,7 +23,10 @@ public function __construct(
 
         /** @var Middleware<\Tempest\Router\HttpMiddleware> */
         public Middleware $middleware = new Middleware(
+            HandleRouteExceptionMiddleware::class,
```
brendt, line 26: `MatchRouteMiddleware` also does some mapping of raw data to request objects, which can throw validation exceptions. That's why `HandleRouteExceptionMiddleware` comes first, to make sure all exceptions are handled

packages/router/src/Routing/Matching/GenericRouteMatcher.php
```diff
@@ -4,9 +4,8 @@
 
 namespace Tempest\Router\Routing\Matching;
 
-use Psr\Http\Message\ServerRequestInterface as PsrRequest;
+use Tempest\Http\Request;
```
brendt, line 7: No more PSR :)

PR #1313, opened 2025-06-26
feat(icon): introduce icon component

innocenzi:
This pull request moves the functionality of the `<x-icon />` component to its own package. This way, it can be used in other places as well.

For instance, it is used in the `tempest/intl` package as one of the supported markup syntaxes (`{#icon-mdi-tsunami/}`), and in the documentation as in custom markdown info blocks.

In userland, icons may now be generated as follow:

```php
use Tempest\Icon;

// functional, quick and easy
Icon\render('mdi:tsunami');

// when the Tempest\Icon\Icon class is injected, call `render` on it
$this->container->get(Icon\Icon::class)->render('mdi:tsunami');
```

Additionally, a new `retryAfter` option has been added to the `IconConfig`. This option controls the duration after which the component may try again to download the icon. Previously, if there was a failure, the icon would never show up for the duration of defined cache expiry.

And last, I added the `IconDownloaded` and `IconDownloadFailed` events.


innocenzi:
@brendt Can I let you create the repository and add it to packagist after merging?

brendt:
Done: https://packagist.org/packages/tempest/icon

PR #1322, opened 2025-06-27
feat(console): allow dynamic arguments for specific console commands

brendt:
This addresses the issue where `t install vite --tailwind` gives an error.


innocenzi:
I don't think we have indeed

innocenzi (approved):

Issue #1323, opened 2025-06-27
500 Internal Server Error when requesting headers


stasadev:
### Tempest version

1.0.1

### PHP version

8.4

### Operating system

Linux

### Description

I see `500 Internal Server Error` using nginx and `php tempest serve` when requesting page headers.

The page itself `curl http://localhost:8000/` is 200 OK.

```
$ curl -I http://localhost:8000/
HTTP/1.1 500 Internal Server Error
Host: localhost:8000
Date: Fri, 27 Jun 2025 20:44:03 GMT
Connection: close
X-Powered-By: PHP/8.4.8
Content-type: text/html;charset=UTF-8

$ php tempest tail
Reading log/debug.log…
exception ""
/home/stas/my-app/vendor/tempest/framework/packages/core/src/ExceptionReporter.php:34
context array:4 [
 "status" => 404
 "message" => ""
 "cause" => Tempest\Http\Responses\NotFound\ {#1159
  +status: Tempest\Http\Status\ {#1536
   +name: "NOT_FOUND"
   +value: 404
  }
  +body: null
  +headers: []
  +session: ~ Tempest\Http\Session\Session
  +cookieManager: ~ Tempest\Http\Cookie\CookieManager
  +view: null
 }
 "previous" => 
null
]
```

### Steps to reproduce

```
composer create-project tempest/app my-app
cd my-app
php tempest serve
curl -I http://localhost:8000/
```

innocenzi:
@brendt I believe this is because we register `GET` routes as-is without registering their `HEAD` counterpart, which we should do

PR #1324, opened 2025-06-27
fix(router): handle `HEAD` requests

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1323

There are probably things that are wrong here, as I didn't dive into the topic:
- not sure if `fastcgi_finish_request` should still be called when handling `HEAD` requests
- not sure for which routes we should register matching `HEAD` routes
- not sure if we should actually register them or if we should tweak the route matcher

@brendt or @blackshadev, you can take over this PR or fix the issue in another one, as you wish :D

### DIFF
--- a/packages/router/src/GenericResponseSender.php
+++ b/packages/router/src/GenericResponseSender.php
@@ -7,6 +7,8 @@ namespace Tempest\Router;
 use Generator;
 use Tempest\Http\ContentType;
 use Tempest\Http\Header;
+use Tempest\Http\Method;
+use Tempest\Http\Request;
 use Tempest\Http\Response;
 use Tempest\Http\Responses\Download;
 use Tempest\Http\Responses\EventStream;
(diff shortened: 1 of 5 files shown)


packages/router/src/GenericResponseSender.php
```diff
@@ -27,7 +30,11 @@ public function send(Response $response): Response
         ob_start();
         $this->sendHeaders($response);
         ob_flush();
-        $this->sendContent($response);
+
+        if (! $this->shouldSendContent()) {
```
andrew-demb:
```suggestion
        if ($this->shouldSendContent()) {
```

innocenzi: Woops good catch

brendt:
Yeah I'm actually going to take another approach. We want HEAD requests to be implicitly handled by the GET request, unless there's a specific `#[Head]` route registered. I'll submit a new PR

PR #1326, opened 2025-06-28
feat(process): introduce process component

innocenzi:
This pull request introduces a new process component, wrapping `symfony/process`. It is heavily inspired by [Laravel's own process wrapper](https://laravel.com/docs/12.x/processes).

## Usage

```php
$executor = $this->container->get(ProcessExecutor::class);

// run a single process
$executor->run('echo "hello world");

// wait for a process
$process = $executor->start('echo "hello world");
$process->wait();

// multiple processes
[$first, $second] = $executor->concurrently([
  'echo "hello"',
  'echo "world"',
]);


$pool = $executor->pool([
  'echo "hello"',
  'echo "world"',
]);

$pool->run();
```

## Testing

By default, executing processes is completely disabled during tests. This means developers will have to explicitly call `registerProcessResult` or `allowRunningActualProcesses`.

This is so that actual processes that could cause issues in local environments would not be executed by default. Developers would get failures in their tests until they explicitly mock processes with `registerProcessResult`, or specifically allow processes from actually running.

```php
// fake a process result
$this->process->registerProcessResult('echo *', 'hello');

// allow processes to be actually executed, this is opt-in!
$this->process->allowRunningActualProcesses('echo *', 'hello');

// prevent processes from being actually executed, needs `registerProcessResult` calls after
$this->process->preventRunningActualProcesses();

// assertions
$this->process->assertCommandRan('echo *');
$this->process->assertCommandDidNotRun('echo *');
$this->process->assertRan(fn (PendingProcess $process, ProcessResult $result) => /* .. */);
```


packages/process/src/Exceptions/ProcessExecutionHasTimedOut.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Tempest\Process\Exceptions;
+
+use Exception;
+use Symfony\Component\Process\Exception\ProcessTimedOutException as SymfonyProcessTimedOutException;
+use Tempest\Process\ProcessResult;
+
+final class ProcessExecutionHasTimedOut extends Exception implements ProcessException
```
brendt: Couldn't we say `PorcessHasTimedOut`

As in: doesn't the existence of a process mean something is being executed?

packages/process/src/ProcessExecutor.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Tempest\Process;
+
+interface ProcessExecutor
+{
+    /**
+     * Runs the given process.
+     *
+     * @param string[]|string|PendingProcess $command
+     */
+    public function run(array|string|PendingProcess $command): ProcessResult;
```
brendt: Could we change it to `string|PendingProcess ...$processes` instead? And the same `start` as well. Using the rest operator is more consistent with Tempest's design

packages/process/src/Testing/InvokedProcessDescription.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $this->exitCode = $exitCode;
+
+        return $this;
+    }
+
+    /**
+     * Specify how many times the "isRunning" method should return "true".
+     */
+    public function withIterations(int $iterations): self
```
brendt: Most people will expect a clone to be made when a method is called `with`. I'd change it to not have the `with` prefix

packages/process/src/Testing/RestrictedProcessExecutorInitializer.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Tempest\Process\Testing;
+
+use Tempest\Container\Container;
+use Tempest\Container\Initializer;
+use Tempest\Container\Singleton;
+use Tempest\Process\ProcessExecutor;
+
+final class RestrictedProcessExecutorInitializer implements Initializer
```
brendt: Should we add `SkipDiscovery`? I reckon this one should only be used during testing

packages/process/src/Testing/TestingProcessExecutor.php
```diff
@@ -0,0 +1,154 @@
...
+            processExecutor: $this,
+        );
+    }
+
+    public function concurrently(iterable $pool): ProcessPoolResults
+    {
+        return $this->pool($pool)->start()->wait();
+    }
+
+    private function findRegisteredProcessResult(array|string|PendingProcess $command): ?ProcessResult
```
brendt: I feel like `MockedProcessResult` is more clear than `RegisteredProcessResult`?

packages/support/src/Arr/ArrayInterface.php
```diff
@@ -18,7 +17,7 @@
  *
  * @internal This interface is not meant to be used in userland.
  */
-interface ArrayInterface extends Iterator, ArrayAccess, Serializable, Countable
+interface ArrayInterface extends Iterator, ArrayAccess, Countable
```
brendt, line 20: Won't this be a breaking change?

tests/Integration/Console/Scheduler/ScheduleRunCommandTest.php
```diff
@@ -14,6 +14,8 @@ final class ScheduleRunCommandTest extends FrameworkIntegrationTestCase
 {
     public function test_invoke(): void
     {
+        $this->process->registerProcessResult('*', '');
```
brendt: Same here with `mockProcessResult`

brendt: Maybe it could even be called `mockProcess()`?

packages/process/src/Exceptions/ProcessExecutionHasTimedOut.php
innocenzi: You're right!

packages/process/src/Testing/InvokedProcessDescription.php
innocenzi: I would argue that since this is an expectation, that would read "expect process with exit code 0 and with N iterations"

I renamed those to remove `with` though, as I agree that generally-speaking this is for immutable builder methods

packages/process/src/ProcessExecutor.php
innocenzi: The array here is for commands of the format `['ls', '-la']`, not for multiple processes.

Maybe that's too confusing and we should drop support (built in `symfony/process`) for this?

packages/process/src/Testing/RestrictedProcessExecutorInitializer.php
innocenzi: Woops.. this is actually dead code. The process tester binds the `RestrictedProcessExecutor` manually now. Good catch!

packages/support/src/Arr/ArrayInterface.php
innocenzi: It shouldn't be, since we're removing `Serializable` (deprecated) and not adding it. Would ideally be in another PR though.

tests/Integration/Console/Scheduler/ScheduleRunCommandTest.php
innocenzi: I don't have a hard opinion on this so we can change it. My initial reasoning was that we were, well, "registering the result of a process". It felt natural.

packages/process/src/ProcessExecutor.php
brendt: Yes I would drop it, I've always found it really confusing with the symfony component

tests/Integration/Process/ProcessTesterAssertRanTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->process->mockProcess('echo *', "hello\n");
+        $this->executor->run('echo "hello"');
+        $this->process->assertCommandRan('echo *', function (ProcessResult $result) {
+            return $result->output !== "hello\n";
+        });
+    }
+
+    public function test_expectation_succeeds_when_callback_returns_nothing(): void
+    {
+        $this->process->mockProcess('echo *', "hello\n");
```
brendt: Wouldn't it be possible to chain all of this together?

```php
$this->process
    ->mock('echo *', "Hello\n")
    ->run('echo "Hello"')
    ->assertCommandRan('echo *', function (): void {});
```

Thinking about it some more, I think the `mockProcess` method can even be renamed simply to `mock`.

tests/Integration/Process/ProcessTesterTest.php
```diff
@@ -0,0 +1,147 @@
...
+
+final class ProcessTesterTest extends FrameworkIntegrationTestCase
+{
+    private ProcessExecutor $executor {
+        get => $this->container->get(ProcessExecutor::class);
+    }
+
+    public function test_prevents_execution_by_default(): void
+    {
+        $this->expectException(ProcessExecutionWasForbidden::class);
```
brendt, line 20: Just FYI, we have an `assertException` method that's more convenient to use than PHPUnit's built in ones. You don't need to change it here, I just wanted to let you know

brendt (changes requested):

tests/Integration/Process/ProcessTesterAssertRanTest.php
innocenzi: I went ahead and renamed `mockProcess` to `mock` and `mockProcesses` back to `mockProcessResults`. I also made the `mock` arguments optional, as I was using `->mock('*', '')` a few times and found this useful.

It would be possible to chain the calls btw, I just usually prefer not to—I find this not super pretty a lot of the time lol

We can update it if you prefer though, I don't mind that much

tests/Integration/Process/ProcessTesterTest.php
innocenzi: Neat, I didn't know! In this case I prefer not to use a closure, but I had many use cases where this one would have been better

tests/Integration/Process/ProcessTesterAssertRanTest.php
brendt: > It would be possible to chain the calls btw, I just usually prefer not to—I find this not super pretty a lot of the time lol
>
> We can update it if you prefer though, I don't mind that much

Yeah I definitely prefer the chained version.

brendt: > I went ahead and renamed mockProcess to mock and mockProcesses back to mockProcessResults

Hang on, are there two ways to mock processes? What's the difference between them?

innocenzi:
```php
// mock a single process
$this->process->mock('echo *', "hello world\n");

// mock multiple ones
$this->process->mockProcessResults([
  'echo "foo", "foo\n",
  'echo "bar", "bar\n",
]);
```

brendt: The `mockProcessResults` API is really confusing to me. Do we need it at this point or can we skip it for now and improve on it later?

brendt: Oh I think you actually wrote a typo, no?

```php
$this->process->mockProcessResults([
  'echo "foo" => "foo\n",
  'echo "bar" => "bar\n",
]);
```

(fat arrows instead of `,`?)

brendt: But why not something like 

```php
$this->process
    ->mockResult('echo "foo"', "foo\n")
    ->mockResult('echo "bar"', "bar\n");
```

This feels much cleaner to me than having to pass in an array where you have to _guess_ what the contents should be

innocenzi: > Oh I think you actually wrote a typo, no?

Oops, indeed!

> But why not something like

It's honestly easier to type an array and subjectively cleaner (I know you prefer chained methods, I don't), so why not have both?

I still prefer `mockProcessResult` over `mock`, the former is super explicit about what it does, the later can let users guessing what it's about

tests/Integration/Process/ProcessTesterAssertRanTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->process->assertCommandRan($pattern);
+    }
+
+    public function test_expectation_succeeds_when_command_is_ran_and_callback_returns_true(): void
+    {
+        $this->process->mock('echo *', "hello\n");
+        $this->executor->run('echo "hello"');
+        $this->process->assertCommandRan('echo *', function (ProcessResult $result) {
+            return $result->output === "hello\n";
+        });
```
brendt, line 34: I still find this very cumbersome and don't understand why we can't have this API

```php
$this->process
    ->mockResult('echo *', "hello\n")
    ->mockResults([
        'echo *' => $this->process
            ->describe()
            ->output('hello')
            ->output('world')
            ->iterations(2)
            ->exitCode(0),
    ])
    ->run('echo "hello"')
    ->assertCommandRan('echo *', function (ProcessResult $result) {
        return $result->output === "hello\n";
    });
```


- `mock` renamed back to `mockResult` for consistency as you proposed
- `mockResults` instead of `mockProcessResults`
- everything chained on the process tester

brendt (changes requested):

tests/Integration/Process/ProcessTesterAssertRanTest.php
innocenzi: The testing API is meant to be used when process executor calls are deep inside application code, hence no `run` on it. We could proxy it, but I don't see a use case for it?

As for the name, I prefer `mockProcessResult` over `mockResult`, aesthetically speaking. If you prefer the latter, it's not that big a deal to me—let me know

innocenzi: I realize this is what you asked me in the first place (the `->run` call)

I think you misunderstood how it's meant to be called. Let's say you have some command handler that calls a process under the hood, but you don't want this process to be actually called in your tests. That's what the API is designed for—it's the fake pattern

brendt: Ooooooh I get it. I was dumb. Sorry for the noise! Let's stick with `mockProcessResult` as well, it's fine.

PR #1331, opened 2025-06-30
chore(core): improve discovery performance by excluding specific directories

### LINKED ISSUE #1328: Discovery is very slow each request when there is a `node_modules` somewhere
When a `node_modules` is somewhere in the tree of scanned directories, discovery will scan `node_modules` and take a very long time to finish.

To reproduce, add this `package.json` in the main directory of a Tempest app and install dependencies while being in the same directory, eg. `bun install`.

```json
{
	"name": "tempest-debug-server-ui",
	"type": "module",
	"private": true,
	"scripts": {
		"dev": "vite build --watch",
		"build": "vue-tsc -b && vite build"
	},
	"devDependencies": {
		"@nuxt/ui": "^3.1.2",
		"@tailwindcss/vite": "^4.1.2",
		"@vitejs/plugin-vue": "^5.2.1",
		"@vue/tsconfig": "^0.7.0",
		"tailwindcss": "^4.1.2",
		"typescript": "~5.7.2",
		"vite": "^6.2.0",
		"vue": "^3.5.13",
		"vue-router": "^4.5.1",
		"vue-tsc": "^2.2.4"
	},
	"trustedDependencies": [
		"@tailwindcss/oxide"
	]
}
```

Navigate to any URL of that app, and observe seconds of delay due to discovery. Note that even when caching discovery, eg. when `node_modules` is somewhere in a scanned vendor package, discovery slows down the app significantly, though less than normal.

brendt:
Instead of relying on `RecursiveDirectoryIterator`, we now manually loop over all directories, giving us an entrypoint to entirely skip specific sub-directories (like `node_modules`). We can improve on this in the future by making it configurable, although for now a hard-coded check is fine. 

Closes #1328

### DIFF
--- a/packages/core/src/Kernel/LoadDiscoveryClasses.php
+++ b/packages/core/src/Kernel/LoadDiscoveryClasses.php
@@ -4,10 +4,6 @@ declare(strict_types=1);
 
 namespace Tempest\Core\Kernel;
 
-use FilesystemIterator;
-use RecursiveDirectoryIterator;
-use RecursiveIteratorIterator;
-use SplFileInfo;
 use Tempest\Container\Container;
 use Tempest\Core\DiscoveryCache;
 use Tempest\Core\DiscoveryCacheStrategy;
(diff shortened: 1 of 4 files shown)


brendt:
By making these changes, a hidden bug surfaced where the `JsonToArray` mapper would always get precedence over `JsonToObject` mapper. 

A while ago when refactoring the mapper, we forgot to disable auto mapping for the `JsonToArray` mapper, so I fixed that with this PR as well.

PR #1332, opened 2025-06-30
feat(validation): add HexColor validation rule

tizianozonta:
This pull request adds a new hex color validation rule.
```
use Tempest\Validation\Rules;

final class Book
{
    #[Rules\HexColor]
    public string $coverColor;

}
```


packages/validation/tests/Rules/HexColorTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $rule = new HexColor();
+
+        $this->assertSame('Value should be a valid hexadecimal color.', $rule->message());
+
+        $this->assertFalse($rule->isValid('string_123'));
+        $this->assertFalse($rule->isValid([])); // Should return false, not a TypeError from preg_match
+        $this->assertFalse($rule->isValid('ffffff'));
+        $this->assertFalse($rule->isValid('#fffffff'));
+
+        $this->assertTrue($rule->isValid('#ffffff'));
```
brendt, line 27: I think it'd be best to also support `#FFFA` and `#FFFFFFAA` where the last one or two letters represent the RGBA alpha channel: https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color

tizianozonta: Sure, good point! Regex updated

brendt:
Also: are you on the Tempest Discord @tizianozonta ? Just checking so that I can mark you as contributor in case you are :)

PR #1333, opened 2025-06-30
perf(core): improve overall discovery performance

brendt:
- [x] It's common practice to symlink packages during development. This PR makes it so that symlinked package's vendor directories are excluded from discovery
- [x] Attempt to only have one directory scan instead of multiple


brendt:
I did a pretty big refactor and benchmarked the changes.

Screenshot 2025-07-01 at 09 07 01

The new implementation performs better with discovery cache disabled and partial discovery cache on. Ie: better performance during local development. It also performs slightly worse with full discovery cache enabled. 

I believe the reason for the decreased performance with full discovery cache comes from the fact that we're fragmenting our cache now more: we used to cache all items per discovery class, now we cache them per location per discovery class. Doing so allows us to only do two passes over the filesystem, instead of doing a pass per discovery class. 

I want to look into further improving this part before merging though. However, I think this approach is still the way to go, since the more discovery classes you get, the more we used to iterate the filesystem, which definitely isn't ideal. With this approach, we always limit file system passes on 2: once to discover other discovery classes, once to run all those discovered classes.

brendt:
Did some more benchmarks

Screenshot 2025-07-01 at 11 06 26

`dev-discovery-improvements-A` is the version that's pushed on this PR, `dev-discovery-improvements-B` was a local experiment where I cached all of discovery into one file, instead of splitting it into a cache entry per location. 

While it had the best performance for partial caching, it had worse performance for production cache and no cache, so I decided to stick with option A

packages/console/src/Exceptions/ConsoleExceptionHandler.php
```diff
@@ -6,6 +6,7 @@
 
 use Tempest\Console\Console;
 use Tempest\Console\ExitCode;
+use Tempest\Console\GlobalFlags;
```
brendt, line 9: This was an oversight I ran into while debugging. Decided to add it real quick

packages/core/src/DiscoveryCache.php
```diff
@@ -32,24 +33,40 @@ public function __construct(
         );
     }
 
-    public function restore(string $className): ?DiscoveryItems
+    /**
+     * @return array<class-string<\Tempest\Discovery\Discovery>, DiscoveryItems>
+     */
+    public function restore(DiscoveryLocation $location): ?array
```
brendt, line 39: The core idea of this PR is that we cache discovery based on location instead of based on discovery class. Switching this around means we only have to loop twice over the while filesystem, instead of doing it once per discovery class.

packages/core/src/DiscoveryCacheInitializer.php
```diff
@@ -22,7 +22,7 @@ public function initialize(Container $container): DiscoveryCache
 
     private function resolveDiscoveryCacheStrategy(bool $isProduction): DiscoveryCacheStrategy
     {
-        if ($this->isDiscoveryGenerateCommand()) {
+        if ($this->isDiscoveryGenerateCommand() || $this->isDiscoveryClearCommand()) {
```
brendt, line 25: Fixed an oversight

packages/core/src/Kernel/LoadDiscoveryClasses.php
```diff
@@ -43,65 +44,70 @@ public function __invoke(): void
     public function build(): array
```
brendt, line 44: This part was refactored so that we start from the locations instead of discovery classes.

packages/discovery/src/DiscoveryItems.php
```diff
@@ -17,6 +17,20 @@ public function __construct(
         private array $items = [],
     ) {}
 
+    public function addForLocation(DiscoveryLocation $location, array $values): self
```
brendt, line 20: Discovery items were already grouped by location, but those weren't exposed to the outside.

PR #1335, opened 2025-07-02
feat(support): add HasMemoization trait

brendt:


brendt:
Not going to do this one, because I don't want to introduce a dependency between reflection and support

PR #1336, opened 2025-07-02
chore(reflection): cleanup memoization trait

brendt:
I didn't want this trait to be part of the public API. I removed it for now since it's only used in one place.


packages/reflection/src/ClassReflector.php
```diff
@@ -167,4 +169,13 @@ public function implements(string $interface): bool
     {
         return $this->isInstantiable() && $this->getType()->matches($interface);
     }
+
+    private function memoize(string $key, Closure $closure): mixed
+    {
+        if (! isset($this->memoize[$key])) {
```
homersimpsons, line 175: Should we use `array_key_exists`  to not call the `$closure` again in case it returns null?

brendt: Ah, yes, good point 😅

brendt: https://github.com/tempestphp/tempest-framework/pull/1360 :)

PR #1338, opened 2025-07-02
feat(session): add reflash method to session

NeoIsRecursive:
I am building an inertia adapter for tempest and to make it work like the laravel one, I need some way of reflashing when "frontend-asset-mismatch" happens and should reload the page for the user.

The test might not be in the correct location, but I couldn't really find a better place :/

### DIFF
--- a/packages/http/src/Session/Session.php
+++ b/packages/http/src/Session/Session.php
@@ -35,6 +35,16 @@ final class Session
         $this->getSessionManager()->set($this->id, $key, new FlashValue($value));
     }
 
+    public function reflash(): void
+    {
+        foreach ($this->getSessionManager()->all($this->id) as $key => $value) {
+            if (! ($value instanceof FlashValue))
+                continue;
+
+            unset($this->expiredKeys[$key]);
+        }
+    }
+
     public function get(string $key, mixed $default = null): mixed
     {
         $value = $this->getSessionManager()->get($this->id, $key, $default);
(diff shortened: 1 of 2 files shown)


brendt:
I wonder if there's a cleaner way of solving this? A middleware is my first idea?

NeoIsRecursive:
> I wonder if there's a cleaner way of solving this? A middleware is my first idea?

Yeah there probably is...

It is already a middleware actually ([source](https://github.com/NeoIsRecursive/inertia-tempest/blob/328274e6b709183d5f5b7305ff5038d302d69f3e/src/Http/Middleware.php#L42)), but to mimic how it is implemented for laravel ([source](https://github.com/inertiajs/inertia-laravel/blob/49521d3b325d09d9e1bce0f308a28bbe9c55cfe0/src/Middleware.php#L108)) it should happen after the request has been processed, and thus the reflash might be nesecary (?).

I don't think doing "reflashing" is possible in "userland" since the `$expiredKeys` property is private?

Perhaps doing the version check before processing the request and just reloading might be better. (negatives: not 1-1 with the laravel adapter).

Or did you mean the reflashing functionality? laravel does it like [this](https://github.com/laravel/framework/blob/74811d462fe6ff9ed762c2b924ec1594d84d4742/src/Illuminate/Session/Store.php#L487), where I took some inspiration (not that well versed with sessions, so sorry about that).

Thanks :)

innocenzi:
@brendt I think `reflash` is a useful functionality to have and not specific to Inertia's protocol, so having it built-in feels like the proper thing to do.

Plus, `expiredKeys` feels internal so I think it'd be weird to expose it

brendt:
Ok, I'm happy to follow you on this one :)

brendt:
Thanks @NeoIsRecursive ! btw are you on Discord so that I can mark you as contributor?

brendt:
Awesome, I found you :)

PR #1340, opened 2025-07-02
fix(core): add missing config cache initializer

lukaskleinschmidt:


brendt:
Could you remind me the reason for this change?

brendt:
Good point! Thanks for fixing :)

PR #1342, opened 2025-07-02
refactor(core)!: update exception processors to no longer return a throwable

innocenzi:
This pull request updates the `ExceptionProcessor` interface to no longer return an updated throwable, which wasn't taken into account anyway.


brendt:
Soooo… this is technically a breaking change 😬

innocenzi:
Yes (notice the exclamation mark in the commit, that says it is). In theory, this should go to a `2.x` branch. 😬

PR #1346, opened 2025-07-03
feat(cryptography): introduce cryptography component

innocenzi:
This pull request adds a new package, `tempest/cryptography`. It contains utilities to work with hashing, signing and encrypting. Later, we may add support for other cryptography-related features, though the three major ones are implemented.

## Signing key

Note that after this pull request, we should encourage adding a `SIGNING_KEY` environment variable to `.env` files. However, no exception would be triggered unless the `Signer` or `Encrypter` are resolved from the container without it. Currently, nothing use them.

I thus added a familiar command `key:generate`. This generates a key using the currently-configured algorithm (`AES_256_GCM` by default) and adds it to `.env` (or updates it if it already exists).

## Password hasher

The `PasswordHasher` interface is used to hash and verify passwords. The provided implementation can use either Argon2id or bcrypt, the former being the default.

```php
$hasher = new GenericPasswordHasher(new ArgonConfig());


$hash = $hasher->hash($password);
$hasher->verify($password, $hash);
```

## Signing

The `Signer` interface may be used to sign some data using a signing key, which would be configured globally. It could also verify the integrity of some data using a provided signature against the globally-configured key.

Under the hood, the provided implementation uses PHP's `hmac` functions and a new `Timelock` class for preventing against timing attacks.

```php
$signer = new GenericSigner(new SigningConfig(
    algorithm: SigningAlgorithm::SHA256,
    key: env('SIGNING_KEY'),
    minimumExecutionDuration: false,
));

$signature = $signer->sign($data);
$signer->verify($data, $signature);
```

## Encrypting

The `Encrypter` interface uses OpenSSL to encrypt and decrypt data, usually with the same key as the signing key. Under the hood, it also verifies the authenticity of the decrypted data using the signer.

```php
$this->container->config(new EncryptionConfig(
    algorithm: EncryptionAlgorithm::AES_256_GCM,
    key: $key = EncryptionKey::generate(EncryptionAlgorithm::AES_256_GCM),
));

$this->container->config(new SigningConfig(
    algorithm: SigningAlgorithm::SHA256,
    key: $key,
    minimumExecutionDuration: false,
));

$encrypted = $this->encrypter->encrypt('important data');
$this->encrypter->decrypt($encrypted);
```

## Timelock

This is a class inspired by Laravel's `Timebox`. It works exactly the same, except it uses `tempest/clock` under the hood.

```php
$timelock->invoke(
      callback: $callback,
      duration: Duration::milliseconds(300),
);
```


packages/cryptography/src/Encryption/EncryptionConfig.php
```diff
@@ -0,0 +1,16 @@
...
+final class EncryptionConfig
+{
+    /**
+     * @param EncryptionAlgorithm $algorithm The algorithm used for encrypting and decrypting values.
+     * @param non-empty-string $key A private, secure encryption key.
+     */
+    public function __construct(
+        public EncryptionAlgorithm $algorithm,
+        #[\SensitiveParameter]
+        public string $key,
```
brendt: Would it make sense to make this one readonly? Just to be absolutely sure the key cannot be changed within the context of a request or invocation?

packages/cryptography/src/Encryption/EncryptionKey.php
```diff
@@ -0,0 +1,48 @@
...
+
+        if (strlen($value) !== $algorithm->getKeyLength()) {
+            throw EncryptionKeyWasInvalid::becauseLengthMismatched($algorithm);
+        }
+    }
+
+    /**
+     * Generates a new cryptographically secure key using the specified algorithm.
+     */
+    public static function generate(EncryptionAlgorithm $algorithm): self
```
brendt, line 26: To me `generate` implies that the key is actually generated. But is that really the case here? We're setting up random bytes and store the associate algo, but isn't the key itself then generated somewhere else?

packages/cryptography/src/Encryption/EncryptionAlgorithm.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        return openssl_cipher_key_length($this->value);
+    }
+
+    public function getIvLength(): int
+    {
+        return openssl_cipher_iv_length($this->value);
+    }
+
+    public function isAead(): bool
```
brendt, line 34: Can we document what `isAhead` means exactly? I assume it's to signify some kind of encryption strenght?

packages/cryptography/src/Encryption/encryption.config.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+use Tempest\Cryptography\Encryption\EncryptionAlgorithm;
+use Tempest\Cryptography\Encryption\EncryptionConfig;
+
+return new EncryptionConfig(
+    algorithm: EncryptionAlgorithm::AES_256_GCM,
+    key: Tempest\env('SIGNING_KEY', default: ''),
```
brendt: Should the default be `''`? Making it `null` seems more explicit to me. `null` meaning that the value isn't set and that we cannot encrypt.

brendt (changes requested):
Todo after this is merged:

- Mention in the docs that this is an experimental package

packages/cryptography/src/Encryption/EncryptionKey.php
innocenzi: No, the key is actually generated here. `random_bytes` generate a key of the length of the specified algorithm. We then return an `EncryptionKey` VO

packages/cryptography/src/Encryption/EncryptionConfig.php
innocenzi: I'm not sure. There could be use cases but I agree we can do this for now

packages/cryptography/src/Encryption/EncryptionAlgorithm.php
innocenzi: Yeah sure, I'll add a comment. `AE` is "authenticated encryption" means that it's encrypted and cannot be tampered with, but the important part is `AD` which means "associated data", which means that the message may contain de-encodable but non-tamperable arbitrary data

packages/cryptography/src/Encryption/encryption.config.php
innocenzi: Indeed! I think I felt that `EncryptionConfig`'s `key` should not be `null`, but an empty string is far worse. I fixed that, thanks!

PR #1352, opened 2025-07-04
fix(database): fix datetime serialization for mysql database

### LINKED ISSUE #1345: [MySQL] Incorrect datetime value, the new timezone Z is the problem
### Tempest version

1.0-alpha.6

### PHP version

8.4

### Operating system

macOS

### Description

When i use the `datetime` data field i am unable to store a `Datetime::now()` to the data base column.

Error:
```
SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '2025-07-03T10:30:10.050Z' for column 'created_at' at row 1 INSERT INTO books (name, created_at) VALUES (?, ?) bindings: [ "Testing Datetime", "2025-07-03T10:30:10.050Z" ]
```

It feels like that something in the string cast of the Datetime has changed. I never saw the Z timezone.
I have the same issue with php 8.4.8 and Laravel.

### Steps to reproduce

```php
// BookMigration.php
<?php

namespace App\Migrations;

use Tempest\Database\DatabaseMigration;
use Tempest\Database\QueryStatement;
use Tempest\Database\QueryStatements\CreateTableStatement;
use Tempest\Database\QueryStatements\DropTableStatement;

final class BooksMigration implements DatabaseMigration
{
    public string $name = '2025-07-02_books';

    {
        return new CreateTableStatement(
            tableName: 'books',
        )
            ->primary()
            ->text('name')
            ->datetime('created_at');
    }

    public function down(): QueryStatement
    {
        return new DropTableStatement(
            tableName: 'books',
        );
    }
}
```

```php
// Book Model
 <?php

namespace App\Models;

use Tempest\DateTime\DateTime;
use Tempest\Validation\Rules\Length;

final class Book
{
    use IsDatabaseModel;

    public function __construct(
        #[Length(min: 1, max: 120)]
        public string $name,
        public DateTime $created_at,
    ) {
    }
}
```

Create Code in Controller or where ever:
```php
 use Tempest\DateTime\DateTime;
Book::create(
            name: 'Testing Datetime',
            created_at: DateTime::now(),
        );
```

brendt:
Closes #1345


packages/mapper/src/SerializerFactoryInitializer.php
```diff
@@ -48,6 +49,7 @@ public function initialize(Container $container): SerializerFactory
             ->addSerializer(JsonSerializable::class, SerializableSerializer::class)
             ->addSerializer(BackedEnum::class, EnumSerializer::class)
             ->addSerializer(DateTime::class, DateTimeSerializer::fromProperty(...))
+            ->addSerializer(TempestDateTime::class, DateTimeSerializer::fromProperty(...))
```
innocenzi: This is already done, the `DateTime` above is ours

packages/mapper/src/Serializers/DateTimeSerializer.php
```diff
@@ -16,26 +16,35 @@
...
 
-        return $input->format($this->format);
+        $format = $this->format;
+
+        if ($format === null) {
+            $format = match (true) {
+                $input instanceof TempestDateTimeInterface => FormatPattern::SQL_DATE_TIME,
+                default => 'Y-m-d H:i:s',
+            };
+        }
```
innocenzi, line 46: I don't get what you're doing here. `$input` is not supposed to ever be anything else than our `DateTime`. We just need to use the `FormatPattern::SQL_DATE_TIME` by default, right?

packages/mapper/src/SerializerFactoryInitializer.php
brendt: I see, thanks for mentioning

packages/mapper/src/Serializers/DateTimeSerializer.php
brendt: What's preventing users from making a model class that uses `\DateTime`?

innocenzi: Nothing, but the code above would convert it back to our own :)

CleanShot 2025-07-04 at 12 47 20

brendt:
Closing in favor of #1383

PR #1354, opened 2025-07-04
feat(database): add database seeder support

### LINKED ISSUE #1152: Database seeders
### Description

Add support for a seeding mechanism in TempestPHP that allows developers to populate their databases with raw SQL data or using PHP classes — primarily for local development, testing, and bootstrapping environments.

Suggestion using php class:
```php
final readonly class UserSeeder implements DatabaseSeeder
{
    public function seed(): ??
    {
        
    }
}
```

Suggestion using raw sql (discovered by the `seeder.sql` name)
```sql
# user.seeder.sql

INSERT INTO `users` ...
```

This would be useful even without the concept of "Factories" or something similar.

There could be a dedicated command, like `./tempest seed` or an argument in the migrations, like `./tempest migrate:up --seed`

### Benefits

It would be great to be able to set up some local-test data that you can recreate after freshly-migrating your database.

brendt:
Closes #1152


packages/database/src/DatabaseSeeder.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Tempest\Database;
+
+use UnitEnum;
+
+interface DatabaseSeeder
+{
+    public function run(null|string|UnitEnum $database): void;
```
brendt, line 9: Need to think about this. I don't like the name "database", but it is most consistent within the rest of the framework

tests/Integration/Database/MultiDatabaseTest.php
```diff
@@ -186,14 +188,14 @@ public function test_migrate_up_command(): void
     public function test_migrate_fresh_command(): void
     {
         $this->console
-            ->call('migrate:fresh --database=main')
+            ->call('migrate:fresh --database=main --all')
```
brendt: Shouldn't have added this flag here

tests/Integration/Database/MultiDatabaseTest.php
```diff
@@ -186,14 +188,14 @@ public function test_migrate_up_command(): void
...
             ->assertSuccess();
 
         $this->assertTableExists(Migration::class, 'main');
         $this->assertTableDoesNotExist(Migration::class, 'backup');
 
         $this->console
-            ->call('migrate:fresh --database=backup')
+            ->call('migrate:fresh --database=backup --all')
```
brendt: Shouldn't have added this flag here

PR #1362, opened 2025-07-06
chore(view): remove unused `illuminate/view`

mmachatschek:
The package is loaded via the tempest/blade package anyway so there is no need to add it as an explicit dependency on the root composer.json


brendt:
Thanks!

Issue #1359, opened 2025-07-05
QueryBuilder issues


brendt:
- GROUP BY comes at the wrong place

```php
query(PackageDownloadsPerDay::class)
            ->select('MAX(total)')
            ->whereField('package', 'framework')
            ->raw('GROUP BY date')
            ->orderBy('date DESC')
            ->limit(30)
            ->toSql()
```

- Adding `whereField` twice breaks the query (should add `AND` instead)
- Broken queries don't show the generated SQL in the HTTP exception handler

brendt:
For reference

```php
            query(PackageDownloadsPerDay::class)
                ->select('MAX(total)')
                ->whereField('package', 'framework')
                ->raw('GROUP BY date')
                ->orderBy('date DESC')
                ->limit(30)
                ->all()
```

PR #1367, opened 2025-07-06
refactor(database)!: query builder refactor

brendt:
Related to #1359


brendt:
This PR makes a breaking change by changing the `BuildsQuery` interface (adding the `bind` method). Since the database component is experimental, we can make this change in a new minor.

PR #1368, opened 2025-07-06
chore: upgrade libs to fix php 8.4 deprecation warnings

mmachatschek:


mmachatschek:
@brendt as tempest is pushing towards modern PHP, is it of interest to "force" higher package versions known to have compatibility with PHP 8.4.
Even though a lot of packages allow higher PHP versions, not all of them are actually tested against the newest version and therefore may lead to deprecation warnings or non intended behaviour. 

e.g. if releases of guzzle which have specifically been tested https://github.com/guzzle/guzzle/releases/tag/7.8.2 and optimized for PHP 8.4, should those be required by tempest directly?

brendt:
That makes sense, yeah :)

PR #1375, opened 2025-07-07
feat(vite): make which extensions vite discovers configurable

NeoIsRecursive:
I want to be able to discover `.tsx` files, so this PR makes it configurable what extensions vite should discover.

Or maybe it is just better to add `.tsx` and `.jsx` to the "allow list" so people can just run the default vite config? Or just match whatever has `.entrypoint.`?

What do you think? (@innocenzi)


NeoIsRecursive:
It seems like there is a composer issue making the test runner fail 🤔

brendt:
Yeah that wasn't your fault :) Thanks for the PR!

PR #1377, opened 2025-07-07
feat(core): list experimental features

### LINKED ISSUE #1373: `#[Experimental]` attribute
At one point we contemplated the idea of having an `#[Experimental]` attribute. We decided not to introduce it back then, but I believe it would actually add some benefits:

- We could discover all classes marked as `#[Experimental]` and list them in `tempest about` or a new command `tempest experimental`. 
- We can auto-generate a list of experimental features on the docs

brendt:
Closes #1373


innocenzi:
It's a tough one because if we just use the config, it's basically the same as just documenting them in the docs, we're just moving the problem

For now I don't think it's that much of an issue to just keep the docs up-to-date manually to be honest. Experimental features are an important feature that we should not just forget

PR #1384, opened 2025-07-08
fix(vite): exclude `.tempest` from vite's file watcher

ewingd:
When the `.tempest` directory is being watched by Vite for changes, any javascript fetch requests trigger Vite to automatically refresh the page.

This excludes `.tempest/*` from the file watcher to prevent page refreshes on session or cache updates.


innocenzi:
Thank you!

PR #1405, opened 2025-07-11
feat(auth): add class-level permission support

### LINKED ISSUE #1403: Provide global `#[Allow(...)] for a whole class
### Description

Currently, we can only use `Allow` on functions of a class.

```php
    #[Allow('admin')]
    #[Get('/admin')]
    public function index(SessionAuthenticator $session): View|Redirect
    {
        // ...
    }

    #[Allow('admin')]
    #[Get('/admin/edit/{slug}')]
        public function edit(string $slug): View
    {
        // ...
    }

    #[Allow('admin')]
    #[Get('/admin/delete/{slug}')]
        public function destroy(string $slug): View
    {
        // ...
    }

    // etc.
)
```

But sometimes, all functions of a class need certain privileges.

So it will be pleasant to have this global `Allow`.

And Psychopath have made the following suggestion : 

> It'd be pretty nice if we could have a class-level permission rule with method-level overrides.
So you could mark my whole class as manager-only, but one method in it as admin-only, in which case every method would be accessible by managers, but that one method would only be accessible for admins - the more specific scope would take precedence.

```php
    #[Allow('writer')]
    #[Get('/admin')]
    public function index(SessionAuthenticator $session): View|Redirect
    {
        // ...
    }

    #[Allow('writer')]
    #[Get('/admin/edit/{slug}')]
        public function edit(string $slug): View
    {
        // ...
    }

    #[Allow('admin')]
    #[Get('/admin/delete/{slug}')]
        public function destroy(string $slug): View
    {
        // ...
    }

    // etc.
)
```

- Admin role can list, edit and delete
- Writer role can only list and edit but not delete

I think it's a good idea for granularity.

### Benefits

- Avoid duplicating same `allow` in the code
- Fine tuning Allow with Psychopath's proposal

Bapawe:
Closes #1403

### DIFF
--- a/packages/auth/src/Allow.php
+++ b/packages/auth/src/Allow.php
@@ -7,7 +7,7 @@ namespace Tempest\Auth;
 use Attribute;
 use UnitEnum;
 
-#[Attribute(Attribute::TARGET_METHOD)]
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD)]
 final readonly class Allow
 {
     public function __construct(
--- a/packages/auth/src/AuthorizerMiddleware.php
+++ b/packages/auth/src/AuthorizerMiddleware.php
@@ -24,10 +24,11 @@ final readonly class AuthorizerMiddleware implements HttpMiddleware
 
     public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
     {
-        $attribute = $this->matchedRoute
+        $handler = $this->matchedRoute
             ->route
-            ->handler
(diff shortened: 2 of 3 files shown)


brendt:
Could you write a test for this?

Bapawe:
@brendt i've changed the admin route in the AdminController fixture to a class permission so it will be tested in AuthorizerTest with the other methods as overrides. Or do you need more tests?

brendt:
Ah, yes perfect

PR #1410, opened 2025-07-17
fix(mail): fix typos

brendt:


innocenzi:
That's a chore btw, we don't want that in the changelog :p

PR #1417, opened 2025-07-20
feat(database): add pagination support

innocenzi:
This pull request adds support for pagination in the select query builder:

```php
$users = query(Chapter::class)
    ->select()
    ->paginate(itemsPerPage: 20, currentPage: $request->currentPage);
```

Pagination is powered by a new `Tempest\Support\Paginator` class, which works by first definition parameters, then applying it on a dataset:

```php
// Database pagination
$paginator = new Paginator(
    totalItems: $total,
    itemsPerPage: $itemsPerPage,
    currentPage: $currentPage,
    maxLinks: $maxLinks,
);

return $paginator->paginateWith(
    callback: fn (int $limit, int $offset) => $this->limit($limit)->offset($offset)->all(),
);

// Array pagination
$paginator = new Paginator(
    totalItems: $total,
    itemsPerPage: $itemsPerPage,
    currentPage: $currentPage,
    maxLinks: $maxLinks,
);

return $paginator->paginate(['foo', 'bar', 'baz']);
```

### DIFF
--- a/packages/database/src/Builder/ModelInspector.php
+++ b/packages/database/src/Builder/ModelInspector.php
@@ -31,8 +31,9 @@ final class ModelInspector
 
     private(set) object|string $instance;
 
-    public function __construct(object|string $model)
-    {
+    public function __construct(
+        private(set) object|string $model,
+    ) {
         if ($model instanceof HasMany) {
             $model = $model->property->getIterableType()->asClass();
             $this->reflector = $model;
--- a/packages/support/src/Paginator/Exceptions/InvalidArgumentException.php
+++ b/packages/support/src/Paginator/Exceptions/InvalidArgumentException.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Tempest\Support\Paginator\Exceptions;
+
+use InvalidArgumentException as PhpInvalidArgumentException;
+
+final class InvalidArgumentException extends PhpInvalidArgumentException implements PaginationException
+{
+}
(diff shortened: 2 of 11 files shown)


packages/support/src/Paginator/Exceptions/InvalidArgumentException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Tempest\Support\Paginator\Exceptions;
+
+use InvalidArgumentException as PhpInvalidArgumentException;
+
+final class InvalidArgumentException extends PhpInvalidArgumentException implements PaginationException
```
brendt: Keep in mind our naming conventions for exceptions: https://tempestphp.com/1.x/extra-topics/contributing#exception-classes

packages/database/src/Builder/ModelInspector.php
```diff
@@ -31,8 +31,9 @@ final class ModelInspector
 
     private(set) object|string $instance;
 
-    public function __construct(object|string $model)
-    {
+    public function __construct(
+        private(set) object|string $model,
```
brendt, line 35: Why this change? I don't mind it, but I don't think you do anything with it?

brendt (changes requested):
Solid! Two minor remarks

packages/database/src/Builder/ModelInspector.php
innocenzi: I do! https://github.com/tempestphp/tempest-framework/pull/1417/files#diff-c89fc844f3c59f968175abaf32ac8fa8c8cf84fbf3e1eaa2f1fe4d0f4d7d9a58R86

packages/support/src/Paginator/Exceptions/InvalidArgumentException.php
innocenzi: Ooh, good catch, thank you!

PR #1418, opened 2025-07-22
fix(http): respect file session locks during reads

ewingd:
This fix prevents session data from being lost when one process attempts to read the session file while another process has only partially written the session data.


innocenzi (changes requested):
Hey, this would affect a lot of other places. You could update `read_file` to accept more options, or simply refactor [where it's used](https://github.com/tempestphp/tempest-framework/blob/main/packages/http/src/Session/Managers/FileSessionManager.php#L84) instead.

ewingd:
The read file locks have been moved to only interact with session data.  It felt like it would be too messy to add a flag in the general case before there was another use case for it.

innocenzi (changes requested):
Hey so, the underlying `file_get_contents` will not use that `flock` since this function works in an advisory way (meaning that only other `flock`s will wait for the lock). So I think you should use that pointer directly (maybe using `stream_get_contents`?)

Also, I assume the persist method has to lock the file as well for the race condition to be entirely solved, right?

ewingd:
Correct, `file_get_contents` won't use the lock directly, but the request to acquire the lock will wait for the lock to become available before calling `file_get_contents`, after which the lock is released.

The persist method is already set up to use the lock with the `LOCK_EX` option passed into `file_put_contents`.

PR #1420, opened 2025-07-23
feat(router): json serializable as response body

NeoIsRecursive:
I think this would be a nice QoL thing. Sort of like `JsonResource`.

And in tests we could "typecast" the body instead of having to work with arrays

```php
$page = $response->body;

static::assertInstanceOf(Response::class, $response);

static::assertSame(
      expected: 'User/Edit',
      actual: $page['component'],
);

// with this PR

/** @var PageData */
$page = $response->body;

static::assertInstanceOf(Response::class, $response);

static::assertSame(
      expected: 'User/Edit',
      actual: $page->component,
);
```

### DIFF
--- a/packages/http/src/Responses/Created.php
+++ b/packages/http/src/Responses/Created.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tempest\Http\Responses;
 
+use JsonSerializable;
 use Tempest\Http\IsResponse;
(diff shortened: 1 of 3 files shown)


brendt:
I like it! However, it's a breaking change. @innocenzi how do you feel about this one? Should it target 2.0 or should we merge into 1.x?

PR #1421, opened 2025-07-23
fix(vite): prevent syntax errors on prefetching script

Maarten-Dekker:
### Description
This PR prevents two potential runtime errors in the browser:

1. An `Uncaught SyntaxError: Unexpected token ']'` when no assets are available to prefetch.
2. An `Uncaught TypeError: Cannot read properties of undefined (reading 'forEach')` when the page's assets are minified.

### Changes Made
**Prevent empty prefetch script injection:**
`resolvePrefetchingScript` now returns early if the list of prefetchable assets is empty, preventing the `SyntaxError`.

**Prevent JS TypeError on minification:**
Added a terminating semicolon to the `new DocumentFragment()` instantiation within the `AGGRESSIVE` strategy's script template to ensure correct statement separation after minification, preventing the `TypeError`.

### Note on Testing
I believe the fixes are straightforward enough not to require dedicated test cases, but I'm happy to add them if you think it's necessary.

### DIFF
--- a/packages/vite/src/TagsResolver/ManifestTagsResolver.php
+++ b/packages/vite/src/TagsResolver/ManifestTagsResolver.php
@@ -177,10 +177,16 @@ final readonly class ManifestTagsResolver implements TagsResolver
             return $assets;
         };
 
-        $assets = Json\encode(array_values(array_map(
+        $assets = array_values(array_map(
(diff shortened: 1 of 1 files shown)


innocenzi:
Good PR, thank you very much!

PR #1423, opened 2025-07-23
feat(response): add a new Response class for json responses

Treggats:
While getting my feet wet in creating a web app with Tempest I noticed that there wasn't a `Response` class for json responses.

While it is possible to do the following
```php
#[Get('/books')]
public function books(): Response
{
    $books = Book::all(['author']);

    return new Ok(map($books)->toJson())
        ->addHeader('Accept', 'application/json')
        ->addHeader('Content-Type', 'application/json');
}
```
Why not make it easier with a new `Response` class
```php
#[Get('/books')]
public function books(): Response
{
    $books = Book::all(['author']);

    return new Json($books);
}
```

### DIFF
--- a/packages/http/src/Responses/Json.php
+++ b/packages/http/src/Responses/Json.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Responses;
+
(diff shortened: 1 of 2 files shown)


brendt:
Perfect :)

PR #1424, opened 2025-07-24
feat(view): add meta command for view components

brendt:


src/Tempest/Framework/Commands/MetaViewComponentCommand.php
```diff
@@ -0,0 +1,121 @@
...
+
+final class MetaViewComponentCommand
+{
+    use HasConsole;
+
+    public function __construct(
+        private readonly ViewConfig $viewConfig,
+    ) {}
+
+    #[ConsoleCommand(name: 'meta:view-component')]
```
innocenzi: This should probably be a hidden command

brendt: Indeed it should!

PR #1426, opened 2025-07-24
feat(vite): support passing configuration to containerized environments

Maarten-Dekker:
This PR introduces a new `config_file_path` option to the Vite plugin.

When this option is provided in the `vite.config.ts` file, the plugin will read the configuration directly from the specified JSON file, completely bypassing the execution of the `php tempest vite:config` command. This allows the Node container to start successfully while still getting the necessary configuration generated by the PHP container, enabling a smooth workflow in containerized environments.

### Reasoning

The current `php tempest vite:config` command is a great feature that ensures the PHP backend and the Vite frontend share the same configuration automatically. However, this approach presents a significant challenge in containerized development environments (like Docker) where the Node.js service runs in a separate container without access to PHP. When `vite` starts, it tries to execute the `php` command, which fails and crashes the Node container. This prevents developers from maintaining dev/prod parity.

### How to Use

1. In your PHP environment, generate the configuration file:

```bash
php tempest vite:config > .tempest/vite.json
```

2. Then, in your `vite.config.ts`, use the new config_file_path option to point to the generated file:

```ts
import tempest from 'vite-plugin-tempest'

export default {
    plugins: [
        tempest({
            config_file_path: '.tempest/vite.json',
        }),
    ],
}
```

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -3,18 +3,18 @@
   "workspaces": {
     "": {
       "devDependencies": {
-        "@types/bun": "latest",
-        "bumpp": "^10.0.1",
-        "dprint": "^0.50.0",
-        "typescript": "^5.7.3",
+        "@types/bun": "1.2.19",
+        "bumpp": "10.2.1",
+        "dprint": "0.50.1",
+        "typescript": "5.8.3",
         "unbuild": "^3.3.1",
         "vite-plugin-tempest": "workspace:*",
-        "vitest": "^3.0.4",
+        "vitest": "3.2.4",
       },
     },
     "packages/vite-plugin-tempest": {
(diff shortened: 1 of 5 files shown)


innocenzi (changes requested):
This is currently already available as `TEMPEST_PLUGIN_CONFIGURATION_OVERRIDE`. However, I understand that it might not be as convenient as having the configuration in a file.

In that case, I still suggest implementing this as an environment variable, as it is indeed an environment-specific configuration. For instance, if you're working on the same project as someone who doesn't use containers, then the configuration becomes a conflict.

I would suggest `TEMPEST_PLUGIN_CONFIGURATION_PATH` or something similar.

PR #1433, opened 2025-07-25
feat(testing): add assert response json assertions

### LINKED ISSUE #1427: feat(tests): add json/array assertions onto the `TestResponseHelper`
### Description

Writing HTTP tests for example application, but was missing json/array assertions on the `TestResponseHelper` class.

### Benefits

Adding these assertions would improve developer experience.

Treggats:
fixes #1427

# Main changes
- added json assertions on the `TestResponseHelper`;
  - `assertJson()`
  - `assertJsonByKeys()`
  - `assertHasJsonValidationError()`
  - 'assertHasNoJsonValidationErrors()`
 - added post route for books

# Secondary changes
- added a refresh method to refresh a model after it was updated
- organised the Request classes


src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,9 +221,77 @@ public function assertNotSee(string $search): self
...
+        Assert::assertNotInstanceOf(Invalid::class, $this->response);
+
+        return $this;
+    }
+
     public function dd(): void
     {
-        // @phpstan-ignore disallowed.function
+        /**
+         * @noinspection ForgottenDebugOutputInspection
```
Treggats, line 476: also disable the inspection. I know this works in PHPStorm, not sure about other editors.

tests/Fixtures/Controllers/ValidationController.php
```diff
@@ -21,8 +25,42 @@ public function get(): Response
     }
 
     #[Post('/test-validation-responses')]
-    public function store(RequestForValidationController $request): Response // @mago-expect best-practices/no-unused-parameter
+    public function store(ValidationRequest $request): Response // @mago-expect best-practices/no-unused-parameter
```
Treggats, line 28: moved the class to the Request namespace, and simplified the name

tests/Fixtures/Requests/ValidationRequest.php
```diff
@@ -2,18 +2,16 @@
...
 use Tempest\Http\IsRequest;
 use Tempest\Http\Request;
 use Tempest\Validation\Rules\Between;
 
-final class RequestForValidationController implements Request
+final class ValidationRequest implements Request
 {
     use IsRequest;
 
-    public RequestItemForValidationController $item;
```
Treggats, line 15: did not seem to be used

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,9 +221,77 @@ public function assertNotSee(string $search): self
...
 
+    public function assertJson(array $expected): self
+    {
+        Assert::assertNotNull($this->response->body);
+        Assert::assertEquals($expected, $this->response->body);
+
+        return $this;
+    }
+
+    public function assertJsonByKeys(array $expected, array $keys): self
```
innocenzi: Some docblocks with usage example would be nice, to avoid to refer to the docs

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,9 +221,77 @@ public function assertNotSee(string $search): self
...
+        Assert::assertNotEmpty($validationErrors, message: 'no validation errors found');
+
+        if ($test !== null) {
+            $test($validationErrors);
+        }
+
+        return $this;
+    }
+
+    public function assertHasNoJsonValidationErrors(): self
```
innocenzi, line 464: I'm not sure if this is enough—what if we redirect to the previous page with validation errors? I think we define headers when this happens

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,9 +221,77 @@ public function assertNotSee(string $search): self
...
+        return $this;
+    }
+
+    public function assertHasJsonValidationError(string $key, ?Closure $test = null): self
+    {
+        Assert::assertInstanceOf(Invalid::class, $this->response);
+
+        $validationErrors = array_map(function ($failingRules) use ($key) {
+            try {
+                $errors = json_decode($failingRules, true, 512, JSON_THROW_ON_ERROR);
```
innocenzi: Please use the `Tempest\Support\Json` util here

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
Treggats: Do you have an example? So I can keep the same structure.

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
Treggats: Can you point me in the right direction where this is done?

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
innocenzi: Sure: https://github.com/tempestphp/tempest-framework/blob/main/packages/support/src/Arr/functions.php#L743-L758

Treggats: Just curious about the `### Example` part, is that parsed somewhere? Or is it just the way we write these things?

(I've used the `@example` docblock initially, but will update it to use this structure)

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
brendt: This happens in \Tempest\Http\Responses\Invalid

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,9 +222,96 @@ public function assertNotSee(string $search): self
...
+    }
+
+    /**
+     * This assertion lets you assert that the response contains the given keys and values.
+     * The first array is the expected values, the second array is the keys to check.
+     *
+     * ### Example
+     * ```php
+     * $this->http->get(uri([BookController::class, 'index']))
+     *     ->assertJsonByKeys(['id' => 1, 'title' => 'The Odyssee'], ['id', 'title'])
```
brendt: I don't really understand the need of `$keys`? Is it to be able to validate that some fields exist WITH specific values, and other fields exists, regardless of the value? In that case I rather see this method split into two:

```php
->assertJsonHasKeys('id', 'author.name', …)
->assertJsonContains(['title' => 'Timeline Taxi', 'author' => ['name' => 'Brent']])
```

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
brendt: We plan on parsing some docblock classes on the docs site at one point, but that's TODO

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
Treggats: The `$keys` is to explicit say what keys are being checked in the `$expected` array.

But maybe I can solve it in a different way, the two separate assertions are actually a good idea. Will be implementing that

brendt:
Thanks! Nice PR :)

PR #1434, opened 2025-07-26
chore: give more accurate information about the internal storage path if permissions aren't met

mmachatschek:
This will additionally check the cases and throw an exception for the cases if

- the internal storage directory doesn't exist but is a file
  image
  image

- the internal storage directory doesn't exist but the user has no permission to create the dir
   image
  
- the internal storage directory exist but isn't writeable (e.g. when the `.tempest` dir was created with the root user but is executed with a less privileged user
   image
   image


brendt:
Awesome, really appreciate it!

PR #1438, opened 2025-07-27
feat(discovery): discover `autoload-dev` namespaces

Treggats:
Working on some tests and wanted to use a different database than default. Which is SQLite.
I tried adding the database config under the tests directory, only it was never discovered.

After digging it turned out that only the namespaces from `autoload.psr-4` where loaded and not `autoload-dev.psr-4`

This PR updates processing the namespaces from the `composer.json` file and adds `autoload-dev.psr-4` to it.

### DIFF
--- a/packages/core/src/Composer.php
+++ b/packages/core/src/Composer.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Tempest\Core;
 
+use Tempest\Support\Arr\ImmutableArray;
 use Tempest\Support\Filesystem;
-use Tempest\Support\Json;
 use Tempest\Support\Namespace\Psr4Namespace;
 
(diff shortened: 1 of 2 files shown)


brendt:
Sorry for not getting back on you via GitHub. We used to discover dev locations, but that ended up causing a lot of issues. Instead, you should manually add your testing discovery locations during test setup. Here's what that looks like:

```php
    protected function setUp(): void
    {
        $this->discoveryLocations = [
            new DiscoveryLocation('Tests\\Tempest\\Integration\\Console\\Fixtures', __DIR__ . '/Console/Fixtures'),
            new DiscoveryLocation('Tests\\Tempest\\Fixtures', __DIR__ . '/../Fixtures'),
        ];

        parent::setUp();
```

PR #1439, opened 2025-07-28
feat(view)!: support overriding vendor view components

brendt:
This PR is breaking because:

- It removes the `ViewComponent` interface
- It removes the ability to make class-based view components; file-based view components must always be used
- It removes the ability to use `<x-component name="">`; file-based view components must always be used

### DIFF
--- a/packages/view/src/Components/x-base.view.php
+++ b/packages/view/src/Components/x-base.view.php
@@ -15,7 +15,7 @@
 
     <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>
 </head>
-<body class="relative antialiased flex flex-col grow font-sans text-(--ui-text) bg-(--ui-bg) !overflow-visible !pr-0">
+<body class="flex flex-col h-full antialiased">
 <x-slot/>
 <x-slot name="scripts"/>
 </body>


packages/console/src/Components/Interactive/MultipleChoiceComponent.php
```diff
@@ -73,7 +73,7 @@ private function getControls(): array
             ...(
                 $this->bufferEnabled
                     ? [
-                        'esc' => 'select',
+                        'esc or /' => 'select',
```
brendt, line 76: Had to add this because `esc` is a reserved key in PhpStorm (and I reckon also in VSC): it closes the terminal pane. So you end up not being able to quit the search bar.

packages/discovery/src/DiscoveryLocation.php
```diff
@@ -23,7 +23,7 @@ public function __construct(
 
     public function isVendor(): bool
     {
-        return str_contains($this->path, '/vendor/') || str_contains($this->path, '\\vendor\\');
+        return str_contains($this->path, '/vendor/') || str_contains($this->path, '\\vendor\\') || str_starts_with($this->namespace, 'Tempest');
```
brendt, line 26: This was added because of local symlinked development: these packages don't register as vendor packages anymore because we use `realpath` on our discovery paths, which means they don't include `vendor` anymore.

packages/view/src/Parser/TempestViewCompiler.php
```diff
@@ -145,22 +145,14 @@ private function applyAttributes(array $elements): array
         $previous = null;
 
         foreach ($elements as $element) {
-            $isDynamicViewComponent = $element instanceof ViewComponentElement && $element->getViewComponent() instanceof DynamicViewComponent;
-
-            if (! $isDynamicViewComponent) {
-                $children = $this->applyAttributes($element->getChildren());
-                $element->setChildren($children);
-            }
+            $children = $this->applyAttributes($element->getChildren());
```
brendt, line 148: Cleanup thanks to dynamic view components now being a normal view component

packages/view/src/ViewComponentDiscovery.php
```diff
@@ -25,98 +23,43 @@ public function __construct(
...
-                contents: $contents,
-            );
-        } elseif ($contents->contains('<x-component name="')) {
-            $this->registerElementComponent(
-                location: $location,
-                path: $path,
-                contents: $contents,
-            );
+        if (! Filesystem\is_file($path)) {
+            return;
```
brendt, line 42: I decided to remove the `<x-component name="">` syntax. Technically we can keep it in, but there's no reason to do so. It's a remnant from the very early days of the framework, and I figured now was a good time to remove it, since I was doing major refactors to view components anyway.

packages/console/src/Components/Interactive/MultipleChoiceComponent.php
innocenzi: Not in VS Code, but it's fine to have an alias

packages/view/src/Components/x-base.view.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * @var string|null $title The webpage's title
+ */
+?>
+
+<html lang="en" class="h-dvh flex flex-col scroll-smooth">
```
innocenzi, line 7:
```suggestion
<html lang="en" class="h-dvh flex flex-col">
```

packages/view/src/Components/x-base.view.php
```diff
@@ -0,0 +1,22 @@
...
+    <title>{{ $title ?? 'Tempest' }}</title>
+
+    <meta charset="UTF-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+
+    <x-slot name="head"/>
+
+    <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>
+</head>
+<body class="relative antialiased flex flex-col grow font-sans text-(--ui-text) bg-(--ui-bg) !overflow-visible !pr-0">
```
innocenzi:
```suggestion
<body class="flex flex-col h-full antialiased">
```

innocenzi (changes requested):

PR #1440, opened 2025-07-28
feat(testing): add view data assertions

Xammie:
This adds a couple of assertions related to views.

- `assertViewData`: Assert if view data key exists and optionally assert the value
- `assertViewDataMissing`: Assert if view data key is missing
- `assertViewDataAll`: Assert the view data array
- `assertView`: Assert the view file path
- `assertViewModel`: Assert the instance of the view model object


src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,6 +219,78 @@ public function assertNotSee(string $search): self
         return $this;
     }
 
+    public function assertViewData(string $key, ?Closure $test = null): self
```
innocenzi: Let's call this `$callback` for consistency. Can you also add docblocks to the closure so users can know what arguments it takes?

```suggestion
    public function assertViewData(string $key, ?Closure $callback = null): self
```

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,6 +219,78 @@ public function assertNotSee(string $search): self
...
+            array: $data,
+            message: sprintf(
+                'No view data was set for [%s], available view data keys: %s',
+                $key,
+                implode(', ', array_keys($data)),
+            ),
+        );
+
+        if ($test !== null) {
+            $test($data, $value);
```
innocenzi: We probably want to fail assertion if the callback returns `false` specifically (not `null`, so failing is opt-in)

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -219,6 +219,78 @@ public function assertNotSee(string $search): self
...
+        Assert::assertArrayNotHasKey(
+            key: $key,
+            array: $data,
+            message: sprintf('Failed asserting that view data key [%s] was not set', $key),
+        );
+
+        return $this;
+    }
+
+    public function assertViewDataAll(Closure $test): self
```
innocenzi: Can you add docblocks to the closure so users can know what arguments it takes?

```suggestion
    public function assertViewDataAll(Closure $callback): self
```

innocenzi (changes requested):
I like it. I would love some comments explaining what each method does, with examples if relevant, so users don't have to check the documentation.

PR #1444, opened 2025-07-30
feat(validator)!: add localization support for validation error messages

innocenzi:
This pull request updates the validator to make validation error messages provide more context and be localizable.

A few APIs change in this pull request, so it is breaking. The first one is that the `Validator` now needs a `Translator` instance. The recommended way of working with the validator is now to use dependency injection, instead of manually instantiating it.

The `Rule` class no longer requires a `message()` method. Instead, rule classes that also implement the `HasErrorMessage` may have a hardcoded validation error message, and more importantly, rule classes that implement `HasTranslationVariables` must return translation variables used in translation files.

Tempest now provides a `localization.en.yml` translation file with default translation messages. This builds on the recently-introduced [localization](https://tempestphp.com/docs/features/localization) feature, using MessageFormat as the translation message syntax. For instance, validation error message for `AfterDate` looks as follows:

```yml
after_date: |
  .input {$field :string}
  .input {$inclusive :boolean}
  .input {$date :datetime date_style=long time_style=none}
  .match $inclusive
    true {{{$field} must be after or equal to {$date}}}
    false {{{$field} must be after {$date}}}
```

This expects a `inclusive` and `date` translation variable, which `AfterDate` provides through its `HasTranslationVariables` implementation.

To get the actual error message, the `Validator` class now has a `getErrorMessage` method, which accepts a `Rule` instance and an optional field name.

```php
$after = new AfterDate(date: '2024-01-01', inclusive: true);

$this->validator->getErrorMessage($after, field: 'Creation date');
// Creation date must be after or equal to January 1, 2024
```

This pull request also has some minor fixes regarding our intl message formatter, it should be now more compliant and I added support for the `:boolean` function, which includes a boolean selector.


Treggats:
@innocenzi can you create an issue so I can get started?

Would help a lot if you have before and after code snippets

innocenzi:
@Treggats I opened https://github.com/tempestphp/tempest-framework/issues/1445, but I don't have the definitive list of breaking changes, I'm waiting for Brent's input regarding the rule class names.

For now, the main breaking change is that `new Validator` now requires a `Translator` instance. I don't think there's other breaking changes than that, but will confirm after PR review

innocenzi:
I'm thinking the migration CLI can be done after this PR is merged? Should we target `2.x` for this one or is `main` fine?

Treggats:
@innocenzi I can use this to create a rule to rename these instances in a codebase :)
> @Treggats here is a rector rule if it helps:
> 
> ```
> ->withConfiguredRule(RenameClassRector::class, [
> //
> ])
> ```

brendt (approved):
Looks like a solid PR, I rebased it on 2.x, feel free to double-check and merge

PR #1447, opened 2025-08-01
feat(http)!: automatically encrypt cookies

innocenzi:
This PR builds on https://github.com/tempestphp/tempest-framework/pull/1346 to automatically encrypt cookies. This PR does not include an option to have plain-text cookies.

What required the most attention with this PR was the internal tests—we now really need to have a `SIGNING_KEY`. For this reason, the tests' bootstrap calls `key:generate` under the hood.

An alternative could have been to add hardcoded configs (like `tests/Integration/encryption.config.testing.php` and `tests/Integration/signing.config.testing.php`), but the HTTP client tests start an actual server outside of the testing environment, to which I couldn't pass through the `SIGNING_KEY` environment variable.


packages/core/src/Kernel/LoadConfig.php
```diff
@@ -45,7 +45,7 @@ public function find(): array
         $suffixes = [
             'production' => ['.production.config.php', '.prod.config.php', '.prd.config.php'],
             'staging' => ['.staging.config.php', '.stg.config.php'],
-            'testing' => ['.test.config.php'],
+            'testing' => ['.test.config.php', '.testing.config.php'],
```
innocenzi, line 48: My initial attempt at using a dedicated testing config for the encryption key made me realize we were missing this config suffix. I could open another PR but figured it was fine to keep it in this one

packages/cryptography/src/GenerateSigningKeyCommand.php
```diff
@@ -13,23 +13,28 @@
 
 use function Tempest\root_path;
 
-final readonly class CreateSigningKeyCommand
+final readonly class GenerateSigningKeyCommand
```
innocenzi, line 16: The class name was inconsistent with the actual command, so I took the opportunity to fix it

packages/http/src/Cookie/Cookie.php
```diff
@@ -78,4 +87,49 @@ public function getExpiresAtTime(): ?int
 
         return null;
     }
+
+    /**
+     * Creates acookie from the `Set-Cookie` header.
+     */
+    public static function createFromString(string $string): self
```
innocenzi, line 94: This is taken from https://github.com/httpsoft/http-cookie/blob/master/src/CookieCreator.php, a sibling file of what this class was taken from as well

src/Tempest/Framework/Testing/Http/TestResponseHelper.php
```diff
@@ -108,16 +135,34 @@ public function assertStatus(Status $expected): self
         return $this;
     }
 
-    public function assertHasCookie(string $key, ?Closure $callback = null): self
+    public function assertHasCookie(string $key, null|string|Closure $value = null): self
```
innocenzi, line 138: I fixed this assertion: previously, it was using the cookie manager, which indeed is used to provide cookies to the response—however, the correct assertion would be to actually ensure that the cookie headers are defined properly, which is what this refactor achieves. It's also the reason for introducing `Cookie::createFromString`

tests/Integration/Cryptography/GenerateSigningKeyCommandTest.php
innocenzi, line 1: GitHub doesn't show this as a rename because I added more tests to cover the new `--no-override` flag

brendt:
Shouldn't we just force the SIGNING_KEY to be present at all times by checking it in the kernel and throwing an exception if it's not present.

Maybe even better, why not generate it automatically if it's not set? Is that too much magic?

PR #1448, opened 2025-08-02
feat(database): add `tap` to query builders

innocenzi:
This pull request adds a `tap` method to all query builders. This is a good way to create reusable scopes:

```php
query(Pilot::class)
	->select()
	->tap(new HasAssignmentScope(AircraftType::PC12))
	->get();
```

Happy to tweak the naming here if you got better ideas.


brendt:
How about just `clone`?

innocenzi:
`clone` in PHP is usually a method that returns a clone of the current object, that would break expectations :/

I could think of `apply` or `transform` as other names, but honestly `with` is the one that makes the most sense to me. However, due to Laravel, not using `with` for relations would kind of break expectations as well

brendt:
I like `transform` a lot more than `tap`! Let's do that?

PR #1449, opened 2025-08-02
feat(mapper): implement serialization mapping for DTOs

### LINKED ISSUE #1329: Add support for mapping persisted DTOs FQCN to arbitrary values
Currently, the FQCN is persisted in `jsonb` columns when serializing/deserializing DTOs using the `DtoCaster` and `DtoSerializer`. This is something that can make it super easy to break an application due to a refactor.

I'm not sure of what the API could be. Maybe in the database config? Or a dedicated config?

innocenzi:
Closes #1329

This pull request introduces the ability to map a DTO class name to an arbitrary value for serialization. This would be a practice we need to encourage to avoid leaking class names in the database. This is done by adding the `SerializeAs` attribute to a data class:

```php
#[SerializeAs('settings')]
final class Settings
{
    public function __construct(
        public Theme $theme,
    ) {}
}
```

Additionally, [I typed all the query builders](https://github.com/tempestphp/tempest-framework/commit/82837a423252de63da88b7d4a14e6ce241e958c3), because the `query` function did not return proper types before.

And more importantly, when a model was queried without `IsDatabaseModel`, it would raise a warning due to the access to the `id` property in the [select model mapper](https://github.com/tempestphp/tempest-framework/commit/201e98103803d60b24846ca6d62eb9dec7c0cd29), which this PR fixes.

### DIFF
--- a/packages/database/src/Builder/ModelInspector.php
+++ b/packages/database/src/Builder/ModelInspector.php
@@ -149,7 +149,7 @@ final class ModelInspector
             return null;
         }
 
-        if ($property->hasAttribute(SerializeWith::class) || $property->getType()->asClass()->haSAttribute(SerializeWith::class)) {
+        if ($property->hasAttribute(SerializeWith::class) || $property->getType()->asClass()->hasAttribute(SerializeWith::class)) {
             return null;
         }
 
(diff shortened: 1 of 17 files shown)


innocenzi:
I don't have a hard opinion on this. I initially called it `SerializationName`, and when adding the attribute on a class later for testing, `SerializeAs` is what came naturally, not I renamed it that way

I think it describes well what it does. We can change it before releasing if we find a better name though!

PR #1451, opened 2025-08-03
fix(view): set doctype in `x-base`

happytodev:
To avoid this warning in console : This page is in Quirks Mode. Page layout may be impacted. For Standards Mode use “<!DOCTYPE html>”.


innocenzi:
Thank you!

PR #1453, opened 2025-08-03
fix(router): do not discover controller stub

### LINKED ISSUE #1446: Stub files are picked up during discovery
### Tempest version

1.4.0

### PHP version

8.4

### Operating system

Linux

### Description

Running `./tempest routes` reveals that the `ControllerStub` class used by the `make:controller` command is picked up during discovery and the route is registered.

Should the stub files be skipped during discovery by default?

Image

### Steps to reproduce

Run `./tempest routes` and observe the `dummy-path` showing up in the route listing.

Bapawe:
Closes #1446


innocenzi:
Thanks! Does the related `make` command remove the attribute?

Bapawe:
> Thanks! Does the related `make` command remove the attribute?

It does now... 😓

innocenzi:
Thank you!

PR #1454, opened 2025-08-03
feat(database): add `string` method on `CreateTableStatement`

innocenzi:
This pull request is a minor quality of life regarding the creation of tables. I added a `string` method to create a `VARCHAR` column (as an alias to the existing `varchar` method). 

I also renamed `dto` to `object`, because `dto` is very generic. I'm also considering renaming `DtoCaster` and `DtoSerializer` in another PR, let me know.

And last, I added comments for each method to give a little more context to developers. :)

### DIFF
--- a/packages/database/src/QueryStatements/CreateTableStatement.php
+++ b/packages/database/src/QueryStatements/CreateTableStatement.php
@@ -83,6 +83,16 @@ final class CreateTableStatement implements QueryStatement, HasTrailingStatement
         return $this;
     }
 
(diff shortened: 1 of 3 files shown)


brendt:
nice, but can you revert the DTO change? I know it makes sense, but it's a breaking change

innocenzi:
Sure, I'll revert it. Thinking more about it, it's perfectly fine to have such a specific method. We can always have less specific aliases actually. But I'll do that in an upcoming PR instead.

Issue #1286, opened 2025-05-27
Support `map(...)->toJson()` with collections


innocenzi:
My use case is using the mapper to serialize an array of models (from the database) to JSON.

```php
// get items
$items = query(DebugItem::class)
  ->select()
  ->all();

// convert to json
map($items)->toJson();
```

The above example works but the registered serializers are not used. Under the hood, `json_encode` is used. I'd like this to work:

```php
map($items)->collection()->toJson();
```

My current workaround is something like that:
```php
$result = [];

foreach ($items as $item) {
    $result[] = map($item)->toArray();
}

dd(Json\encode($result));
```

Issue #1087, opened 2025-03-28
HasMany and HasOne support while updating or inserting models


brendt:
Queries like these currently don't work: in case of update a `CannotUpdateHasManyRelation` is thrown. In case of insert, `CannotInsertHasManyRelation` should be thrown, but that doesn't work.

```php
query(Book::class)
    ->insert(
        title: 'Timeline Taxi',
        chapters: [
            Chapter::new(title: 'Chapter 01'),
        ],
    )
    ->build();
```

```php
query($book)
    ->update(
        chapters: [
            Chapter::new(title: 'Chapter 01'),
        ],
    )
    ->build();
```

Ideally we make it so that HasMany and HasOne relations get inserted/updated as well. The difficulty is that in their cases, they'll have to be inserted/updated _after_ the original insert or update of the parent model. For that to work, we'd need a mechanism to run queries _after_ another query (the parent query) has run.

brendt:
For now, I decided to prevent this behaviour by throwing an exception. It's truly a can of worms to automate this, and in order to properly support it, we first need #1086

Issue #1079, opened 2025-03-26
Optional ids in ORM


brendt:
The model query builders currently assume there's always an Id field on model classes. This assumption should be made optional, because Ids aren't actually required by the database

innocenzi:
Yes, thank you. This assumption also discarded the fact that many RDBMS support compound primary keys.

erikaraujo:
The framework may still need to make the `id` "assumption" somehow for tracking relationships and whatnot *if not specified otherwise*.

I propose we add a `Key` and/or `PrimaryKey` attributes. One for the class, which would allow specifying compound primary keys and one if you just want to specify it directly on the property.

Something like:

```php
#[PrimaryKey(['license_plate', 'state'])]
final class Car
{
    public string $licensePlate;

    public string $state;

    public string $make;

    //...
}
```

or

```php
final class Book
{
    #[Key]
    public string $uuid;

    public string $title;

    //...
}
```

PR #1458, opened 2025-08-04
feat(database)!: improve orm experience

innocenzi:
Okay, this pull request is huge. It's the result of me trying to build things with the ORM and stumbling upon more and more little things.

It's probably the largest I ever PR'd feature wise, and hopefully the largest you'll ever have to review @brendt!

## Arguments for a hybrid ORM

So, I know you designed the ORM with raw SQL in mind. I personally think it's very valuable to have the ability to map the result of a raw SQL query to an array of models, especially if relationships are mapped as well. That design is great.

However, I also think there is a lot of value in more abstraction, notably when it comes to building queries dynamically _or_ writing very small queries. Here is an example:

```php
// Quite convenient
->whereToday('release_date');

// Works but way less convenient
$start = DateTime::now()->startOfDay()->format('Y-m-d');
$end = DateTime::now()->endOfDay()->format('Y-m-d');
->where('release_date BETWEEN ? and ?', $start, $end);
```

Now, this is a basic example, but imagine having to implement a page with dozens of filters. You'd have to maintain a `$where` and a `$bindings` variable and conditionally concatenate the conditions as well as make sure the final result does not contain syntax errors. Not a lot of developers would be pleased with that.

I also get you wanting to promote raw SQL as the default to embrace it, to promote developer to work more with SQL and think SQL. Personally, I think an ORM is there to make things more convenient. If I wanted to write raw SQL all the time, I wouldn't use one. And maybe that's what you wanted Tempest's ORM to be; an ORM that's almost not one, just there to help with too tedious stuff like relations. However, I don't think this is realistic and I am positive that it would hurt adoption. Most developers will not want to write boilerplate to get their common queries working, they want a quick and nice way to write a query.

The good thing is, thanks to the fact that the ORM is based on the mapper, having an hybrid approach that supports both raw and abstracted SQL is "easy" (putting that in quote because supporting relationships in inserts and updates was not easy lol).

So the initial design was very good in my opinion, but there's an opportunity to have a hybrid approach that would be a win-win to cater to even more developers.

## Changes

- I improved working with `where`
  - I added support for nested `where` using callbacks
  - I added support for `where*` convenience methods for all operations I could think of
  - `where` expected raw SQL previously, now it's `whereRaw`
  - Same for `orderBy`
- I added inline documentation to almost the whole query builder, and the table creation methods as well
- I added some aliases for column creation
- I added a `foreignKey` method for creating a foreign key
- I added a `toRawSql` method (mainly for debugging)
- I fixed serialization of `true`/`false` on MySQL/SQLite, somehow we didn't have that
- I refactored model queries outside of `IsDatabaseModel`
  - `IsDatabaseModel` still exists but no longer feels like it changes the actual model, it's just methods now
  - You can now use the `model` function to get a query builder with methods like `findOrNew` if you don't want to use the trait (though `load`/`save`/`delete`/`update` are only on the trait)
  - The trait no longer forces having an `id` column (see below)
- I improved working with IDs in the ORM:
  - [Models without IDs are now first-class citizens](https://github.com/tempestphp/tempest-framework/issues/1079)
  - Performing operations that require IDs on models that don't have them throw explicit exceptions
  - The ORM no longer assumes the primary key is `id`
  - I renamed `Id` to `PrimaryKey`, which is used to identify the name of the primary key
- [I improved working with DTOs in the ORM](https://github.com/tempestphp/tempest-framework/issues/1286):
  - DTO serialization now works recursively (that was the whole reason I started working on the database, actually)
  - DTO (de)serialization handles a lot more cases, including working with arrays and top-level arrays
- I also reorganized a lot of database tests, removed the legacy `ORM` test directory, and moved most fixtures inside test files directly so it's a lot easier to just jump into a test and find out what's happening

## Considerations

- I wanted to update `Database#fetch` to always return an `ImmutableArray` instead of an array, which would make `fetchFirst` useless as well. I think this would be great but I'm thinking some people might be put off by this. I'm still in favor of it though, so let me know if I should look into that
- Adding first-party UUID generation, would be cool but we can always add that later with no breaking change (I think), and the PR is already big enough

## New updates (06/08)

- I ended up tackling [relation support during inserts/updates](https://github.com/tempestphp/tempest-framework/issues/1087)
  - Doing this still throw explicit exceptions for models without IDs


packages/database/src/Builder/ModelQueryBuilder.php
```diff
@@ -0,0 +1,332 @@
...
+     * Finds a model instance by its ID.
+     *
+     * **Example**
+     * ```php
+     * model(User::class)->resolve(1);
+     * ```
+     *
+     * @return TModel
+     */
+    public function resolve(string|int|PrimaryKey $id): object
```
brendt: I don't like having two aliases for the same method.

packages/database/src/Builder/ModelQueryBuilder.php
```diff
@@ -0,0 +1,332 @@
...
+use Tempest\Database\PrimaryKey;
+
+use function Tempest\Database\inspect;
+use function Tempest\Database\query;
+use function Tempest\make;
+
+/**
+ * @template TModel of object
+ */
+final readonly class ModelQueryBuilder
```
brendt: Overall I think this is a good refactor

packages/database/src/Builder/QueryBuilders/CountQueryBuilder.php
```diff
@@ -59,19 +67,34 @@ public function distinct(): self
...
+     */
+    public function toSql(): ImmutableString
     {
         return $this->build()->toSql();
     }
 
+    /**
+     * Returns the SQL statement with bindings. This method may generate syntax errors, it is not recommended to use it other than for debugging.
+     */
+    public function toRawSql(): ImmutableString
```
brendt, line 93: I'm fine with simply changing to `toSql` method to include the bindings tbh. The only reason it's not there is because I never bothered to add it. I rather have only one debugging method (which `toSql` already is) than two

brendt: Edit: I just looked into it and noticed that we actually use `toSql` in `GenericDatabase`, `MigrationManager`, and probably some other places, to compile the query. My suggestion would be to keep `toSql` as a debugging method (and include the bindings with it), and then have a new `Query::compile` method as a replacement for the current `toSql` implementation. WDYT?

packages/database/src/Builder/QueryBuilders/HasConvenientWhereMethods.php
```diff
@@ -0,0 +1,498 @@
...
+ * Shared methods for building WHERE conditions and convenience WHERE methods.
+ */
+trait HasConvenientWhereMethods
+{
+    /**
+     * Builds a SQL condition with bindings for a given field, operator, and value.
+     *
+     * @return array{ sql: string, bindings: array<mixed> }
+     */
+    protected function buildCondition(string $fieldDefinition, WhereOperator $operator, mixed $value): array
```
brendt, line 26: Ideally, the actual building of the query string is done by `QueryStatement` classes. That's because they are dialect-aware.

packages/database/src/Builder/QueryBuilders/HasWhereQueryBuilderMethods.php
```diff
@@ -2,53 +2,172 @@
...
 
 /**
- * @template TModelClass
+ * @template TModel of object
  * @phpstan-require-implements \Tempest\Database\Builder\QueryBuilders\BuildsQuery
+ * @uses \Tempest\Database\Builder\QueryBuilders\HasConvenientWhereMethods<TModel>
  */
 trait HasWhereQueryBuilderMethods
 {
+    use HasConvenientWhereMethods;
```
brendt, line 21: If  we include it here, then why not just combine the two traits into one?

packages/database/src/Builder/QueryBuilders/HasWhereQueryBuilderMethods.php
```diff
@@ -2,53 +2,172 @@
...
     abstract private function getStatementForWhere(): HasWhereStatements;
 
-    /** @return self<TModelClass> */
-    public function where(string $where, mixed ...$bindings): self
+    /**
+     * Adds a where condition to the query.
+     *
+     * @return self<TModel>
+     */
+    public function where(string $field, mixed $value, string|WhereOperator $operator = WhereOperator::EQUALS): self
```
brendt: To be discussed. I don't like this refactor and rather have it working the old way where `where` accepts raw SQL

brendt: Let's discuss in Discord though :)

innocenzi:
I've noticed things I need to fix, so I'll put it back on draft

innocenzi:
The CI failure is due to `tempest/cryptography` not being published yet on packagist

packages/database/src/Builder/ModelInspector.php
```diff
@@ -378,17 +384,45 @@ public function getName(): string
...
 
-    public function getPrimaryKeyValue(): ?Id
+    public function getPrimaryKeyProperty(): ?PropertyReflector
+    {
+        if (! $this->isObjectModel()) {
+            return null;
+        }
+
+        $idProperties = arr($this->reflector->getProperties())
+            ->filter(fn (PropertyReflector $property) => $property->getType()->getName() === PrimaryKey::class);
```
brendt: use `getType()->matches(PrimaryKey::class)`

packages/database/src/Builder/QueryBuilders/HasConvenientWhereMethods.php
```diff
@@ -0,0 +1,498 @@
...
+    /**
+     * Adds an `OR WHERE` condition for records created before a specific date.
+     *
+     * @return static<TModel>
+     */
+    public function orWhereBefore(string $field, DateTimeInterface|string $date): self
+    {
+        return $this->orWhere($field, DateTime::parse($date), WhereOperator::LESS_THAN);
+    }
+
```
brendt, line 496: Can we add `whereHas` as well for relations?

packages/database/src/Builder/QueryBuilders/HasWhereQueryBuilderMethods.php
brendt: What about making both possible in `where`? We can detect whether the expression passed in is a single word, and if that's the case, it should call `whereField` behind the scenes?

packages/database/src/Builder/QueryBuilders/SelectQueryBuilder.php
```diff
@@ -136,23 +161,47 @@ public function chunk(Closure $closure, int $amountPerChunk = 200): void
...
     }
 
-    /** @return self<T> */
-    public function orderBy(string $statement): self
+    /**
+     * Orders the results of the query by the given field name and direction.
+     *
+     * @return self<TModel>
+     */
+    public function orderBy(string $field, Direction $direction = Direction::ASC): self
```
brendt, line 169: Not convinced I want this change ;)

packages/database/src/Builder/QueryBuilders/WhereGroupBuilder.php
```diff
@@ -0,0 +1,196 @@
...
+use Tempest\Database\QueryStatements\WhereStatement;
+
+use function Tempest\Support\arr;
+use function Tempest\Support\str;
+
+/**
+ * @template TModel of object
+ * @uses \Tempest\Database\Builder\QueryBuilders\HasConvenientWhereMethods<TModel>
+ */
+final class WhereGroupBuilder
```
brendt, line 19: Very high level note here, I think it would be better if instead of a `GroupBuilder`, we embrace QueryStatements and create a `GroupedWhereStatement`

brendt: Oh I just noticed there already is `WhereGroupStatement`. Can't we add all the methods in this class directly over there and skip the class in between?

packages/database/src/GenericDatabase.php
```diff
@@ -124,6 +128,14 @@ private function resolveBindings(Query $query): array
         $bindings = [];
 
         foreach ($query->bindings as $key => $value) {
+            // Database handle booleans differently. We might need a database-aware serializer at some point.
+            if (is_bool($value)) {
```
brendt, line 132: Should this be in `GenericDatabase`? Can't we add that serializer to the serializer factory?

brendt: Ah, probably not because it needs the dialect?

packages/database/src/functions.php
```diff
@@ -2,24 +2,42 @@
...
-     * @param class-string<T>|string|T $model
-     * @return ModelInspector<T>
+     * Provides model-related convenient query methods.
+     *
+     * @template TModel of object
+     * @param class-string<TModel> $modelClass
+     * @return ModelQueryBuilder<TModel>
      */
-    function model(string|object $model): ModelInspector
+    function model(string $modelClass): ModelQueryBuilder
```
brendt: I understand this change… kind of. I prefer `query` because it's not about models, it's about building queries.

brendt: Ah no I understand there's a difference between `ModelQueryBuilder` and `QueryBuilder`. Do we really need these two classes? I rather have one query builder (the query builders always assumed some kind of `model`, so I'm fine with merging the two)

packages/database/src/functions.php
```diff
@@ -2,24 +2,42 @@
...
+
+    /**
+     * Inspects the given model or table name to provide database insights.
+     *
+     * @template TModel of object
+     * @param class-string<TModel>|string|TModel $model
+     * @return ModelInspector<TModel>
+     * @internal
+     */
+    function inspect(string|object $model): ModelInspector
```
brendt, line 27: I do think that this rename is good

packages/database/src/Query.php
```diff
@@ -68,7 +76,95 @@ public function toSql(): string
...
         }
 
-        return $sql;
+        return new ImmutableString($sql);
+    }
+
+    /**
+     * Returns the SQL statement with bindings. This method may generate syntax errors, it is not recommended to use it other than for debugging.
+     */
+    public function toRawSql(): ImmutableString
```
brendt, line 85: Maybe refactor the replace binding stuff to a separate class, just to keep this one clean?

packages/database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -221,6 +278,20 @@ public function dto(string $name, bool $nullable = false, ?string $default = nul
         return $this;
     }
 
+    /**
+     * Adds a JSON column for storing serializable objects. This is an alias to the `json()` method.
+     */
+    public function object(string $name, bool $nullable = false, ?string $default = null): self
```
brendt, line 284: Is the `dto` method then still available?

packages/database/src/QueryStatements/HasWhereStatements.php
```diff
@@ -6,7 +6,9 @@
 
 interface HasWhereStatements
 {
+    /** @var ImmutableArray<WhereStatement> */
     public ImmutableArray $where {
         get;
+        set;
```
brendt: Why do we need `set`?

packages/database/src/Builder/QueryBuilders/HasWhereQueryBuilderMethods.php
innocenzi: Because `WhereGroupBuilder` doesn't use methods in `HasWhereQueryBuilderMethods`. `HasConvenientWhereMethods` is specifically providing generic convenient `where*` methods

packages/database/src/Builder/ModelInspector.php
innocenzi: Good catch

packages/database/src/Builder/QueryBuilders/WhereGroupBuilder.php
innocenzi: I think the separation of concerns has value here, the statement compiles the query and the builder has, well, builder methods (note the user of `HasConvenientWhereMethods`)

packages/database/src/GenericDatabase.php
innocenzi: Yes :/

I don't like this here but it's because it requires more engineering, we can refactor later

packages/database/src/functions.php
innocenzi: This is a good point actually, I might explore this

packages/database/src/Query.php
innocenzi: I agree, I'll explore this

packages/database/src/QueryStatements/CreateTableStatement.php
innocenzi: Yes, I think convenient/specific aliases like that are fine

packages/database/src/Builder/QueryBuilders/HasConvenientWhereMethods.php
innocenzi: I agree, this deserves a refactor, but that method is also aware of bindings so it's not that easy

packages/database/src/Builder/ModelQueryBuilder.php
innocenzi: We can remove this, I thought that was a method you liked so I kept it 😅

I still have both `get` and `findById` (I find `get` to be confusing personally, but fortunately I added nice comments)

innocenzi: Actually this method is required by the `Bindable` interface!

packages/database/src/QueryStatements/HasWhereStatements.php
innocenzi: Probably not!

packages/database/src/Builder/QueryBuilders/CountQueryBuilder.php
innocenzi: Sounds good, done!

packages/database/src/functions.php
innocenzi: Implemented! Everything use `query` now. Way better.

packages/database/src/Query.php
innocenzi: Done, there's a new dedicated `RawSql` class now

packages/database/src/Builder/QueryBuilders/HasConvenientWhereMethods.php
innocenzi: Someone wanted to look into this, so let's get it after this PR

packages/database/src/Builder/QueryBuilders/HasWhereQueryBuilderMethods.php
innocenzi: So, as said on stream: due to PHP limitations, this is far from ideal. We would not be able to type the arguments properly due to the variadic argument :/

packages/database/src/Builder/QueryBuilders/SelectQueryBuilder.php
innocenzi: Same arguments than `where`, and we still have the ability to use a raw order by 😌

innocenzi:
> Can you run me through what changed there because it was a bit overwhelming.

## Tests

- I moved around tests, especially in `Builder`, `DtoSerialization`, and `ModelInspector`
- Most of these tests were at the root of `Database` and were a bit tricky to navigate in
- I moved tests from `ORM` to `Database` as well, merging cases in related files
- I added a lot more cases in general
- And I moved fixtures directly in test files so I didn't have to wonder if I was going to break other tests when updating fixtures for new cases

## `InsertQueryBuilder` and `UpdateQueryBuilder`

Most of the changes have to do with supporting relationships insertion/updating in different scenarios, as well as handling situations where there is no primary key. I also took some time to try and clean up the logic in separate methods because it was becoming complex.

I think the best way to review is to checkout the PR locally and look around these files yourself 😬

innocenzi:
Thanks for the thoughts @moebrowne! I think you're trying to find a nice middleground between this PR and Brent's opinion, but I'm convinced the default `where` experience should be the most convenient one. Otherwise, this will increase friction and will be detrimental to Tempest.

I sincerely believe that `whereRaw` is the best option. It's really not that bad to type. It's consistent with expectations (not that this is a criteria for Tempest's features), it has clear naming and the same functionality as before.

PR #1459, opened 2025-08-05
feat(http): offer more control over server sent events format

thoresuenert:
The current implementation of the `GenericResponseSender` hardcoded the server sent event protocoll to:

```php
 echo "event: {$event}\n";
 echo "data: {$data}";
 echo "\n\n";
```

We cannot use multiple data lines or add something like `id` or `retry` to the fields list https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#fields.

I introduced a Interface `ServerSentEvent` with a hooked property `$datalines` to move the shape of the data transmitted with a single message into the Event itself.

This change allows us to implement the datastar protocoll for example:
```txt
event: datastar-patch-elements
data: elements <div id="hal">
data: elements     I’m sorry, Dave. I’m afraid I can’t do that.
data: elements </div>
```

### DIFF
--- a/packages/http/src/ServerSentEvent.php
+++ b/packages/http/src/ServerSentEvent.php
@@ -2,9 +2,45 @@
 
 namespace Tempest\Http;
 
+use JsonSerializable;
+use Stringable;
+use Tempest\DateTime\Duration;
+
 interface ServerSentEvent
 {
-    public array $datalines {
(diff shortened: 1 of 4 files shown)


innocenzi:
Thank you!

PR #1461, opened 2025-08-06
feat(auth): Adds basic OAuth support

gturpin-dev:


shaffe-fr:
Great work on this implementation! 

This OAuth implementation should allow **provider-specific security checks**.

- **In `generateAuthorizationUrl`**:  
  Add extra query parameters (e.g., a `nonce` stored in session).

- **In `getUserDataFromResponse`** or in a security check method:  
  1. Validate refresh/ID token:  
     - Decode JWT and verify payload  
     - Check known public key (`kid`)  
     - Match client ID (`aud`)  
     - Match issuer URL (`iss`)  
     - Match nonce with session (`nonce`)  
     - Ensure not expired (`exp`)  
     - Verify token integrity (`sha256(access_token)` matches `at_hash`)  
     - …  
  2. Store access token (e.g., for logout)  
  3. Store expiration timeout (`expires_in`)  
  4. …

gturpin-dev:
@shaffe-fr Thanks for those great ideas 🎉 

Is the `nonce` you mentioned the equivalent of the `state` validation I mentioned in the TODO ? ( I think yes, but to be sure :) )

I think all security methods you mentioned can be done depending on the provider, but I'll look into them to see if I can implement some of these to be easier to use or abstracted in requests like the `stateless` which prevent sending state and decoding it

FYI, I've looked into Socialite implementation to check important things and DX

brendt:
I haven't looked through this in depth yet. 

> does not uses phpleague wrapper

I do wonder why not? I've been using the phpleaguea wrapper and it's really intuitive + has support for so many providers. Why start from scratch in this case?

innocenzi:
I agree that we should use the PHP League's OAuth2 client. As Brent said, there are many providers available.

We should wrap the API to make it Tempest-y, and maybe wrap the provider API while letting the ability to use PHP League wrappers (like we do for email and storage providers).

The only issue I can foresee is with the HTTP client under the hood, but I assume they provide a way to specify a PSR-compliant one

gturpin-dev:
SO, I've updated the PR to use PHPLeague by wrapping it to get benefits of it internally but having the Tempest DX  
I also design it to create our own official providers as I suggest in last comment  

Now, here's how it looks to perform OAuth now :  
```php
final class OAuthController
{
    #[Get('/auth/google')]
    public function redirect(
        #[Tag('from_env_variables')] GoogleProvider $provider // There's a built-in initializer to load providers from env
    ): Response
    {
        return new Redirect(
            $provider->generateAuthorizationUrl()
        );
    }

    #[Get('/auth/google/callback')]
    public function callback(
        #[Tag('from_env_variables')] GoogleProvider $provider,
        Request $request,
        Authenticator $authenticator,
        Session $session
    ): Response
    {
        if ( ($requestState = $request->get('state')) === null ) {
            throw new InvalidStateException('Missing state parameter in request');
        }

        if ( ($sessionState = $session->consume($provider->stateSessionSlug)) === null ) {
            throw new InvalidStateException('Missing state in session');
        }

        if ( $requestState !== $sessionState ) {
            throw new InvalidStateException('State mismatch between request and session');
        }

        $accessToken = $provider->generateAccessToken(
            code: $request->get('code') ?? throw new InvalidCodeException('Missing code parameter in request'),
        );

        $userData = $provider->fetchUserDataFromToken($accessToken);
        $user = User::findOrNew(
            find: [
                'email' => $userData->email
            ],
            update: [
                'email' => $userData->email,
                'password' => str()->random(40)->toString() // Random password as OAuth users don't need a password
            ]
        );

        $user->save();
        $authenticator->authenticate($user);

        return new Redirect('/auth/dashboard');
    }
}
```

If you want to switch provider you must only use different provider in route params like so :  
```php
#[Get('/auth/test')]
public function redirect(
    #[Tag('from_env_variables')] GithubProvider $provider
): Response
```

You can also configure the provider from user input or DB data or something else :  
```php
#[Get('/auth/test')]
public function redirect(
    GoogleProvider $provider,
    Request $request,
): Response
{
    $provider->configure(
        clientId: $request->get('client_id'),
        clientSecret: $request->get('client_secret'),
        redirectUri: $request->get('redirect_uri'),
    );
    
    return new Redirect(
        $provider->generateAuthorizationUrl()
    );
}
```

Finally, You also can use a `GenericProvider` to use a non-built-in provider or make your own : 
```php
#[Get('/auth/test')]
public function redirect(
    GenericProvider $provider,
): Response
{
    $provider->configure(
        clientId: env('GOOGLE_CLIENT_ID'),
        clientSecret: env('GOOGLE_CLIENT_SECRET'),
        redirectUri: env('GOOGLE_REDIRECT_URI'),
        defaultScopes: ['openid', 'profile', 'email'],
        authorizeEndpoint: 'https://accounts.google.com/o/oauth2/v2/auth',
        accessTokenEndpoint: 'https://accounts.google.com/o/oauth2/v2/token',
        userDataEndpoint: 'https://openidconnect.googleapis.com/v1/userinfo',
    );

    return new Redirect(
        $provider->generateAuthorizationUrl()
    );
}
```

shaffe-fr:
Hi @gturpin-dev,
If I'm correct, custom providers will not be able to use the tag `from_env_variables` and implement OauthProvider? 
Maybe we should add an extra interface like BuiltinOauthProvider so we can restrict BuiltinOauthProvider Initializer to core providers? 

Did you had something in mind?

gturpin-dev:
👋 @shaffe-fr I have to try non-built-in providers with the new implementation.  
But I think for custom ones it still possible to have a userland initializer to do the same with or without a tag.  
But maybe the `#[Tag('from_env_variables')]` doesn't allow it yet.

If possible, I'll try to avoid two interfaces to distinguish between a classic provider and a built-in one ( this is why I've done an enum to list them, but still don't 100% convinced by that btw )

brendt:
I think I prefer @innocenzi 's approach in #1577

That being said, I really appreciate you looking into this @gturpin-dev , it was a first step in the right direction, and even though it won't get merged, it has been really valuable 💯

PR #1464, opened 2025-08-07
feat(core): automatic upgrade

Treggats:
On the roadmap there was automatic upgrade with a new release.

This PR makes this possible by using Rector to rename the `Id` class, as proposed in #1458 Though it is disabled until v2 which that PR targets.

I also created a new console command which runs specific Rector rules based on the Tempest version a user want to upgrade to.

### DIFF
--- a/packages/console/src/Testing/ConsoleTester.php
+++ b/packages/console/src/Testing/ConsoleTester.php
@@ -320,4 +320,11 @@ final class ConsoleTester
 
         return $this;
     }
+
+    public function dump(): self
+    {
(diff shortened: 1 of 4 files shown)


brendt:
So yeah, I think we need a different approach. I'm going to make a new repository called `tempest/upgrades` where we bundle rectors for each major upgrade. Whenever people want to update their Tempest project, they should require that repository, and run rector directly with the correct config (we'll provide it for them via the docs).

I think that's a more viable approach.

Issue #1470, opened 2025-08-09
Unable to create POST route for API without triggering CsrfTokenDidNotMatch


misbahansori:
### Tempest version

^1.2

### PHP version

8.4

### Operating system

macOS

### Description

I’m trying to build a CRUD functionality using an API. When I attempt to create a POST route to store data, I encounter the `CsrfTokenDidNotMatch`
```php
#[Post('/hotels')]
public function store(Request $request)
{
    dd($request->body);
}
```
Since this is an API request, I don’t think we need CSRF token validation.
However, I’m not sure how to disable the CSRF middleware for specific routes in Tempest.

### Steps to reproduce

1. Create a POST route.
2. Send a request to the endpoint using Postman or something similar.
3. Observe `CsrfTokenDidNotMatch` the error.

innocenzi:
We should document this/improve the experience for this, but in the meantime you can disable the CSRF middleware like this:

```php
use Tempest\Core\KernelEvent;
use Tempest\EventBus\EventHandler;
use Tempest\Http\Session\VerifyCsrfMiddleware;
use Tempest\Router\RouteConfig;

final class DisableCsrfMiddleware
{
    public function __construct(
        private RouteConfig $routeConfig,
    ) {
    }

    #[EventHandler(KernelEvent::BOOTED)]
    public function __invoke(): void
    {
        $this->routeConfig->middleware->remove(VerifyCsrfMiddleware::class);
    }
}
```

brendt:
We actually already have `DiscoveryConfig` where you could add `VerifyCsrfMiddleware`: https://tempestphp.com/1.x/internals/discovery#excluding-files-and-classes-from-discovery

PR #1472, opened 2025-08-09
feat(router): add ability to skip middleware per route

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1470

This pull request adds the ability to skip a middleware from a particular route. This works by updating the middleware callable to not call the actual middleware if the route specifies that the middleware should be skipped.

This is done through the new `without` property on route attributes. Happy to change the name. `withoutMiddleware` is more explicit but I felt like it was too long. Would be nice to have `with` and `without` but it's more generic and would be a breaking change.

This offers the ability to have attributes dedicated to group of routes, for instance APIs:

```php
#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD)]
final readonly class Api implements Route
{
    public function __construct(
        public Method $method,
        public string $uri,
        public array $middleware = [],
        public array $without = [],
    ) {
        $this->without[] = [
            ...$without,
            VerifyCsrfMiddleware::class,
            SetCookieMiddleware::class
        ];
    }
}

```


brendt:
As I mentioned in the issue: we already have `DiscoveryConfig` taking care of removing global middleware, but I think this feature also makes sense on its own.

packages/router/src/Route.php
```diff
@@ -20,4 +20,9 @@ interface Route
     public array $middleware {
         get;
     }
+
+    /** @var class-string<HttpMiddleware>[]  */
+    public array $without {
```
brendt, line 25: Technically this is a breaking change. Since we intend on tagging 2.x soon, I think it's best to move this to 2.x

brendt (changes requested):

innocenzi:
Are you fine with the naming (`without`)?

brendt:
Yes, sounds good

PR #1487, opened 2025-08-13
fix(mapper): don't overwrite request fields

aazsamir:
If you send request
```json
{
    "body": "hello world"
}
```
it will treat request body as `hello world`, but it should be `['body' => 'hello world']`

I think this splat operator have no meaning here? Eventually it could be
```php
[
    ...$body,
   'body' => $data,
]
```


brendt:
Thank you! Don't mind those failing tests, they are unrelated.

PR #1514, opened 2025-08-15
feat(database): add `#[Hashed]` and `#[Encrypted]` attributes

innocenzi:
This pull request adds support for a new `#[Hashed]` and `#[Encrypted]` attributes, expected to be used on database models.

```php
final class User
{
    public PrimaryKey $id;

    public string $email;

    #[Hashed]
    public ?string $password;

    #[Encrypted]
    public ?string $accessToken;
}
```

To ease their implementation, I came up with a new `ProvidesCaster` and `ProvidesSerializer` interfaces. These interfaces can be implemented by attributes, which may provide specific serializers and casters.

```php
#[Attribute(Attribute::TARGET_PROPERTY)]
final class Hashed implements ProvidesSerializer
{
    public string $serializer {
        get => HashedSerializer::class;
    }
}

#[Attribute(Attribute::TARGET_PROPERTY)]
final class Encrypted implements ProvidesSerializer, ProvidesCaster
{
    public string $serializer {
        get => EncryptedSerializer::class;
    }

    public string $caster {
        get => EncryptedCaster::class;
    }
}
```

This is an alternative to `#[CastWith]` and `#[SerializeWith]`, but they are still useful—their purpose is just slightly different.

### DIFF
--- a/packages/mapper/src/ProvidesCaster.php
+++ b/packages/mapper/src/ProvidesCaster.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Tempest\Mapper;
+
+/**
+ * Implemented on an attribute, this interface specifies that a specific caster must be used to serialize the associated property.
+ */
+interface ProvidesCaster
+{
+    /** @var class-string<Caster> */
+    public string $caster {
+        get;
+    }
+}
--- a/packages/cryptography/src/Password/GenericPasswordHasher.php
+++ b/packages/cryptography/src/Password/GenericPasswordHasher.php
@@ -44,9 +44,14 @@ final class GenericPasswordHasher implements PasswordHasher
         return password_needs_rehash($hash, $this->algorithm->value, $this->config->options);
     }
 
-    public function analyze(#[\SensitiveParameter] string $hash): Hash
+    public function analyze(#[\SensitiveParameter] string $hash): ?Hash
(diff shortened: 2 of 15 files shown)


packages/mapper/src/ProvidesCaster.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Tempest\Mapper;
+
+/**
+ * Implemented on an attribute, this interface specifies that a specific caster must be used to serialize the associated property.
+ */
+interface ProvidesCaster
+{
+    /** @var class-string<Caster> */
+    public string $caster {
```
brendt, line 11: Not a dealbreaker but I tend to suffix variables with `class` when they represent a class string. So `$casterClass` and `$serializerClass`.

brendt (changes requested):
I like it! 

I have one remark, and was also wondering if you could add the docs for it within this PR?

packages/mapper/src/ProvidesCaster.php
innocenzi: I tend to avoid the double-context because it's already described by the type (`class-string`)

innocenzi:
The docs are sparse but I plan on updating the db docs anyway, since the ORM refactor

PR #1516, opened 2025-08-15
feat(auth)!: overhaul authentication and access control

innocenzi:
> [!WARNING]
> ~~This pull request depends on https://github.com/tempestphp/tempest-framework/pull/1513, https://github.com/tempestphp/tempest-framework/pull/1514 and https://github.com/tempestphp/tempest-framework/pull/1515, so CI is expected to fail until I rebase it.~~

---

**This pull request is a complete overhaul of the auth package.**

## Authentication

My main goal was removing the assumption that an authenticated model is a user (it could be an access token, an external application, or anything else that is not necessarily a "user").

Moreover, the previous implementation's `SessionAuthenticator` assumed that users were in a database. While true in most cases, this new implementation offers the ability to easily override this logic by providing an `AuthenticatableResolver` interface:

```php
interface AuthenticatableResolver
{
    /**
     * Resolves an authenticatable entity by the given ID.
     */
    public function resolve(int|string $id): ?CanAuthenticate;

    /**
     * Resolves an identifier for the given authenticatable entity.
     */
    public function resolveId(CanAuthenticate $authenticatable): int|string;
}
```

The default implementation is a database implementation and requires the model to have a primary key (not necessarily named `id`). But it is possible to override it, for instance if the application interfaces with LDAP.

The concept of authenticator is preserved, with a default session-based implementation that depends on the `AuthenticatableResolver` interface to resolve a unique ID from the authenticatable model or resolve the authenticatable model instance given the authenticated ID.

~~**Note to myself**: I don't think it is possible, with this implementation, to support different authenticatable models at once (eg. a `User` and a `ExternalApplication` or something). I should explore that before merging.~~ ✅ 

## Access control

My second goal was to not force a RBAC model on Tempest developers, as this is often very limiting in medium-sized+ applications. Instead, I implemented an ABAC (policy-based) model, which can use a RBAC model under the hood if desired. This is more flexible, and familiar with what Laravel and Symfony provide.

It is still possible to implement in userland the `#[Allow]` attribute that the previous implementation provided, but this is not included (for now). If we decide to provide an RBAC implementation, it should be done through an installer that would publish it to the user's application.

The ABAC implementation uses `#[PolicyFor(Resource::class)]` attributes and is quite flexible. It also interfaces with the authenticator to provide the currently-authenticated model if a subject is not specified.

Here are the usual ABAC terms:
- An action is a specific operation that can be performed on a resource, such as `view`, `edit`, or `delete`.
- A resource may be anything represented by a class.
- A subject is the entity that is trying to perform the action, typically the authenticated user.

Here is what creating a policy looks like. They could be in any class (like controller actions or console commands). The `#[PolicyFor]` attribute requires the resource as its first argument and an optional action as its second one. If not provided, the action is the method name.

```php
final class PostPolicy
{
    #[PolicyFor(Post::class)]
    public function view(Post $post): bool
    {
        if (! $post->published) {
            return false;
        }

        return true;
    }
}
```

The method associated to the attribute accepts the resource as its first argument and the subject as the second one. Both are optional: if the resource doesn't exist (eg. when creating it), it's not needed. And if the subject is not authenticated or provided, it's not needed either.

To determine if a subject has access to a resource, the `AccessControl` interface must be used. It has a `denyAccessUnlessGranted` method that throws if access is denied, and an `isGranted` method that returns a boolean.

```php
final readonly class PostsController
{
    public function __construct(
        private AccessControl $accessControl,
    ) {}

    #[Get('/posts/{post}')]
    public function show(Post $post): View
    {
        $this->accessControl->denyAccessUnlessGranted('show', $post);

        // ...
    }

    #[Post('/posts')]
    public function store(CreatePostRequest $request): Redirect
    {
        $this->accessControl->denyAccessUnlessGranted('create', Post::class);

        // ...
    }
}
```

Note that in the `PolicyBasedAccessControl` implementation of `AccessControl`, the methods associated to `#[PolicyFor]` are checked and a nice exception is thrown if the parameters don't accept the required types:

> The type of the resource parameter of the `App\PostPolicy::view` policy does not match the expected type `App\Post`.

## Considerations

- We can consider adding back an RBAC implementation as an installer, for convenience.
- ~~We should consider supporting multiple authenticatable models at once.~~ ✅
- ~~I was thinking about providing different authentication strategies by default.~~
	- ~~For instance, frameworks usually provide a way to login with a username and a password. This implies too many assumptions that get in the way as soon as we use a slightly different authentication paradigm, so it's annoying. So currently, users have to use the `PasswordChecker` and fetch the user from the database manually (which is perfectly fine and not a lot of LoC).~~
	- ~~If we find a way to do that, we could also provide other strategies, like passkeys, magic links, token-based auth, and maybe OAuth2 could fit this scope as well.~~
	- I explored the above, but the different ways to authenticate are too different to have a common interface, I think it's better to provide them though installers

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -225,8 +225,8 @@
     "scripts": {
         "phpunit": "@php -d memory_limit=2G vendor/bin/phpunit --display-warnings --display-skipped --display-deprecations --display-errors --display-notices",
         "coverage": "vendor/bin/phpunit --coverage-html build/reports/html --coverage-clover build/reports/clover.xml",
-        "mago:fmt": "vendor/bin/mago fmt && vendor/bin/mago lint --fix --potentially-unsafe --fmt",
-        "mago:lint": "vendor/bin/mago lint --minimum-level=note",
+        "fmt": "vendor/bin/mago fmt && vendor/bin/mago lint --fix --potentially-unsafe --fmt",
+        "lint": "vendor/bin/mago lint --minimum-level=note",
         "phpstan": "vendor/bin/phpstan analyse src tests --memory-limit=1G",
         "rector": "vendor/bin/rector process --no-ansi",
         "merge": "php -d\"error_reporting = E_ALL & ~E_DEPRECATED\" vendor/bin/monorepo-builder merge",
@@ -236,13 +236,13 @@
             "./bin/release"
         ],
         "qa": [
-            "composer mago:fmt",
+            "composer fmt",
             "composer merge",
             "./bin/validate-packages",
             "./tempest discovery:clear --no-interaction",
             "composer rector",
             "composer phpunit",
-            "composer mago:lint",
+            "composer lint",
             "composer phpstan"
         ]
     }
--- a/docs/1-essentials/07-testing.md
+++ b/docs/1-essentials/07-testing.md
@@ -22,7 +22,7 @@ composer phpunit
 
 ## Test-specific discovery locations
 
-Tempest will only discover non-dev namespaces defined in composer.json automatically. That means that `{:hl-keyword:require-dev:}` namespaces aren't discovered automatically. Whenever you need Tempest to discover test-specific locations, you may specify them within the `discoverTestLocations()` method of the provided `IntegrationTest` class. 
+Tempest will only discover non-dev namespaces defined in composer.json automatically. That means that `{:hl-keyword:require-dev:}` namespaces aren't discovered automatically. Whenever you need Tempest to discover test-specific locations, you may specify them within the `discoverTestLocations()` method of the provided `IntegrationTest` class.
 
 On top of that, Tempest _will_ look for files in the `tests/Fixtures` directory and discover them by default. You can override this behavior by providing your own implementation of `discoverTestLocations()`, where you can return an array of `DiscoveryLocation` objects (or nothing).
 
@@ -46,7 +46,7 @@ final class HomeControllerTest extends IntegrationTest
 If you want to test code that interacts with the database, your test class can call the `setupDatabase()` method. This method will create and migrate a clean database for you on the fly.
 
 ```php
-class TodoControllerTest extends IntegrationTest
+final class TodoControllerTest extends IntegrationTest
 {
     protected function setUp(): void
     {
@@ -60,8 +60,6 @@ class TodoControllerTest extends IntegrationTest
 Most likely, you'll want to use a test-specific database connection. You can create a `database.config.php` file anywhere within test-specific discovery locations, and Tempest will use that connection instead of the project's default. For example, you can create a file `tests/Fixtures/database.config.php` like so:
 
 ```php tests/Fixtures/database.config.php
-<?php
-
 use Tempest\Database\Config\SQLiteConfig;
 
 return new SQLiteConfig(
@@ -72,7 +70,7 @@ return new SQLiteConfig(
 By default, no tables will be migrated. You can choose to provide a list of migrations that will be run for every test that calls `setupDatabase()`, or you can run specific migrations on a per-test basis.
 
 ```php
-class TodoControllerTest extends IntegrationTest
+final class TodoControllerTest extends IntegrationTest
 {
     protected function migrateDatabase(): void
     {
@@ -85,7 +83,7 @@ class TodoControllerTest extends IntegrationTest
 ```
 
 ```php
-class TodoControllerTest extends IntegrationTest
+final class TodoControllerTest extends IntegrationTest
 {
     public function test_create_todo(): void
     {
--- a/packages/container/src/GenericContainer.php
+++ b/packages/container/src/GenericContainer.php
@@ -419,13 +419,15 @@ final class GenericContainer implements Container
         }
 
         foreach ($classReflector->getProperties() as $property) {
-            if ($property->hasAttribute(Inject::class) && ! $property->isInitialized($instance)) {
+            $inject = $property->getAttribute(Inject::class);
+
+            if ($inject && ! $property->isInitialized($instance)) {
                 if ($property->hasAttribute(Proxy::class)) {
                     $property->set($instance, $property->getType()->asClass()->getReflection()->newLazyProxy(
-                        fn () => $this->get($property->getType()->getName()),
+                        fn () => $this->get($property->getType()->getName(), $inject->tag),
                     ));
                 } else {
-                    $property->set($instance, $this->get($property->getType()->getName()));
+                    $property->set($instance, $this->get($property->getType()->getName(), $inject->tag));
                 }
             }
         }
--- a/packages/container/src/Inject.php
+++ b/packages/container/src/Inject.php
@@ -9,4 +9,7 @@ use Attribute;
 #[Attribute(Attribute::TARGET_PROPERTY)]
 final readonly class Inject
 {
+    public function __construct(
+        public ?string $tag = null,
+    ) {}
 }
--- a/packages/container/tests/ContainerTest.php
+++ b/packages/container/tests/ContainerTest.php
@@ -427,11 +427,13 @@ final class ContainerTest extends TestCase
     public function test_inject(): void
     {
         $container = new GenericContainer();
+        $container->singleton(InjectB::class, $bTagged = new InjectB(), 'tagged');
 
         /** @var InjectA $a */
         $a = $container->get(InjectA::class);
 
         $this->assertInstanceOf(InjectB::class, $a->getB());
+        $this->assertSame($bTagged, $a->getBTagged());
     }
 
     public function test_unregister(): void
--- a/packages/container/tests/Fixtures/InjectA.php
+++ b/packages/container/tests/Fixtures/InjectA.php
@@ -11,8 +11,16 @@ final readonly class InjectA
     #[Inject]
     private InjectB $b; // @phpstan-ignore-line
 
+    #[Inject('tagged')]
+    private InjectB $bTagged; // @phpstan-ignore-line
+
     public function getB(): InjectB
     {
         return $this->b;
+    }
+
+    public function getBTagged(): InjectB
+    {
+        return $this->bTagged;
     }
 }
--- a/packages/database/src/Migrations/MigrationManager.php
+++ b/packages/database/src/Migrations/MigrationManager.php
@@ -306,7 +306,7 @@ final class MigrationManager
         );
     }
 
-    private function getMigrationHash(MigratesUp $migration): string
+    private function getMigrationHash(MigratesUp|MigratesDown $migration): string
     {
         $sql = '';
 
--- a/packages/database/src/Migrations/RunnableMigrations.php
+++ b/packages/database/src/Migrations/RunnableMigrations.php
@@ -19,7 +19,7 @@ final class RunnableMigrations implements IteratorAggregate
     public function __construct(
         private array $migrations = [],
     ) {
-        usort($this->migrations, static fn (MigratesUp $a, MigratesUp $b) => $a->name <=> $b->name);
+        usort($this->migrations, static fn (MigratesUp|MigratesDown $a, MigratesUp|MigratesDown $b) => strnatcmp($a->name, $b->name));
     }
 
     public function getIterator(): Traversable
--- a/packages/database/tests/Migrations/RunnableMigrationsTest.php
+++ b/packages/database/tests/Migrations/RunnableMigrationsTest.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\Tests\Migrations;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use Tempest\Database\MigratesUp;
+use Tempest\Database\Migrations\RunnableMigrations;
+use Tempest\Database\QueryStatement;
+use Tempest\Database\QueryStatements\RawStatement;
+
+final class RunnableMigrationsTest extends TestCase
+{
+    #[DataProvider('provide_migrations')]
+    #[Test]
+    public function migration_ordering(array $migrationNames, array $expectedOrder): void
+    {
+        $migrations = array_map(fn (string $name) => $this->createDatabaseMigration($name), $migrationNames);
+
+        $this->assertSame(
+            expected: $expectedOrder,
+            actual: array_map(
+                callback: static fn (MigratesUp $migration) => $migration->name,
+                array: iterator_to_array(new RunnableMigrations($migrations)),
+            ),
+        );
+    }
+
+    public static function provide_migrations(): array
+    {
+        return [
+            [
+                'migrationNames' => [
+                    'migration1_1',
+                    'migration1_10',
+                    '2025-10-12_create_idbn_table',
+                    'migration1_2',
+                    '2025-01-12_create_author_table',
+                    '2025-08-10_create_user_table',
+                    '2025-08-01_create_book_table',
+                    '2025-08-12_create_chapter_table',
+                ],
+                'expectedOrder' => [
+                    '2025-01-12_create_author_table',
+                    '2025-08-01_create_book_table',
+                    '2025-08-10_create_user_table',
+                    '2025-08-12_create_chapter_table',
+                    '2025-10-12_create_idbn_table',
+                    'migration1_1',
+                    'migration1_2',
+                    'migration1_10',
+                ],
+            ],
+        ];
+    }
+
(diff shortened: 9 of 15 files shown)


docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
 ---
-title: Authentication and authorization
-keywords: "Experimental"
```
brendt, line 3: I think we should definitely keep this experimental for one more major

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
+
+## Quick start
+
+Tempest does not assume that all applications have users, but it is the most common case. For this reason, we provide the ability to publish a basic user model and its migration.
+
+```sh sh
+./tempest install auth
+```
+
+After publishing, you may run `./tempest migrate`. You now have the building blocks for your authentication.
```
brendt, line 21: Couldn't we run migrate straight from the installer? If not then let's keep it this way

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
 ```
 
-## Built-in user model
+To instruct Tempest that you want to use your own resolver, you will need to create a dedicated [initializer](../1-essentials/05-container.md#implementing-an-initializer).
 
-Tempest's authenticator and authorizer are compatible with any class implementing the {`Tempest\Auth\CanAuthenticate`} and {`Tempest\Auth\CanAuthorize`} interfaces. However, Tempest comes with a pre-built `User` model that makes it easier to get started. In order to use Tempest's `User` implementation, you must install the auth files:
+```php app/Authentication/LdapAuthenticatableResolverInitializer.php
+use Tempest\Auth\Authentication\AuthenticatableResolver;
 
+final class LdapAuthenticatableResolverInitializer implements Initializer
```
brendt, line 181: Maybe also briefly mention and refer to tagged singletons so that people know they can have multiple auth resolvers?

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
 
-use Tempest\Auth\Allow;
-use Tempest\Http\Response;
+```php app/Authentication/LdapAuthenticatableResolver.php
+use Tempest\Auth\Authentication\AuthenticatableResolver;
+use Tempest\Auth\Authentication\CanAuthenticate;
+use App\Authentication\User;
 
-final readonly class AdminController
+final readonly class LdapAuthenticatableResolver implements AuthenticatableResolver
```
brendt, line 148: Can we call this interface `AuthResolver`? Sound better imo

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
+final class RequestAuthenticator implements Authenticator
+{
+    private ?CanAuthenticate $current = null;
+
+    public function authenticate(CanAuthenticate $authenticatable): void
+    {
+        $this->current = $authenticatable;
+    }
+
+    public function deauthenticate(): void
```
brendt, line 213: Don't really like the name `deauthenticate`, but I also cannot really come up with something else besides `login` and `logout`, which I assume you considered?

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
+The attribute expects the class name of the resource as its first parameter, and an optional action name as the second parameter. If the action name is not provided, the kebab-cased method name is used instead.
+
+```php app/PostPolicy.php
+use Tempest\Auth\AccessControl\PolicyFor;
+use Tempest\Auth\AccessControl\AccessDecision;
+
+final class PostPolicy
+{
+    #[PolicyFor(Post::class)]
+    public function view(Post $post): bool
```
brendt, line 256: We could infer for which the policy is for based on the parameter's type, no?

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
+final readonly class PostController
+{
+    public function __construct(
+        private AccessControl $accessControl,
+    ) {}
+
+    #[Delete('/posts/{post}')]
+    public function delete(Post $post): Redirect
+    {
+        $this->accessControl->denyAccessUnlessGranted('delete', $post);
```
brendt: No more attribute as a shorthand?

docs/2-features/04-authentication.md
```diff
@@ -1,105 +1,315 @@
...
-use App\Auth\User;
-
-$user = new User(
-    name: 'Brent',
-    email: 'brendt@stitcher.io',
-)
-    ->setPassword('password')
-    ->save()
-    ->grantPermission('admin');
+$accessControl->isGranted('create', resource: Post::class, subject: $user);
```
brendt, line 321: I find "subject" a bit confusing. I rather have `user`, but I understand that that wouldn't cover all bases, and `$canAuthenticate` is very awkward

docs/2-features/04-authentication.md
innocenzi: Indeed, that's an oversight!

docs/2-features/04-authentication.md
innocenzi: Technically we can, but for things like migration, which affect the actual environment, I think running it manually is better

docs/2-features/04-authentication.md
innocenzi: I did! "login" is more specific than "authenticate", it refers to user sessions. "authenticate" is broader, it means that you provide an identity to the system. It was one of the goals of this PR: implement an authentication system that doesn't assume everything is a user

Because of that, "deauthenticate" provides the correct meaning. I could suggest stuff like "revoke" or "invalidate", but both are less clear in my opinion

docs/2-features/04-authentication.md
innocenzi: No, because that would require some conventions/magic, in this case the `$post` would have to be forwarded

innocenzi: Not impossible to add an attribute later though

docs/2-features/04-authentication.md
innocenzi: I'm generally not a fan of short names. Auth is ambiguous, it could say "authentication" or "authorization"—you can't guess at a glance what `AuthResolver` would be, as opposed to `AuthenticatableResolver` (a class that resolves something that can authenticate)

docs/2-features/04-authentication.md
innocenzi: Good idea, will do!

docs/2-features/04-authentication.md
innocenzi: This is the usual, conventional term when it comes to PBAC/ABAC—the subject is the identity performing the action (and often there can be no subject at all). Indeed, `user` is not correct here, because the subject could be entirely unrelated to an user

docs/2-features/04-authentication.md
innocenzi: Actually, potentially! Let me explore that

innocenzi: Done! It's nice this way, a bit less typing and not much magic going on, good idea

docs/2-features/04-authentication.md
brendt: You could have it as an option that people have to confirm? I'd expect an installer to take care of all the steps, we also run composer update after the framework installer for example.

docs/2-features/04-authentication.md
brendt: That's fair, and not a deal breaker to me

docs/2-features/04-authentication.md
brendt: I actually don't know if your definition for login is correct:

> Login is the process of an individual entering credentials, such as a username and password, to gain access to a system or application. Authentication is the underlying security process that verifies a user's identity to ensure they are who they claim to be. In essence, logging in is the action, while authentication is the verification that happens as part of that action, often confirming the credentials used for the login are valid.

If login/logout is the action, then I'd say it's actually a more correct name.

docs/2-features/04-authentication.md
brendt: True, let's keep it as is for now

docs/2-features/04-authentication.md
brendt: Yes, let's keep as is

docs/2-features/04-authentication.md
innocenzi: Sounds like a good idea actually

docs/2-features/04-authentication.md
innocenzi: > Login is the process of an individual entering credentials

Keywords here are "process" and "individual". It's not always an individual, and the "process" implied by "login" implies it's a manual action, while authentication can be done under the hood, for instance with Kerberos/NTLM (though to be fair, this is not easy/common to do in our ecosystem, but still)

brendt: Fair. Let's keep it as you proposed.

packages/auth/src/AccessControl/AccessControl.php
```diff
@@ -0,0 +1,34 @@
...
+     * Checks if the action is granted for the given resource and subject. If not, an exception is thrown.
+     *
+     * @template Resource of object
+     * @param UnitEnum|string $action An arbitrary action to check access for, e.g. 'view', 'edit', etc.
+     * @param Resource|class-string<Resource> $resource A model instance or class string of a model to check access for.
+     * @param null|Subject $subject An optional subject to check access against, e.g. a user or service account.
+     *
+     * @throws AccessWasDenied
+     */
+    public function denyAccessUnlessGranted(UnitEnum|string $action, object|string $resource, ?object $subject = null): void;
```
brendt: Personally, I'd use something like `ensureGranted`

packages/auth/src/Authentication/AuthenticatableResolver.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Tempest\Auth\Authentication;
+
+interface AuthenticatableResolver
```
brendt, line 5: Circling back to this name, I dislike that we call it `Authenticatable` in this interface but then return `CanAuthenticate`. I don't have a strong opinion on which option exactly, but I think these two should be called the same.

packages/auth/src/Authentication/CanAuthenticate.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Auth\Authentication;
+
+/**
+ * Represents an entity that may be authenticated.
+ */
+interface CanAuthenticate
```
brendt: I generally dislike empty interfaces. I reckon we need it to register an authenticated "user" in the container?

packages/auth/src/Exceptions/ModelIsNotAuthenticatable.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Tempest\Auth\Exceptions;
+
+use Exception;
+use Tempest\Auth\Authentication\CanAuthenticate;
+
+final class ModelIsNotAuthenticatable extends Exception implements AuthenticationException
```
brendt: `ModelWasNot` instead of `ModelIsNot`? I think `Is` makes more sense indeed, but it breaks with our exception naming guidelines

packages/auth/src/Exceptions/NoPolicyWereFoundForResource.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Tempest\Auth\Exceptions;
+
+use Exception;
+
+final class NoPolicyWereFoundForResource extends Exception implements AuthenticationException
```
brendt, line 7: `NoPoliciesWere` or `NoPolicyWas`

packages/auth/src/Exceptions/PolicyIsInvalid.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Tempest\Auth\Exceptions;
+
+use Exception;
+
+final class PolicyIsInvalid extends Exception implements AuthenticationException
```
brendt: `Was` instead of `Is`

packages/auth/src/Exceptions/PolicyMethodIsInvalid.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Tempest\Auth\Exceptions;
+
+use Exception;
+
+final class PolicyMethodIsInvalid extends Exception implements AuthenticationException
```
brendt: `Was` instead of `Is`

packages/auth/src/Installer/basic-user/StubCreateUsersTableMigration.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Tempest\Auth\Installer;
+
+use Tempest\Database\MigratesUp;
+use Tempest\Database\QueryStatement;
+use Tempest\Database\QueryStatements\CreateTableStatement;
+use Tempest\Discovery\SkipDiscovery;
+
+#[SkipDiscovery]
+final class StubCreateUsersTableMigration implements MigratesUp
```
brendt: We don't have consistent naming here, some files are called `FileStub`, others `File.stub.php`. But I rather not introduce another variant by prefixing with `Stub`.

My proposal is to go with `File.stub.php` because it's has built-in support with PhpStorm to be recognised as a stub file, and mago is already configured to ignore them

packages/auth/src/Installer/basic-user/StubUserModel.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Tempest\Auth\Installer;
+
+use Tempest\Auth\Authentication\CanAuthenticate;
+use Tempest\Database\Hashed;
+use Tempest\Database\PrimaryKey;
+use Tempest\Discovery\SkipDiscovery;
+
+#[SkipDiscovery]
+final class StubUserModel implements CanAuthenticate
```
brendt: Same here: `UserModel.stub.php`

packages/auth/src/Authentication/AuthenticatableResolver.php
innocenzi: I tried to follow your convention on some other interfaces, but I personally prefer `Authenticatable` as a name, so I'll take that!

packages/auth/src/Authentication/CanAuthenticate.php
innocenzi: Yeah, I guess this could be an attribute instead:

```php
#[CanAuthenticate]
final class User
{
    public PrimaryKey $id;

    public function __construct(
        public string $email,
        #[Hashed]
        public ?string $password,
    ) {}
}
```

Not sure which I prefer honestly. An attribute feels less "standalone" somehow, but I wouldn't take that as an argument in favor of an interface. I'll let you decide on this one

packages/auth/src/AccessControl/AccessControl.php
innocenzi: Fine by me! I stole `denyAccessUnlessGranted` from Symfony. I liked the explicitness

packages/auth/src/Exceptions/ModelIsNotAuthenticatable.php
innocenzi: Both make sense but I'm being inconsistent here, you're right

packages/auth/src/Installer/basic-user/StubCreateUsersTableMigration.php
innocenzi: I also prefer that. I actually used `Stub` as a prefix to work around a VSC issue (the stub files were suggested in intellisense), but I could fix that issue by ignoring `*.stub.php` files. Let's use that convention from now on 👍

packages/auth/src/Authentication/CanAuthenticate.php
brendt: I think an interface feels more natural tbh, but I was mainly wondering if it really needed 0 methods or properties? Shouldn't an authenticateble at least provide _something_ ?

innocenzi: Not in this case! Since we provide no *login* implementation (at least not directly within Tempest, though trivial to implement in userland), we don't have to ask the model for a password, an identifier or anything. This is by design, because a model could be authenticated using vastly different strategies (not necessarily email/password, could be an access token, a passkey, a single email for SSO...)

brendt: Gotcha. I think that an interface feels a bit more natural atm, even though it's empty. Let's keep that for now

PR #1517, opened 2025-08-16
feat(database): primary key improvements

### LINKED ISSUE #1482: Convert `Id` (or `PrimaryKey`) to textual variant when passed into `uri`, as well as object with `PrimaryKey`

### LINKED ISSUE #1519: Add `id` and `bindableValue` to IsDatabaseModel
To be discussed: wouldn't it make sense to still include the primary key in the IsDatabaseModel trait. The trait only exists to make the "default implementation" of models easier. I think it's fair to say that having an id is a sensible default.

It would also make the addition of `bindableValue` a lot more convenient (so that models are correctly bound when calling uri without any overhead)

brendt:
- [x] The PrimaryKey is now again included in `IsDatabaseModel`
- [x] Added support for bindable values via attributes
- [x] Update docs

Closes #1482
Closes #1519

### DIFF
--- a/docs/1-essentials/01-routing.md
+++ b/docs/1-essentials/01-routing.md
@@ -103,6 +103,24 @@ final class Aircraft implements Bindable
 }
 ```
 
+By default, `Bindable` objects will be cast to strings when they are passed into the `uri()` function as a route parameter. You can override this default behaviour by tagging a public property on the object with the {b`\Tempest\Router\IsBindingValue`} attribute:
+
+```php
+use Tempest\Router\Bindable;
+use Tempest\Router\IsBindingValue;
+
+final class Aircraft implements Bindable
+{
+    #[IsBindingValue]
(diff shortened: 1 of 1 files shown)


packages/router/src/Bindable.php
```diff
@@ -4,5 +4,9 @@
 
 interface Bindable
 {
+    public int|string $bindingValue {
```
innocenzi: Would love a docblock describing what this is

innocenzi:
What do we think about `bindingValue` being an attribute? I think it feels bad to have a property for this purpose, kind of like we have a `#[Table]` to avoid having a property

For models, we can make `PrimaryKey` the fallback bindable value if the attribute is not provided. It's a bit specialized in the framework code but it would feel better when consuming in userland (and we already have an exception for `BackedEnum` anyway)

PR #1527, opened 2025-08-18
fix(database): update ORM to consider get hooks as virtual properties

### LINKED ISSUE #1518: Virtual properties (only get hook) should be considered virtual by the ORM
```php
    #[Virtual]
    public int|string $bindingValue {
        get => $this->id->value;
    }
```

WendellAdriel:
TBH, IDK if I completely understood the issue, so this is a shot at what I understood #1518 was about.
If I misunderstood the issue, please just ignore this PR. 😅 

Fixes #1518


brendt:
As discussed, let's clean these PRs up and tackle them one by one instead of all at the same time.

PR #1529, opened 2025-08-19
feat(router): add method spoofing support for HTML forms

WendellAdriel:
## Summary

* Supports PUT, PATCH, and DELETE method spoofing
* Only applies to POST requests for security
* Case-insensitive method values
* Preserves all request data
* Automatic support in `x-form` component
* Backward compatible (HEAD→GET fallback preserved)
* Secure by design (doesn't allow GET spoofing)

## Changes

This updates the `MatchRouteMiddleware` to:
* Check for a `_method` parameter in the request body
* Validate and allow only PUT, PATCH, and DELETE methods for security

This also updates the `x-form.view.php` component to:
* Automatically detects when method is PUT, PATCH, or DELETE
* Render the form with method="POST" for browser compatibility
* Automatically insert a hidden `_method` field with the original method value

## Examples

If using the `<form>` HTML tag, you'll need to set the `method` attribute of the form to `POST` and add a hidden input with the method you want:

```html
<form action="/books/{{ $book->id }}" method="POST">
    <input type="hidden" name="_method" value="PUT" />
</form>
```

If you're using the `x-form` component, it's even easier:

```html
<x-form action="/books/{{ $book->id }}" method="PUT">
</x-form>
```

This will render the `<form>` tag, setting the method to `POST` and automatically inserting a hidden `_method` input with the value you set.

Fixes #1510

PR #1530, opened 2025-08-19
feat(event-bus): support to make assertions on dispatched events even when the event bus was not faked

WendellAdriel:
## Summary

* TrackedEventBus: A new decorator class that records dispatched events while delegating to the real EventBus
* Automatic Installation: `EventBusTester` now automatically wraps the EventBus with `TrackedEventBus` on construction
* Dual Mode Support: The tester supports both tracked mode (handlers execute) and faked mode (handlers don't execute)
* Backward Compatibility: Existing tests using `preventEventHandling()` continue to work as before
* Merged Event Tracking: The `EventBusTester` now merges dispatched events from both `TrackedEventBus` and `FakeEventBus`, solving edge cases where components hold references to the old EventBus
* Fixed `assertDispatched()` to properly fail when no events were dispatched

### DIFF
--- a/docs/2-features/08-events.md
+++ b/docs/2-features/08-events.md
@@ -174,9 +174,6 @@ By extending {`Tempest\Framework\Testing\IntegrationTest`} from your test case, 
 These utilities include a way to replace the event bus with a testing implementation, as well as a few assertion methods to ensure that events have been dispatched or are being listened to.
 
 ```php
-// Prevents events from being handled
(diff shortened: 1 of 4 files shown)


WendellAdriel:
@brendt I'm seeing that one test failed, but when I check, it's not related to these changes, and on my local this specific test that's marked as failed is passing. And to get weirder, it's the same one failing from #1529 😅

Issue #1526, opened 2025-08-18
App Specific Configuration Files May Be Overwritten by Default Files During Discovery on Windows


TheRealMrG:
### Tempest version

1.6.0

### PHP version

8.4

### Operating system

Windows

### Description

Since the introduction of priority setting for config files in ```tempest-framework/packages/core/src/Kernel/LoadConfig.php``` the loading of app config files on Windows does not always work. The issue is that the priority for files in the ```/vendor``` directory is not set correctly on Windows. The pattern used in the match statement is ```/vendor/``` but Windows uses ```\``` as a directory separator. This results in the default files sometimes getting a higher priority than app overrides, leading to app specific config files being overwritten by default config files during discovery.

The resolution is to change line 61 of ```tempest-framework/packages/core/src/Kernel/LoadConfig.php```  from:
```Str\contains($path, '/vendor/') => 0,```
to:
```Str\contains($path, DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR) => 0,```

This will use the directory separator native to the platform that PHP is running on.

### Steps to reproduce

1. Create a custom database config file for MySQL as ```app\Config\mysql.config.php``` following the instructions in the Database section of the documentation:
```
<?php

use Tempest\Database\Config\MysqlConfig;
use function Tempest\env;

return new MysqlConfig(
    host: env('DB_HOST', default: 'localhost'),
    port: env('DB_PORT', default: '3306'),
    username: env('DB_USERNAME', default: 'mysql'),
    password: env('DB_PASSWORD', default: 'mysql'),
    database: env('DB_DATABASE', default: 'app'),
);
```
2. Update the Home page controller to query the MySQL database:
```
	#[Get('/')]
	public function __invoke(): View
	{
		$now = $this->database->fetch(new Query(
						sql: <<<SQL
                SELECT NOW();
            SQL,
				));
		return view('home.view.php', now: var_export($now, true));
	}
```
3. Loading the home page returns a QueryWasInvalid exception with the error ```SQLSTATE[HY000]: General error: 1 no such function: NOW```. (The function NOW() is MySQL/MariaDB specific and does not work on sqlite.) 
4. Change the query to ```SELECT sqlite_version();``` and the page now loads successfully, showing that the app specific config is being ignored.

brendt:
Do you want to submit a PR? I'm happy to merge it!

TheRealMrG:
[...]

PR #1535, opened 2025-08-20
fix(core): discover vendor directory first on all platforms

TheRealMrG:
The vendor directory is string matched to give it the lowest priority when loading configuration files during discovery. This change sets the prefixed and suffixed directory separator to the PHP constant DIRECTORY_SEPARATOR rather than hard-coding '/'. This allows the match to work when PHP is running on Windows (which uses '\' as a directory separator), as well as on other platforms.

Resolves tempestphp/tempest-framework#1526


brendt:
Awesome, thanks!

PR #1536, opened 2025-08-20
feat(router): add method spoofing support for HTML forms

### LINKED ISSUE #1510: update/delete verb routes aren't resolved
### Tempest version

1.6.0

### PHP version

8.4

### Operating system

macOS

### Description

This is kinda related to #856 

I am creating a CRUD application, to better understand how Tempest would work in a actual application.

In line with Tempest examples / tests I created a application to manage books, authors and book chapters.
Creating, listing goes well, however when I want to edit or delete a record I run into issues. The route I want to use doesn't resolve. This route is marked with the `#[Post(..)]` or `#{Delete)` attribute.

### Steps to reproduce

**Example implementation**
I think this all that is needed.
<details>
<summary>edit.view.php</summary>

```html
<form action="/books/{{ $book->id }}" method="post">
    <x-csrf-token />
    <input type="hidden" name="_method" value="PUT" />
    <h1 class="text-balance text-5xl font-semibold tracking-tight text-gray-800 sm:text-7xl">
        Edit Book
    </h1>
    <p class="mt-8 text-pretty text-lg font-medium text-gray-500 sm:text-xl/8">
        <input type="text" name="title" value="{{ $originalValues['title'] ?? ($book?->title ?? '') }}" placeholder="Title" class="border placeholder:px-2 px-2" />
        <div :if="isset($errors['title'])" class="text-red-600 text-sm mt-1">
            <div :foreach="$errors['title'] as $error">
                {{ $error->message() }}
            </div>
        </div>
    </p>
    <p class="mt-8 text-pretty text-lg font-medium text-gray-500 sm:text-xl/8">
        <select name="authorId" id="authorId" class="border px-2 py-1">
            <option value="">Select Author</option>
            <option :foreach="$this->authors as $author" :value="$author->id" :selected="($originalValues['authorId'] ?? ($book?->author?->id ?? '')) == $author->id">
                {{ $author->name }}
            </option>
        </select>
        <div :if="isset($errors['authorId'])" class="text-red-600 text-sm mt-1">
            <div :foreach="$errors['authorId'] as $error">
                {{ $error->message() }}
            </div>
        </div>
    </p>
    <p class="mt-8 text-pretty text-lg font-medium text-gray-500 sm:text-xl/8">
        <input type="submit" value="Update" class="bg-blue-500 text-white px-4 py-0.5 rounded-lg" />
    </p>
</form>
```

</details>

<details>
<summary>BookController::edit</summary>

```php
#[Get('/books/{bookId}/edit')]
public function edit(int $bookId): View|Redirect
{
    $book = Book::find(id: $bookId)
        ->with('author.publisher')
        ->first();

    if ($book === null) {
        return new Redirect('/');
    }

    // Get validation errors and original values from session with null safety
    $sessionErrors = $this->session->get(Session::VALIDATION_ERRORS, []);
    $sessionOriginalValues = $this->session->get(Session::ORIGINAL_VALUES, []);

    // Ensure session data is always an array
    $errors = is_array($sessionErrors) ? $sessionErrors : [];
    $originalValues = is_array($sessionOriginalValues) ? $sessionOriginalValues : [];

    return view('Views/Books/edit.view.php',
        title: 'Edit Book',
        book: $book,
        authors: Author::select()->all() ?? [],
        publishers: Publisher::select()->all() ?? [],
        errors: $errors,
        originalValues: $originalValues
    );
}
```

</details>

<details>
<summary>BookController::update</summary>

```php
#[Put('/books/{bookId}')]
public function update(int $bookId, UpdateBookRequest $request): Redirect
{
    $book = Book::find(id: $bookId)->first();

    // Ensure book exists
    if ($book === null) {
        return new Redirect('/');
    }

    $author = Author::find(id: $request->authorId)->first();

    // Ensure author exists
    if ($author === null) {
        // Author not found - redirect back to edit form
        return new Redirect('/books/' . $bookId . '/edit');
    }

    $book->title = $request->title;
    $book->author = $author;
    $book->save();

    return new Redirect('/books/' . $bookId);
}
```
</details>

**Steps**:
1. create a book record, with dependent records
2. visit the edit route
3. update something (without triggering a validation error)
4. submit the form

**Expectation**
I expected the following things to happen upon submitting
1. I would be redirected to the show book page/route
2. the database record would have been updated

**Result**
The browser returns a 404 (outside of the Tempest application)

---

What I've worked out so far is that when resolving the request from the form it looks for a `POST` route, ignoring the `method="PUT"` form element.

WendellAdriel:
## Summary

* Supports PUT, PATCH, and DELETE method spoofing
* Only applies to POST requests for security
* Case-insensitive method values
* Preserves all request data
* Automatic support in `x-form` component
* Backward compatible (HEAD→GET fallback preserved)
* Secure by design (doesn't allow GET spoofing)

## Changes

This updates the `MatchRouteMiddleware` to:
* Check for a `_method` parameter in the request body
* Validate and allow only PUT, PATCH, and DELETE methods for security

This also updates the `x-form.view.php` component to:
* Automatically detects when method is PUT, PATCH, or DELETE
* Render the form with method="POST" for browser compatibility
* Automatically insert a hidden `_method` field with the original method value

## Examples

If using the `<form>` HTML tag, you'll need to set the `method` attribute of the form to `POST` and add a hidden input with the method you want:

```html
<form action="/books/{{ $book->id }}" method="POST">
    <input type="hidden" name="_method" value="PUT" />
</form>
```

If you're using the `x-form` component, it's even easier:

```html
<x-form action="/books/{{ $book->id }}" method="PUT">
</x-form>
```

This will render the `<form>` tag, setting the method to `POST` and automatically inserting a hidden `_method` input with the value you set.

Fixes #1510

### DIFF
--- a/packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
+++ b/packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
@@ -93,13 +93,6 @@ final readonly class PsrRequestToGenericRequestMapper implements Mapper
             return $originalMethod;
         }
 
-        $spoofedMethod = Method::tryFrom(strtoupper($data['_method']));
-        if ($spoofedMethod === null) {
-            return $originalMethod;
-        }
-
-        return $spoofedMethod->isSpoofable()
-            ? $spoofedMethod
-            : $originalMethod;
+        return Method::trySpoofingFrom($data['_method']) ?? $originalMethod;
     }
 }
--- a/packages/view/src/Components/x-form.view.php
+++ b/packages/view/src/Components/x-form.view.php
@@ -14,7 +14,7 @@ if ($method instanceof Method) {
     $method = $method->value;
 }
 
-$needsSpoofing = in_array(strtoupper($method), ['PUT', 'PATCH', 'DELETE'], true);
+$needsSpoofing = Method::trySpoofingFrom($method) !== null;
 $formMethod = $needsSpoofing ? 'POST' : $method;
 ?>
 
--- a/packages/http/src/Method.php
+++ b/packages/http/src/Method.php
@@ -16,11 +16,13 @@ enum Method: string
     case TRACE = 'TRACE';
     case PATCH = 'PATCH';
 
-    public function isSpoofable(): bool
+    public static function trySpoofingFrom(string $method): ?Method
     {
-        return match ($this) {
-            Method::PUT, Method::PATCH, Method::DELETE => true,
-            default => false,
+        $method = Method::tryFrom(strtoupper($method));
+
+        return match ($method) {
+            self::DELETE, self::PATCH, self::PUT => $method,
+            default => null,
         };
     }
 }


packages/router/tests/MatchRouteMiddlewareTest.php
```diff
@@ -0,0 +1,281 @@
...
+        $next = new HttpMiddlewareCallable(
+            fn (Request $_request) => new Ok('Middleware processed'),
+        );
+
+        $response = ($this->middleware)($request, $next);
+
+        $this->assertInstanceOf(NotFound::class, $response);
+    }
+
+    public function test_no_spoofing_without_method_parameter(): void
```
innocenzi: What does this test exactly?

packages/router/tests/MatchRouteMiddlewareTest.php
```diff
@@ -0,0 +1,281 @@
...
+
+        $response = ($this->middleware)($request, $next);
+        $matchedRoute = $this->container->get(MatchedRoute::class);
+
+        $this->assertInstanceOf(Ok::class, $response);
+        $this->assertNotNull($matchedRoute);
+        $this->assertEquals(Method::POST, $matchedRoute->route->method);
+    }
+
+    public function test_head_to_get_fallback_still_works(): void
```
innocenzi: I feel like this test is not needed?

packages/router/tests/MatchRouteMiddlewareTest.php
```diff
@@ -0,0 +1,281 @@
...
+
+        $response = ($this->middleware)($request, $next);
+        $matchedRoute = $this->container->get(MatchedRoute::class);
+
+        $this->assertInstanceOf(Ok::class, $response);
+        $this->assertNotNull($matchedRoute);
+        $this->assertEquals(Method::GET, $matchedRoute->route->method);
+    }
+
+    public function test_method_spoofing_preserves_request_data(): void
```
innocenzi: There is no code in the spoofing implementation that could possibly lead to removal of request data, right? I feel like this test is redundant

packages/router/src/MatchRouteMiddleware.php
```diff
@@ -70,4 +72,40 @@ private function resolveRequest(Request $request, MatchedRoute $matchedRoute): R
...
+            return $request;
+        }
+
+        $allowedMethods = [
+            Method::PUT,
+            Method::PATCH,
+            Method::DELETE,
+        ];
+
+        if (! in_array($spoofedEnum, $allowedMethods, true)) {
```
innocenzi:
```suggestion
        if (! in_array($spoofedEnum, $allowedMethods, strict: true)) {
```

packages/router/src/MatchRouteMiddleware.php
```diff
@@ -70,4 +72,40 @@ private function resolveRequest(Request $request, MatchedRoute $matchedRoute): R
...
+        if (! ($request instanceof GenericRequest)) {
+            return $request;
+        }
+
+        if (! $request->hasBody('_method')) {
+            return $request;
+        }
+
+        $spoofedMethod = $request->get('_method');
+        $spoofedEnum = ($spoofedMethod instanceof Method)
```
innocenzi: I don't think `$request->get` can return a non-scalar value, in which case the `instanceof` check here is not needed

packages/router/src/MatchRouteMiddleware.php
```diff
@@ -70,4 +72,40 @@ private function resolveRequest(Request $request, MatchedRoute $matchedRoute): R
...
         return $request;
     }
+
+    private function applyMethodSpoofing(Request $request): Request
+    {
+        if ($request->method !== Method::POST) {
+            return $request;
+        }
+
+        if (! ($request instanceof GenericRequest)) {
```
innocenzi: Why is this check in place? Why can't we use method spoofing with userland requests?

innocenzi (dismissed):
A few remarks here and there.

@brendt I'm wondering if https://github.com/tempestphp/tempest-framework/blob/2.x/packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php#L55 is not a better place for method spoofing? Just a thought

packages/router/tests/MatchRouteMiddlewareTest.php
WendellAdriel: This was to make sure no spoofing was happening if the spoofing parameter is not sent, but now that I mention it, it seems not needed, so I'm removing it

packages/router/tests/MatchRouteMiddlewareTest.php
WendellAdriel: True

packages/router/tests/MatchRouteMiddlewareTest.php
WendellAdriel: This was mostly because of the clone part, to make sure everything was the same, but yeah. I can remove it.

packages/router/src/MatchRouteMiddleware.php
WendellAdriel: I put it here to be able to use the `->withMethod` that's from the `IsRequest`
Because if I leave without this, it complains that the `Tempest\Http\Request` doesn't have this method, so I thought of adding this.

If you feel like I could remove, should I add the logic from the `withMethod()` here instead?

innocenzi: Ah, I see, to be confirmed by @brendt but `withRequest` should probably be on the `Request` interface then

WendellAdriel: Got it!
Should I apply this change already, or do you want me to wait for @brendt confirmation first?

WendellAdriel: Hey there @innocenzi 
Any news on this? Should I add the `withMethod` in the interface?
Or should I update the code to not use the `withMethod` and clone manually the request and change the method?

WendellAdriel:
Hey @brendt, I updated the PR.
Please take a look to see if it's ok now or if you want me to update anything else on it.
Thanks in advance!

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
```diff
@@ -70,4 +70,31 @@ public function map(mixed $from, mixed $to): GenericRequest
         ])
             ->to(GenericRequest::class);
     }
+
+    private function requestMethod(mixed $from, array $data): Method
```
brendt: `mixed $from` should  be `PsrRequest $from`, we're certain about its type by this point

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
```diff
@@ -70,4 +70,31 @@ public function map(mixed $from, mixed $to): GenericRequest
...
+        if (! isset($data['_method'])) {
+            return $originalMethod;
+        }
+
+        $spoofedMethod = Method::tryFrom(strtoupper($data['_method']));
+        if ($spoofedMethod === null) {
+            return $originalMethod;
+        }
+
+        $allowedMethods = [
```
brendt: Why do we need this check? We've already tried creating a method enum, if it's null, an invalid value was passed. So at this point we're certain it's valid already.

packages/http/tests/Mappers/PsrRequestToGenericRequestMapperTest.php
```diff
@@ -0,0 +1,124 @@
...
+    private PsrRequestToGenericRequestMapper $mapper;
+    private Encrypter $encrypter;
+    private ReflectionMethod $requestMethod;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->encrypter = $this->createMock(Encrypter::class);
+        $this->encrypter->method('decrypt')->willReturnArgument(0);
```
brendt: You don't need to mock the encrypter. Just use an instance of `\Tempest\Cryptography\Encryption\GenericEncrypter`

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
WendellAdriel: This is to allow to spoof only these three methods, so we can't change a request to be a `GET` or a `HEAD` for example. I could transfer this logic by creating a method `isSpoofable` in the Enum if you agree with it.

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
WendellAdriel: Got, it updating now!

packages/http/tests/Mappers/PsrRequestToGenericRequestMapperTest.php
WendellAdriel: Oh, thanks!
I'll change now

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
WendellAdriel: I updated it, I also updated the name to `$request` for the param

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
WendellAdriel: I created the method I said above in the Enum, check if it's ok like this.

packages/http/tests/Mappers/PsrRequestToGenericRequestMapperTest.php
WendellAdriel: I updated this one.
I followed the pattern that we had a `CreatesSigner` trait and created a `CreatesEncrypter` trait for this and updated one other test that was doing the same.

WendellAdriel:
I see that the `PsrRequestToGenericRequestMapper` now has a second parameter for the constructor, I'm fixing this in the tests to make it work!

packages/http/src/Mappers/PsrRequestToGenericRequestMapper.php
brendt: Yeah I think moving it to the enum makes more sense. I'd do it like this:

```php
        $spoofedMethod = Method::trySpoofedFrom(strtoupper($data['_method']));
        if ($spoofedMethod === null) {
            return $originalMethod;
        }
```

And move that in_array check to that method, return `null` if it fails

packages/view/src/Components/x-form.view.php
```diff
@@ -13,10 +13,17 @@
 if ($method instanceof Method) {
     $method = $method->value;
 }
+
+$needsSpoofing = in_array(strtoupper($method), ['PUT', 'PATCH', 'DELETE'], true);
```
brendt: I'd refactor this:

```php
$needsSpoofing = Method::trySpoofingFrom($method) !== null;
```

That way all logic is kept within the Method enum

WendellAdriel:
@brendt I think now it's ok 😅 
Sorry for the back and forth with this.
But if any other change is needed, I'm happy to work on it!
Thanks for taking the time to review and help me with this PR.

brendt (approved):
Very nice! Let's merge :)

PR #1537, opened 2025-08-23
feat(container): support decorators

aazsamir:
[Decorator pattern](https://refactoring.guru/design-patterns/decorator) is one I'm using often and I'm really missing it in Tempest. It can be somehow emulated using initializers but it's not the same.

I added `Decorator` attribute
```php
#[Attribute(Attribute::TARGET_CLASS)]
final readonly class Decorator
{
    public function __construct(
        public string $decorates,
    ) {}
}
```

and to decorate existing implementation
```php
#[Decorator(Repository::class)]
class CacheRepository implements Repository
{
    public function __construct(
        private Repository $decorated,
    ) {}
}
```

-------
That was v1, I'll leave that for the archaeologists.
I added `Decorator` interface
```php
<?php

declare(strict_types=1);

namespace Tempest\Container;

interface Decorator
{
    public function setDecorated(mixed $decorated): mixed;
}
```

and to decorate existing implementation
```php
<?php

declare(strict_types=1);

namespace App;

use Tempest\Container\Decorator;

class CacheRepository implements Decorator, Repository
{
    private Repository $decorated;

    public function setDecorated(mixed $decorated): Repository
    {
        assert($decorated instanceof Repository);
        $this->decorated = $decorated;

        return $decorated;
    }
}
```

As you can see, decorated instance is indicated to discovery by return type of `setDecorated` method (like in Initializers).

### DIFF
--- a/docs/1-essentials/05-container.md
+++ b/docs/1-essentials/05-container.md
@@ -386,6 +386,50 @@ class MyCommand
 
 For these edge cases, it's nicer to make the trait self-contained without having to rely on constructor injection. That's why injected properties are supported.
 
+## Decorators
+
+The container supports the [decorator pattern](https://refactoring.guru/design-patterns/decorator), which allows you to wrap objects and add new behavior to them at runtime without changing their structure. This is particularly useful for adding cross-cutting concerns like logging, caching, validation, or authentication.
+
+To create a decorator, you need to:
+
+1. Use the {b`#[Tempest\Container\Decorates]`} attribute on your decorator class
+2. Implement the same interface as the class you're decorating
+3. Accept the decorated object as a constructor parameter
+
+```php app/Cache/CacheRepository.php
+use Tempest\Container\Decorates;
+
+#[Decorates(Repository::class)]
+final readonly class CacheRepository implements Repository
+{
+    public function __construct(
+        private Repository $repository,
+        private Cache $cache,
+    ) {}
+
+    public function findById(int $id): ?Book
+    {
+        return $this->cache->resolve(
+            key: "book.{$id}",
+            callback: fn () => $this->repository->find($id)
+        );
+    }
+
+    public function save(Book $book): Book
+    {
+        $this->cache->delete("book.{$book->id}");
+
+        return $this->repository->save($book);
+    }
+}
+```
+
+When you request the `Repository` from the container, Tempest will automatically wrap the original implementation with your decorator. The decorated object (the original `Repository`) is injected into the decorator's constructor.
+
+:::info
+Decorators are discovered automatically through Tempest's [discovery](../4-internals/02-discovery.md), so you don't need to manually register them.
+:::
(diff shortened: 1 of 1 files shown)


packages/container/src/GenericContainer.php
```diff
@@ -299,7 +315,28 @@ public function removeInitializer(ClassReflector|string $initializerClass): Cont
...
+        $origin = $this->rawResolve($className, $tag, ...$params);
+
+        if ($this->decorators[$className] ?? null) {
+            $origin = $this->resolveDecorator($className, $origin, $tag, ...$params);
+        }
+
+        return $origin;
+    }
+
+    private function rawResolve(string $className, null|string|UnitEnum $tag = null, mixed ...$params): ?object
```
innocenzi: Something like `resolveDependency` would be clearer, `rawResolve` doesn't have any useful meaning

packages/container/src/Exceptions/DecoratorNotImplementingInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Container\Exceptions;
+
+use Exception;
+
+class DecoratorNotImplementingInterface extends Exception implements ContainerException
```
innocenzi: Class should be final, and the exception class name should follow our standards (event-like name, like `DecoratorDidNotImplementInterface`

packages/container/src/GenericContainer.php
```diff
@@ -299,7 +315,28 @@ public function removeInitializer(ClassReflector|string $initializerClass): Cont
...
+        $decoratedClass = is_string($decoratedClass) ? $decoratedClass : $decoratedClass->getName();
+
+        $this->decorators[$decoratedClass][] = $decoratorClass;
+
+        return $this;
+    }
+
     private function resolve(string $className, null|string|UnitEnum $tag = null, mixed ...$params): ?object
+    {
+        $origin = $this->rawResolve($className, $tag, ...$params);
```
innocenzi: `$origin` -> `$instance`

packages/container/src/GenericContainer.php
```diff
@@ -602,4 +639,35 @@ private function resolveTaggedName(string $className, null|string|UnitEnum $tag)
             ? "{$className}#{$tag}"
             : $className;
     }
+
+    private function resolveDecorator(string $className, mixed $origin, null|string|UnitEnum $tag = null, mixed ...$params): ?object
```
innocenzi: `$origin` -> `$instance`

innocenzi (dismissed):

brendt:
I like it! Although, `#[Decorates()]` is a better name imo than `#[Decorator]`. Can we change that? After that I'll do a code review

brendt:
Very nice! Thanks

PR #1556, opened 2025-09-16
fix(database): handle loading circular eager relations

### LINKED ISSUE #1494: Circular eager loads
When two models reference eachother and both eager load the relation, you end up with an infinite loop

WendellAdriel:
## Summary

Modified `ModelInspector` to track visited model types and prevent infinite recursion:

* Tracking visited model types in an array as we traverse the relationship tree
* Before recursing into a relation, check if the target model type is already in the visited path
* If a circular reference is detected, skip that relation to prevent infinite recursion
* Passing the updated visited paths array to recursive calls

Fixes #1494

### DIFF
--- a/tests/Integration/Database/CircularEagerLoadingTest.php
+++ b/tests/Integration/Database/CircularEagerLoadingTest.php
@@ -4,10 +4,15 @@ declare(strict_types=1);
 
 namespace Tests\Tempest\Integration\Database;
 
+use Tempest\Database\Migrations\CreateMigrationsTable;
+use Tests\Tempest\Fixtures\Migrations\CreateProfileWithEagerTable;
+use Tests\Tempest\Fixtures\Migrations\CreateUserWithEagerTable;
+use Tests\Tempest\Fixtures\Models\ProfileWithEager;
 use Tests\Tempest\Fixtures\Models\UserWithEager;
 use Tests\Tempest\Integration\FrameworkIntegrationTestCase;
 
 use function Tempest\Database\inspect;
+use function Tempest\Database\query;
 
 /**
  * @internal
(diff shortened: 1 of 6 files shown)


WendellAdriel:
Hey @brendt I opened this PR based on an old one I created before.
I'm seeing that it failed one test case.

However, this doesn't seem related to any of the changes I did.
How should I proceed in this case?

tests/Integration/Database/CircularEagerLoadingTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Tests\Tempest\Fixtures\Models\UserWithEager;
+use Tests\Tempest\Integration\FrameworkIntegrationTestCase;
+
+use function Tempest\Database\inspect;
+
+/**
+ * @internal
+ */
+final class CircularEagerLoadingTest extends FrameworkIntegrationTestCase
```
brendt, line 20: I think I'd add one more integration test here that actually performs the database query and maps unto the models, I like to have these high-level tests to make sure we didn't miss anything. 

Apart from that, looking good!

WendellAdriel: Thanks @brendt 
Should I create a migration for each table then to be able to test the inserts and reads?

brendt: Yes :)

WendellAdriel:
Hey @brendt I finished the changes, there's one failing test but IDK if it's realted to my changes tho. Can you help me checking?

brendt (approved):

brendt:
Nice! Thanks for the contribution: I wasn't looking forward doing this myself because I had already spent so much time in the ORM lately :p So I'm happy you took it on to do :)

WendellAdriel:
@brendt I'm happy to help, I'll take a look at other things to help with!
And I like to work with ORMs, so it was nice to work on this one! :partying_face:

PR #1572, opened 2025-09-20
feat(console): support variadic argument

Bapawe:


packages/reflection/src/TypeReflector.php
```diff
@@ -53,12 +53,15 @@
...
 
+    public bool $isVariadic;
+
     public function __construct(
         private PHPReflector|PHPReflectionType|string $reflector,
     ) {
         $this->definition = $this->resolveDefinition($this->reflector);
         $this->isNullable = $this->resolveIsNullable($this->reflector);
         $this->cleanDefinition = str_replace('?', '', $this->definition);
+        $this->isVariadic = $this->reflector instanceof PHPReflectionParameter && $this->reflector->isVariadic();
```
brendt: I'm not sure if this should go in `TypeReflector`, maybe `ParameterReflector` is a better place? I think you did it this way because the `TypeReflector` was already available in `ConsoleArgumentDefinition` ? But you also have access to `ParameterReflector` from within the `ConsoleArgumentDefinition::fromParameter()` function. 

If that's not sufficient, I think a better approach is to refactor `ConsoleArgumentDefinition` a bit to always store the `ParameterReflector` as well. Seems cleaner than having `isVariadic` on the type.

brendt (changes requested):
Cool PR! One relatively small remark :)

packages/reflection/src/TypeReflector.php
Bapawe: Good catch, using `ParameterReflector` is sufficient. Somehow i was to focussed on the type to check if the param is variadic while writing this.

I'll make some more test cases and i'll update the PR description so i can remove the draft status.

brendt (approved):

brendt:
Nice feature! Thanks

PR #1585, opened 2025-10-01
feat(database): add testing utilities

innocenzi:
This pull request adds database testing utilities, such as `assertTableHasRow`, `assertTableHasCount`, `assertTableDoesNotHaveRow`, `assertTableEmpty`, and `assertTableNotEmpty`.

Previously, there was our base test case had a `setupDatabase`, `migrateDatabase` and `migrate` methods. They still exists, but are now deprecated in favor of the ones in the new `DatabaseTester` class.

The behavior is different as well:
- The `setupDatabase` method is now `reset` (with a `setup` alias), and migrates *all registered migrations*. You can set the `migrate` argument to `false` to not migrate anything.
- The `migrate` method migrates all registered migrations if no specific migration is provided.
- I removed `migrateDatabase`.

Additionally, discovery is now configured to scan `tests` instead of `tests/Fixtures`.

### Usage

**New assertions**
```php
$this->database->assertTableHasRow(User::class, name: 'Frieren');
$this->database->assertTableDoesNotHaveRow('users', name: 'Eisen');
$this->database->assertTableHasCount(User::class, count: 1);
$this->database->assertTableEmpty(User::class);
$this->database->assertTableNotEmpty(User::class);
```

**Migrate only specific migrations**

For specific use cases.
```php
$this->database->reset(migrate: false);
$this->database->migrate(CreateMigrationsTable::class, CreateUsersTable::class);

// ...
```

**Migrate everything**

What userland will do most of the time.
```php
$this->database->reset();

// ...
```

### DIFF
--- a/packages/database/src/Testing/DatabaseTester.php
+++ b/packages/database/src/Testing/DatabaseTester.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Tempest\Database\Testing;
+
+use InvalidArgumentException;
+use PHPUnit\Framework\Assert;
+use PHPUnit\Framework\ExpectationFailedException;
+use Tempest\Container\Container;
+use Tempest\Database\Migrations\MigrationManager;
+use TypeError;
+use ValueError;
+
+use function Tempest\Database\query;
+
+final class DatabaseTester
+{
+    public function __construct(
+        private Container $container,
+    ) {}
+
+    /**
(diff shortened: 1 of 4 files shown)


brendt:
Looking good! Will you submit another docs update as well? I already merged this because I want to tag 2.1 early this morning :)

brendt:
As another followup: shall we also provide a rector that fixes all deprecations in tests?

PR #1587, opened 2025-10-02
feat(auth): improve OAuth user flow

brendt:
This PR adds support for 

```php
$user = $this->oauth->authenticate(
            $request,
            function (OAuthUser $oauthUser): Authenticatable {
                return query(User::class)->updateOrCreate([
                    'email' => $discordUser->email,
                ], [
                    'discord_id' => $discordUser->id,
                    'username' => $discordUser->nickname,
                ]);
            });
```

As well as automatic state to session management between OAuth calls.

TODO

- [x] Tests
- [x] Docs

### DIFF
--- a/docs/2-features/17-oauth.md
+++ b/docs/2-features/17-oauth.md
@@ -52,24 +52,22 @@ final readonly class DiscordOAuthController
     #[Get('/auth/discord')]
     public function redirect(): Redirect
     {
-        // Redirects to the OAuth provider's authorization page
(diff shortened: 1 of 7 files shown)


brendt:
@innocenzi I'd like your input on this before I continue: do these changes make sense?

I also wondered what `TestingOAuthClient::authenticate` should look like. Right now I simply assume the oauth state is valid — maybe that's not a good approach?

PR #1595, opened 2025-10-02
refactor(logging)!: dissociate `tempest/debug` from `tempest/log`

innocenzi:
This pull request dissociates `tempest/debug` from `tempest/log`, among other improvements.
- All log files default to being in `.tempest/logs` instead of `./log`
- The debug package has its own config for defining the debug log file
- Debug tailing is now more readable
- Channel properties are now properly commented
- Channels support minimum log levels
- The single `tail` command and the `tail:project` one have been removed, there's just `tail:debug` and `tail:logs` now

This is still a draft because I feel like there could be better APIs for the config, but not sure what exactly. Maybe I'm overthinking and it's fine as-is?

Also, I need to update docs.


innocenzi:
There would be naming conflicts with the new `LogConfig` interface unfortunately. That would be a log of glue code too...

To be honest, the breaking change is minor: if users have a custom logging config, it will be ignored in favor of the default one

brendt:
I'm on the fence about it if we don't have a rector for it. Gonna ask around on Discord for some real life examples.

aazsamir:
`./tempest tail:debug` causes my CPU temperature to rise by 15 degrees :smile: Maybe it's busy waiting? I'm using linux

btw. related XKCD https://xkcd.com/1172/

aazsamir:
> > causes my CPU temperature to rise by 15 degrees
> 
> Do you mean with this branch?

Yes, but now I checked and it's the same on `main`, so not a regression, sorry^^

brendt:
> Gonna ask around on Discord for some real life examples

I believe many people indeed are using custom logging config. I would only be comfortable merging this in 2.x if we have an automated upgrade for it. I'm not sure that'll be trivial to do though 😬

I briefly looked into it, and I'm not sure we can do what we need to do with Rector to automate this upgrade. That being said, maybe my Rector knowledge is lacking. 

Maybe we should simply start preparing for 3.0 and include this one in there?

brendt:
I've given it some more thought and I think I'd like it to be targeted toward 3.x. 

I'm ok tagging 3.x soon though. Let's aim for the beginning of January? That way I can finish my FrankenPHP PR, and then we'll merge 8.5 support in 3.x as well

brendt:
@innocenzi feel free to merge in 3.x when you're ready

PR #1605, opened 2025-10-04
feat(http): support database-based sessions

innocenzi:
This pull request introduces support for database-based sessions. It works by providing a migration file through an installer, a new configuration file, and a built-in database session manager.

The installer publishes the migration file and a `session.config.php` with the following contents:
```php
use Tempest\Http\Session\Config\DatabaseSessionConfig;
use Tempest\DateTime\Duration;

return new DatabaseSessionConfig(
    expiration: Duration::days(30),
);
```

Note that even though the migration is published, it should not be modified, because the database session manager expects the exact columns defined on it. Unfortunately, I don't see a good way to publish the implementation itself without it feeling out of place.

I also fixed an issue with the existing file-based implementation: sessions always expired after the specified expiration based on the session *creation date*, not the last activity. In practice, this results in bad user experience, since users would _have_ to log in even if they're being active. Now, the last activity is tracked, and the expiration is based on that.

### DIFF
--- a/packages/http/src/Session/Session.php
+++ b/packages/http/src/Session/Session.php
@@ -41,6 +41,7 @@ final class Session
     public function __construct(
         public SessionId $id,
         public DateTimeInterface $createdAt,
+        public DateTimeInterface $lastActiveAt,
         /** @var array<array-key, mixed> */
         public array $data = [],
     ) {}
--- a/docs/1-essentials/01-routing.md
+++ b/docs/1-essentials/01-routing.md
@@ -732,22 +732,48 @@ public function store(Todo $todo): Redirect
 
 ### Session configuration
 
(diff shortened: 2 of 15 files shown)


packages/http/src/Session/Session.php
```diff
@@ -41,6 +41,7 @@ final class Session
     public function __construct(
         public SessionId $id,
         public DateTimeInterface $createdAt,
+        public DateTimeInterface $lastActiveAt,
```
brendt, line 44: Will this change be backwards compatible? It's ok if the current session becomes invalid, but there shouldn't be any errors

brendt (approved):
Looks great! I had one question but feel free to merge if it's not a problem

packages/http/src/Session/Session.php
innocenzi: It is! There's no error when upgrading to this PR when sessions already exist

PR #1609, opened 2025-10-05
docs(routing): fix x-error view component snippet

Bapawe:
I just noticed that the x-error view component snippet is not working when i tried to implement my own x-error view component.

This PR updates the snippet so it now uses the validator to fetch the error message.


brendt:
Thank you! Actually that whole section is outdated now that we have form components. I'll merge this for now and update the docs later

PR #1616, opened 2025-10-06
fix(http): xsrf header validation

NeoIsRecursive:
When doing ajax requests users (me atleast) probably expect to use the `xsrf-cookie` value to provide an `X-Xsrf-token` header. 

The issue is that the cookie value is an encrypted version of the csrf uuid, so the comparison will always fail (`uuid` != `encryptedUuid`).

I had to do an `urldecode` on the header value when I tested it in an application, which is why it is there. Should this be done on the client instead?

EDIT: seems like axios does this on the client, so it might be something the client should handle.
Maybe using urlencoded base64 strings by default would be nice? I think webauthn stuff uses that to make it easier sending the data around.

I used laravel as a reference for this fix/change:
https://github.com/laravel/framework/blob/1f0bcbf0941923d7f884459a9f5fcfbd16bb8ac8/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L153

Would this be considered a breaking change?

### DIFF
--- a/packages/auth/src/Installer/AuthenticationInstaller.php
+++ b/packages/auth/src/Installer/AuthenticationInstaller.php
@@ -15,40 +15,42 @@ use function Tempest\root_path;
 use function Tempest\src_path;
 use function Tempest\Support\Namespace\to_fqcn;
 
-final class AuthenticationInstaller implements Installer
-{
-    use PublishesFiles;
-
-    private(set) string $name = 'auth';
-
-    public function __construct(
-        private readonly MigrationManager $migrationManager,
-        private readonly Container $container,
-        private readonly Console $console,
-        private readonly ConsoleArgumentBag $consoleArgumentBag,
-    ) {}
-
-    public function install(): void
+if (class_exists(\Tempest\Console\ConsoleCommand::class, false)) {
+    final class AuthenticationInstaller implements Installer
     {
-        $migration = $this->publish(__DIR__ . '/basic-user/CreateUsersTableMigration.stub.php', src_path('Authentication/CreateUsersTable.php'));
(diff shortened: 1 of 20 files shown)


packages/http/src/Session/VerifyCsrfMiddleware.php
```diff
@@ -67,9 +69,14 @@ private function ensureTokenMatches(Request $request): void
     {
         $tokenFromRequest = $request->get(
             key: Session::CSRF_TOKEN_KEY,
-            default: $request->headers->get(self::CSRF_HEADER_KEY),
         );
 
+        if (! $tokenFromRequest && $request->headers->has(self::CSRF_HEADER_KEY)) {
+            $tokenFromRequest = $this->encrypter->decrypt(
```
innocenzi: We'd need to `try/catch` this because `decrypt` can throw, in which case the header should be ignored instead of the request failing

packages/http/src/Session/VerifyCsrfMiddleware.php
```diff
@@ -67,9 +69,14 @@ private function ensureTokenMatches(Request $request): void
...
+                urldecode($request->headers->get(self::CSRF_HEADER_KEY)),
```
innocenzi: I think it's fine to keep `urldecode` here, I don't see a reason why not if it helps reducing boilerplate on the front-end

NeoIsRecursive: cool! I'll keep it then :)

packages/http/src/Session/VerifyCsrfMiddleware.php
NeoIsRecursive: It is sufficient catching `EncryptionException` right?

Issue #1611, opened 2025-10-06
feat: add ability to specify a custom message for validation rules


azjezz:
### Description

When adding rule validation to a DTO, i would like to be able to specify a custom message.

Example:

```php
use Tempest\Validation\Rules;

final class Book {
    #[Rules\HasLength(
      min: 5,
      minMessage: 'book title is too short.',
      max: 50,
      maxMessage: 'book title is too long!',
    )]
    public string $title;

    #[Rules\IsNotEmptyString(message: 'book description should not be empty!')]
    public string $description;

    #[Rules\HasDateTimeFormat('Y-m-d', 'book publish date is incorrectly formatted!')]
    public ?DateTime $publishedAt = null;
}
```

It would also be nice to be able to specify translation keys instead, allowing users to translate validation error messages depending on the request locale.

```php
use Tempest\Validation\Rules;

final class Book {
    #[Rules\HasLength(
      min: 5,
      minMessage: 'book.error.title.short',
      max: 50,
      maxMessage: 'book.error.title.long',
    )]
    public string $title;

    #[Rules\IsNotEmptyString(message: 'book.error.description.empty')]
    public string $description;

    #[Rules\HasDateTimeFormat('Y-m-d', 'book.error.published_at.format')]
    public ?DateTime $publishedAt = null;
}
```

### Benefits

More customization, and better / more specific error messages to end users.

Issue #1610, opened 2025-10-06
Support `xliff` translation file format


azjezz:
### Description

Tempest localization currenctly supports `yaml` and `json` formats, however, neither of those are really meant for storing translations.

A good format for this would be `xliff`, which is currently supported by Symfony Translation.

### Benefits

XML Localization Interchange File Format is an XML-based file format that standardizes the exchange of translatable content between different localization tools and platforms.

Many professional translation tools support XLIFF, allowing for a seamless workflow for managing multilingual content.

- https://docs.lokalise.com/en/articles/1400816-xliff-xlf-xliff
- https://docs.tolgee.io/platform/formats/xliff
- https://docs.memoq.com/current/en/Workspace/xliff-files.html

brendt:
I'm ok adding it, but it's not high on my personal priority list atm

PR #1618, opened 2025-10-06
feat(validation)!: add ability to specify translation keys for specific properties

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1611

This pull request gives the ability to do that:
```php
final class Book {
    #[Rules\HasLength(min: 5, max: 50)]
    #[TranslationKey('book_management.book_title')]
    public string $title;
}
```

When the `HasLength` rule fails, instead of using the `validation_error.has_length` translation key to get the error message, the validator will use `validation_error.has_length.book_management.book_title`—basically, appending the specified translation key to what would have been the original translation key.

Technically, it's a breaking change, since the returned failing rules are now wrapped in a `FailingRule` VO

### DIFF
--- a/tests/Integration/Validator/TranslationKeyTest.php
+++ b/tests/Integration/Validator/TranslationKeyTest.php
@@ -5,7 +5,6 @@ namespace Tests\Tempest\Integration\Validator;
 use PHPUnit\Framework\Attributes\Test;
 use Tempest\Intl\Catalog\Catalog;
 use Tempest\Intl\Locale;
-use Tempest\Intl\Translator;
 use Tempest\Validation\Exceptions\ValidationFailed;
 use Tempest\Validation\Rules;
 use Tempest\Validation\TranslationKey;


azjezz:
The problem i see with this approach is when using a xliff format to store translations ( ref #1610 ), book related things won't be part of the same group. example:

```xml
<?xml version="1.0"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
    <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
        <body>
            <!-- Group: Book-related translations -->
            <group id="book">
                <!-- Subgroup: Create Book -->
                <group id="book.create">
                    <!-- Form Labels -->
                    <group id="book.create.form">
                        <trans-unit id="book.create.form.title">
                            <source>book.create.form.title</source>
                            <target>Title</target>
                        </trans-unit>

                        <!-- Additional form labels can be added here -->
                    </group>

                    <!-- Error Messages -->
                    <group id="book.create.error">
                        <trans-unit id="book.create.error.title.max_length">
                            <source>book.create.error.title.max_length</source>
                            <target>The book title must not exceed {max} characters.</target>
                        </trans-unit>

                        <trans-unit id="book.create.error.title.min_length">
                            <source>book.create.error.title.min_length</source>
                            <target>The book title must be at least {min} characters long.</target>
                        </trans-unit>
                    </group>
                </group>
            </group>

            <!-- Additional translation units can be added here -->
        </body>
    </file>
</xliff>
```

Compared to the current:

```xml
<?xml version="1.0"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
    <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
        <body>
            <!-- Group: Book-related translations -->
            <group id="book">
                <!-- Subgroup: Create Book -->
                <group id="book.create">
                    <!-- Form Labels -->
                    <group id="book.create.form">
                        <trans-unit id="book.create.form.title">
                            <source>book.create.form.title</source>
                            <target>Title</target>
                        </trans-unit>

                        <!-- Additional form labels can be added here -->
                    </group>
                </group>
            </group>

            <!-- Group: Validation Errors -->
            <group id="validation_error">
                <!-- Has Length Errors -->
                <group id="validation_error.has_length">
                    <!-- Book Management Specific Errors -->
                    <group id="validation_error.has_length.book_management">
                        <trans-unit id="validation_error.has_length.book_management.book_title">
                            <source>validation_error.has_length.book_management.book_title</source>
                            <target>The book title must be between 5 and 50 characters long.</target>
                        </trans-unit>
                    </group>
                </group>
            </group>

            <!-- Additional translation units can be added here -->
        </body>
    </file>
</xliff>
```

Grouping in xliff is useful not just for making things obvious, when using an online translation service where translators will be updating translations, this provides context to them, and makes navigating to the `Book -> Create -> Errors` section much easier.

Also, having parameters for translation here is a MUST, for big projects, people who work on translation never touch the code, nor do they know what is the minimum or maximum value allowed, therefor, it should not be hardcoded in the message.

innocenzi:
I see that the translation units' `id`s are being repeated anyway, so I'm not sure how it's an issue? Because of the tool you use to generate the files? Can you not group book management stuff together, despite the `id`s starting differently?

> Also, having parameters for translation here is a MUST, for big projects, people who work on translation never touch the code, nor do they know what is the minimum or maximum value allowed, therefor, it should not be hardcoded in the message.

There's no code to touch, it's all in the translation message:
```mf2
.input {$field :string}
.input {$min :number default=null}
.input {$max :number default=null}
.local $has_min = {$min :boolean}
.local $has_max = {$max :boolean}
.match $has_min $has_max
  true true {{{$field} must be between {$min} and {$max}}}
  true false {{{$field} must be at least {$min}}}
  false true {{{$field} must be at most {$max}}}
```

The content above would be in the `<target>` tag of your xliff.

See: https://messageformat.unicode.org/
Example: https://messageformat.unicode.org/playground/#ICAgIC5pbnB1dCB7JGZpZWxkIDpzdHJpbmd9CiAgICAuaW5wdXQgeyRtaW4gOm51bWJlciBkZWZhdWx0PW51bGx9CiAgICAuaW5wdXQgeyRtYXggOm51bWJlciBkZWZhdWx0PW51bGx9CiAgICAubG9jYWwgJGhhc19taW4gPSB7JG1pbiA6Ym9vbGVhbn0KICAgIC5sb2NhbCAkaGFzX21heCA9IHskbWF4IDpib29sZWFufQogICAgLm1hdGNoICRoYXNfbWluICRoYXNfbWF4CiAgICAgIHRydWUgZmFsc2Uge3t7JGZpZWxkfSBtdXN0IGJlIGF0IGxlYXN0IHskbWlufX19CiAgICAgIGZhbHNlIHRydWUge3t7JGZpZWxkfSBtdXN0IGJlIGF0IG1vc3QgeyRtYXh9fX0KICAgICAgKiAqIHt7eyRmaWVsZH0gbXVzdCBiZSBiZXR3ZWVuIHskbWlufSBhbmQgeyRtYXh9fX0.ewogICJmaWVsZCI6ICJCb29rIHRpdGxlIGxlbmd0aCIsCiAgIm1pbiI6IDAsCiAgIm1heCI6IDEwMAp9.ZW4tVVM
(don't mind the errors, Tempest has custom helper selectors/functions)

azjezz:
The problem is i want a different message depending on what failed validation, not depending on constraint configuration.

If `min: 5, max: 20`, and user supplied a message that is 4 in length, i want to show an error message specifically about minimum length constraint. not `Field should be between 5 and 20`.

My assumption is the current way this works, user will always see `Field should be between 5 and 20` even if their message is below `20`.

PR #1629, opened 2025-10-08
feat(router): support giving back json when a request fail

### LINKED ISSUE #1619: Json responses from exceptions and validation errors
### Description

When a request wants json (`Accept` header === `'application/json'` or something), request validation errors should be a 424 json-response with the errors as the body.

Also when an exception occurs in the "json" context, it would be preferable to not get the error view, but a json response with the exception.

```php
new Response(
  body: [
      'message' => $failure->getMessage(),
      'code' => $failure->getCode(),
      'file' => $failure->getFile(),
      'line' => $failure->getLine(),
      'trace' => $failure->getTrace(),
]),
```

(a problem details response for all of these scenarios would be cool, but I guess that might be out-of-scope).

Right now you get a redirect or a view, which isn't that nice for api's.

### Benefits

We can handle errors in apis better 👍

NeoIsRecursive:
Fixes #1619

Okay, so this is very wip, but I wanted to get it out to see if I'm thinking about this correctly before doing more work on this 🙂

The error "shape" is taken from Laravel, which I think it is pretty easy to work with:
[invalid](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Foundation/Exceptions/Handler.php#L795) & [generic](https://github.com/laravel/framework/blob/df826dfbb63e8894ea9915b2ba9ce06e407547b0/src/Illuminate/Foundation/Exceptions/Handler.php#L1019).

But it should probably be discussed how it should look (maybe [problem details](https://datatracker.ietf.org/doc/rfc9457/) or some other easier "standard").

EDIT:

Another thing im thinking about is this:

```php
if ($response->status->isServerError() || $response->status->isClientError()) {
    throw new HttpRequestFailed(
        request: $request,
        status: $response->status,
        cause: $response,
    );
}
```

Maybe the errors that are handled by the middleware (in `forward`) shouldn't be cast to `HttpRequestFailed`?
`ConvertsToResponse`, `ValidationFailed` and `RouteBindingFailed`.


packages/router/src/HandleRouteExceptionMiddleware.php
```diff
@@ -38,15 +40,33 @@ public function __invoke(Request $request, HttpMiddlewareCallable $next): Respon
         return $this->forward($request, $next);
     }
 
+    private function wantsJson(Request $request): bool
```
brendt: Maybe we should move this to the request itself somehow. But I'm also weary of adding all kinds of `wants*` methods on the interface. Maybe something like `$request->accepts()` could work as a small improvement already? 

@innocenzi do you have an opinion about it? 

Also, I'm writing this here, but we don't necessarily need to do it in this PR. Just wanted to take the opportunity to discuss. If we agree on something, let's make a separate issue for it.

packages/router/src/Exceptions/HttpExceptionHandler.php
```diff
@@ -49,20 +54,30 @@ public function handle(Throwable $throwable): void
 
     private function renderErrorResponse(Status $status, ?HttpRequestFailed $exception = null): Response
     {
+        if (get(Request::class)->headers->get('Accept') === 'application/json') {
+            return new Json(
```
kniziol: It seems like an excellent fit for a separate class called `JsonResponseCreator`, `JsonResponseFactory`, or something similar 😇

packages/router/src/HandleRouteExceptionMiddleware.php
kniziol: How about we introduce a new method in the `Request` interface?

```php
Tempest\Http\Request::isJson(): bool;
```

NeoIsRecursive: Yes!

It would be quite nice if one could so something like

```php
$request->accepts(ContentType::Json);
```

To minimize the risk for spelling errors.

innocenzi:
```php
$request->accepts(ContentType::Json);
```

I like this API!

packages/router/src/Exceptions/HttpExceptionHandler.php
```diff
@@ -26,19 +28,20 @@ public function __construct(
...
             $this->exceptionReporter->report($throwable);
 
             $response = match (true) {
-                $throwable instanceof ConvertsToResponse => $throwable->toResponse(),
-                $throwable instanceof AccessWasDenied => $this->renderErrorResponse(Status::FORBIDDEN),
-                $throwable instanceof HttpRequestFailed => $this->renderErrorResponse($throwable->status, $throwable),
-                $throwable instanceof CsrfTokenDidNotMatch => $this->renderErrorResponse(Status::UNPROCESSABLE_CONTENT),
-                default => $this->renderErrorResponse(Status::INTERNAL_SERVER_ERROR),
+                $request->accepts(ContentType::HTML, ContentType::XHTML) => $this->handleHtml($throwable),
+                $request->accepts(ContentType::JSON) => $this->jsonHandler->render($throwable),
```
brendt, line 38: Shouldn't we have `HtmlHttpExceptionRenderer` then for consistency?

brendt (changes requested):
Can you also add tests for this?

packages/router/src/Exceptions/HttpExceptionHandler.php
NeoIsRecursive: Yes, absolutely!

Would it make sense to remove the `DeveloperExceptionHandler` and in the `HtmlHttpExceptionRenderer` handle to show either whooosh or the custom view?

Tests for this (entire PR) is on my todo 🙂

brendt: > Would it make sense to remove the DeveloperExceptionHandler and in the HtmlHttpExceptionRenderer handle to show either whooosh or the custom view?

Maybe, but I'd say out of scope for this PR?

NeoIsRecursive:
> Wanted to check up: what needs to be done for this one, and can I help? Just want to make sure we don't loose track: it's a good feature and I'd like to see it merged 👍

Hi, I haven't had a lot of time (nor energy tbh) the last few weeks, but I do have some stuff locally that I'll clean up and push up asap :)

Whats left:
- Tests for the "exceptionRenderers"
- Router tests
- Make sure that the `accepts` guard has all the correct html content types

And I did try to remove the `throwHttpExceptions` thinggy but it triggered a lot of other components to fail so I'll leave that to someone else 😅

brendt:
Let's close for now since the PR seems inactive. Happy to revive it when there's time :)

innocenzi:
I might take a look at this at the same time I take a look at https://github.com/tempestphp/tempest-framework/issues/1805

NeoIsRecursive:
> I might take a look at this at the same time I take a look at https://github.com/tempestphp/tempest-framework/issues/1805

That would be great!
I have unfortunatly not had the time to work more on this.

Though ive been thinking about it and Im wondering if doing it like dotnet does might be nice? (register a exception handler, if it returns false, try the next, until the default one is ran)

PR #1637, opened 2025-10-11
ci: try rector ^2.2.5 to allow global phpunit

samsonasik:
@TomasVotruba @brendt here I try latest rector dev-main to allow global PHPUnit usage.


samsonasik:
^2.2.3 is fine, or just wait for next release as include latest patch for PHPUnit 12+ version check on bootstrap.

PR #1638, opened 2025-10-12
feat(http): add accepts helper method

NeoIsRecursive:
So we can do this:

```php
$request->accepts(ContentType::JSON);
```

Should we handle wildcards? (`Accepts: text/*`)

### DIFF
--- a/packages/http/src/IsRequest.php
+++ b/packages/http/src/IsRequest.php
@@ -154,7 +154,7 @@ trait IsRequest
 
             $mediaTypes[] = [
                 'mediaType' => $acceptedType->before('/')->toString(),
-                'subType' => $acceptedType->afterFirst('/')->toString(),
+                'subType' => $acceptedType->afterFirst('/')->beforeLast(';q')->toString(),
             ];
         }
 
--- a/packages/http/tests/GenericRequestTest.php
+++ b/packages/http/tests/GenericRequestTest.php
@@ -150,7 +150,21 @@ final class GenericRequestTest extends TestCase
         $this->assertTrue($request->accepts(ContentType::XML));
     }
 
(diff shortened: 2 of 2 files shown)


packages/http/src/IsRequest.php
```diff
@@ -136,6 +137,26 @@ public function hasQuery(string $key): bool
         return has_key($this->query, $key);
     }
 
+    public function accepts(ContentType $contentType): bool
```
innocenzi: Would improve the DX to accept multiple types here
```suggestion
    public function accepts(ContentType ...$contentTypes): bool
```

packages/http/src/Request.php
```diff
@@ -57,4 +57,6 @@ public function get(string $key, mixed $default = null): mixed;
     public function getSessionValue(string $name): mixed;
 
     public function getCookie(string $name): ?Cookie;
+
+    public function accepts(ContentType $contentType): bool;
```
innocenzi: Could you document the method in a docblock? We don't have that in the others, but we try to document new stuff, this improves the DX

packages/http/src/IsRequest.php
NeoIsRecursive: Yes! I agree 😄

NeoIsRecursive: @innocenzi  Do you think it should be `acceptsAll` or `acceptsAny`?

I changed it to be "or", that seems like it would be more useful but I'm not sure.

innocenzi: No, `accepts` sounds explicit enough to me, as long as it's an `or`—which I agree with you, it sounds like the most useful check

PR #1643, opened 2025-10-14
feat(intl): add `current_locale` util

Bapawe:
@innocenzi as mentioned in the discussion on Discord, this fixes a bug where validation error messages containing dates were not being localized correctly. Dates were always formatted using the system locale instead of respecting the application's configured currentLocale setting.

The test configuration change ensures consistent behavior across all integration tests regardless of the developer's system language while running the test suite.
 
- Added `Intl\current_locale` helper function to determine the current locale (from `IntlConfig` config or system default as fallback)
- Used `Intl\current_locale` helper function in `create_intl_date_formatter` to determine the current locale when none is provided
- Add i18n.config.php to test fixtures to set default locale for integration tests to Locale::ENGLISH_UNITED_STATES
- Remove ini_set('intl.default_locale', 'en-US') workaround from TranslatorTest
- Remove test_default_locale test as ENGLISH_UNITED_STATES is now the default for integration tests


innocenzi (approved):

innocenzi:
Thank you @Bapawe, awesome work 👌

PR #1656, opened 2025-10-22
feat(http)!: add `--crawl` flag to `static:generate` command

brendt:
I've decided to make link crawling opt-in. It's very resource-heavy on real websites, which I only now noticed after converting stitcher to Tempest.


brendt:
Just for reference: running the static:generate command on stitcher without the crawler runs in 30s, _with_ the crawler takes minutes

brendt:
Gonna merge this because I need it :p , feel free to shout at me if you don't agree 😁

Issue #1663, opened 2025-10-27
Database adapters don't allow to use SSL connection


laylatichy:
### Tempest version

2.4.0

### PHP version

8.4

### Operating system

Linux

### Description

$ ./tempest migrate
// VALIDATING MIGRATION FILES
× // PDOException
× // SQLSTATE[HY000] [1105] unknown error: Code: UNAVAILABLE
      server does not allow insecure connections, client must use SSL/TLS

there is no option to use secure db connection which breaks planetscale for example

same as being unable to pass persistent connection like

$options = [
    PDO::ATTR_PERSISTENT   => true,
    PDO::MYSQL_ATTR_SSL_CA => '/etc/ssl/certs/ca-certificates.crt',
];

### Steps to reproduce

establish SSL connection with db

PR #1664, opened 2025-10-27
fix(database): fix dsn format for mysql connection

laylatichy:
fix for mysql dsn connection string

secure connections require host=x;port=x instead of host=x:x


brendt:
Thanks! I assume this fixes #1663?

laylatichy:
> Thanks! I assume this fixes #1663?

@brendt  that's only a part, because ssl connections require specific dsn, 

what's needed too is a way to pass PDO options to db connector but wasnt sure how you want to implement that or if you want some defaults etc

laylatichy:
@brendt hmm sure, but could you let me know what was the cause?

as per spec https://www.php.net/manual/en/ref.pdo-mysql.connection.php so Im curious a bit ;)

PR #1672, opened 2025-10-31
feat(core): support booting in phar

aazsamir:
I want to pack my application as a PHAR file. However, using the `realpath()` function causes problems because phar files are located in a virtual file system, and tempest does not boot at all.
Therefore, I added the `real_path()` function, which is safe for phar files, and replaced all calls to `realpath()` with `real_path()`.
In addition, tempest's internal storage is now configurable, as we cannot create files inside a PHAR file, so we need to specify an external location instead.

### DIFF
--- a/packages/support/src/Filesystem/functions.php
+++ b/packages/support/src/Filesystem/functions.php
@@ -507,7 +507,7 @@ function read_symbolic_link(string $path): string
 /**
  * Returns the real path for the specified $path or null if it doesn't exist.
  */
-function real_path(string $path): ?string
+function normalize_path(string $path): ?string
 {
     if (class_exists(\Phar::class) && \Phar::running(false) !== '' && str_starts_with($path, 'phar:')) {
         return $path;
--- a/packages/console/src/Highlight/TempestConsoleLanguage/Injections/FileInjection.php
+++ b/packages/console/src/Highlight/TempestConsoleLanguage/Injections/FileInjection.php
@@ -22,9 +22,9 @@ final readonly class FileInjection implements Injection
             pattern: '/(?<match>\<file=(?<quote>[\"\'])(?<file>.+)\k<quote>\s*\/?>)/',
             callback: function (array $matches) {
                 $href = $matches['file'];
-                $exists = Filesystem\real_path($href) !== null;
+                $exists = Filesystem\normalize_path($href) !== null;
                 $file = $exists
-                    ? str(Filesystem\real_path($href))->replace('\\', '/')->stripStart(root_path())->stripStart('/')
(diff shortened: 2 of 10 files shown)


packages/support/src/Filesystem/functions.php
```diff
@@ -503,3 +503,15 @@ function read_symbolic_link(string $path): string
 
     return $result;
 }
+
+/**
+ * Returns the real path for the specified $path or null if it doesn't exist.
+ */
+function real_path(string $path): ?string
+{
+    if (\class_exists(\Phar::class) && \Phar::running(false) !== '' && \str_starts_with($path, 'phar:')) {
```
innocenzi:
```suggestion
    if (class_exists(\Phar::class) && \Phar::running(false) !== '' && str_starts_with($path, 'phar:')) {
```

packages/support/src/Path/functions.php
```diff
@@ -146,6 +146,11 @@ function normalize(null|Stringable|string ...$paths): string
             $path .= '/';
         }
 
+        // Restore virtual phar prefix
+        if (\str_starts_with($path, 'phar:')) {
```
innocenzi:
```suggestion
        if (str_starts_with($path, 'phar:')) {
```

packages/support/src/Path/functions.php
```diff
@@ -146,6 +146,11 @@ function normalize(null|Stringable|string ...$paths): string
...
+            $path = \str_replace('phar:', 'phar://', $path);
```
innocenzi:
```suggestion
            $path = str_replace('phar:', 'phar://', $path);
```

innocenzi (changes requested):
Can you use the `Filesystem\real_path` syntax instead?

Also, have you considered more descriptive names? Maybe `normalize_path`?

packages/support/src/Filesystem/functions.php
aazsamir: hmm, I thought that `real_path` is quite nice because it sounds the same as native `realpath` and behaves similar (but does not contain footguns). It also fits with `Path\xxx_path` functions naming.

Also, there already is a `Path\normalize(...)` function, so I think it would be a bit misleading (for the lack of better word) to name it `normalize_path`.

When typing this I also started thinking if `Path` wouldn't be a better namespace for it, but `Filesystem` acts like sort of wrapper for native filesystem functions, so I think this fits here.

However, I'm not insisting on this. If you think that `normalize_path` (or something else) would be better, I'll change it.

brendt: I think `normalize_path` is more in line with the rest of our naming functions. Let's change it. After that I think we can merge :)

PR #1674, opened 2025-11-04
feat(auth): add OAuth installer

### LINKED ISSUE #1582: oauth installer
`./tempest install oauth` which will show a list of available providers, and automatically:

- create a config file
- create a controller
- install the required league dependencies
- add empty values in `.env.example` and `.env`

For each selected one.

Bapawe:
Fixes #1582

OAuth can be installed optionally during `./tempest install auth` or directly via `./tempest install auth --oauth`.

Not sure how to best handle the UserModel for different OAuth providers, feedback is welcome!

### DIFF
--- a/packages/auth/src/Installer/OAuthInstaller.php
+++ b/packages/auth/src/Installer/OAuthInstaller.php
@@ -137,7 +137,7 @@ final class OAuthInstaller
             ->filter();
 
         if ($packages->isNotEmpty()) {
-            $this->processExecutor->run("composer require {$packages->implode(', ')}");
+            $this->processExecutor->run("composer require {$packages->implode(' ')}");
         }
     }
 


brendt:
By the way, apart from the question about controllers, I think this is a very solid PR and I'm happy to merge it once we settled on an answer 👍

Bapawe:
> 
> I think it's fair to keep a `// todo` in the generated controller, where people should fill in their own logic. Personally, I have one oauth controller action that handles all callbacks, but that's only possible because I only use it for simple logins:

That works for me. I'll add a // TODO comment in the handler and comment out the current implementation in the stub file, so users have a reference to start from.

> Maybe having one such controller is enough? OTOH, I do see value in having one per oauth provider as well. WDYT?

I initially considered combining multiple providers into one controller, but switched to the current approach (one controller per provider) because that's how it was described in #1582.

This approach also makes the OAuth installer easier to maintain since the stub file remains straightforward and doesn't require the installer to generate PHP code dynamically (which would be necessary for a multi-provider controller).

brendt:
True. Let's keep it as is then. Ping me when this is ready for review :)

Bapawe:
@brendt should be ready now.

packages/auth/src/Installer/OAuthInstaller.php
```diff
@@ -0,0 +1,177 @@
...
+use Tempest\Support\Str\ImmutableString;
+
+use function Tempest\root_path;
+use function Tempest\src_path;
+use function Tempest\Support\arr;
+use function Tempest\Support\Filesystem\read_file;
+use function Tempest\Support\Namespace\to_fqcn;
+use function Tempest\Support\str;
+
+final class OAuthInstaller
```
brendt, line 21: Awesome PR. The only thing missing are some tests. Take a look at these tests for inspiration:

- `\Tests\Tempest\Integration\Console\Commands\MakeGeneratorCommandCommandTest`
- `\Tests\Tempest\Integration\Console\Commands\MakeConfigCommandTest`
- `\Tests\Tempest\Integration\Console\Commands\MakeMiddlewareCommandTest`

brendt (changes requested):

packages/auth/src/Installer/OAuthInstaller.php
Bapawe: @brendt You're right, I'll add some tests this weekend.

Also, the OAuth docs should be changed regarding this new OAuth installer, not too sure how much should be changed. Maybe you can do that or give me some directions, then I'll update those this weekend.

PR #1685, opened 2025-11-06
test(core): cover `normalize_path` inside phar

aazsamir:
Follow up to https://github.com/tempestphp/tempest-framework/pull/1672

Added test that builds phar file with `Filesystem\functions.php` and stub file that calls `normalize_path`. Then test executes builded phar and checks output.

I tried different shenanigans with `require XXX` instead of `shell_exec` but without success.


brendt:
Thanks!

PR #1686, opened 2025-11-06
feat(view): standalone `TempestViewRenderer` support

brendt:
This PR allows `TempestViewRenderer` to run truly standalone without any dependencies on discovery or the container

### DIFF
--- a/packages/view/src/Renderers/TempestViewRenderer.php
+++ b/packages/view/src/Renderers/TempestViewRenderer.php
@@ -17,7 +17,6 @@ use Tempest\View\GenericView;
 use Tempest\View\Parser\TempestViewCompiler;
 use Tempest\View\View;
 use Tempest\View\ViewCache;
-use Tempest\View\ViewCachePool;
 use Tempest\View\ViewConfig;
 use Tempest\View\ViewRenderer;
 use Throwable;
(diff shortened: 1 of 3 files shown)


packages/view/src/Renderers/TempestViewRenderer.php
```diff
@@ -26,9 +30,41 @@ public function __construct(
...
+        $compiler = new TempestViewCompiler(
+            elementFactory: $elementFactory,
+            attributeFactory: new AttributeFactory(),
+        );
+
+        $elementFactory->setViewCompiler($compiler);
+
+        $viewCache = new ViewCache(
+            enabled: $cache,
+            pool: new ViewCachePool(__DIR__ . '/../../.tempest/cache'),
```
innocenzi: Should we not use `internal_storage_path` here?

innocenzi (commented):
Took a quick look, and besides the cache path, nothing specific caught my attention

packages/view/src/Renderers/TempestViewRenderer.php
brendt: That's not possible because we're running without the container or kernel, both are required for `internal_storage_path`.

innocenzi: @brendt I think we should still use it and catch the lack of container and _then_ fall back to the hardcoded path

brendt: Ah, yes, that makes sense, I'll follow up: #1694

PR #1692, opened 2025-11-07
feat(router): add `#[Stateless]` attribute

### LINKED ISSUE #1691: Add `#[Stateless]` attribute
This attribute can be added to routes to indicate that they are stateless and should not start sessions

So instead of this

```php
    #[Get('/rss', without: [SetCurrentUrlMiddleware::class, VerifyCsrfMiddleware::class])]
```

You'd write this

```php
    #[Stateless]
    #[Get('/rss')]
```

Or

```php
    #[Stateless, Get('/rss')]
```

brendt:
Closes #1691

### DIFF
--- a/docs/1-essentials/01-routing.md
+++ b/docs/1-essentials/01-routing.md
@@ -633,7 +633,7 @@ final class ErrorResponseProcessor implements ResponseProcessor
 
 ## Stateless routes
 
-When you're building API endpoints, RSS pages, or any other kind of page that does not require any cookie or session data, you mark a route as stateless, skipping all cookie-related logic:
+When you're building API endpoints, RSS pages, or any other kind of page that does not require any cookie or session data, you may use the `{#[Tempest\Router\Stateless]}` attribute, which will remove all state-related logic:
 
 ```php
 use Tempest\Router\Stateless;


docs/1-essentials/01-routing.md
```diff
@@ -631,6 +631,23 @@ final class ErrorResponseProcessor implements ResponseProcessor
 }
 ```
 
+## Stateless routes
+
+When you're building API endpoints, RSS pages, or any other kind of page that does not require any cookie or session data, you mark a route as stateless, skipping all cookie-related logic:
```
innocenzi:
```suggestion
When you're building API endpoints, RSS pages, or any other kind of page that does not require any cookie or session data, you may use the `{#[Tempest\Router\Stateless]}` attribute, which will remove all state-related logic:
```

packages/router/src/GenericRouter.php
```diff
@@ -70,12 +71,32 @@ private function getCallable(): HttpMiddlewareCallable
...
+                        needle: $middlewareClass->getName(),
+                        haystack: $matchedRoute->route->without,
+                        strict: true,
+                    )) {
+                        return $callable($request);
+                    }
+
+                    // Skip middleware that sets cookies or session values when the route is stateless
+                    if (
+                        $matchedRoute->route->handler->hasAttribute(Stateless::class)
```
innocenzi, line 89: QoL idea: make it work on the class itself as well?

packages/router/src/Stateless.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Tempest\Router;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+final class Stateless
```
innocenzi, line 11: We should add some docblocks for handy inline documentation

packages/router/src/GenericRouter.php
brendt: I considered it, but we currently don't support class-based route attributes, so this would cause some inconsistency. I think it might a good idea, but we should tackle the problem as a whole

innocenzi: Makes perfect sense

PR #1693, opened 2025-11-07
feat(auth): mark password property with `#[SensitiveParameter]`

sander-bol:
User passwords should not end up in stack traces or logs.

Most important is adding it to the UserModel stub that is used by the `tempest install auth`.

In the documetation for Hashed this example also shows up, but since it would distract from the purpose of the example there, I've chosen to omit it.

### DIFF
--- a/docs/2-features/04-authentication.md
+++ b/docs/2-features/04-authentication.md
@@ -24,7 +24,7 @@ After publishing, you may run `./tempest migrate`. You now have the building blo
 
 Tempest's authentication is flexible enough not to assume that an authenticatable model is a user. If your application uses a different system for authentication, such as an API key or a service account, you have the ability to create such a model while preserving the correct nomenclature.
 
-To register an authenticatable model, you may create a class that implements the {b`Tempest\Auth\Authentication\Authenticatable`} interface. Tempest automatically discovers this interface. If you run `tempest install auth`, a basic `User` model like below can be created for you.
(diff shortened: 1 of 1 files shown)


docs/2-features/04-authentication.md
```diff
@@ -24,7 +24,7 @@ After publishing, you may run `./tempest migrate`. You now have the building blo
 
 Tempest's authentication is flexible enough not to assume that an authenticatable model is a user. If your application uses a different system for authentication, such as an API key or a service account, you have the ability to create such a model while preserving the correct nomenclature.
 
-To register an authenticatable model, you may create a class that implements the {b`Tempest\Auth\Authentication\Authenticatable`} interface. This interface is automatically discovered by Tempest.
+To register an authenticatable model, you may create a class that implements the {b`Tempest\Auth\Authentication\Authenticatable`} interface. Tempest automatically discovers this interface. If you run `tempest install auth`, a basic `User` model like below can be created for you.
```
innocenzi: This is specified in the section just above 😬

innocenzi (changes requested):

docs/2-features/04-authentication.md
sander-bol: Fair enough :) I'll scrub this change. I just noticed that the documentation-code was the same as what the stub generated.

PR #1695, opened 2025-11-07
feat(router)!: add route decorators

brendt:
Technically this is a breaking change because of how the `Route` interface changed (added `set` hooks). 

I don't think many people are relying on this, but I'm asking around on Discord first to be sure


packages/router/src/RouteDiscovery.php
```diff
@@ -26,15 +26,21 @@ public function discover(DiscoveryLocation $location, ClassReflector $class): vo
             $routeAttributes = $method->getAttributes(Route::class);
 
             foreach ($routeAttributes as $routeAttribute) {
-                $this->discoveryItems->add($location, [$method, $routeAttribute]);
+                $decorators = [
```
brendt, line 29: Moving all of this to the discovery phase, meaning it will be cached.

PR #1709, opened 2025-11-13
fix(eventbus): change dispatched assertion from not null to not empty

laylatichy:
current assert for findDispatches never fails


brendt:
Thanks!

PR #1713, opened 2025-11-14
refactor(container): remove redundant null return types

brendt:


packages/container/src/GenericContainer.php
```diff
@@ -173,7 +173,7 @@ public function config(object $config): self
      * @param class-string<TClassName> $className
      * @return null|TClassName
```
aazsamir: `null` also needs to be removed from docblock here and in `Container` interface

PR #1721, opened 2025-11-18
test(core): support paratest

aazsamir:
Use with
`./vendor/bin/paratest` instead of `./vendor/bin/phpunit`
4min20s -> 50s

So now, tests have custom internalStorage set per worker. Paratest guarantees to set unique TEST_TOKEN.
Tests that modify files MUST use unique names for them (or use lock).

There were some kinda random errors due to container being statically set, so on tests `tearDown` I'm setting it to `null`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ packages/database.sqlite
 packages/database/src/database.sqlite
 src/Tempest/database.sqlite
 tests/Fixtures/database.sqlite
+tests/Fixtures/database*.sqlite
 tests/Unit/Console/test-console.log
 src/Tempest/Database/src/database.sqlite
 .env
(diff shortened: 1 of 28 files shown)


aazsamir:
> I am running into 70-ish database related errors on my machine
> 
> ```
> 62) Tests\Tempest\Integration\Database\ConvenientWhereMethodsTest::test_where_not
> Tempest\Database\Exceptions\QueryWasInvalid: SQLSTATE[HY000]: General error: 1 no such table: users
> ```
> 
> Any ideas to what might cause them? (They only happen with paratest)

hmm, no idea at the moment, I just pulled fresh main and on my machine there is 100% success rate

I suspect these lines, seems suspicious
https://github.com/tempestphp/tempest-framework/blob/main/tests/Integration/FrameworkIntegrationTestCase.php#L51
```php
$databaseConfigPath = __DIR__ . '/../Fixtures/Config/database.config.php';

if (! file_exists($databaseConfigPath)) {
    copy(__DIR__ . '/../Fixtures/Config/database.sqlite.php', $databaseConfigPath);
}
```

Issue #1722, opened 2025-11-18
uri generator not picking up class attributes


laylatichy:
### Tempest version

2.9.1

### PHP version

8.4

### Operating system

Linux

### Description

uri(action: [UserController::class, 'list']) is not working correctly

class attribute for example Prefix is never picked up by normalizeActionToUri

 

### Steps to reproduce

```php
#[Prefix('/manage')]
final readonly class UserController {
    

    #[Get(uri: '/users')]
    public function list(): void {
        ...
    }
}
```

```php
uri(action: [UserController::class, 'list']) // will return /users instead of /manage/users
```

brendt:
Fixed in 2.9.3

PR #1724, opened 2025-11-18
fix(router): use route registry to generate uris

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/1722

This PR introduces a lookup by controller method for all routes, so the URI generator can use it instead of reproducing the logic inside the route discovery.

It also handles an edge case where a controller method has multiple different routes: in this case, an exception is thrown to suggest the user to specify the path manually, to avoid any confusion since the previous behavior would be confusing (before, it would use the first defined route).

### DIFF
--- a/tests/Integration/Route/UriGeneratorTest.php
+++ b/tests/Integration/Route/UriGeneratorTest.php
@@ -12,7 +12,6 @@ use Tempest\Http\Method;
 use Tempest\Router\Exceptions\ControllerActionDoesNotExist;
 use Tempest\Router\UriGenerator;
 use Tests\Tempest\Fixtures\Controllers\ControllerWithEnumBinding;
-use Tests\Tempest\Fixtures\Controllers\ControllerWithRepeatedRoutes;
 use Tests\Tempest\Fixtures\Controllers\EnumForController;
 use Tests\Tempest\Fixtures\Controllers\PrefixController;
 use Tests\Tempest\Fixtures\Controllers\TestController;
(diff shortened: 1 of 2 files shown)


tests/Integration/Route/UriGeneratorTest.php
```diff
@@ -249,4 +262,44 @@ public function cannot_add_custom_signature(): void
...
+        $this->assertSame(
+            '/prefix/endpoint',
+            uri(PrefixController::class),
+        );
+    }
+
+    #[Test]
+    public function is_current_uri_with_prefix_decorator(): void
+    {
+        $this->http->get('/prefix/endpoint')->assertOk();
```
brendt: Can you add a test for when both the controller class and method have a prefix? In my branch I changed the order of class and method decorators in discovery, because that was still a bug

brendt (changes requested):

packages/router/src/Exceptions/ControllerMethodHasMultipleRoutes.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Tempest\Router\Exceptions;
+
+use Exception;
+use Tempest\Support\Arr;
+
+final class ControllerMethodHasMultipleRoutes extends Exception implements RouterException
+{
+    /** @param string[] $routes */
+    public function __construct(string $controllerClass, string $controllerMethod, array $routes)
```
brendt: We don't need an exception here. We just need to use the first route for URI generation.

brendt (changes requested):

PR #1729, opened 2025-11-21
fix(view): php 8.5 null offset deprecation warning

buismaarten:
```
Deprecated: Using null as an array offset is deprecated, use an empty string instead in .../app/vendor/tempest/framework/packages/view/src/Parser/TempestViewParser.php on line 12
```


brendt:
Thanks!

PR #1790, opened 2025-12-07
feat(session): add redis session manager

### LINKED ISSUE #1700: Redis session support

thoresuenert:
Close #1700 

Questions:
- Do we need actions in cleanup function when we set the expiration on the redis stored data?
- Is the RedisSessionConfig the right way to set a prefix for session keys? 
- Do we need multiple tagged Redis setups?
- How can we setup the Redis client to a different database for sessions in this scenario? If we want to

Laravel uses multiple configuration an treat redis as database, to separate cache from other things they configure multiple connections to different databases. 
The Redis Client itself allows to set the prefix for the client as a whole. From my understanding to support that we need a tagged(`'session'`) redis client. I dont understand how to do that without touchung the kv-store.

For example `RedisCacheConfig`:
The cache uses the same RedisConfig as the RedisSessionConfig and the data will be written in the same database.

### DIFF
--- a/packages/http/src/Session/Managers/RedisSessionManager.php
+++ b/packages/http/src/Session/Managers/RedisSessionManager.php
@@ -139,7 +139,6 @@ final readonly class RedisSessionManager implements SessionManager
         do {
             $result = $this->redis->command('SCAN', $cursor, 'MATCH', $this->getKey(new SessionId('*')), 'COUNT', '100');
             $cursor = $result[0];
-            ld($result);
             foreach ($result[1] as $key) {
                 $sessionId = $this->getSessionIdFromKey($key);
 
--- a/packages/http/src/Session/Managers/RedisSessionManager.php
+++ b/packages/http/src/Session/Managers/RedisSessionManager.php
@@ -11,6 +11,7 @@ use Tempest\Http\Session\SessionDestroyed;
 use Tempest\Http\Session\SessionId;
 use Tempest\Http\Session\SessionManager;
 use Tempest\KeyValue\Redis\Redis;
+use Tempest\Support\Str\ImmutableString;
 use Throwable;
 
 use function Tempest\event;
@@ -122,10 +123,29 @@ final readonly class RedisSessionManager implements SessionManager
         return sprintf('%s_%s', $this->sessionConfig->prefix, $id);
     }
 
+    private function getSessionIdFromKey(string $key): SessionId
+    {
+        return new SessionId(new ImmutableString($key)->afterFirst('_')->toString());
+    }
+
     public function cleanup(): void
     {
-        // what should we do here?
-        // on persist we set the expiration (ttl) for the session in the redis store
-        // in theory all session data should be expire by itself
+        $cursor = '0';
+
+        do {
+            $result = $this->redis->command('SCAN', $cursor, 'MATCH', $this->getKey(new SessionId('*')), 'COUNT', '100');
+            $cursor = $result[0];
+
+            foreach ($result[1] as $key) {
+                $sessionId = $this->getSessionIdFromKey($key);
+
+                if ($this->isValid($sessionId)) {
+                    continue;
+                }
+
+               $this->destroy($sessionId);
+            }
+
+        } while( $cursor != '0');
     }
 }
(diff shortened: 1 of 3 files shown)


packages/http/src/Session/Managers/RedisSessionManager.php
```diff
@@ -0,0 +1,131 @@
...
+        $data = $this->getData($id);
+
+        unset($data[$key]);
+
+        $this->persist($id, $data);
+    }
+
+    public function destroy(SessionId $id): void
+    {
+        $this->redis->getClient()->del($this->getKey($id));
```
innocenzi: This should use the `command` method here so all underlying clients are supported. `getClient` is meant as a userland escape hatch

packages/http/src/Session/Managers/RedisSessionManager.php
```diff
@@ -0,0 +1,131 @@
...
+
+        return $session;
+    }
+
+    private function getKey(SessionId $id): string
+    {
+        return sprintf('%s_%s', $this->sessionConfig->prefix, $id);
+    }
+
+    public function cleanup(): void
```
innocenzi, line 135: Even though TTL guarantees cleanup, this is a public method that we should be able to call manually, so it needs to be implemented as well for manual cleanup

innocenzi (changes requested):

packages/http/src/Session/Managers/RedisSessionManager.php
thoresuenert: I use UNLINK instead of DEL because it is not blocking and redis handles the garbage collection in the background.

packages/http/src/Session/Managers/RedisSessionManager.php
thoresuenert: Done. Used the scan command. The keys will include the prefix, this will open up room for errors if the provided prefix includes `_` itself. 

Dont know how to handle this. I dont tested it if this is a none issue when the prefix is set on the redis client itself.

packages/http/src/Session/Managers/RedisSessionManager.php
```diff
@@ -0,0 +1,154 @@
...
+        if ($data !== null) {
+            $session->data = $data;
+        }
+
+        $this->redis->set($this->getKey($id), serialize($session), $this->sessionConfig->expiration);
+
+        return $session;
+    }
+
+    private function getKey(SessionId $id): string
```
thoresuenert, line 121: Maybe this SessionId to Key transformation is not needed if we implement tagged redis clients and we set the prefix on the client.
Same for getSessionIdFromKey.

Didnt find a way to overwrite the setOption from the SessionManager to test my idea.

tests/Integration/Http/RedisSessionTest.php
```diff
@@ -0,0 +1,286 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Integration\Http;
```
xHeaven: Shouldn't this be `namespace Tests\Tempest\Integration\Http;`?

packages/http/src/Session/Managers/RedisSessionManager.php
```diff
@@ -0,0 +1,153 @@
...
+            other: $session->lastActiveAt->plus($this->sessionConfig->expiration),
+        );
+    }
+
+    private function resolve(SessionId $id): ?Session
+    {
+        try {
+            $content = $this->redis->get($this->getKey($id));
+            return unserialize($content, ['allowed_classes' => true]);
+        } catch (Throwable $e) {
```
xHeaven:
```suggestion
        } catch (Throwable) {
```

Drop the unused variable.

tests/Integration/Http/RedisSessionTest.php
thoresuenert: You are totally right.

brendt:
For now I don't think we should widen the scope of this PR, and just assume one Redis instance. We can add tagged Redis configs later, but let's not complicate this PR with it. 

Apart from that, all looks fine!

brendt (approved):

brendt:
@innocenzi did you still have an open change request?

innocenzi (approved):

PR #1791, opened 2025-12-08
feat(mapper): support contextual serializers and casters

### LINKED ISSUE #1522: Support serializing DTOs for front-end consumption
Here is the setup. I have a user model with a DTO, which gets serialized properly to the database, and casted back as well. This works and it looks like that:

```json
{
  "full_name": "Jon Doe"
  "metadata": "{\"type\":\"metadata\",\"data\":{\"theme\":\"dark\"}}"
}
```

However, I have another serialization context. Instead of serializing for the database, I want to serialize for front-end consumption. This means that the DTO should be serialized as an actual array, not as a JSON string. This is what I need, instead of the above:

```json
{
  "full_name": "Jon Doe"
  "metadata": {
    "type": "metadata",
    "data": {
      "theme": "dark",
    }
  }
}
```

I assume I could do `map($objectFetchedFromDatabase)->with(SomeDedicatedMapper::class)->do()`, but I don't think that's ideal. Imagine having to do that for every API or Inertia endpoint.

So maybe the specific JSON-string mapping (example 1) should only be done in a database context, and the normal mapping (example 2) should be the default behavior.

innocenzi:
Depends on #1789 
Closes #1522 

This pull requests adds the concept of contexts to casters and serializers. It works by adding the ability to the mapper to use a context (string, enum or `Context` class that might hold data).

The mapper will inject the context to serializers/casters/mappers when resolved from the container, and will use serializers/casters registered for that context (or fall back to the default context).

This allows, for instance, the database to use dedicated serializers/casters. Before this pull request, we had to manually add a condition for boolean values to be transformed properly depending on the current database dialect. After this PR, this functionality works through a serializer:

```php
#[Context(DatabaseContext::class)]
final class BooleanSerializer implements Serializer
{
    public function __construct(
        private DatabaseContext $context,
    ) {}

    public static function for(): array
    {
        return ['bool', 'boolean'];
    }

    public function serialize(mixed $input): string
    {
        if (! is_bool($input)) {
            throw new ValueCouldNotBeSerialized('boolean');
        }

        return match ($this->context->dialect) {
            DatabaseDialect::POSTGRESQL => $input ? 'true' : 'false',
            default => $input ? '1' : '0',
        };
    }
}
```

Serializers and casters are now discovered by Discovery instead of being registered through an initializer. Since ordering is important for them, they support priority ordering through `Tempest\Core\Priority`.

It's also possible for userland to register their own casters and serializers this way. To use a specific context, we use the `in` method on `ObjectFactory` (or `SerializerFactory` or `CasterFactory`):

```php
$array = map($object)
    ->in('inertia')
    ->toArray();
```

To pass data to a caster/serializer/mapper, an object extending `Context` may be used instead of a string/enum. The database package now uses a `DatabaseContext` under the hood:

```php
return map($query->fetch())
    ->with(SelectModelMapper::class)
    ->in(new DatabaseContext(dialect: $this->dialect))
    ->to($this->model->getName());
```

As seen above, the `BooleanSerializer` expects a `DatabaseContext` in its constructor. Note that for now, the injection works by parameter name, not parameter type.

### DIFF
--- a/packages/debug/composer.json
+++ b/packages/debug/composer.json
@@ -5,9 +5,11 @@
     "minimum-stability": "dev",
     "require": {
         "php": "^8.5",
-        "tempest/console": "dev-main",
         "tempest/highlight": "^2.11.4",
         "symfony/var-dumper": "^7.1"
+    },
+    "require-dev": {
+        "tempest/console": "dev-main"
     },
     "autoload": {
         "psr-4": {


innocenzi:
This is ready for an initial review! What do we think about this approach?

packages/mapper/src/Serializer.php
```diff
@@ -4,7 +4,20 @@
...
+    /**
+     * Declares what this serializer can handle.
+     * Can return:
+     * - A string type name: 'bool', 'int', 'string', etc.
+     * - A class name: DateTime::class
+     * - A Closure for complex matching: fn (TypeReflector $type) => $type->matches(SomeClass::class)
+     *
+     * @return false|string[]|string|(Closure(TypeReflector):bool)
+     */
+    public static function for(): false|array|string|Closure;
```
innocenzi: I'm considering moving this to another interface, because caster/serializers that don't need to be discovered and registered in the factories are forced to implement this and return `false`. Not a huge deal, but not the best DX for end users.

innocenzi: That could be a `DiscoverableSerializer`, not sure about the name. Currently, the return value of `for()` is checked in the discovery classes, and if it's `false`, the serializer is not registered.

With a different interface, discovery would just look for that interface.

docs/1-essentials/03-database.md
```diff
@@ -304,34 +304,41 @@ final class User
 
 The encryption key is taken from the `SIGNING_KEY` environment variable.
 
-### DTO properties
+### Data transfer object properties
 
-Sometimes, you might want to store data objects as-is in a table, without there needing to be a relation to another table. To do so, it's enough to add a serializer and caster to the data object's class, and Tempest will know that these objects aren't meant to be treated as database models. Next, you can store the object's data as a json field on the table (see [migrations](#migrations) for more info).
+It's possible to store objects as-is in a `json` column, when they don't need to be part of your database schema.
+
+To do this, you may simply mark these objects with the `#[Tempest\Mapper\SerializeAs]` and specify a name for this object to be serialized as. Not that this name must uniquely correspond to a single class.
```
brendt: Note* instead of not I reckon?

docs/1-essentials/03-database.md
```diff
@@ -304,34 +304,41 @@ final class User
...
+        #[Hashed, SensitiveParameter]
+        public ?string $password,
+        public Settings $settings,
+    ) {}
 }
 
-#[CastWith(DtoCaster::class)]
-#[SerializeWith(DtoSerializer::class)]
-final class Backtrace
+#[SerializeAs('user_settings')]
```
brendt, line 328: I don't really understand what `user_settings` is referring to? Is it the database column name? Don't we already have `#[MapTo]` that we could maybe build on further? 

I understand the need for this feature, but the current naming is confusing to me

packages/database/src/Builder/QueryBuilders/InsertQueryBuilder.php
```diff
@@ -40,6 +43,14 @@ final class InsertQueryBuilder implements BuildsQuery
 
     public ModelInspector $model;
 
+    private Database $database {
+        get => get(Database::class, $this->onDatabase);
```
brendt, line 47: If we can avoid service location, we should. I still have to read through the rest of the code, so maybe it'll become clear why we need this, but just sharing my first thoughts up front.

I'm also ok keeping it for now, but refactoring these query builders at a later stage to support proper DI. That's outside the scope of this PR, of course. So it's not a blocker, just wanted to share my thoughts.

packages/database/src/DatabaseContext.php
```diff
@@ -0,0 +1,15 @@
...
+
+use Tempest\Database\Config\DatabaseDialect;
+use Tempest\Mapper\Context;
+
+final class DatabaseContext implements Context
+{
+    private(set) string $key = self::class;
+
+    public function __construct(
+        private(set) DatabaseDialect $dialect,
```
brendt, line 13: Do we really need the dialect here? Because this is what's causing the need for service location in the query builders. Can we avoid it somehow?

packages/database/src/Casters/DataTransferObjectCaster.php
```diff
@@ -1,21 +1,36 @@
...
+use Tempest\Reflection\TypeReflector;
 use Tempest\Support\Arr;
 use Tempest\Support\Json;
 
 use function Tempest\Mapper\map;
 
-final readonly class DtoCaster implements Caster
+#[Priority(Priority::HIGHEST)]
+#[Context(DatabaseContext::class)]
+final readonly class DataTransferObjectCaster implements Caster
```
brendt: I'm a little concerned about naming collisions when we'd have another DTO caster for another context.

packages/mapper/src/Context.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Tempest\Mapper;
+
+/**
+ * Provides context information to casters, serializers and mappers.
+ */
+interface Context
```
brendt, line 8: I know it's more verbose, but `MapperContext` sounds more clear to me. `Context` is rather vague.

packages/mapper/src/Context.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    /**
+     * A unique identifier for this context.
+     */
+    public string $key {
```
brendt: I think `contextName` or `name` makes more sense?

packages/mapper/src/Casters/ObjectCaster.php
```diff
@@ -4,21 +4,39 @@
...
+use Tempest\Mapper\Context;
+use Tempest\Mapper\DynamicCaster;
 use Tempest\Mapper\Mappers\ArrayToObjectMapper;
+use Tempest\Reflection\PropertyReflector;
 use Tempest\Reflection\TypeReflector;
 
 use function Tempest\Mapper\map;
 
-final readonly class ObjectCaster implements Caster
+#[Priority(Priority::HIGH)]
```
brendt, line 18: I like that we're reusing priority :D

packages/mapper/src/DynamicCaster.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Tempest\Mapper;
+
+use Tempest\Reflection\PropertyReflector;
+
+interface DynamicCaster
```
brendt, line 10: Maybe add a docblock explaining what this interface is used for?

packages/mapper/src/DynamicSerializer.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Tempest\Mapper;
+
+use Tempest\Reflection\PropertyReflector;
+use Tempest\Reflection\TypeReflector;
+
+interface DynamicSerializer
```
brendt, line 10: Maybe add a docblock explaining what this interface is used for?

packages/mapper/src/Caster.php
```diff
@@ -4,7 +4,22 @@
...
+     * Declares what this caster can handle.
+     *
+     * Can return:
+     * - A string type name: 'bool', 'int', 'string', etc.
+     * - A class name: DateTime::class
+     * - A Closure for complex matching: fn (TypeReflector $type) => $type->matches(SomeClass::class)
+     *
+     * @return false|string|array|(Closure(TypeReflector):bool)
+     */
+    public static function for(): false|array|string|Closure;
```
brendt: Why do we need this and is there another way? 

If we really need such a method, I think it's much more intuitive that it always takes some input, and returns a boolean: `public function canCast(mixed $input): bool`

packages/mapper/src/Mappers/ObjectToArrayMapper.php
```diff
@@ -65,10 +67,14 @@ private function resolvePropertyValue(PropertyReflector $property, object $objec
             return $propertyValue;
         }
 
-        if ($propertyValue !== null && ($serializer = $this->serializerFactory->forProperty($property)) !== null) {
+        if ($propertyValue !== null && ($serializer = $this->serializerFactory->in($this->context)->forProperty($property)) !== null) {
             return $serializer->serialize($propertyValue);
         }
 
+        // if (is_object($propertyValue)) {
```
brendt: Cleanup?

packages/mapper/src/MappingContext.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Tempest\Mapper;
+
+use Stringable;
+use UnitEnum;
+
+final class MappingContext implements Context, Stringable
```
brendt, line 8: What is this one used for?

brendt: Is it the "default" one?

packages/mapper/src/ObjectFactory.php
```diff
@@ -27,12 +28,21 @@ final class ObjectFactory
 
     private bool $isCollection = false;
 
+    private Context|UnitEnum|string|null $context = null;
```
brendt, line 31: I wish we'd already have type aliases :/

packages/mapper/src/Serializer.php
brendt: Definitely better to have it in another interface (I assume the same is true for the caster interface?)

I also don't like the current implementation, but I already mentioned that in another comment

docs/1-essentials/03-database.md
innocenzi: Basically, `#[SerializeAs]` is discovered and stored in the `MapperConfig`. It's what  allows for this class to be serialized like this:

```json
{
  "type": "user_settings",
  "data": {
    "theme": "dark",
    "hide_sidebar_by_default": true
  }
}
```

Without a name here, the FQCN would be used. This is a bad thing because it prevents refactoring class names in code, since at de-serialization time, the old FQCN stored in this JSON string would not correspond to any class.

We already had this for a while: https://github.com/tempestphp/tempest-framework/pull/1449

packages/database/src/Builder/QueryBuilders/InsertQueryBuilder.php
innocenzi: I definitely agree, I don't like service location. The issue is that since query builders have the ability to refer to another database on-the-fly (through `OnDatabase`, the database being referred to can't just be passed in the constructor.

Would love to see this refactored at some point too

packages/database/src/DatabaseContext.php
innocenzi: Yes, we do—this information is needed for the boolean serializer to serialize booleans in the proper format (true/false for postgres, 1/0 for mysql/sqlite).

Even if we avoid having data in the context, we'd still need the query builder to be aware of the dialect in order to provide a different context to the serializer

packages/database/src/Casters/DataTransferObjectCaster.php
innocenzi: I actually ended up removing the mapper package's `DataTransferObjectCaster` and `DataTransferObjectSerializer` as I found out they are not needed.

That being said, there _will_ be potential name collisions for other serializers/casters depending on their package. I don't think we'll have this in Tempest though. It would be a userland thing.

packages/mapper/src/Context.php
innocenzi: Man, I struggled with naming there 😬 There's also a `packages/mapper/src/Attributes/Context.php` attribute!

And a `packages/mapper/src/MappingContext.php` value object too.

That being said, I kind of dislike naming classes with their own namespace, it's redundant?

packages/mapper/src/Context.php
innocenzi: Fine by me!

packages/mapper/src/MappingContext.php
innocenzi: Yes it's the default context, which also serves as a factory. I could separate the `from` method in another `ContextFactory` class 🤔

packages/mapper/src/ObjectFactory.php
innocenzi: Agreed. 😔

packages/mapper/src/DynamicCaster.php
innocenzi: I'm not even sure of the name honestly. It's used for casters/serializers that need to be created from a property or a type, and can't just be resolved from the container.

innocenzi: I've opted for `ConfigurableCaster` and `ConfigurableSerializer`, renamed the `make` method in `configure` and added docblocks. I think this is better.

packages/mapper/src/Caster.php
innocenzi: We need this because casters and serializers need to dynamically match properties and values.

This is actually the same as your original design; before it was in the initializer classes, you could specify `string`, or `DateTime::class`, or a closure to dynamically match the property/value to the caster/serializer.

This is the exact same, but since casters and serializers are now discovered, this needs to be declared in the class itself.

I'll work on separating this to another interface and see if it can be a normal method instead of having to return a closure

innocenzi: Managed to make it cleaner! Though you can't just specify types or array of types, it's a bit more verbose.

```
public static function accepts(PropertyReflector|TypeReflector $input): bool
{
    $type = $input instanceof PropertyReflector
        ? $input->getType()
        : $input;

    return $type->matches(Address::class);
}
```

brendt: Can't we only have `TypeReflector`?

brendt: This is already a lot cleaner btw

packages/database/src/Builder/QueryBuilders/InsertQueryBuilder.php
brendt: Great, let's keep it as is for now

packages/mapper/src/Caster.php
innocenzi: The serializer factory has a `forProperty` used in the ORM mostly, so serializers do have to be aware of the property

brendt (approved):

PR #1792, opened 2025-12-08
feat(core): support frankenphp worker mode

brendt:


public/index.php
```diff
@@ -1,14 +1,37 @@
...
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-HttpApplication::boot(__DIR__ . '/../', discoveryLocations: [
+$discoveryLocations = [
     new DiscoveryLocation('Tests\\Tempest\\Fixtures\\', __DIR__ . '/../tests/Fixtures/'),
-])->run();
+];
+
+if (function_exists('frankenphp_handle_request')) {
```
xHeaven, line 13: This won't be enough, as this method exists even when FrankenPHP is running in non-worker mode, and the call below to this method will throw an exception.

This would be a proper check:

```suggestion
$isWorkerMode = $_SERVER['FRANKENPHP_WORKER'] ?? false;

if ($isWorkerMode && function_exists('frankenphp_handle_request')) {
```

xHeaven (changes requested):

xHeaven:
I'm not sure this is the proper way to do this. The entire point of containerizing Tempest would be that the user don't need PHP on their system at all - so having to invoke the `tempest` binary is not exactly what we want here.

We could just ship with a `Dockerfile`:
```Dockerfile
FROM serversideup/php:8.5-frankenphp

USER root

RUN install-php-extensions intl ftp

USER www-data

WORKDIR /var/www/html
```

And then have 2 configurations, one in simple mode, one in worker mode:
```yml
x-volume: &app-volume .:/var/www/html

services:
  php:
    build: .
    container_name: tempest
    ports:
      - "80:8080"
      - "443:8443"
    volumes:
      - *app-volume
    environment:
      SSL_MODE: "mixed"
```

```yml
x-volume: &app-volume .:/var/www/html
x-worker: &worker-config "worker /var/www/html/public/index.php"

services:
  php:
    build: .
    container_name: tempest
    ports:
      - "80:8080"
      - "443:8443"
    volumes:
      - *app-volume
    environment:
      SSL_MODE: "mixed"
      FRANKENPHP_CONFIG: *worker-config
```

aidan-casey:
I am in agreement with @xHeaven here. This should be a simple Dockerfile that bundles the application and allows a worker mode + non-worker mode.

I'm not sure if this should be part of the framework so much as an application scaffold/package that can publish to the application.

xHeaven:
> I am in agreement with @xHeaven here. This should be a simple Dockerfile that bundles the application and allows a worker mode + non-worker mode.
> 
> I'm not sure if this should be part of the framework so much as an application scaffold/package that can publish to the application.

It could be an installable package like Sail for Laravel IMHO. We could also provide our own binary to run things as well (just like `sail` again), we could even call it `dockest`. 👀 

Then you could do:
```bash
./dockest serve # simple server
./dockest serve --mode=worker # worker mode
```

brendt:
@aidan-casey @xHeaven I'm super open for your feedback here as I'm a total noob in this area. My only goal is to get Tempest properly working with FrankenPHP worker mode, I'm totally fine with whatever approach. 

The reason I briefly stopped working on this PR is because I ran into framework-specific errors that were impossible to read in worker-mode output. 

I talked about it with Kevin Dunglas, who suggested writing errors to files and then tail them, but then I ran into not knowing how to do that with a Docker container 😅

Long story short: I could use some help :p

innocenzi:
Note: https://github.com/tempestphp/tempest-framework/issues/1834#issuecomment-3719257223

henderkes:
@brendt I gave this a spin because I wanted to branch out our benchmarking to not be limited to Symfony only (adding Laravel and Tempest) but found some very unexpected results for Tempest:

Tempest no worker mode:
```wrk
Running 15s test @ http://localhost/blog/discovery-explained/
  4 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   281.69ms   49.70ms 913.73ms   90.38%
    Req/Sec    18.28      8.31    40.00     76.58%
  1055 requests in 15.02s, 35.12MB read
Requests/sec:     70.24
Transfer/sec:      2.34MB
❯❯ tempest-docs git:(main) 22:40 curl -o /dev/null -s -w 'Total time: %{time_total}s\n' http://localhost/blog/
Total time: 0.196324s
```

Tempest with worker mode:
```wrk
Running 15s test @ http://localhost/blog/discovery-explained/
  4 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   381.14ms  385.88ms   1.95s    83.91%
    Req/Sec    15.39     10.28    60.00     67.72%
  746 requests in 15.02s, 24.83MB read
  Socket errors: connect 0, read 0, write 0, timeout 21
Requests/sec:     49.66
Transfer/sec:      1.65MB
❯❯ tempest-docs git:(main) 22:41 curl -o /dev/null -s -w 'Total time: %{time_total}s\n' http://localhost/blog/
Total time: 0.028908s
```

While for Laravel and Symfony it instead increased RPS:

Symfony no worker mode:
```wrk
Running 15s test @ http://localhost/en/
  4 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    42.94ms   28.90ms 362.86ms   97.90%
    Req/Sec   127.35     14.39   151.00     80.78%
  7486 requests in 15.01s, 400.24MB read
Requests/sec:    498.61
Transfer/sec:     26.66MB
```

Symfony worker mode:
```wrk
Running 15s test @ http://localhost/en/
  4 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    22.84ms    1.88ms  50.90ms   79.60%
    Req/Sec   219.72     12.64   252.00     76.00%
  13146 requests in 15.03s, 703.00MB read
Requests/sec:    874.53
Transfer/sec:     46.77MB
```

Do you have any explanation what in the code could cause this? For what it's worth, all benchmarks taken with FrankenPHP as the webserver, stock configuration. I didn't want to bother optimising nginx+fpm and stock configuration is terrible, so I didn't include it, as it was hovering around 1/4th of FrankenPHP's RPS.

brendt:
We haven't made any optimizations yet for worker mode, I'm even surprised it worked. We want to work on this after 3.0 has been release to truly optimize worker-mode support.

Meanwhile, I'm going to close this PR because it was mainly some experimentation from my side and others have said they want to tackle the problem for real.

Issue #1805, opened 2025-12-11
404 page not showing - server panics instead


adamkiss:
### Tempest version

2.9.1

### PHP version

8.4

### Operating system

macOS

### Description

404 Pages are not showing, instead Router panics and throws 500. You can see it even on the Tempest Framework site.

Additionally, if I set `ENVIRONMENT` to `production` the missing route returns `200 OK`, but the content leaks paths/errors:

```
<b>Fatal error</b>:  Uncaught Tempest\Http\HttpRequestFailed: Failed request: 404 Not Found in /[PATH]/vendor/tempest/framework/packages/router/src/HandleRouteExceptionMiddleware.php:28
```
### Steps to reproduce

Open https://tempestphp.com/asdasdasd

innocenzi:
Caused by https://github.com/tempestphp/tempest-framework/pull/1624 it seems, cc @brendt

innocenzi:
Specifying `false` prevents autoloading, so `class_exists` may just return `false`. Hence, we will not register any exception handler.

brendt:
Should be fixed in 2.14.0

adamkiss:
@brendt is tempestphp.com supposed to show error 500 page on 404?

brendt:
Yes, that's because of a catchall route which we need to fix on the project level (or maybe the framework, but unrelated to this)

adamkiss:
@brendt Cool! I'm on the road, so I couldn't double check http code

adamkiss:
@brendt @innocenzi This is definitely still not working as intended.

Following the [docs on custom error page](https://tempestphp.com/2.x/features/exception-handling#customizing-the-error-page), I created a [minimal repro where home returns `NotFound`](https://github.com/adamkiss/tempest-404-fail), and instead of `GenericResponse` with 404 status, I still get an uncaught `HttpRequestFailed` exception thrown.  

Is this something else?

Edit: If I switch from local to production environment, I actually do get 404 page with 404 status (wooo!), but it's the default Tempest 404, not my custom `GenericResponse`

xHeaven:
> [@brendt](https://github.com/brendt) [@innocenzi](https://github.com/innocenzi) This is definitely still not working as intended.
> 
[...]

PR #1806, opened 2025-12-11
revert: fix: make package dependencies optional where possible (#1624)

innocenzi:
This pull request reverts #1624. It doesn't work as expected (see https://github.com/tempestphp/tempest-framework/issues/1805).


brendt:
Oh that's annoying :(

Did you do `git revert` or manually removed the checks? Because I fear this bulk change will have unforeseen sideeffects. It should also target main for now. I'll submit a new PR with the checks removed. 

Unfortunately it also means we have no proper solution for our independent packages then 😬

Issue #619, opened 2024-10-24
Support UUIDs in ORM


brendt:
- Postgres has the UUID type: https://www.postgresql.org/docs/current/datatype-uuid.html
- MySQL advices to store UUIDs as binary values because of index performance: https://dev.mysql.com/blog-archive/storing-uuid-values-in-mysql-tables/
- SQLite could use BLOB to store the binary representation

In case we need to convert from UUID to binary (MySQL and SQLite), the easiest way to do it would be in PHP:

- That way we don't have to worry about special inserts per database-dialect. 
- Selecting should be done using the binary format of the UUID, so we'd have to convert it before building the select query
- The `CreateTableStatement` and `AlterTableStatement` should simply have a `uuid()` method that creates the field with a unique index
- We might have an option to store the "human readable" version in the database as well

An example of how this was done with a Laravel package (before Laravel had support for it): https://github.com/spatie/laravel-binary-uuid/tree/master

Having thought it through, I think we're in the clear about breaking changes, and can move this to post-1.0

iamdadmin:
Just adding, it would be good if the UUID supported was specifically UUIDv7 so it's sortable https://uuid7.com - coming from Laravel I've been using ULID for the advantages, but UUIDv7 makes ULID unnecessary and obviously there's wider support then.

PR #1807, opened 2025-12-11
feat(database): support uuids as primary columns

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/619

This pull request adds support for generating UUID v7 primary keys during insertion. Note that since `IsDatabaseModel` provides a `id` property, which cannot be extended to receive the `#[Uuid]` attribute, this feature is technically not compatible with `IsDatabaseModel`.

```php
final class User
{
    #[Uuid]
    public PrimaryKey $id;

    public function __construct(
        public string $name,
    ) {}
}
```

### DIFF
--- a/packages/database/src/QueryStatements/UuidPrimaryKeyStatement.php
+++ b/packages/database/src/QueryStatements/UuidPrimaryKeyStatement.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\QueryStatements;
+
+use Tempest\Database\Config\DatabaseDialect;
+use Tempest\Database\QueryStatement;
+
+final readonly class UuidPrimaryKeyStatement implements QueryStatement
+{
+    public function __construct(
+        private string $name = 'id',
+    ) {}
+
+    public function compile(DatabaseDialect $dialect): string
+    {
+        return match ($dialect) {
+            DatabaseDialect::MYSQL => sprintf('`%s` VARCHAR(36) PRIMARY KEY', $this->name),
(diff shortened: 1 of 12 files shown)


packages/database/src/QueryStatements/UuidPrimaryKeyStatement.php
```diff
@@ -0,0 +1,24 @@
...
+final readonly class UuidPrimaryKeyStatement implements QueryStatement
+{
+    public function __construct(
+        private string $name = 'id',
+    ) {}
+
+    public function compile(DatabaseDialect $dialect): string
+    {
+        return match ($dialect) {
+            DatabaseDialect::MYSQL => sprintf('`%s` VARCHAR(36) PRIMARY KEY', $this->name),
```
homersimpsons, line 19: (I read this while reading the Tempest 3.0 announcement, thanks for your work on this guys)

`VARCHAR(36)` is not the optimal datatype. I do not know what is feasible with tempest.

- A `CHAR(36)` should be better
- An old article https://dev.mysql.com/blog-archive/storing-uuid-values-in-mysql-tables/ using `BINARY(16)`.
- UUID Support in MySQL8 https://dev.mysql.com/blog-archive/mysql-8-0-uuid-support/ (still uses `BINARY(16)`

Also, I do not know how MariaDB is handled, but they have an UUID type: https://mariadb.com/docs/server/reference/data-types/string-data-types/uuid-data-type

(I do not know about sqlite)

brendt: Thanks! I made an issue for it: https://github.com/tempestphp/tempest-framework/issues/1966

PR #1809, opened 2025-12-12
fix(http): cleanup session flash values

### LINKED ISSUE #1734: Session::cleanup() is not called before shutdown
### Tempest version

2.9.3

### PHP version

8.4

### Operating system

macOS

### Description

### Steps to reproduce

brendt:
Fixes #1734


brendt:
Closed in favor of #1829

PR #1819, opened 2025-12-15
feat(core)!: overhaul exception handling

### LINKED ISSUE #1619: Json responses from exceptions and validation errors

### LINKED ISSUE #1805: 404 page not showing - server panics instead

innocenzi:
Supersedes #1629
Fixes #1805
Fixes #1619

This pull request attempts to properly fix the error handling to allow for more flexibility regarding how responses are returned when there are errors.

### Exception renderers

Previously, we had the rendering logic in our exception handler. The exception handler is now responsible for finding the right exception renderer to render an exception as a response.

Exception renderers implement the `ExceptionRenderer` interface, which require a `canRender` and a `render` method.

We now provide a `JsonExceptionRenderer` and a `HtmlExceptionRenderer`, which check for the `Accept` header to decide whether they can be used. For now, the HTML renderer has priority because browsers send `*/*` as an `Accept` value, but I feel like we should have the equivalent of a `wantsJson` and also use that as a heuristic.

Users may now implement their own 404 or other error pages by implementing their own renderer:

```php
final class NotFoundExceptionRenderer implements ExceptionRenderer
{
    public function canRender(Throwable $throwable, Request $request): bool
    {
        if (! $request->accepts(ContentType::HTML)) {
            return false;
        }

        if (! $throwable instanceof HttpRequestFailed) {
            return false;
        }

        return $throwable->status === Status::NOT_FOUND;
    }

    public function render(Throwable $throwable): Response
    {
        return new NotFound(
            body: view('./404.view.php'),
        );
    }
}
```

Note that while used internally, `HttpRequestFailed` is now also intended for userland usage. It's not always convenient to return a response (like `NotFound`) for errors due to nesting or return types.

The optional message that this exception accepts will be displayed in the error pages in production, too. When generating our own `HttpRequestFailed` internally, we use generic error messages and don't leak exception details.

&nbsp;

### Exception page

I got rid of Whoops and implemented our own exception page (highly inspired by Laravel's design). It's not feature-complete, but it's a good basis with more features and quality of life than Whoops.

CleanShot 2025-12-17 at 23 02 55

> [!NOTE]
> Its front-end makes a lot of noise in the diff. It's in `packages/router/src/Exceptions/local`, so unless you want to review Vue stuff, you can ignore that. Also note that the distribution files must be built whenever a change is made to the front-end.

&nbsp;

### `ExceptionsConfig`

It adds a convenient way to disable the default `LoggingExceptionReporter`, instead of having to write a kernel boot event listener to remove it manually.

It also contains the discovered exception reporters instead of the `AppConfig`. They needed their own home.

&nbsp;

### Removed `throwHttpExceptions` in tests

This feature was a bit of a work around the inability to catch exceptions in tests. The issue is that it introduced inconsistent behavior during tests and in production, and it was weird to reason about.

The new workaround is simply to catch exceptions during `HttpRouterTester#sendRequest`. When an exception is thrown, the exception handler is manually called to render the response, which is passed down to the `TestResponseHelper`.

Users that used `throwExceptions()` manually before will need to adapt their tests.

&nbsp;

### Removed `Invalid`

I removed `Invalid` as a response. It felt out of place and made it hard to have proper code for HTML and JSON handling.

Plus, its constructor expected "internal" properties (request, failed rules, failing class FQCN). Its logic is now in dedicated exception renderers when `ValidationFailed` is thrown.

&nbsp;

### Renamed `ExceptionProcessor` and `ExceptionReporter`

I swapped their names because it makes more sense:

- The `ExceptionProcessor` is responsible for processing exceptions. It's not discoverable, replacing it requires an initializer.

- The `ExceptionReporter` classes are discoverable userland classes meant for reporting exceptions. By default, we still offer a `LoggingExceptionReporter` which write exceptions in log files.

It's a breaking change only if users implemented their own `ExceptionProcessor`, which they have to rename now.

&nbsp;

### Renamed `HasContext`

I renamed it to `ProvidesContext`. The purpose of the interface is to be able to determine if the exception can provide more context, so I wanted to clarify its name.

&nbsp;

## To-dos

- [x] Get rid of `throwHttpExceptions`
- [x] Support JSON responses
- [x] Make HTML and JSON exception renders customizable in userland
- [x] Ensure ability to override error pages
- [x] Documentation
- [x] More tests
- [x] Custom development exception page
- [ ] Add an option to display vendor code snippets
- [ ] Add "copy as markdown" / "open with ChatGPT/Claude/Gemini"


packages/router/src/Exceptions/HtmlExceptionRenderer.php
```diff
@@ -0,0 +1,110 @@
...
+    ) {}
+
+    public function canRender(Throwable $throwable, Request $request): bool
+    {
+        return $request->accepts(ContentType::HTML, ContentType::XHTML);
+    }
+
+    public function render(Throwable $throwable): Response
+    {
+        if ($throwable instanceof ConvertsToResponse) {
```
NeoIsRecursive: This is in the match bellow too, one is sufficient right?

NeoIsRecursive (commented):
One thing I was thinking about, and I might have missed, translations on the "built-in" error messages?

Really looking forward to using this 👌

packages/router/src/Exceptions/HtmlExceptionRenderer.php
innocenzi: Whoops, good catch. Artifact of a refactor 😅

innocenzi:
> One thing I was thinking about, and I might have missed, translations on the "built-in" error messages?

That made me go through an unexpected clean-up rabbit hole! It's now implemented for HTML responses 👍

packages/router/src/RouteConfig.php
```diff
@@ -24,15 +24,16 @@ public function __construct(
         /** @var class-string<\Tempest\Router\ResponseProcessor>[] */
         public array $responseProcessors = [],
 
+        /** @var array<int,class-string<\Tempest\Router\Exceptions\ExceptionRenderer>> */
```
xHeaven:
```suggestion
        /** @var array<int, array<class-string<\Tempest\Router\Exceptions\ExceptionRenderer>>> */
```

packages/router/src/Exceptions/HttpExceptionHandler.php
```diff
@@ -2,73 +2,54 @@
...
-                    Status::NOT_FOUND => 'This page could not be found on the server',
-                    Status::FORBIDDEN => 'You do not have permission to access this page',
-                    Status::UNAUTHORIZED => 'You must be authenticated in to access this page',
-                    Status::UNPROCESSABLE_CONTENT => 'The request could not be processed due to invalid data',
-                    default => null,
-                },
-            ]),
-        );
-    }
+        ksort($this->routeConfig->exceptionRenderers);
```
xHeaven, line 42: Shouldn't we sort during discovery instead?

innocenzi: I'd rather sort as the last step in case a renderer is added during runtime for any reason

docs/1-essentials/01-routing.md
```diff
@@ -299,7 +299,7 @@ final class PasswordlessAuthenticationController
     public function __invoke(Request $request): Response
     {
         if (! $this->uri->hasValidSignature($request)) {
-            return new Invalid();
+            throw new HttpRequestFailed(Status::UNPROCESSABLE_CONTENT);
```
brendt, line 302: I'll need to get used to `Invalid` being gone, but I also understand why. I'm ok with this change

packages/core/src/AppConfig.php
```diff
@@ -15,9 +15,6 @@ public function __construct(
 
         ?string $baseUri = null,
```
brendt, line 17: We'd need to provide a Rector for this change

packages/core/src/Exceptions/ExceptionProcessor.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Tempest\Core\Exceptions;
+
+use Throwable;
+
+/**
+ * Responsible for processing exceptions when they are thrown.
+ * By default, Tempest implements {@see GenericExceptionProcessor}, which calls all discovered {@see ExceptionReporter}.
+ */
+interface ExceptionProcessor
```
brendt, line 11: We'd need a Rector for this

packages/core/src/Exceptions/ExceptionReporter.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Tempest\Core\Exceptions;
+
+use Throwable;
+
+/**
+ * Classes implementing this interface are automatically discovered and may report all thrown exceptions in some way.
+ */
+interface ExceptionReporter
```
brendt, line 10: Also a Rector for this one

packages/core/src/ProvidesContext.php
```diff
@@ -2,10 +2,10 @@
 
 namespace Tempest\Core;
 
-interface HasContext
+interface ProvidesContext
```
brendt, line 5: Rector for this one

brendt (commented):
The logo in `packages/router/src/Exceptions/local/src/assets/tempest-logo.svg` is the old one. The current one is here: https://github.com/tempestphp/.github/tree/main/logo_current

packages/router/src/RouteConfig.php
```diff
@@ -24,15 +24,16 @@ public function __construct(
...
+
         /** @var Middleware<\Tempest\Router\HttpMiddleware> */
         public Middleware $middleware = new Middleware(
             HandleRouteExceptionMiddleware::class,
             MatchRouteMiddleware::class,
             SetCookieMiddleware::class,
             HandleRouteSpecificMiddleware::class,
         ),
-
-        public bool $throwHttpExceptions = true,
```
brendt, line 35: Ideally we'd have a rector for this. Simply removing the property suffices

brendt (changes requested):

packages/core/src/AppConfig.php
innocenzi: I honestly don't think it's needed because those were discovered, I highly doubt anyone would manually be filling this property

packages/router/src/RouteConfig.php
innocenzi: Same thing, `RouteConfig` is not documented and is actually rather internal, I don't think a rector is needed here

packages/core/src/Exceptions/ExceptionReporter.php
innocenzi: This one is completely new, the previous reporter was a simple class. The Rector rule that corresponds to this one is `ExceptionProcessor` -> `ExceptionReporter`, I'll add it

innocenzi:
Do you have the logo in SVG?

I added two Rector rules for the interface changes, but didn't add the config ones (harder, and I don't believe they are relevant). If that's fine by you, we can go ahead and merge :D

xHeaven:
> Do you have the logo in SVG?

https://tempestphp.com/img/tempest-logo.svg

packages/core/src/AppConfig.php
brendt: Ok, fair enough

packages/router/src/RouteConfig.php
brendt: Ok

brendt (approved):

brendt:
Awesome!!

PR #1823, opened 2025-12-15
fix(auth): support running migrations for the first time during auth installation

### LINKED ISSUE #1684: Running migrations as part of `tempest install auth` does not create migrations table
### Tempest version

2.5

### PHP version

8.4

### Operating system

macOS

### Description

Another one of my "rocky start" bugs (I'm really good at doing stuff you did not expect new users to do, it seems :D).

If you have not run migrations before, answering Yes to the migration question in `tempest install auth` ends up with an error due to the migrations table not yet existing. Running `migrate:up` before doing the install works, but is not documented as a required step for getting started.

Perhaps the `migrate:up` command does preflight checks that are skipped when going via `tempest install` ?

### Steps to reproduce

1. `composer create-project tempest/app`
2. `php tempest install auth`
3. Yes to all the steps, specifically "Do you want to execute migrations?" (defaults to No).

```
× // Tempest\Database\Exceptions\QueryWasInvalid
× // SQLSTATE[HY000]: General error: 1 no such table: migrations
      
      INSERT INTO `migrations` (`name`, `hash`) VALUES ('0000-00-00_create_users_table', '8176018ac1ea405ef12fad65424a0c72')
```

4. Trying to recover from this error by running `php tempest migrate:up` now also fails, because now the `users` table already exists, without the migration being tracked by the previous attempt.

Tried to dive into the code, but didn't have enough time yet to pinpoint the cause of this.

MohammadAlhallaq:
fixes #1684

### DIFF
--- a/packages/auth/src/Installer/AuthenticationInstaller.php
+++ b/packages/auth/src/Installer/AuthenticationInstaller.php
@@ -9,7 +9,6 @@ use Tempest\Console\Input\ConsoleArgumentBag;
 use Tempest\Container\Container;
 use Tempest\Core\Installer;
 use Tempest\Core\PublishesFiles;
-use Tempest\Database\Database;
 use Tempest\Database\Migrations\CreateMigrationsTable;
 use Tempest\Database\Migrations\MigrationManager;
 
--- a/packages/auth/src/Installer/AuthenticationInstaller.php
+++ b/packages/auth/src/Installer/AuthenticationInstaller.php
@@ -9,6 +9,8 @@ use Tempest\Console\Input\ConsoleArgumentBag;
 use Tempest\Container\Container;
 use Tempest\Core\Installer;
 use Tempest\Core\PublishesFiles;
+use Tempest\Database\Database;
+use Tempest\Database\Migrations\CreateMigrationsTable;
 use Tempest\Database\Migrations\MigrationManager;
 
 use function Tempest\root_path;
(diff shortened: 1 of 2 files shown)


innocenzi:
Hey, thanks! I don't think hardcoding the migration here is the right approach. I'd probably look into fixing the migration command inside the auth installer instead.

MohammadAlhallaq:
> Hey, thanks! I don't think hardcoding the migration here is the right approach. I'd probably look into fixing the migration command inside the auth installer instead.

yup, I agree, will be done 👍🏻

MohammadAlhallaq:
> Hey, thanks! I don't think hardcoding the migration here is the right approach. I'd probably look into fixing the migration command inside the auth installer instead.

@innocenzi  Hey, maybe a combination of checking in the manager and taking action in the installer, keeping the manager as our source of truth for whether the migrations table exists

MohammadAlhallaq:
@brendt @innocenzi what do you guys think about this one ?

brendt:
Makes me wonder: why don't we just run `migrate:up` from within that installer? That way we don't have to check for anything at all. The migration file should have been published in the step before, right?

muhammadHallaq:
correct, will change that

MohammadAlhallaq:
> Makes me wonder: why don't we just run `migrate:up` from within that installer? That way we don't have to check for anything at all. The migration file should have been published in the step before, right?

just updated it, maybe something like this ?

brendt:
No I don't think this is the right approach because migrate:up will run all migrations, and you shouldn't have to manually run another after that. I'll take a look at this myself

Issue #1775, opened 2025-12-03
Validation rules based on closures


brendt:
Closures are allowed in attributes as of 8.5

MohammadAlhallaq:
@innocenzi If no one is working on this yet, would you mind assigning it to me ?

innocenzi:
Sure, go ahead!

innocenzi:
Though to be honest, I don't particularly like the idea. We already support closure-based rules in places where it makes sense (eg. console), but closures in attributes (which I assume Brent meant here) looks hella ugly and bad practice to me :D

MohammadAlhallaq:
> Though to be honest, I don't particularly like the idea. We already support closure-based rules in places where it makes sense (eg. console), but closures in attributes (which I assume Brent meant here) looks hella ugly and bad practice to me :D

Yeah, I totally see your point and I kind of share it with you, but you know Brent and attributes 😄, he always tries to utilize them to their fullest.

Issue #1007, opened 2025-03-13
`#[ValidateWith]`


brendt:
Add a `#[ValidateWith]` attribute which is used by the validator. It works similar to casters. There should also be the possibility to have built-in validators mapped to types (like for example, `DateTimeRule` should map to `DateTime` type

Related to #999

MohammadAlhallaq:
@brendt do you think we can tackle this next after #1828  got merged? Like having it behave the same way as casters but with built-in validation rules?

brendt:
Ah yes, we could still improve this support, indeed. Not sure if it's worth pursuing right now though

brendt:
Actually, I'm going to close this issue. If needed we can reiterate later, but I don't think we should spend time on it now.

PR #1828, opened 2025-12-21
feat(validation): add closure-based validation

MohammadAlhallaq:
#1775


innocenzi:
A few things:
- The name of the attribute/rule should be something more explicit that "Closure"
- There should be a test with a class and a property actually using the rule
- As I said in the issue, I really don't think this is a good pattern that we should promote by supporting it

MohammadAlhallaq:
> A few things:
> 
> * The name of the attribute/rule should be something more explicit that "Closure"
> * There should be a test with a class and a property actually using the rule
> * As I said in the issue, I really don't think this is a good pattern that we should promote by supporting it

Yeah, I was planning to revisit it today, I’ll take care of the first two points, and we can wait for brendt's review.

packages/validation/src/Rules/Predicate.php
```diff
@@ -0,0 +1,45 @@
...
+use ReflectionFunction;
+use Tempest\Validation\Rule;
+
+/**
+ * Custom validation rule defined by a closure.
+ *
+ * The closure receives the value and must return true if it is valid, false otherwise.
+ */
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
+final readonly class Predicate implements Rule
```
brendt: I do think this is a valuable feature, but I don't think this is a good name for it. I'd even go for something as

```php
    #[Validate(static function (string $value): bool {
        return str_contains((string) $value, '@');
    })]
```

brendt (changes requested):

packages/validation/src/Rules/Predicate.php
innocenzi: Maybe `#[ValidateWith]`? See: https://github.com/tempestphp/tempest-framework/issues/1007

brendt: Ah yes, that's good

MohammadAlhallaq: yup, will be done

MohammadAlhallaq: do you think we can tackle #1007 next? Like having it behave the same way as casters with built-in validation rules?

brendt:
Perfect, thanks!

PR #1829, opened 2025-12-21
refactor(http)!: improve session management and CSRF protection

### LINKED ISSUE #1699: Database Sessions Improved Perfomance
### Description

I don't mark it as a bug, because it's not a bug per se but DatabaseSession is unusable in current state if you have > 10ms latency

each set/create/remove send a persist call in DatabaseSessionManager, on each request there is easily > 10 calls to either of them and if you are running db not locally or not on the same close network each call is easily in 10s of ms, slowing down the response by 200-500ms

now I rolled out my own session manager because of that and I added response parser that simply persists the state if it has to, some operations set some value then remove it later on within the same requests so persistence in that case is not needed. Not sure if thats the best way to do so, but wasnt able to get middleware working for that. My first idea was middleware that will run as last in the chain but even with priority juggling I wasnt able to do so

Redis/Valk support would be nice too ;)

### Benefits

Remove pointless calls to db within the same request

Add redis/valk support

### LINKED ISSUE #1734: Session::cleanup() is not called before shutdown

innocenzi:
Closes #1734
Closes #1699
Replaces #1809

This pull request improves the implementation details of multiple parts of the `http` package, separates the "previous URL", the validation errors and the form original values from the session, and implements CSRF protection through `Sec-Fetch-*` headers instead of a CSRF token. 

## `Session` and `SessionManager`

So, I was never a fan of dependency location in `Session`, especially given the fact that this class is meant to be serialized. My initial goal with this PR was to refactor it to make it a singleton value object and to improve performance of session managers.

`Session` now only holds its data and do nothing else. It doesn't have methods to save the session, destroy it or anything. Those are now the sole responsibility of `SessionManager` implementations, which have also been simplified.

I improved the performance of session manager by limiting their IO calls. Previously, they would load/save when retrieved and on any change. Now, they load once and save once, ad the start of the request and at the end. The `ManageSessionLifecycleMiddleware` middleware is responsible for the last part, but also implicitly the first part since `Session` is lazy-loaded using an initializer.

&nbsp;

## CSRF protection

I replaced the CSRF protection from a classic CSRF token approach to using [`Sec-Fetch-Site`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site) and [`Sec-Fetch-Mode`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Mode), which are [baseline](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).

This cleans up a bit of code, removes the hard-coded CSRF token from the session, removes the `<x-csrf />` component and its need in `<x-form>`. And it's modern. 

Precedents: https://github.com/rails/rails/pull/56350
Discord discussion: https://discord.com/channels/1236153076688359495/1236153079603269766/1449963794515361896

&nbsp;

## Previous URL

In an effort to decouple specific feature from the `Session` object, I ended up with a nice isolated `PreviousUrl` class and kept a middleware like before.

The `PreviousUrl` class depends on the session and has utilities to track the previous and intended URLs:

```php
// Saves the current URL from the request if needed
$this->previousUrl->track($request);

// Gets the previous URL, with an optional default
$this->previousUrl->get(default: '/');

// Gets the intended URL (eg. to set before a redirect to the login page), with an optional default
$this->previousUrl->getIntended(default: '/');
```

&nbsp;

## Validation errors and original values

In the same effort to decouple features from the `Session` object, I created a `FormSession` class, similar to the `PreviousUrl` class: it isolates features related to forms, their validation errors and their original values.

The `HtmlExceptionRenderer` introduced in https://github.com/tempestphp/tempest-framework/pull/1819 is still responsible for filling the proper values when validation errors occur. The components that were using the session directly now use `FormSession`.

I'm not sure about this name though. I thought of `HtmlFormState`, or just `FormState`, but none of those sound particularly good.

These changes should be transparent to users unless they manually reached for `Session::VALIDATION_ERRORS` or `$session->getErrorsFor()`, which was generally not required thanks to the view components.

```php
// get the list of failing rules for a field
$formSession->getErrorsFor('name');

// get the original value of a field before validation failure
$formSession->getOriginalValueFor('name', default: '');
```

---

- [x] Make `Session` a standalone object
  - [x] Clean up its serialization
  - [x] Remove dependency on `SessionManager`
- [x] Make `SessionManager` interface more focused
- [x] Improve performance of all session managers
  - [x] Less IO for all managers
  - [x] Don't save session on every change, only at the end of the request
- [x] Decouple "previous url" feature from 
- [x] Decouple validation errors and original values (make it a HTML form-specific feature)
- [x] Replace CSRF protection with `Sec-Fetch-*` implementation
- [ ] Support trusted origin config (in another PR later)


packages/core/src/FrameworkKernel.php
```diff
@@ -245,6 +245,7 @@ public function registerExceptionHandler(): self
 
         $handler = $this->container->get(ExceptionHandler::class);
 
+        ini_set('display_errors', 'Off'); // @mago-expect lint:no-ini-set
```
brendt, line 248: Why?

packages/http/src/Responses/Back.php
```diff
@@ -7,7 +7,7 @@
 use Tempest\Http\IsResponse;
 use Tempest\Http\Request;
 use Tempest\Http\Response;
-use Tempest\Http\Session\Session;
+use Tempest\Http\Session\PreviousUrl;
```
brendt, line 10: I really like this change!  💯

packages/http/src/Responses/Back.php
```diff
@@ -22,20 +22,14 @@ final class Back implements Response
...
         $this->status = Status::FOUND;
-        $request = get(Request::class);
-
-        $url = $request->headers['referer'] ?? $request->getSessionValue(Session::PREVIOUS_URL);
 
-        if ($url) {
-            $this->addHeader('Location', $url);
-            return;
-        }
+        $tracker = get(PreviousUrl::class);
```
brendt: Just call the variable `$previousUrl`

packages/http/src/Session/FormSession.php
```diff
@@ -0,0 +1,105 @@
...
+namespace Tempest\Http\Session;
+
+use Tempest\Container\Singleton;
+use Tempest\Validation\FailingRule;
+
+/**
+ * Manages form validation errors and original input values in the session.
+ */
+#[Singleton]
+final readonly class FormSession
```
brendt, line 14: Really like it!

packages/http/src/Session/ManageSessionLifecycleMiddleware.php
```diff
@@ -0,0 +1,32 @@
...
+use Tempest\Http\Request;
+use Tempest\Http\Response;
+use Tempest\Router\HttpMiddleware;
+use Tempest\Router\HttpMiddlewareCallable;
+
+/**
+ * This middleware is responsible for creating the session and saving it on response.
+ */
+#[Priority(Priority::FRAMEWORK - 20)]
+final readonly class ManageSessionLifecycleMiddleware implements HttpMiddleware
```
brendt: Could we rename to something simpler like `WithSessionMiddleware` or `HasSessionMiddleware` ?

packages/http/src/Session/SessionManager.php
```diff
@@ -6,19 +6,28 @@
...
-
-    public function destroy(SessionId $id): void;
-
-    public function isValid(SessionId $id): bool;
-
-    public function cleanup(): void;
+    /**
+     * Retrieves or creates a session based on its identifier.
+     */
+    public function getOrCreate(SessionId $id): Session;
```
brendt, line 12: This might be a personal preference, but I usually use `resolve` when a function gets or creates something.

packages/http/src/Session/SessionManager.php
```diff
@@ -6,19 +6,28 @@
...
+
+    /**
+     * Determines whether the session is still valid.
+     */
+    public function isValid(Session $session): bool;
+
+    /**
+     * Removes all expired sessions from the server.
+     */
+    public function deleteExpiredSessions(): void;
```
brendt, line 32: I wonder if we really need this here, or should simply have some kind of action class that deletes expired sessions?

packages/http/src/Response.php
```diff
@@ -8,41 +8,70 @@
...
+     * Adds a header to the response.
+     */
     public function addHeader(string $key, string $value): self;
 
+    /**
+     * Removes a header from the response.
+     */
     public function removeHeader(string $key): self;
 
-    public function addSession(string $name, mixed $value): self;
```
brendt, line 33: Why remove these but then keep the cookie ones? I think I rather keep them for now.

packages/router/src/RouteConfig.php
```diff
@@ -31,7 +31,7 @@ public function __construct(
         public Middleware $middleware = new Middleware(
             HandleRouteExceptionMiddleware::class,
             MatchRouteMiddleware::class,
-            SetCookieMiddleware::class,
+            SetCookieHeadersMiddleware::class,
```
brendt, line 34: How about `HasCookiesMiddleware` or `WithCookiesMiddleware` so that it's the same as the session middleware?

brendt (dismissed):

packages/core/src/FrameworkKernel.php
innocenzi: This shouldn't belong in that PR honestly, but I found out that sometimes, errors were directly displayed in the browser. This could leak stuff so it's important to disable this. I checked and Laravel does that too

packages/http/src/Responses/Back.php
innocenzi: Refactor artifact, whoops

packages/http/src/Session/ManageSessionLifecycleMiddleware.php
innocenzi: I really like the explicit name, what do you think of `ManageSessionMiddleware`?

packages/http/src/Session/SessionManager.php
innocenzi: I actually renamed that from `resolve` because I kept wondering what it did. If you really like `resolve` we can rename it back but I think it's a confusing name

packages/http/src/Session/SessionManager.php
innocenzi: We could have an action but I do believe this belongs to the manager class

packages/http/src/Response.php
innocenzi: Cookies make sense in a response class (they will be sent with the response), but session values will not—they are a backend concern

However, I concede that we can keep them as QoL

packages/router/src/RouteConfig.php
innocenzi: Maybe `ManageCookiesMiddleware`?

packages/core/src/FrameworkKernel.php
brendt: Check!

packages/http/src/Session/ManageSessionLifecycleMiddleware.php
brendt: `ManageSessionMiddleware` Good!

packages/http/src/Session/SessionManager.php
brendt: I'm fine with this as well, let's keep

packages/router/src/RouteConfig.php
brendt: Yes, perfect

packages/http/src/Session/SessionManager.php
brendt: Let's keep it for now then.

brendt:
Feel free to merge when you're done :)

PR #1831, opened 2025-12-23
feat(database): add foreign key support for sqlite dialect

### LINKED ISSUE #1830: Support foreign keys with sqlite dialect
### Description

Context: SQLite database with the following configuration:

```php
return new SQLiteConfig(
    path: root_path('database.sqlite'),
);

```

I created an entity with a `belongsTo` statement:

```php
    {
        return new CreateTableStatement('xxxxx')
            // ...
            ->belongsTo(...);
    }
```

When running migrations with `vendor/bin/tempest migrate:up`, the table is created without foreign keys.

After inspecting the tempest source code, I found out that the sqlite dialect is not compatible with this feature.
In CreateTableStatement:

```php
// Remove BelongsTo for sqlLite as it does not support those queries
->filter(fn (QueryStatement $queryStatement) => ! ($dialect === DatabaseDialect::SQLITE && $queryStatement instanceof BelongsToStatement))
```

Is it planned to add foreign key support for sqlite?
Foreign keys are implemented in sqlite (https://sqlite.org/foreignkeys.html).

### Benefits

Foreign keys are mandatory to ensure data consistency.

guvra:
Fixes #1830

This PR adds foreign key support for SQLite dialect (cf. https://sqlite.org/foreignkeys.html for specification).

## Changes

The following code now successfully creates a foreign key with sqlite:

```php
return new CreateTableStatement('books')
    ->primary()
    ->belongsTo('books.author_id', 'authors.id', OnDelete::CASCADE)
    ->text('title');
```

At first, it didn't work because in SQLite, foreign keys must be defined after table columns.

So I used the `sortByCallback` function on the statements array in order to make sure that foreign keys are always defined at the end of the create table statement.

## Limitations

The following code fails (QueryWasInvalid exception): 

```php
{
    return new AlterTableStatement('authors')
        ->add(new BelongsToStatement('books.author_id', 'authors.id', OnDelete::CASCADE));
}
```

This is perfectly normal because it is not possible to add a foreign key on an existing table in sqlite.

It would be better if tempest threw a DialectWasNotSupported exception, but I'm not sure if it's possible to implement easily, because the BelongsToStatement is wrapped inside a AlterAddColumnStatement object.

## Warning

I updated the unit tests but I didn't do a lot of real world testing.
I only tested a migration with a create statement and an alter statement.

### DIFF
--- a/packages/database/src/QueryStatements/BelongsToStatement.php
+++ b/packages/database/src/QueryStatements/BelongsToStatement.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Tempest\Database\QueryStatements;
 
 use Tempest\Database\Config\DatabaseDialect;
-use Tempest\Database\DialectWasNotSupported;
 use Tempest\Database\QueryStatement;
 
 final readonly class BelongsToStatement implements QueryStatement
@@ -44,7 +43,7 @@ final readonly class BelongsToStatement implements QueryStatement
                     'ON UPDATE ' . $this->onUpdate->value,
                 )),
             ),
-            DatabaseDialect::POSTGRESQL => new ConstraintStatement(
+            DatabaseDialect::POSTGRESQL , DatabaseDialect::SQLITE => new ConstraintStatement(
                 $constraintName,
                 new RawStatement(sprintf(
                     'FOREIGN KEY(%s) REFERENCES %s(%s) %s %s',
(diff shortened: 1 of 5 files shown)


guvra:
I'm not sure why tests are failing on the CI, am I targeting the wrong branch?

guvra:
Closing because I forgot to create a feature branch on my fork repository...
I'll create a new PR that isn't using the main branch.

PR #1832, opened 2025-12-23
feat(database): add foreign key support for sqlite dialect

### LINKED ISSUE #1830: Support foreign keys with sqlite dialect

guvra:
Fixes #1830

This PR adds foreign key support for SQLite dialect (cf. https://sqlite.org/foreignkeys.html for specification).

## Changes

The following code now successfully creates a foreign key with sqlite:

```php
return new CreateTableStatement('books')
    ->primary()
    ->belongsTo('books.author_id', 'authors.id', OnDelete::CASCADE)
    ->text('title');
```

At first, it didn't work because in SQLite, foreign keys must be defined after table columns.

So I used the `sortByCallback` function on the statements array in order to make sure that foreign keys are always defined at the end of the create table statement.

## Limitations

The following code fails (QueryWasInvalid exception): 

```php
{
    return new AlterTableStatement('authors')
        ->add(new BelongsToStatement('books.author_id', 'authors.id', OnDelete::CASCADE));
}
```

This is perfectly normal because it is not possible to add a foreign key on an existing table in sqlite.

It would be better if tempest threw a DialectWasNotSupported exception, but I'm not sure if it's possible to implement easily, because the BelongsToStatement is wrapped inside a AlterAddColumnStatement object.

## Warning

I updated the unit tests but I didn't do a lot of real world testing.
I only tested a migration with a create statement and an alter statement.

### DIFF
--- a/packages/database/src/QueryStatements/ConstraintNameStatement.php
+++ b/packages/database/src/QueryStatements/ConstraintNameStatement.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Tempest\Database\QueryStatements;
 
 use Tempest\Database\Config\DatabaseDialect;
-use Tempest\Database\DialectWasNotSupported;
 use Tempest\Database\QueryStatement;
 
 final readonly class ConstraintNameStatement implements QueryStatement
@@ -21,9 +20,6 @@ final readonly class ConstraintNameStatement implements QueryStatement
 
     public function compile(DatabaseDialect $dialect): string
     {
-        return match ($dialect) {
-            DatabaseDialect::MYSQL, DatabaseDialect::POSTGRESQL => sprintf('CONSTRAINT %s', $this->name->compile($dialect)),
-            default => throw new DialectWasNotSupported(),
-        };
+        return sprintf('CONSTRAINT %s', $this->name->compile($dialect));
     }
 }
--- a/packages/database/src/QueryStatements/BelongsToStatement.php
+++ b/packages/database/src/QueryStatements/BelongsToStatement.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Tempest\Database\QueryStatements;
 
 use Tempest\Database\Config\DatabaseDialect;
-use Tempest\Database\DialectWasNotSupported;
 use Tempest\Database\QueryStatement;
 
 final readonly class BelongsToStatement implements QueryStatement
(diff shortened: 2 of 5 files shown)


innocenzi:
Could you rebase this on `3.x` and target that branch?

guvra:
@innocenzi Rebase done 👍

brendt:
Are you sure this actually works? Because the sqlite docs don't mention this CONSTRAINT syntax? https://sqlite.org/foreignkeys.html

I'm probably misreading it, but wanna be sure.

guvra:
@brendt I don't have a dev environment currently, but if I run the following query in an online sqlite browser (e.g. sqliteonline.com), it works:

```sql
CREATE TABLE `books` (
  `id` SERIAL PRIMARY KEY, 
  `author_id` INTEGER  NOT NULL, 
  `name` VARCHAR(255) NOT NULL, 
  CONSTRAINT `fk_authors_books_author_id` FOREIGN KEY(author_id) REFERENCES authors(id) ON DELETE CASCADE ON UPDATE NO ACTION
);
```

The command `PRAGMA foreign_key_list('books')` lists the foreign key, however this key doesn't have a name.
It doesn't seem to be possible to name foreign keys in sqlite.

Ideally it would be better to not name the foreign key if the current dialect is sqlite.

brendt:
> Ideally it would be better to not name the foreign key if the current dialect is sqlite.

Yeah so I think we should drop that `CONSTRAINT fk_authors_books_author_id` part in case of SQLite instead. That can be done in the `ConstraintNameStatement` statement: just return an empty string when it's SQLite

packages/database/src/QueryStatements/ConstraintNameStatement.php
```diff
@@ -21,9 +20,6 @@ public function __construct(
 
     public function compile(DatabaseDialect $dialect): string
     {
-        return match ($dialect) {
-            DatabaseDialect::MYSQL, DatabaseDialect::POSTGRESQL => sprintf('CONSTRAINT %s', $this->name->compile($dialect)),
-            default => throw new DialectWasNotSupported(),
-        };
+        return sprintf('CONSTRAINT %s', $this->name->compile($dialect));
```
brendt, line 23: Return an empty string for SQLite

Issue #1657, opened 2025-10-22
SkipDiscovery based on closure


brendt:
In PHP 8.5, we'll be able to pass closures to attributes. This opens up the possibility to make `SkipDiscovery` a lot more powerful. 

```php
#[SkipDiscovery(static function (Container $container): bool {
    return ! $container->get(Application::class) instanceof ConsoleApplication;
})]
final class BlogPostEventHandlers
```

Or even use DI in the closure:

```php
#[SkipDiscovery(static function (Application $application): bool {
    return ! $application instanceof ConsoleApplication;
})]
final class BlogPostEventHandlers
```

PR #1837, opened 2025-12-28
feat(discovery): support closures in `#[SkipDiscovery]`

MohammadAlhallaq:
handle #1657


MohammadAlhallaq:
hey @brendt @innocenzi , just wanted to confirm if this is what we are looking for, because I was thinking of adding some tests there

innocenzi:
I think Brent wants the container to work here. As for me, just like https://github.com/tempestphp/tempest-framework/issues/1775, I don't like the idea.

There's also the performance and caching questions regarding closures in Discovery context

PR #1838, opened 2025-12-29
feat(core)!: make `Environment` its own source of truth

innocenzi:
Because injecting `Environment` is more discoverable and a better DX than injecting `AppConfig`. 

```php
public function __construct(
    private readonly Environment $environment,
) {}
```

**EDIT**: so actually, as specified in my comment below, I completely dissociated `AppConfig` from `Environment`:
- The booting process uses `Environment::fromEnv` indirectly through `AppConfig`'s constructor when resolved from the container
- Specifying the `environment` property of the `AppConfig` in userland will not behave as expected due to that
- It makes sense that `Environment` is its own source of truth in the container, it's even easier to spoof in tests

### DIFF
--- a/docs/1-essentials/06-configuration.md
+++ b/docs/1-essentials/06-configuration.md
@@ -28,30 +28,30 @@ return new PostgresConfig(
 
 The configuration object above instructs Tempest to use PostgreSQL as its database, replacing the framework's default database, SQLite.
 
-## Accessing configuration objects
+### Accessing configuration objects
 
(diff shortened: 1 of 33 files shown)


innocenzi:
Actually, I think there is an opportunity to refactor the environment outside of `AppConfig`. The kernel booting process uses the `ENVIRONMENT` variable itself indirectly anyway, because `AppConfig` gets resolved every time with null values, and its `environment` property gets initialized with `::fromEnv`, so even if users set a custom environment value in their config, that wouldn't work.

PR #1839, opened 2025-12-29
fix(discovery): handle the discovery test case corectlly

MohammadAlhallaq:
as I was working on tests for another PR, I came across a specific test case that kept asserting as true. this happens because assertNotContains() checks for value equality, not type


innocenzi:
Thanks!

PR #1850, opened 2026-01-03
fix(performance): enahnce container memory consumption

MohammadAlhallaq:


brendt:
I believe what you're trying to do here is already covered by singletons. Non-singleton dependencies shouldn't be cached in memory, that's why they are non-singleton.

Take a look at `GenericContainer:342-351`: 

```php
        if ($instance = $this->singletons[$dependencyName] ?? null) {
            if ($instance instanceof Closure) {
                $instance = $instance($this);
                $this->singletons[$className] = $instance;
            }

            $this->resolveChain()->add($class);

            return $instance;
        }
```

Singleton definitions are stored as a closure, when they are resolved for the first time, that closure is executed, and the resulting object is stored in `$this->singletons`, which is what I think you're trying to do with this PR (but then for all dependencies, which isn't correct).

Let me know if I misunderstood anything :)

PR #1852, opened 2026-01-06
perf(mapper): pre-resolve mapper classes

### LINKED ISSUE #1834: [Performance] ORM and DI performance issues due to lack of cache
### Tempest version

2.14

### PHP version

8.4

### Operating system

Linux

## TLDR

Tempest has huge performance issues related to the DI container and the database mapper/inspector.

Currently, when interacting with the container or the ORM, the data they resolve isn't kept in memory.

For example, if the method `get` of the container is called 1000 times with the same arguments, it will do the same calculations 1000 times instead of returning a cached value.

Recommended solution:

- ORM: caching the database metadata in memory (php) by default in dev mode, and in filesystem by default in production mode.
- Container: caching resolve dependencies in memory (php) at the very least, ideally in filesystem if it's technically possible.
- Being able to customize the cache adapter to use depending on the environment.

For example, Symfony uses the following caching system:

- Symfony uses a cache for Doctrine metadata and queries (documented [here](https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-orm-configuration)). The configuration allows to use any cache provider. I think it's none by default in dev mode, and filesystem by default in production mode.
- Symfony compiles a cached version of the DI container (documented [here](https://symfony.com/doc/current/components/dependency_injection/workflow.html#working-with-a-cached-container)).

## DI Component Performance

I used xdebug to profile the request:

Image

The slowest part of the code is the `get` method of the GenericContainer class.
This method is called **hundreds of thousands of time** in this reproduction scenario.

This wouldn't be an issue if the container kept resolved dependencies in memory, but it doesn't.

Conclusion: **the DI container must keep in memory (or store in cache) anything that it resolves.**
The constant recalculations have a huge impact on performance.

To confirm this, I added some cache around the `get` method (by building a hash from the function arguments), the HTTP request is now served in ~200 ms instead of 650 ms (70% improvement) with 677 entities loaded.

## ORM/Mapper Performance

The SelectModelMapper object uses the ModelInspector object to fetch information from the database.

These two objects don't keep anything in memory, so for example when mapping a collection of 677 entities, everything is recalculated 677 times.

Conclusion: **these objects (and more generally all database components) need to keep in memory (or store in cache) anything that they resolve**.

For example, I added some cache around the methods `getRelations`, `getBelongsTo`, `getHasOne`, `getAsMany` and `getTableDefinition`. The HTTP response is now served in 60ms instead of 200ms (and 650ms before the DI container optimization). This is now nearly as fast as when I do not use the ORM! (only 10 ms slower)

Unfortunately I can't provide a PR because I only did some quick and dirty hacks in order to confirm that these components were indeed the reason behind the slow performance.

### Steps to reproduce

Create two tables that depend on each other.
For example:

- product: id name, description, price, category_id
- category: id name, description

Populate the product table with hundreds of items, the category table doesn't need to have many rows.

Then use the ORM to fetch all products:

```php
public function __invoke(): Json
{
    $products = $this->productRepository->findByCategoryId(1);

    return new Json(['products' => $products]);
}
```

And the function in the repository:

```php
public function findByCategoryId(int $categoryId): array
{
    /** @var Product[] */
    return query(Product::class)
        ->select()
        ->with('categories')
        ->where('category_id = ?', $categoryId)
        ->all();
}
```

On my dev environnement, the HTTP response time is ~650 ms with 677 rows.

If i don't use the ORM (e.g. by building the SQL query manually and not using the mapper), the response time is 50 ms instead (-92%).

brendt:
Closes #1834

### DIFF
--- a/packages/mapper/src/ObjectFactory.php
+++ b/packages/mapper/src/ObjectFactory.php
@@ -6,6 +6,7 @@ namespace Tempest\Mapper;
 
 use Closure;
 use Tempest\Container\Container;
+use Tempest\Container\Singleton;
 use Tempest\Mapper\Exceptions\DataCouldNotBeMapped;
 use Tempest\Mapper\Exceptions\MapperWasMissing;
 use Tempest\Mapper\Mappers\ArrayToJsonMapper;
@@ -18,6 +19,7 @@ use Tempest\Support\Json;
 use UnitEnum;
 
 /** @template ClassType */
+#[Singleton]
 final class ObjectFactory
 {
     private mixed $from;
@@ -30,10 +32,15 @@ final class ObjectFactory
 
     private Context|UnitEnum|string|null $context = null;
 
+    /** @var \Tempest\Mapper\Mapper[] */
+    private array $mappers;
+
     public function __construct(
         private readonly MapperConfig $config,
         private readonly Container $container,
-    ) {}
+    ) {
+        $this->mappers = $this->resolveMappers();
+    }
 
     /**
      * Sets the target class for mapping operations.
@@ -105,8 +112,11 @@ final class ObjectFactory
      */
     public function in(Context|UnitEnum|string|null $context): self
     {
-        $clone = clone $this;
-        $clone->context = $context;
+        $clone = clone($this, [
+            'context' => $context,
+        ]);
+
+        $clone->mappers = $clone->resolveMappers();
 
         return $clone;
     }
@@ -316,10 +326,7 @@ final class ObjectFactory
 
         $context = MappingContext::from($this->context);
 
-        foreach ($this->config->mappers as $mapperClass) {
-            /** @var Mapper $mapper */
-            $mapper = $this->container->get($mapperClass, context: $context);
-
+        foreach ($this->mappers as $mapper) {
             if ($mapper->canMap($from, $to)) {
                 return $mapper->map($from, $to);
             }
@@ -359,5 +366,23 @@ final class ObjectFactory
 
         /** @var Mapper $mapper */
         return $mapper->map($from, $to);
+    }
+
+    /**
+     * We cache mapper instances within the factory so that we prevent mappers being resolved on every mapping call.
+     * Whenever a mapping context changes, we'll have to re-resolve the mapper classes with the new context.
+     */
+    private function resolveMappers(): array
+    {
+        $mappers = [];
+
+        $context = MappingContext::from($this->context);
+
+        foreach ($this->config->mappers as $mapperClass) {
+            /** @var Mapper $mapper */
+            $mappers[] = $this->container->get($mapperClass, context: $context);
+        }
+
+        return $mappers;
     }
 }


packages/mapper/src/ObjectFactory.php
```diff
@@ -360,4 +367,22 @@ private function mapWith(mixed $mapper, mixed $from, mixed $to): mixed
...
+     */
+    private function resolveMappers(): array
+    {
+        $mappers = [];
+
+        $context = MappingContext::from($this->context);
+
+        foreach ($this->config->mappers as $mapperClass) {
+            /** @var Mapper $mapper */
+            $mappers[] = $this->container->get($mapperClass, context: $context);
```
innocenzi, line 383: Wouldn't this cache make it so that a mapper could potentially receive a dependency with a stale state?

xHeaven: Can you give me a scenario for this?

innocenzi: Imagine a value object registered as a singleton at the start of the request. The mappers are resolved and cached. The value object is changed in the container for any reason, the next time the mapper that uses it is called, it will have the previous value?

Might be niche though, but could be a pain to debug

xHeaven: Sounds pretty nice indeed, but possible. Perhaps we could just add a `refresh()` method that'd essentially do:
```php
public function refresh(): self
{
    $this->mappers = $this->resolveMappers();

    return $this;
}
```

The above is a global refresh. We could also play with the idea of giving fresh mappers to the caller only:
```php
public function refresh(): self
{
    $clone = clone($this);
    $clone->mappers = $clone->resolveMappers();

    return $clone;
}
```

brendt: I'm not sure I completely understand the problem? 

> Wouldn't this cache make it so that a mapper could potentially receive a dependency with a stale state?

Mappers themselves aren't singletons (they can't be because they require that context variable), so whenever they are retrieved from the container, it's a fresh new object.

brendt:
Gonna merge, we can further reiterate on it if needed.

packages/mapper/src/ObjectFactory.php
xHeaven: It's more about dependencies than mappers being singletons or not.

Let's say you have this:
```php
class SomeMapper implements Mapper {
    public function __construct(private SomeSingletonService $service) {}
}
```

1. Request start, `ObjectFactory` is created, mappers are resolved and cached
2. `SomeSingletonService` gets re-registered for whatever reason mid-request
3. `SomeMapper` gets called, but it still holds the old `SomeSingletonService` instance

It's very niche and antipattern, but a possibility.

brendt: Gotcha. Yeah I don't think that we need to consider that edge case because, as you say, that's a clear antipattern. I can only imagine this happening during internal framework testing, and in that case there would be better solutions.

PR #1866, opened 2026-01-08
fix(intl): add missing dependency to ext-intl

TimWolla:
The native intl extension is required for the `Locale` and the `Number` components functions, making it a pretty fundamental requirement to be able to make use of the package's functionality, since the `Number` component is also required by the `MessageFormat` component.

### DIFF
--- a/packages/cache/composer.json
+++ b/packages/cache/composer.json
@@ -5,13 +5,13 @@
         "php": "^8.5",
         "psr/cache": "^3.0",
         "symfony/cache": "^7.3",
-        "tempest/core": "dev-main",
-        "tempest/clock": "dev-main",
-        "tempest/kv-store": "dev-main",
-        "tempest/container": "dev-main"
+        "tempest/core": "2.14.0",
+        "tempest/clock": "2.14.0",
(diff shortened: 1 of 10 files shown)


TimWolla:
I believe the test failures are a pre-existing issue. They also happen to me locally when on main:

```
[timwolla@/t/t/p/validation (main)]composer update --prefer-lowest --prefer-dist --no-interaction
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires tempest/support dev-main -> satisfiable by tempest/support[dev-main].
    - tempest/container v2.14.0 requires tempest/reflection 2.14.0 -> found tempest/reflection[v2.14.0] but it conflicts with your root composer.json require (dev-main).
    - tempest/support dev-main requires tempest/container 2.14.0 -> satisfiable by tempest/container[v2.14.0].
```

brendt:
Thanks! Yeah those failing tests only happen on 2.x, I've rebased your branch on 3.x and will merge it there (we plan on tagging 3.0 soon)

PR #1871, opened 2026-01-08
fix(database): wrap enums in quotes during raw sql serialization

brendt:


brendt:
Enum values need to be wrapped in double quotes for a query to be valid

PR #1873, opened 2026-01-08
ci: add code owners

innocenzi:


innocenzi:
@aidan-casey do you want to be asked for a review for any package?

aidan-casey:
@innocenzi - this is helpful! Thanks for putting this together.

I'd love to be _added_ to the following if Brent agrees:

- Command bus
- Event bus
- Reflection
- Router
- Http
- Discovery

I feel like http-client should get re-assigned my way entirely, but not sure if you had a different conversation with Brent.

brendt:
I agree :)

PR #1877, opened 2026-01-08
fix(database): use singleton serializer factory instead of creating a new one

brendt:


packages/database/src/GenericDatabase.php
```diff
@@ -142,10 +142,12 @@ private function resolveBindings(Query $query): array
     {
         $bindings = [];
 
+        $serializerFactory = $this->serializerFactory->in($this->context);
```
brendt, line 145: This is a small optimization so that we don't have to clone the factory for every binding

innocenzi (approved):

PR #1878, opened 2026-01-08
feat(event-bus)!: support enum events in `EventBus::listen()`

TimWolla:
Currently when attempting to listen to a enum event, such as `KernelEvent` with an ad-hoc listener, one needs to manually construct the event key that is generated internally in `GenericEventBus::resolveHandlers()`, which is fragile:

    $bus->listen(…, KernelEvent::class . '::' . KernelEvent::SHUTDOWN->name);

Support taking a UnitEnum directly, which is already supported for the `EventHandler` attribute.

### DIFF
--- a/packages/event-bus/src/EventBus.php
+++ b/packages/event-bus/src/EventBus.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Tempest\EventBus;
 
 use Closure;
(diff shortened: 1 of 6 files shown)


innocenzi (approved):

brendt:
Very nice, thanks!

PR #1885, opened 2026-01-13
fix(http-client): support unknown status codes

innocenzi:


aidan-casey (approved):
Looks good to me. I'm partial to `Status::code`, but I also get why you made the change.

PR #1896, opened 2026-01-17
refactor(generation): move php generation into its own namespace

innocenzi:
Because I have plans for TypeScript code generation.


ek9r:
👍

PR #1897, opened 2026-01-17
feat(generation): add support for TypeScript types generation

innocenzi:
This pull request adds support for generating TypeScript definitions out of PHP classes and enums.

Documentation: https://github.com/tempestphp/tempest-framework/blob/feat/typescript-codegen/docs/2-features/18-typescript.md

---

Running the `generate:typescript-types` command will create a `types.d.ts` file at the root of the project, with namespaced definitions:

```ts
/*
|----------------------------------------------------------------
| This file contains TypeScript definitions generated by Tempest.
|----------------------------------------------------------------
*/

export namespace App {
  export interface Settings {
    theme: Theme.Theme;
    sidebar_open: boolean;
    last_edited_at: string;
  }
  export interface Permission {
    name: string;
  }
  export interface User {
    settings: Settings;
    permissions: Permission[];
  }
}

export namespace App.Theme {
  export type Theme = 'dark' | 'light';
}
```

This is configurable by creating a config returning `NamespacedTypeScriptGenerationConfig` or `DirectoryTypeScriptGenerationConfig`. The latter generates a directory structure in `./types` that's importable instead of `.d.ts` files:

```
types
└── app
    ├── index.ts
    └── theme
        └── index.ts
```

To mark a class for generation, it needs to have a `#[AsType]` attribute. However, I plan to implement heuristics (eg. the class only has public readonly properties) to automatically convert classes without having to add an attribute.


NeoIsRecursive:
Would route generation a la wayfinder be something that could be included in this package?

I built a very POC command for that while ago https://github.com/NeoIsRecursive/tempest-passkey-demo/blob/main/app/Generation/GenerateTypescriptRoutes.php.

Super cool feature this!

innocenzi:
> Would route generation a la wayfinder be something that could be included in this package?

Definitely, it's something I'm planning to do as well 👍 though I don't like leaking the controller names in the front-end for some reason, and I haven't thought of an alternative.

Thanks for the PoC, I'll take a look!

NeoIsRecursive:
> > Would route generation a la wayfinder be something that could be included in this package?
> 
> Definitely, it's something I'm planning to do as well 👍 Thanks for the PoC, I'll take a look!

Awesome! Let me know if there is anything I can help with 😄

shaffe-fr:
>However, I plan to implement heuristics (eg. the class only has public readonly properties) to automatically convert classes without having to add an attribute.

One thing I like about Spatie's TypeScript transformer, and dislike about Wayfinder,  is the granular control over what gets exported. Some simple DTOs are strictly for internal use and shouldn't be exposed to the frontend. If you implement automatic conversion based on heuristics, I hope there will be a way to 'opt-out' by default via config (my personal favorite) or a specific attribute (e.g., #[NoTypescript])

NeoIsRecursive:
> If you implement automatic conversion based on heuristics, I hope there will be a way to 'opt-out' by default via config (my personal favorite) or a specific attribute (e.g., #[NoTypescript])

Would something strategy based make sense for configuring this?

HeuresticsGenerationStrategy => opt out with an attribtue
SelectedGenerationStrategy => only stuff with the attribute
NamespaceGenerationStrategy => everything in some namespace

NeoIsRecursive:
> though I don't like leaking the controller names in the front-end for some reason, and I haven't thought of an alternative.

Didn't see this part until now.

Perhaps exposing an `uri/url` function with a bunch of overloads would do the trick - though not as "clean".

Personally I am happy as long as the available routes and their params can be typed somehow :)

EDIT:

wont vite in the minification step rename "everything", so in the build output it wont be the controller names?

innocenzi:
> If you implement automatic conversion based on heuristics, I hope there will be a way to 'opt-out' by default via config (my personal favorite) or a specific attribute (e.g., #[NoTypescript])

We'll definitely have an opt-out or something like that, no worries! That being said, it's only types, so nothing would be leaked in production anyway.

> wont vite in the minification step rename "everything", so in the build output it wont be the controller names?

Good point! If we use something similar to your proof of concept, it's entirely possible indeed, since its code that should be minified. I'll definitely try this out, but that will be in another PR later 👍

brendt (approved):
Just a very solid PR. Feel free to merge when you're ready because I have nothing to add to it :)

PR #1899, opened 2026-01-19
feat(auth): update Twitch OAuth to use Helix API

iamdadmin:
composer.json included library depotwarehouse/oauth2-twitch for OAuth with Twitch.

This library uses the Twitch Kraken API, which is no longer available, as of 2022 ([source](https://dev.twitch.tv/docs/product-lifecycle/))

1. Removed this and selected the replacement TwitchHelix wrapper from ThePHPLeague vertisan/oauth2-twitch-helix
2. Added this to the installer and stubs
3. Implemented the wrapper for TwitchHelix in the auth package
4. Updated documentation
5. Updated the authinstaller test

I've run composer qa successfully, and also performed a test of the method in an app I'm writing.

PR #1900, opened 2026-01-19
feat(auth): various improvements to auth/oauth installers

### LINKED ISSUE #1888: Improve installer by adding extra OAuth fields to the model/migration stubs when OAuth is installed
### Description

Per the documentation, the following fields are exposed from an OAuth user

```php
$user = $this->oauth->fetchUser($code);

$user->id;         // The unique identifier for the user from the OAuth provider
$user->email;      // The user's email address
$user->name;       // The user's name
$user->nickname;   // The user's nickname/username
$user->avatar;     // The user's avatar URL
$user->provider;   // The OAuth provider name
$user->raw;        // Raw user data from the OAuth provider
```

When launching `./tempest install auth --oauth` and the user chooses `yes` for migrations and user model, we could supply a stub with these additional fields.

Making the question `Do you want to install OAuth? Yes/no` the first question in the installer may be appropriate also, in the case of the `--oauth` value not being specifically passed.

### Benefits

Smoother developer experience, as invariably this code would be required immediately anyway.

iamdadmin:
Closes #1888 various improvements to auth/oauth installers, as follows

- (Installer) Asks whether to install OAuth first and provides different UserModel and Migration stubs with pre-populated fields matching the OAuthController
- Adds a very basic MustBeAuthenticated middleware stub, I've tested it in my own app, it's something that can be built upon but should give a working system
- Adds a very basic LoginController stub to show the login view (view NOT PROVIDED, comment explains this), post login (database query NOT PROVIDED, comment explains this), handle logout
- Implements Tempest\Http\Session\PreviousUrl to set/get the intended URL and redirect there after auth

Creating as a draft so I can add some docs to this as well.

### DIFF
--- a/packages/auth/src/Installer/AuthenticationInstaller.php
+++ b/packages/auth/src/Installer/AuthenticationInstaller.php
@@ -20,6 +20,8 @@ if (class_exists(\Tempest\Console\ConsoleCommand::class)) {
     {
         use PublishesFiles;
 
+        public bool $installOAuth = false;
+
         private(set) string $name = 'auth';
 
         public function __construct(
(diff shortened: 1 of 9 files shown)


iamdadmin:
I've checked the docs and they didn't deep-dive into the existing stubs and system, so there's nothing to really add. The comments within the new stubs should tie everything together. I am interested in doing a separate PR later to extend the authentication, authorization and oauth docs, but I will need to get a bit further on in my project before I'm confident enough with the topic. Hope the comments and current docs are sufficient for now, sending for review!

iamdadmin:
This and #1899 got merged and made a mess, I am re-submitting both cleanly.

PR #1906, opened 2026-01-20
feat(console): add ability to prevent built-in console commands from loading

innocenzi:
This PR updates `ConsoleApplication::boot` and `ConsoleConfig` to add options to prevent the built-in console commands from being loaded, which is useful in a context where we are building an external CLI application on top of `temnpest/console`.

The loading prevention is done through `DiscoveryLocation`, which already had support for detecting if the class came from Tempest directly (by checking the namespace).

The console command discovery uses the new option in `ConsoleConfig` and this heuristic to determine whether to load a specific command.

This _may_ cause issues if we ever build an external but first-party console application which namespace starts with `Tempest`, so we might need to update how this work later. However, for now I think it is good enough.

### DIFF
--- a/packages/console/src/ConsoleApplication.php
+++ b/packages/console/src/ConsoleApplication.php
@@ -43,7 +43,7 @@ final readonly class ConsoleApplication implements Application
 
         $consoleConfig = $container->get(ConsoleConfig::class);
         $consoleConfig->name ??= $name;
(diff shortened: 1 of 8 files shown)


brendt (approved):

brendt:
Looks good :)

PR #1908, opened 2026-01-21
chore: subsplit wip — Enzo will have to change this title :)

brendt:


brendt:
Still need to fix the tests, also discovered a bug in the container that I fixed

app/Subsplit/TempestSubsplitCommand.php
```diff
@@ -0,0 +1,86 @@
...
+
+                $this->writeln(' - Pulling latest changes');
+
+                try {
+                    $git->pull();
+                } catch (GitOperationFailed) {
+                    // Remote branch does not exist yet
+                }
+
+                // Don't we have a recursive directory copy function somewhere??
```
innocenzi, line 67: We do now: https://github.com/tempestphp/tempest-framework/pull/1909

However, this is not the right approach. We need to use actual `git subtree split` (or an equivalent alternative) to keep the history and tags in sync with the main monorepo.

composer.json
```diff
@@ -202,6 +202,7 @@
     },
     "autoload-dev": {
         "psr-4": {
+            "App\\": "app/",
```
innocenzi, line 205: We definitely need to use another directory than `app` 😬 

Something like `steward/`, `release/`, `cd/`, or something more specific like that.

tempest
innocenzi, line 1: This needs to be a separate entry point, to not be confused with what's strictly related to testing the framework itself

innocenzi (changes requested):

app/Subsplit/TempestSubsplitCommand.php
brendt: > to keep the history and tags in sync with the main monorepo

Ah, I was under the impression the release script took care of keeping tags in sync, actually, but that was probably a wrong assumption

When it comes to history: I really don't care about keeping it in sync. However, tags are important indeed.

tempest
brendt: TBH I never use `./tempest` anymore for testing, that's a remnant from way back in the day. If I need to test the framework with real apps, I have a separate project for that where I symlink tempest/framework. 

So I'm fine for removing all the testing stuff, and only keep one binary for these kinds of tasks. 

I don't have a strong opinion on whether the binary itself should be renamed, but it would be a minor inconvenience for me if it did

composer.json
brendt: Agree, I used app as a placeholder. Any preferences?

app/Subsplit/TempestSubsplitCommand.php
aidan-casey: You need to care about history. This will break Packagist if you don't.

tempest
innocenzi: > So I'm fine for removing all the testing stuff, and only keep one binary for these kinds of tasks.

That'd be great—I was never really a fan of all the test-time discovery and this binary

> I don't have a strong opinion on whether the binary itself should be renamed

I would be fine not renaming it if we ended up cleaning up all the stuff around it

PR #1919, opened 2026-01-26
feat(auth): implement Twitch OAuth provider wrapper

iamdadmin:
(Replaces broken PR 1899...)

Prior to this PR, composer.json included library depotwarehouse/oauth2-twitch for OAuth with Twitch, but it did not have a wrapper.

However this library uses the Twitch Kraken API, which is no longer available, as of 2022 ([source](https://dev.twitch.tv/docs/product-lifecycle/)) - PHPLeague also need to deprecate the package, I'll let them know to remove it.

1. Removed the old provider and selected the replacement TwitchHelix wrapper from ThePHPLeague vertisan/oauth2-twitch-helix
2. Added this to the installer and stubs
3. Implemented the wrapper for TwitchHelix in the auth package
4. Updated documentation
5. Updated the authinstaller test
6. Created an Enum to provide the canonical name of the providers to keep naming conventions consistent throughout

### DIFF
--- a/packages/auth/src/OAuth/SupportedOAuthProvider.php
+++ b/packages/auth/src/OAuth/SupportedOAuthProvider.php
@@ -33,7 +33,7 @@ enum SupportedOAuthProvider: string
      *
      * @return string|null The canonical name, or null if the provider is generic.
      */
-    public function getName(): ?string
+    public function getName(): string
     {
         return match ($this) {
             self::APPLE => 'Apple',
--- a/packages/auth/src/OAuth/SupportedOAuthProvider.php
+++ b/packages/auth/src/OAuth/SupportedOAuthProvider.php
@@ -11,6 +11,7 @@ use League\OAuth2\Client\Provider\Google;
 use League\OAuth2\Client\Provider\Instagram;
 use League\OAuth2\Client\Provider\LinkedIn;
 use Stevenmaguire\OAuth2\Client\Provider\Microsoft;
+use Vertisan\OAuth2\Client\Provider\TwitchHelix;
 use Wohali\OAuth2\Client\Provider\Discord;
 
 enum SupportedOAuthProvider: string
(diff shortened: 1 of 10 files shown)


iamdadmin:
That one was an utter nightmare to chase down, I had to xdebug and step in the code. Turns out, changing the installer to handle outputting 'Twitch' instead of 'TwitchHelix' also broke the null -> 'generic' default case, so I had to put 'Generic' specifically in `SupportedOAuthProvider->getName()` for the `self::GENERIC` case.

The tests pass locally. But I can see it complaining about rate-limiting from docker in the integration test on GH actions so that's fun.

packages/auth/src/OAuth/SupportedOAuthProvider.php
```diff
@@ -25,7 +26,35 @@ enum SupportedOAuthProvider: string
...
     case MICROSOFT = Microsoft::class;
     case SLACK = Slack::class;
+    case TWITCH = TwitchHelix::class;
 
+    /**
+     * Returns the canonical name for the given OAuth provider. Required because some of the providers have mixed-case names.
+     *
+     * @return string|null The canonical name, or null if the provider is generic.
+     */
+    public function getName(): ?string
```
innocenzi:
```suggestion
    public function getName(): string
```

packages/auth/src/OAuth/SupportedOAuthProvider.php
```diff
@@ -25,7 +26,35 @@ enum SupportedOAuthProvider: string
     case LINKEDIN = LinkedIn::class;
     case MICROSOFT = Microsoft::class;
     case SLACK = Slack::class;
+    case TWITCH = TwitchHelix::class;
 
+    /**
+     * Returns the canonical name for the given OAuth provider. Required because some of the providers have mixed-case names.
+     *
+     * @return string|null The canonical name, or null if the provider is generic.
```
innocenzi:
```suggestion
```

innocenzi (changes requested):
Thanks for taking time to fix the CI 🙏 just a minor change before merging

iamdadmin:
> Thanks for taking time to fix the CI 🙏 just a minor change before merging

No worries, makes sense, so have committed both. Another test run is on it's way!

iamdadmin:
Tests all passing!

innocenzi (approved):

innocenzi:
Thanks!

PR #1922, opened 2026-01-26
chore: add governance page

brendt:
Still WIP and up for discussion


brendt:
Slept on it for another night:

- I think the 90 days inactive should be 180 days; at least for as long as Tempest is a hobby project
- I think the same time limit should also be set for both Council Members and and Core Contributors
- We need to define what it means to be "inactive". It shouldn't be just code contributions, but then what?

innocenzi:
> We need to define what it means to be "inactive". It shouldn't be just code contributions, but then what?

Tough to properly define it. I'd say "activity" should be:
- committing code
- opening/reviewing/commenting on PRs
- opening issues/commenting/triaging
- sending messages in Discord

With those criteria, it would effectively be pretty hard to be "inactive"

iamdadmin:
> > We need to define what it means to be "inactive". It shouldn't be just code contributions, but then what?
> 
> Tough to properly define it. I'd say "activity" should be:
> 
>     * committing code
> 
>     * opening/reviewing/commenting on PRs
> 
>     * opening issues/commenting/triaging
> 
>     * sending messages in Discord
> 
> 
> With those criteria, it would effectively be pretty hard to be "inactive"

It may be hard to have a single criteria; you might have a team member whose sole commitment is to triaging and progressing issues and doing code reviews. They'd not need to submit their own code in thise role. So people could be asked to commit to a level of activity based on what they are bringing to the project, and then moderately hold them to it, for 'inactivity' to then be considered on a case-by-case-basis. If there are multiple people covering the duties, then it could be a 60 or 90 days "no activity on the repo" with a removal of access, with a swift path for them to return.

I think it's important to define something, for the times you NEED to invoke it, even if overall you don't invoke it often.

aidan-casey:
@brendt - a few things...

First, the easy one. Inactivity is the failure to perform any of their responsibilities defined under the roles section. My assumption, though I feel it necessary to leave provision for other scenarios, is that most of the time this will be without any communication to us as well - total disappearance.

Second, I pretty strongly disagree with your 180 days comments. On a few fronts:

Part of the BDFL's responsibilities include signing off on releases and setting the vision for the next release cycle. Remember, the governance document is intended to ensure a healthy project. Are you really comfortable with that not happening for half a year? I'm not. I think it'd kill Tempest and is very unhealthy.

The intention, also, is that the BDFL could come back within 180 days and resume their duties. Beyond that (181 days) the officer is permanently made BDFL.

Regarding community members roles, my intention was that they _may_ be removed after 90 days. It's not an automatic removal, but suggested and ratified by the council.

This feels fair, because we don't want to introduce a situation where a member is inactive for 179 days, makes a contribution, and then goes inactive for 179 days again. That's not consistent with the privilege of this role.

brendt:
Thanks @aidan-casey for the clarification, indeed when we're talking about inactivity on all fronts, I'm fine with 90 days. I've reworded that in the document. 

What we still need before we can merge this is:

- "manner of benevolence" must be clarified: @aidan-casey did you already have something defined for this?
- "link to core values": @aidan-casey same question
- Appointing the first officer

brendt:
@innocenzi @aidan-casey I made the final adjustments. Can you take a look? We can of course always fine tune this document along the way. I'd like to get it published though so that I can include it in the 3.0 release post

aidan-casey:
It looks like you struck the date milestones in favor of votes, which I'm not big on. This creates opportunity for council members to drag their feet or attempting to cut out the BDFL too soon. I still think 90/180 day milestones is a good idea.

I think this is missing a definition for how the governance document can change, because the BDFL should not be able to simply adjust it anytime.

Given that, I would also suggest moving current team members to a separate file.

It puts us in a bind today, but I don't like having the BDFL a member of the council given what decisions may need a 2/3 majority.

It might be worth adding a section about how projects and milestones are defined.

brendt:
> I think this is missing a definition for how the governance document can change, because the BDFL should not be able to simply adjust it anytime.

Ah good point. I want to keep the names in the doc for consistency though. I actually like that making a change here would also require approval in GitHub, as it makes the voting process public.

I've added a section outlining this process.

> It looks like you struck the date milestones in favor of votes, which I'm not big on. This creates opportunity for council members to drag their feet or attempting to cut out the BDFL too soon. I still think 90/180 day milestones is a good idea.

Do you mean this section?

> The decision to mark the current BDFL as "inactive" must be approved by a 2/3 majority vote of the Standing Council

brendt:
As discussed in Discord: we're all good with it now.

In the future we'll have to explicitly approve each PR that makes changes to this document.

PR #1923, opened 2026-01-26
chore: add milestone section on the contribution page

brendt:


docs/5-extra-topics/04-contributing.md
```diff
@@ -282,3 +282,34 @@ Pull request titles and descriptions should be as explicit as possible to ease t
 Contributors are not required to respect conventional commits within pull requests, but doing so will ease the review process by removing some overhead for core contributors.
 
 All pull requests will be renamed to the conventional commit convention if necessary before being squash-merged to keep the commit history and changelog clean.
+
+## Release cycles
+
+Tempest current does not follow a fixed release cycle. In general, bug fixes and minor features can be released as soon as possible. For breaking changes, though, we aim to bundle as many as possible in a single major release. 
+
+### Milestones
```
innocenzi, line 290: I struggle with seeing "2.x" or "3.x" as a milestone. For me it's supposed to be a goal to reach, not something you track continuously.

I would suggest we use specific milestones for minor versions. eg. milestone `3.4` should contain the specific fixes and features assigned for its milestone.

That way, we have a very specific goal to work towards.

If we don't like that, I'd suggest leaving out the concept of "current minor milestone" entirely, and just work on stuff as we see fit and occasionally release patches and minors.

brendt:
@aidan-casey , @innocenzi , I've slept on the matter another night and I want to ask you to give this system a chance :) 

I know "a milestone" is technically something else, but to me it's really important to have a overview of what things can be worked on actively, which things are planned, and which things we don't have on our roadmap. 

There are undoubtably other ways of achieving this, but I find it important to work within the boundaries of one platform, and that it's clearly visible for everyone to see. 

I considered using project boards, but I like the ease of filtering on milestones in the general issue list, the fact that you have a nice progress bar, and that you can attach deadlines to milestones. Also, we don't need swimlanes, so in that regard project boards would be overkill. 

I say: we try it like this for a full major release cycle (so until 4.0 is tagged), and then we reconsider.

PR #1938, opened 2026-01-30
feat(view): extend isset to handle multiple conditions

### LINKED ISSUE #1937: Extend :isset element handling in view to support multiple vars
### Description

Would there be any interest in a PR to refactor `tempest/view` `:isset` element handling so it can handle complex expressions? The :isset element literally wraps whatever it's given in isset() expecting a true or false

`<x-template :isset="$var1"` < currently works

`<x-template :isset="$var1 || $var2"` < currently does not

`<x-template :if="isset($var1) || isset($var2)"` < what you have to do right now for the equivalent

I therefore propose to extend this so that when multiple $vars are detected, it will wrap each instance of $var and as long as it returns true or false, according to developer needs, it will issue true or false.

It should also support multiple cases, as long as the expression can be written to correctly return a true or false using `isset()` statements only.

For more complex cases, then `:if` should still be used.

### Benefits

Developer experience and efficiency in code.

iamdadmin:
Closes #1937

Updated `/packages/view/src/Elements/PhpIfElement.php`

```php
     public function compile(): string
     {
         if ($condition = $this->wrappingElement->consumeAttribute(':isset')) {
-            $condition = sprintf('isset(%s)', $condition);
+            $condition = preg_replace('/\$\w+/', 'isset($0)', $condition);
```
Updated docs and tests, and real-world tested it in an app, both in single-case and multi-case.

Expression logic is untested, because all the function does is wrap isset around a `$var`, so it remains the responsibility of a developer to produce expressions which evaluate to true or false as needed.


iamdadmin:
This should be ready for workflows to run when someone has a chance! Many thanks.

packages/view/src/Elements/PhpIfElement.php
iamdadmin, line 1: Updated the regex as it was mis-handling array keys and $object->property

tests/Integration/View/TempestViewRendererTest.php
iamdadmin, line 1: Added additional test to catch these cases also

brendt:
I like it. It's a bit out of the box, but that's exactly what Tempest is about. And this feels very intuitive to me.

brendt (approved):

brendt:
Also I appreciate the docs update 💯

Gonna merge this when CI is done

iamdadmin:
Epic thanks!

PR #1944, opened 2026-02-05
feat(testing): add `assertJsonSubset` to test response helper

Maarten-Dekker:
This PR adds a new testing helper, `assertJsonSubset()`. It provides recursive JSON subset matching, allowing tests to assert only specific nested keys without requiring full JSON equality. This is useful when responses include additional framework‑generated data and you only want to verify the fields relevant to the test.


brendt:
Great feature! Thank you :)

PR #1952, opened 2026-02-09
fix(router): windows output buffer flushing

brendt:


aidan-casey:
Hmm... I'm somewhat suspect on this. We were calling `ob_flush` in two different places? Wondering if that was our issue more than the actual call?

brendt:
Yeah I only created this PR to verify whether it solves the issue, I actually don't plan on merging it until I understand what's going on

brendt:
Update: it doesn't solve the problem: https://github.com/tempestphp/tempest-framework/actions/runs/21821632954/job/62956151577

PR #1962, opened 2026-02-12
fix(database): prevent infinite loop when creating a model with a HasOne relationship

### LINKED ISSUE #1961: Infinite loop when saving a model with a HasOne relation
### Tempest version

2.14/3.1

### PHP version

8.4

### Operating system

Linux

### Description

When creating/saving a new model along with a related model in a HasOne relationship, you get an infinite loop.

### Steps to reproduce

Using the book/isbn models from the current test suite:

```php
$book = Book::create(
    title: 'Book Title',
    isbn: Isbn::new(value: '123-456-789'),
);
```

The isbn query saves the parent book which saves the isbn which saves the book which saves the isbn which saves the book.... etc.

nunntom:
When creating a model along with a child with a HasOne relationship, you get an infinite loop:
closes #1961

- Added a failing (more like hanging) test.
- Excluded the foreign key property from the child model query data.


brendt:
Thanks! This seems to break `Tests\Tempest\Integration\Database\Builder\IsDatabaseModelTest::test_create_with_hasone_relation`, any ideas?

nunntom:
That was the test I added! Hopefully should be fixed now. Local tests were running with sqlite which doesn't care about foreign key tables not being present on table creation.

brendt:
Awesome, thanks!

PR #1967, opened 2026-02-13
refactor(console): improve command completion

xHeaven:
This PR replaces the `_complete` completion handler usage with a PHP helper (`tempest-complete`) that reads a pre-built JSON metadata file. The shell scripts for Bash and Zsh have been rewritten to delegate completion logic to this helper instead of booting the PHP runtime on every TAB press.

Shell completion previously required a full Tempest boot cycle with discovery for every TAB press, which introduced noticeable latency. By shifting the matching logic to a small binary that reads a static JSON file, completions become near-instant.

Currently, the supported platforms are macOS and Linux. Windows users should use WSL.


docs/1-essentials/04-console-commands.md
```diff
@@ -250,28 +252,46 @@ Run the install command and follow the prompts:
...
+```bash
+# Zsh: add to ~/.zshrc
+source ~/.tempest/completion/tempest.zsh
+
+# Bash: add to ~/.bashrc
+source ~/.tempest/completion/tempest.bash
 ```
 
-For Bash, source the completion file in your `~/.bashrc`:
+### Keeping completions up to date
```
brendt, line 271: I wonder if we can improve this flow? Maybe we can write the output `completion:generate` command every time a command is invoked during dev? Of course if it's not too heavy, but I reckon writing a JSON file shouldn't be noticeable?

`completion:update-bin` could be run on `composer up/install` ?

packages/console/completion-helper/src/main.rs
```diff
@@ -0,0 +1,395 @@
+use std::collections::{BTreeMap, HashSet};
```
brendt: I think this is a good approach, but before going the Rust way, I'd like to know how much time a plain PHP script + the JSON file would take? Even if it's 50-100ms, that should still be ok I reckon; and shipping PHP would be a lot less hassle.

packages/console/src/Actions/BuildCompletionMetadata.php
```diff
@@ -0,0 +1,103 @@
...
+
+        ksort($commands);
+
+        return [
+            'version' => 1,
+            'commands' => $commands,
+        ];
+    }
+
+    private static function buildFlagNotation(ConsoleArgumentDefinition $definition): string
```
brendt: Why static functions?

packages/console/src/Actions/EnsureCompletionHelperBinary.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Tempest\Console\Actions;
+
+use RuntimeException;
+use Tempest\Console\CompletionRuntime;
+use Tempest\Support\Filesystem;
+
+use function Tempest\Support\box;
+
+final readonly class EnsureCompletionHelperBinary
```
brendt: I feel like this name is missing something? What are we ensuring?

packages/console/src/Actions/EnsureCompletionHelperBinary.php
```diff
@@ -0,0 +1,59 @@
...
+        $context = stream_context_create([
+            'http' => [
+                'follow_location' => 1,
+                'max_redirects' => 10,
+                'timeout' => 30,
+                'user_agent' => 'tempest-completion-installer',
+            ],
+        ]);
+
+        [$contents, $errorMessage] = box(static fn (): false|string => file_get_contents($downloadUrl, false, $context));
```
brendt: If we continue this route, we should probably do some hash comparison to make sure the binary wasn't tempered with

packages/console/src/Commands/CompletionGenerateCommand.php
```diff
@@ -0,0 +1,47 @@
...
+        description: 'Generate shell completion metadata as JSON',
+    )]
+    public function __invoke(
+        #[ConsoleArgument(
+            description: 'Optional output path for the completion metadata JSON',
+            aliases: ['-p'],
+        )]
+        ?string $path = null,
+    ): ExitCode {
+        if (! CompletionRuntime::isSupportedPlatform()) {
```
brendt: I prefer to inject a singleton instance of CompletionRuntime over static methods

brendt (changes requested):

packages/console/src/Actions/BuildCompletionMetadata.php
xHeaven: Dumbness, fixed.

packages/console/src/Actions/EnsureCompletionHelperBinary.php
xHeaven: It ensures that the completion helper binary exists and is executable. It indeed feels a bit off. How would you feel about `ResolveCompletionHelperBinary`? Maybe `InstallCompletionHelperBinary`?

packages/console/src/Actions/EnsureCompletionHelperBinary.php
xHeaven: It may not add too much value here. The files are downloaded from the GitHub releases page, where the CI adds the builds. If the download URL changed in the helper in a malicious way, it's safe to believe that the hash also changed. Thoughts?

packages/console/src/Actions/EnsureCompletionHelperBinary.php
brendt: I think `ResolveCompletionHelperBinary` is good :)

packages/console/src/Actions/EnsureCompletionHelperBinary.php
brendt: I think people might be weary of running a binary downloaded from anywhere, even if it's from a GH release. 

Let's continue discussing after we've determined whether plain PHP is a viable approach or not

packages/console/completion-helper/src/main.rs
xHeaven: Replaced with a PHP approach.

packages/console/src/Actions/EnsureCompletionHelperBinary.php
xHeaven: I've removed this entirely with a new PHP approach.

packages/console/src/Actions/EnsureCompletionHelperBinary.php
xHeaven: I've implemented the PHP approach, not relevant anymore.

docs/1-essentials/04-console-commands.md
xHeaven: So `completion:update-bin` is gone with the PHP version.

However, the `completion:generate` on every command run is still up for debate. It's around ~250ms for one run.
Technically, I can hook up a middleware so it runs on every command run, I'm not sure if it's a bit excessive, though.

brendt: That sounds excessive indeed, but having to remember to do it manually feels like going back to the dark ages like with Laravel's IDE helper. 

Would there be a way we could detect whether it needs updating? Technically it would be possible to keep track of it during discovery (assuming we can write to a file of some sorts), but that would make discovery a bit slower then… 

One thing we could do for sure is to run the command after calling `make:command`, but not everyone will use it… 

I'd also run it automatically as a composer hook post install and update, at least it's something

xHeaven: I believe we can't truly automate this part, as the autocompletion needs the new command in the file as well, which can only happen in 2 ways: either you run `completion:generate` or you run any other command, and that generates the metadata for you.

We could also keep track of whether we need an update or not. We could fingerprint the builds. However, that also requires rebuilding the entire thing each time in memory, at which point not writing it out to a file seems to be... a bit pointless?

I was thinking of using Tempest's in-built `_complete` command for completion, but that would be _really_ slow (250+ ms), it would make the autocompletion feel very wonky. That's probably the only way where we wouldn't need any file at all, even.

brendt: Let's stick with the JSON file for now :) I do think running it with composer hooks will have merit (but actually that should be changed in tempest/app, not on the framework level)

brendt (approved):

aidan-casey (approved):

PR #1969, opened 2026-02-13
fix(view): add props for width and height to x-icon

### LINKED ISSUE #1955: View component x-icon defaults to 1em x 1em and doesn't include method to override this
### Tempest version

3.x

### PHP version

8.5

### Operating system

Linux

### Description

Reference: https://iconify.design/docs/usage/svg/no-code/

By default the Iconify API returns a SVG with `height` and `width` attributes set to `1em`.

The `tempest/icon` class doesn't currently expose a prop to override this.

The `tempest/view` `x-icon` component does add a `$class` prop, but does not override the `width` and `height` attributes.

I've found this to be incompatible with tailwind height and width utilities in the class string, instead defaulting to the width and height attributes. That could be a precedence or browser rendering issue, but nonetheless highlights the default behaviour is fixed and not developer-controllable.

### Steps to reproduce

Use `<x-icon` to render any icon and you will see the returned HTML includes the `width` and `height` attributes

### Possible fixes

1. Add `?height=none` to the download url in `icon.config.php` and allow the `<x-icon class` attribute defined by the developer to control height and width, with perhaps a default of 1em applied here for easy control.
2. Add the `width` and `height` props to `tempest/icon` class which in turn is filled from `<x-icon height="" width=""` attributes. More extensive changes to `tempest/icon` and possibly not proportionate given the simplicity of option 1 and little added benefit.

iamdadmin:
Closes #1955 

- Add props for width and height to x-icon, and ability to disable it entirely so it can be set in style or class

- Add prop for style

- Update docs

- Update tests


iamdadmin:
Weird, I thought that checks were set for manual trigger, I know they are going to fail as it stands as the tests aren't finished, hence leaving it in draft. Sorry, had I known they'd run I wouldn't have raised it as a draft until I had clean tests!

iamdadmin:
Ready for review, I'd suggest squashing this commit as prettier which wasn't supposed to run on markdown files, ran on markdown files and really messed up the formatting, so there's some regression commits that are over the top.

docs/1-essentials/02-views.md
```diff
@@ -554,10 +554,58 @@ Includes the CSRF token in a form
...
+    />
+</svg>
+```
+
+This behaviour can be overridden in a couple of ways to suit your needs.
+
+Firstly, you can disable the width and height attributes entirely, by setting the `noSize` prop, and then provide dimensions using Tailwind w- and -h utilities, or your own custom class.
+
+```html
+<x-icon name="material-symbols:php" noSize class="w-[24px] h-[24px] text-indigo-400" />
```
brendt: Should be `no-size`

docs/1-essentials/02-views.md
```diff
@@ -554,10 +554,58 @@ Includes the CSRF token in a form
...
+
+<svg class="w-[24px] h-[24px] text-indigo-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
brendt, line 594: This is the generated output, right? Add a comment so that that's clear:

docs/1-essentials/02-views.md
brendt: Is there are reason we'd ever want the default without `no-size`? We already control the icon component, right? Couldn't we always use `no-size` underneath, and then just set the default size in the icon component, which could then be overwritten?

brendt (changes requested):
Some thoughts I'd like some input on

docs/1-essentials/02-views.md
iamdadmin: ### TL;DR

We potentially always want a default of 1em x 1em as we also ship with Tailwind in the app template, and it's required for compatibility with Tailwind font-size scaling. Anyone using Tailwind will likely never use `no-size`. Removing the default entirely forces a dev to choose each time instead of relying on a default which isn't necessarily bad but obviously requires a re-factor of existing. I implemented `no-size` as a shorter and declarative alternative to `width="" height=""`.

### Longer:

1. With iconify specifically, they default to providing 1em x 1em as per https://iconify.design/docs/icon-components/svg-framework/dimensions.html so that the icon scales with font-size. Keeping the default is consistent with the dev-ux anyone already familiar with Iconify will have had previously.

2. With a default, we're requiring less code in the view templates

`<x-icon name="material-symbols:php" class="size-8" />` can be called in the view template

and you will get

`<svg class="size-8" width="1em" height="1em">...</svg>`

Without a default size-8 wouldn't do anything, and you'd have to put the height and width in the view template

`<x-icon name="material-symbols:php" class="size-8" width="1em" height="1em" />`

3. If we set the default in class, then we need a merge utility in the framework to handle it.

4. if we set the default in width and height attributes then we need to pick a default we like, and given Iconify sets 1em x 1em and tailwind scales with font-size then that's a sensible default.

5. Tailwind font-size utility classes (text-xs, text-sm, etc) scale fonts and is the approach of Nuxt-UI for icon scaling for example. Other frameworks may vary of course, but then we have the option to override the base.

### Alternative and possibly better than what I've already come up with

We can have an optional `themedefaults.config.php` with various default values, starting with default icon width & height. When present `x-icon` can fill width and height with this default, and also still have the optional props. This would default not-rendered so still be breaking, but enables a default for Tailwind users without growing each individual usage of `x-icon` to add mandatory props.

We can have a Publish command to publish the default file and obviously give it in docs for manual creation, for anyone not using Vite specifically. The vite installer in tempest/vite can call this command when the 'Tailwind' option is chosen.

I'm sure that in time other defaults within this config file will also make sense, and this approach is borrowed from the nuxt / nuxt-ui appConfig approach for tweaking defaults throughout the view templates.

### Post your decision

I'm happy with either, and can benefit only from your direction here. I just opted for non-breaking in my suggestion as a starting point since I had to pick one way of doing it :)

Also the `no-size` (as it would be after rename you requested) prop is arbitrary, since that's clearer to be programmatically truthy versus relying on typos and preferences between `none|None|NONE|no|No|false|False|FALSE` although it could be implemented by checking if `width="" height=""` are simply non-values.

docs/1-essentials/02-views.md
iamdadmin: Will fix that up with the commit to fix the rest!

docs/1-essentials/02-views.md
brendt: I guess what I meant was that, say you have this today:

`<x-icon name="material-symbols:php" class="size-8" />`

Can't we make it so that this still works, even though we're using `no-size` underneath? By adding `width="1em" height="1em"` as fallbacks when nothing is set?

iamdadmin: I've gotten really turned around on this, so I did some testing.

This is what I found:

1. If you rely on Tailwind to scale your icons with font-size i.e. `text-2xl`, then `width="1em" height="1em"` is required to make the font scale with it.

2. If you're using Tailwind `size` i.e. `size-8` then it scales with the `--spacing` var which doesn't require `width="1em" height="1em"`.

3. If you use the Tailwind `height` and `width` `w-` and `h-` utilities i.e. `h-[64px] w-[64px]` they also don't require `width="1em" height="1em"`.

4. Outside of Tailwind you can do it whichever way you want as long as it produces valid CSS of course.

With that in mind I recommend:

1. No default height and width
2. Unless no style, class, or width/height props are set at all, in which case default to 1em

This would mean if no-one does anything it'll still render a sane default, but as soon as someone starts passing props it's over to them to be declarative.

brendt: That sounds reasonable, right? So that means we can also ditch the no-size attribute, right?

iamdadmin: Commits to follow!

iamdadmin: Resolved by removing as no longer required.

docs/1-essentials/02-views.md
iamdadmin: Resolved by adjusting docs to include a comment explaining the second line is the result.

docs/1-essentials/02-views.md
iamdadmin: Resolved by refactor.

iamdadmin:
Since there was a pile of regression commits due to some formatting errors, I rebased, and re-applied the changes to make this a cleaner commit.

Ready for review again, thanks for the feedback!

packages/view/src/Components/x-icon.view.php
```diff
@@ -1,37 +1,63 @@
...
+        replace: '',
+    )
+    ->when(
+        $style ?? null,
+        fn (ImmutableString $s): ImmutableString => $s->replace(
+            search: '<svg',
+            replace: "<svg style=\"{$style}\"",
+        ),
+    )
+    ->when(
```
brendt, line 37: What if both class and style attributes are provided?

brendt (changes requested):

packages/view/src/Components/x-icon.view.php
iamdadmin: If my understanding is correct, `->when` should match multiple cases in order; I intended that if both are provided, both will be written into the resulting component, pushing class on at the start.

brendt: Ah yes, of course. That was a brainfart on my end :p

brendt (approved):

brendt:
Can you rebase and resolve the conflicts? Then it's good to merge :)

iamdadmin:
Awesome, rebased and conflicts resolved :)

brendt:
Thank you!

PR #1971, opened 2026-02-15
refactor: re-enable phpstan project-wide

xHeaven:
# Changes that might be breaking

## API changes

- `Tempest\Core\Insight`
  - Removed constants: `Insight::ERROR`, `Insight::SUCCESS`, `Insight::NORMAL`, `Insight::WARNING`.
  - Constructor now takes `Tempest\Core\InsightType` instead of a string constant.
- `Tempest\Router\Exceptions\ConvertsToResponse`
  - Now extends `\Throwable`.
  - Non-throwable implementations are no longer valid.
- `Tempest\Auth\Authentication\DatabaseAuthenticatableResolver`
  - Unused constructor with `Database` dependency was removed.
  - `new DatabaseAuthenticatableResolver($database)` no longer works.
- `Tempest\Console\Exceptions\UnknownArgumentsException`
  - Unused constructor parameter `consoleCommand` was removed.
- `Tempest\Http\Session\Managers\FileSessionManager`
  - Constructor dependency changed from `SessionConfig` to `FileSessionConfig`.
- `Tempest\Reflection\EnumReflector`
  - Constructor changed from `string|object` to `string|UnitEnum|self|ReflectionEnum`.
  - Passing arbitrary objects is no longer supported.
- `Tempest\View\Elements\PhpDataElement`
  - Class removed as it was entirely dead code.
- `Tempest\View\ViewCachePool::getItem()`
  - Return type changed from `Symfony\Component\Cache\CacheItem` to `Psr\Cache\CacheItemInterface`.

## Positional argument changes

- `Tempest\KeyValue\Redis\Config\RedisConfig`
  - Added missing `$persistentId` before `$timeout`/`$options` parameters. If you are using named parameters, you are unaffected.
- `Tempest\Log\Config\SlackLogConfig`
  - Added missing `$channels` before `$prefix` and `$tag`. If you are using named parameters, you are unaffected.

## Behavioral changes

- `packages/mail/src/mail.config.php`
  - Invalid `MAIL_SMTP_SCHEME` now throws `InvalidArgumentException` instead of silently defaulting.
- `Tempest\DateTime\DateTime::fromParts()` and `Tempest\DateTime\create_intl_calendar_from_date_time()`
  - Out-of-range years now consistently throw `OverflowException`.
- `Tempest\Storage\Config\AzureStorageConfig::createAdapter()`
  - Azure client wiring changed from `BlobRestProxy` to `BlobServiceClient`.


xHeaven:
Okay, so the PR title is failing because I'm incapable of creating good PR titles. Enzo will hopefully save me from eternal pain regarding this.
The coding convention is failing due to a regression in PHPStan which was introduced in `2.1.34` - I've already managed to send a fix in there, should be fine by `2.1.40`.
~~The datetime tests... I have no idea.~~

iamdadmin:
For the title the check wants you to use a tag of one of the following

[build, chore, ci, docs, feat, fix, perf, refactor, revert, style, test]

Given phpstan is being "put back" it's probably overall a chore, opposed to "introduced for the first time" which might make it a refactor.

/me awaits enzo-bot :D

brendt:
Ok we can merge all of these in 3.x without any issues. I consider all of these bug fixes for oversights, not proper breaking changes.

Also: for the config files, our documentation always shows the named argument approach, so those are fine as well.

packages/mapper/src/SerializerFactory.php
```diff
@@ -113,20 +108,16 @@ public function forValue(mixed $value): ?Serializer
...
         }
 
         return null;
     }
 
     /**
-     * @param Closure|class-string<Serializer|ConfigurableSerializer> $serializerClass
+     * @param class-string<Serializer|ConfigurableSerializer> $serializerClass
      */
-    private function resolveSerializer(string $serializerClass, PropertyReflector|TypeReflector|string $input): ?Serializer
```
brendt, line 129: I don't think we can be sure we'll always get a serializer, can we?

packages/view/src/Components/x-component.view.php
```diff
@@ -22,7 +22,9 @@
 </%s>
 HTML, $is, $attributeString, $content, $is);
 
-$html = get(TempestViewRenderer::class)->render(view($template, ...$this->data));
```
brendt, line 25: Why this change?

brendt (changes requested):
@innocenzi can you double-check the pluralizer stuff?

packages/view/src/Components/x-component.view.php
xHeaven: PHPStan is screaming about `$this->data`, because it's accessed magically and it's impossible to properly type it.

`$this` in this context would be `TempestViewRenderer`, but then `->data` resolves to a `?View::data` using magic.

The updated approach doesn't break anything (or if it does, then we need more tests!) and PHPStan seems to like it, too.

brendt: `$_data` is indeed a valid alternative, but I don't understand why `$scopedVariables` should be included, these are a private property on the view component element, never directly exposed to the view (if I remember correctly)

xHeaven: As discussed on Discord, it was a previously existing bug. Added regression tests in [`492c1de`](https://github.com/tempestphp/tempest-framework/pull/1971/commits/492c1de3550a31652844baaf7a501aaef98d6f87).

packages/mapper/src/SerializerFactory.php
xHeaven: This doesn't change behavior, the previously existing checks for `null` were dead code. You either get a serializer, or a `TypeError`.

packages/database/src/Connection/Connection.php
```diff
@@ -5,9 +5,14 @@
 namespace Tempest\Database\Connection;
 
 use PDOStatement;
+use Tempest\Database\Config\DatabaseConfig;
 
 interface Connection
 {
+    public DatabaseConfig $config {
```
brendt: hat was the reason for adding this one?

packages/database/src/IsDatabaseModel.php
```diff
@@ -21,76 +22,84 @@ trait IsDatabaseModel
...
+    protected static function queryBuilder(): QueryBuilder
+    {
+        return query(static::class);
+    }
+
     /**
      * Returns a builder for selecting records using this model's table.
      *
-     * @return SelectQueryBuilder<self>
+     * @return SelectQueryBuilder<static>
```
brendt, line 36: Static + traits tend to give a lot of problems in PhpStorm, I rather keep self

brendt (changes requested):

packages/database/src/IsDatabaseModel.php
xHeaven: Keeping `self` breaks the type inference entirely. What kind of issues are you referring to? I can try to reproduce and see if there's any problem. Calling/chaining these methods doesn't break PHPStorm from what I can see.

packages/database/src/Connection/Connection.php
xHeaven: https://github.com/tempestphp/tempest-framework/blob/06302c7cfd0ad28613e54b01fc01e6949f1ddd4f/packages/database/src/GenericDatabase.php#L26-L26

https://github.com/tempestphp/tempest-framework/blob/06302c7cfd0ad28613e54b01fc01e6949f1ddd4f/packages/database/src/GenericDatabase.php#L30-L30

`GenericDatabase` reads this property.

packages/database/src/IsDatabaseModel.php
brendt: Then it should be fine, actually. Maybe it was already fixed in PhpStorm actually :)

packages/database/src/Connection/Connection.php
brendt: I suggest we do it the other way for now: add a docblock in `GenericDatabase` to indicate that `$connection` is `PDOConnection`. I know that's not entirely correct, but this change is a bit too breaking to my liking

xHeaven: Fixed in [`a6aa54a`](https://github.com/tempestphp/tempest-framework/pull/1971/commits/a6aa54a547ddb72819b42da7cebae2ad5acc9e4f), left a TODO for when we can break things.

brendt (approved):

PR #1972, opened 2026-02-15
feat(database): use different text type declaration with mysql according to field length

### LINKED ISSUE #1941: Text types in database schema
### Description

Hello dear Tempest team,

I think the database system is very good. However, I have noticed one thing that can lead to problems, especially in connection with MySQL/MariaDB.

SQLite and PostgreSQL are not affected by this. The data type ‘TEXT’ is so large in SQLite and PostgreSQL that it is unlimited. In MySQL, however, there are several types of TEXT, and the version selected by the dialect is limited to 65,535 characters. At first glance, this number may seem large, but in a CMS it can quickly become too small if markup language is used.

It would therefore be worth considering whether to change the dialect to MEDIUMTEXT or LONGTEXT. This would make it comparable with PostgreSQL and SQLite.

DevPandi

### Benefits

The advantage for developers would be that the database behaviour would be aligned at this point and SQLite, PostgreSQL and MySQL/MariaDB would really ‘behave the same’ - to a large extent at least.

SQLite supports approx. 1 GB, PostgreSQL has no limit for text. MySQL is the exception here. This could result in different behaviour for products that are also offered for others.

MrYamous:
Fixes #1941 

While working on it, I realized this may can be applied to `Integer` too


packages/database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -143,12 +143,13 @@ public function foreignKey(string $local, string $foreign, OnDelete $onDelete =
     /**
      * Adds a `TEXT` column to the table.
      */
-    public function text(string $name, bool $nullable = false, ?string $default = null): self
+    public function text(string $name, bool $nullable = false, ?string $default = null, ?int $length = null): self
```
brendt: I like that you can just specify a number here, and it will determine the correct type for you. However, I think I'd rather see it a bit more flexible:

1. instead of constants on MySQLConfig, let's refactor to an enum (not sure about the name yet)

```php
enum TextLength: int
{
    case TINY = 255;
    case DEFAULT = 65535;
    case MEDIUM = 16777215;
    case LONG = 4294967295;
}
```

2. Then users could provide either an int or this enum: `int|TextLength|null $length = null`
3. We could make the field non-optional for clarity: `int|TextLength $length = TextLength::TEXT`
4. Improve the enum so that all logic is contained in one place:

```php
enum TextLength: int
{
    case TINY = 255;
    case DEFAULT = 65535;
    case MEDIUM = 16777215;
    case LONG = 4294967295;

    public static function fromLength(int $length): self
    {
        return match (true) {
            $length > TextLength::LONG => TextLength::LONG,
            $length > TextLength::MEDIUM => TextLength::MEDIUM,
            $length > TextLength::DEFAULT => TextLength::DEFAULT,
            default => TextLength::TINY,
        };
    }

    public function toString(): string
    {
        return match($this) {
            TextLength::TINY => 'TINYTEXT',
            TextLength::DEFAULT => 'TEXT',
            TextLength::MEDIUM => 'MEDIUMTEXT',
            TextLength::LONG => 'LONGTEXT',
        };
    }
}
```

WDYT?

I'm only a bit concerned about names, because as far as I understand, this is only necessary for MySQL — right?

brendt (changes requested):

packages/database/src/QueryStatements/CreateTableStatement.php
DevPandi: In SQLite, the limit of TEXT can be set in  SQlite config, standard are 1 GB (1.000.000.000 Bytes): https://sqlite.org/limits.html

In PostgreSQL TEXT is without limit. https://www.postgresql.org/docs/current/datatype-character.html

From your three supported database types, only MySQL has TINYTEXT, TEXT, MEDIUMTEXT, LONGTEXT. https://dev.mysql.com/doc/refman/8.4/en/blob.html

MySQL is some times little bit frustating with this.

MrYamous: I see your point and looks good to me, it's probably more friendly to use if we allow an Enum here.
But I think `TextLength` is too generic, WDYT about `DatabaseTextLength`, `FieldTextLength` or `MySQLTextLength`  ?

brendt: `DatabaseTextLength` sounds good 👍

MrYamous: Great, i'll finish this PR asap 🙂

brendt:
Perfect, ping me when it's done :)

MrYamous:
> Perfect, ping me when it's done :)

It's updated @brendt 
Do I need to squash and/or rebase ?

brendt (approved):

brendt:
Can you check the failing tests?

packages/cryptography/tests/Password/PasswordHasherTest.php
```diff
@@ -29,7 +29,7 @@ public function test_config_options(): void
 
         $this->assertSame(
             ['cost' => 10],
-            (new BcryptConfig(cost: 10))->options,
+            new BcryptConfig(cost: 10)->options,
```
MrYamous, line 32: Is that normal if this change is in this PR @brendt ?

MrYamous:
> Can you check the failing tests?

Fixed 👍

brendt:
Thanks! Appreciate it :)

PR #1978, opened 2026-02-16
feat(database): add integer type sizes

DevPandi:
Extends #1972

[SQLite](https://www.sqlite.org/datatype3.html): "INTEGER. The value is a signed integer, stored in 0, 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value."

[PostgreSQL](https://www.postgresql.org/docs/current/datatype-numeric.html) supports SMALLINT, INTEGER and BIGINT.

[MySQL](https://dev.mysql.com/doc/refman/8.4/en/integer-types.html) supporty TINY*, SMALL*, MEDIUM*, INTEGER and BITINT.

For simplicity's sake: SMALLINT, INTEGER, BIGINT. Less code, less complexity. 

I also picked up on the ideas from the suggestion for text types.


packages/database/tests/QueryStatements/CreateTableStatementTest.php
```diff
@@ -18,6 +18,10 @@
  */
 final class CreateTableStatementTest extends TestCase
 {
+    public function test_integer_table_database_dialect(): void
+    {
+
+    }
```
MrYamous: This test is empty, is that intentional ?

DevPandi: No, it was more due to my exhaustion that I didn't reset the file and only noticed it yesterday evening, so to speak. I sat down in the evening and just did it to get back into development even better. Sorry about that. But I won't really get around to it before Sunday because I have a lot of other things to do. :)

DevPandi: and removed.

packages/database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -199,12 +199,13 @@ public function char(string $name, bool $nullable = false, ?string $default = nu
     /**
      * Adds an `INTEGER` column to the table.
      */
-    public function integer(string $name, bool $unsigned = false, bool $nullable = false, ?int $default = null): self
+    public function integer(string $name, bool $unsigned = false, bool $nullable = false, int $bytes = 4, ?int $default = null): self
```
brendt: Can we do it with a union type? `int|IntergerBytes $size`?

packages/database/src/QueryStatements/IntegerBytes.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Database\QueryStatements;
+
+enum IntegerBytes: int
```
brendt: To be consistent, with the other PR, let's call it `DatabaseIntegerSize`

brendt (changes requested):

packages/database/src/QueryStatements/CreateTableStatement.php
DevPandi: Good idea, done with next commit.

packages/database/src/QueryStatements/IntegerBytes.php
DevPandi: Good idea, done with next commit.

brendt (approved):

brendt:
For future reference: run `composer qa` locally before pushing, that way code style is properly fixed. I did it for you just now :)

brendt:
Awesome!

PR #1980, opened 2026-02-17
refactor(view)!: improved view component rendering

### LINKED ISSUE #1194: Improved lexer context
If possible, it would be nice to store the current file and cursor position in every token, so that whenever we throw token or parser related exceptions, we can attach the file, line number, and column position

### LINKED ISSUE #1380: Colliding class names in view imports
Let's say one view imports `App\Book` and another view imports `App\Models\Book`. Both imports will be combined in the final rendered view, and will then collide. The solution is for the user to import one or both as aliases, but there should also be a clear error explaining them what to do

brendt:
This PR fixes some long-standing issues with tempest/view. Technically it does not introduce any breaking changes, but it might lead to failed template rendering in some occasions. First, an overview of what's changing:

- Compiled templates will now use PHP's `include` instead of compiling everything to one file. This fixes #1380 when two view components import classes with the same names from different namespaces
- Namespaces can still collide, because they are passed from the parent views or view component to child view components. We provide a clear error in this case
- We've added source mapping, so that now we can refer to view source files with line number when a view compilation error occurs, fixes #1194

## Breaking changes

With this PR comes a refactor to the `Element` interface, which now requires a `getImports` method. However, a default implementation is provided by the `IsElement` trait, and thus should have no impact.

Apart from that, this PR may still lead to invalid view files with missing imports.

Let's say you have these two components:

```php
// x-a.view.php

<?php 
use App\HomeController::class;
?>

{{ uri(HomeController::class) }}
```

```php
// x-b.view.php

<?php 
use App\BlogController::class; 
use function Tempest\Router\uri;
?>

{{ uri(HomeController::class) }}
```

When these two view components were referenced in the same file, they would render fine before this PR:

```html
<x-a />
<x-b />
```

That's because tempest views used to be compiled into one big file, and all import statements would be grouped together. However, now that we've switched to including compiled view components, the above code would break, since there's no `use function Tempest\Router\uri;` within the scope of `<x-a>`

The fix is to include `uri` in `<x-a>` as well:

```php
// x-a.view.php

<?php 
use App\HomeController::class;
use function Tempest\Router\uri;
?>

{{ uri(HomeController::class) }}
```

Thanks to source-mapping, you will now get a clear exception pointing you to the source of the problem, making fixing these issues simple.


packages/view/src/Elements/ViewComponentElement.php
```diff
@@ -162,7 +164,16 @@ public function compile(): string
             },
         );
 
-        return $this->compiler->compile($compiled->toString());
+        $compiled = $this->compiler->compile($compiled->toString());
+
+        return $compiled;
```
brendt: Comment out this line to switch to the new `include` strategy

packages/view/src/Elements/ElementFactory.php
```diff
@@ -42,15 +42,7 @@ public function withIsHtml(bool $isHtml): self
...
-    public function make(Token $token): ?Element
-    {
-        return $this->makeElement(
-            token: $token,
-            parent: null,
-        );
-    }
-
-    private function makeElement(Token $token, ?Element $parent): ?Element
+    public function make(Token $token, Element $parent): ?Element
```
brendt, line 45: `ElementFactory` now always requires a parent/root element

packages/view/src/Elements/ElementFactory.php
```diff
@@ -123,23 +116,15 @@ private function makeElement(Token $token, ?Element $parent): ?Element
             );
         }
 
-        $children = [];
+        $element->setParent($parent);
```
brendt, line 119: Refactored parent/child setting because not all elements have children, but all have a parent.

packages/view/src/Elements/IsElement.php
```diff
@@ -121,6 +121,8 @@ public function setParent(?Element $parent): self
     {
         $this->parent = $parent;
 
+        $this->parent->setChildren([...$this->parent->getChildren(), $this]);
```
brendt, line 124: See before: not all elements have children but all have a parent

packages/view/src/Elements/PhpElement.php
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types=1);
+
+namespace Tempest\View\Elements;
+
+use Tempest\View\Element;
+use Tempest\View\HasImports;
+use Tempest\View\Parser\Token;
+use Tempest\View\WithToken;
+
+final class PhpElement implements Element, WithToken, HasImports
```
brendt: Represents a block of PHP code, currently only used to extract imports

packages/view/src/Elements/PhpElement.php
```diff
@@ -0,0 +1,32 @@
...
+    ) {}
+
+    public function compile(): string
+    {
+        return $this->content;
+    }
+
+    public function getImports(): array
+    {
+        preg_match_all('/^\s*use .*;/m', $this->content, $matches);
```
brendt, line 27: Currently we're still using a regex, we could improve if needed once this refactor works

packages/view/src/Elements/RootElement.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Tempest\View\Elements;
+
+use Tempest\View\Attributes\PhpAttribute;
+use Tempest\View\Element;
+use Tempest\View\HasImports;
+
+final class RootElement implements Element, HasImports
```
brendt: This new RootElement is used as… the root.

It ensures a `ViewComponentElement` has a parent where it can gather imports from

packages/view/src/Elements/RootElement.php
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        return implode($compiled);
+    }
+
+    public function getImports(): array
+    {
+        $imports = [];
+
+        foreach ($this->children as $child) {
```
brendt, line 31: We only want to resolve imports from `PhpElement`s, not from sibling `ViewComponent` elements. That's why we check `instanceof PhpElement` here

packages/view/src/Elements/ViewComponentElement.php
```diff
@@ -281,4 +280,15 @@ private function exportAttributesArray(): string
 
         return sprintf('new \%s([%s])', ImmutableArray::class, implode(', ', $entries));
     }
+
+    public function getImports(): array
```
brendt, line 282: View components can be nested, and we need to recursively resolve all imports.

Writing this reminds me: we also need to check for sibling `PhpElements` in the same file still. That's still TODO

packages/view/src/Parser/TempestViewCompiler.php
```diff
@@ -66,13 +67,13 @@ public function compileWithSourceMap(string|View $view, ?string $sourcePath = nu
         $ast = $this->parseAst($template, $sourcePath);
 
         // 4. Map to elements
-        $elements = $this->mapToElements($ast);
+        $rootElement = $this->mapToElements($ast);
```
brendt, line 72: Refactoring to accomodate the new `RootElement`

brendt:
Incredible feature! Thanks Mark!

xHeaven:
> Incredible feature! Thanks Mark!

Thank you too, it was an amazing collab with you!

PR #1988, opened 2026-02-18
refactor(auth): remove insecure `patrickbussmann/oauth2-apple` suggestion

brendt:


brendt:
This PR doesn't break anything, it only removes the suggestion of using an insecure package for Apple OAuth, instead the generic provider is suggested.

brendt:
I made an issue for it here: https://github.com/patrickbussmann/oauth2-apple/issues/63

For now though, we completely remove the suggestion. People can still use it at their own risk, but Tempest won't suggest it anymore.

brendt:
Update: we can't remove the optional dependency in a non-breaking way :( For now we'll ignore the audit, but we should come up with a better solution. Let's await https://github.com/patrickbussmann/oauth2-apple/issues/63 for a while, if nothing happens, we could fork it.

PR #1993, opened 2026-02-19
fix(cache): clear the default cache by name

brendt:


brendt:
When passing `default` to the cache clear command, it wasn't used as a filter, because the default config isn't a tagged config, and thus did not have `#default` in its name.

PR #1996, opened 2026-02-20
feat(core): support frankenphp worker mode

aazsamir:
Initial support for frankenphp worker mode. 
I wanted it to be as opaque for framework users as possible, plug it in and everything (well, as most as possible) should just work. Main glue code is done by `OpaqueRequest` and `OpaqueSession` - their state will be distinct for every request, even if `Request` or `Session` were injected as class dependencies.

# Changes
- `Resetable` and `ResetableStatic` interfaces, for clearing state after request
- rename `Session` as `GenericSession`, add `Session` interface
- `OpaqueRequest` / `OpaqueSession` - request/session classes that may be injected as class dependencies, but their content will be different for every request in worker mode due to their internal state mangling
- `WorkerApplication` - http application with `frankenphp_handle_request` support
- `WorkerRouter` - almost the same as `GenericRouter`, however it sets `OpaqueRequest` in container, and matches route before any middleware is ran. ~~Maybe it can be merged with `GenericRouter`, not sure for now~~
- session serialization - as `OpaqueRequest` cannot be unserialized, I had to do a few changes in session serialization
- ~~Change middleware priorities to prevent session creation unless necessary~~

# How it works

1. We run `WorkerApplication`, boot framework and wait in loop for request. During boot, we register `OpaqueSession` as a `Session` lazy proxy - framework is highly coupled with `Session` being present at any time.
2. Frankenphp resets superglobals like `$_GET` and calls `frankenphp_handle_request` handle on request
3. We build `Request` instance, and pass it to `WorkerRouter`
4. `WorkerRouter` sets `Request` into `RequestHolder` singleton and sets `OpaqueRequest` as a `Request` in container
5. `WorkerRouter` matches route and all the middleware stack begins
6. When some service depends on `Request`, `OpaqueRequest` will be injected. It is a proxy that will get current request from `RequestHolder` underneath. A bit like `RequestStack` in symfony.
7. When some service depends on `Session`, `OpaqueSession` will be injected. Session will be resolved at call time. This process also depends on `Request`.
8. When response is returned, `ResetHandler` is called and every service implementing `Resetable` will reset its internal state. Along with `RequestHolder` unsetting its request it holds.

# Limitations

This approach seems to work, however it wouldn't work with async servers like Swoole. For that, request/session and related shouldn't be used as class dependencies. Probably should be passed with something like `$requestContext`, but it needs **a lot** of work and breaking changes.

# Future work

- I believe session lifecycle should be refactored, but I feel that this Pull Request is already too big, and decided to revert any changes related to that. It should be done in new PR.
- `WorkerRouter` matches route at the beginning of request lifecycle, it breaks a few assumptions about it being matched later. I don't think it is a must-have to resolve it now, and this can be done in new PR.
- I'm not sure that there should be two distinct routers - `WorkerRouter` and `GenericRouter`, especially that they are almost the same. However, it's a bit simpler that way and maybe it will be easier to "stabilize" worker mode if they will be kept separate.
- `DeferredTasks` and `KernelEvent::SHUTDOWN` - they are currently being ran on worker exit, it probably needs a bit more thought put into how it should behave, I don't think it is a must-have for first take.

# Testing

For testing/debugging, I'm using such dockerfile
```Dockerfile
FROM dunglas/frankenphp:1.11.2-php8.5
RUN install-php-extensions intl
```
then `docker build -t tempest-worker .` and
```bash
docker run \                                                                                                                                                                                                                                                                                
    -e FRANKENPHP_CONFIG="worker ./public/index.php" \
    -v $PWD:/app \
    -p 8000:80 -p 4430:443 \
    tempest-worker:latest
```

Also, built such repository for testing performance https://github.com/aazsamir/tempest-worker-test/

I will write some unit tests after first round of code review if that's okay, I suspect that still a lot of things can change.

### DIFF
--- a/packages/core/src/GenericResetHandler.php
+++ b/packages/core/src/GenericResetHandler.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Core;
+
+use Tempest\Container\Container;
+
+final readonly class GenericResetHandler implements ResetHandler
+{
+    public function __construct(
+        private ResetableContainer $resetableContainer,
+    ) {}
+
+    public function reset(Container $container): void
+    {
+        foreach ($this->resetableContainer->resetableClasses as $class) {
+            if (! $container->has($class)) {
+                continue;
+            }
+
+            $container->get($class)->reset();
+        }
+
+        foreach ($this->resetableContainer->resetableStaticClasses as $class) {
+            $class::resetStatic();
+        }
+    }
+}
--- a/packages/core/src/ResetableStatic.php
+++ b/packages/core/src/ResetableStatic.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Core;
+
+interface ResetableStatic
+{
+    public static function resetStatic(): void;
+}
--- a/packages/http/src/RequestHolder.php
+++ b/packages/http/src/RequestHolder.php
@@ -5,9 +5,10 @@ declare(strict_types=1);
 namespace Tempest\Http;
 
 use Tempest\Container\Singleton;
+use Tempest\Core\Resetable;
 
 #[Singleton]
-final class RequestHolder
+final class RequestHolder implements Resetable
 {
     private(set) Request $request;
 
@@ -16,7 +17,7 @@ final class RequestHolder
         $this->request = $request;
     }
 
-    public function clear(): void
+    public function reset(): void
     {
         unset($this->request);
     }
--- a/packages/http/src/RequestHolder.php
+++ b/packages/http/src/RequestHolder.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http;
+
+use Tempest\Container\Singleton;
+
+#[Singleton]
+final class RequestHolder
+{
+    private(set) Request $request;
+
+    public function setRequest(Request $request): void
+    {
+        $this->request = $request;
+    }
+
+    public function clear(): void
+    {
+        unset($this->request);
+    }
+}
--- a/packages/http/src/Session/GenericSession.php
+++ b/packages/http/src/Session/GenericSession.php
@@ -9,7 +9,7 @@ use Tempest\DateTime\DateTimeInterface;
 use Tempest\Support\Str;
 use UnitEnum;
 
-class GenericSession implements Session
+final class GenericSession implements Session
 {
     private array $expiredKeys = [];
 
--- a/packages/http/src/Session/GenericSession.php
+++ b/packages/http/src/Session/GenericSession.php
@@ -1,0 +1,118 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Http\Session;
+
+use Tempest\DateTime\DateTime;
+use Tempest\DateTime\DateTimeInterface;
+use Tempest\Support\Str;
+use UnitEnum;
+
+class GenericSession implements Session
+{
+    private array $expiredKeys = [];
+
+    public function __construct(
+        private(set) SessionId $id,
+        private(set) DateTimeInterface $createdAt,
+        public DateTimeInterface $lastActiveAt,
+        /** @var array<array-key,mixed> */
+        private(set) array $data = [],
+    ) {}
+
+    public static function unserialize(array $data): self
+    {
+        $session = new self(
+            id: new SessionId($data['id']),
+            createdAt: DateTime::fromTimestamp($data['created_at']),
+            lastActiveAt: DateTime::fromTimestamp($data['last_active_at']),
+            data: $data['data'],
+        );
+
+        $session->expiredKeys = $data['expired_keys'];
+
+        return $session;
+    }
+
+    public function set(string|UnitEnum $key, mixed $value): void
+    {
+        $this->data[Str\parse($key)] = $value;
+    }
+
+    public function flash(string|UnitEnum $key, mixed $value): void
+    {
+        $this->data[Str\parse($key)] = new FlashValue($value);
+    }
+
+    public function reflash(): void
+    {
+        foreach ($this->data as $key => $value) {
+            if (! $value instanceof FlashValue) {
+                continue;
+            }
+
+            unset($this->expiredKeys[$key]);
+        }
+    }
+
+    public function get(string|UnitEnum $key, mixed $default = null): mixed
+    {
+        $key = Str\parse($key);
+        $value = $this->data[$key] ?? $default;
+
+        if ($value instanceof FlashValue) {
+            $this->expiredKeys[$key] = $key;
+            $value = $value->value;
+        }
+
+        return $value;
+    }
+
+    public function consume(string|UnitEnum $key, mixed $default = null): mixed
+    {
+        $key = Str\parse($key);
+        $value = $this->get($key, $default);
+
+        $this->remove($key);
+
+        return $value;
+    }
+
+    public function all(): array
+    {
+        return $this->data;
+    }
+
+    public function remove(string|UnitEnum $key): void
+    {
+        $key = Str\parse($key);
+
+        if (isset($this->data[$key])) {
+            unset($this->data[$key]);
+        }
+    }
+
+    public function cleanup(): void
+    {
+        foreach ($this->expiredKeys as $key) {
+            $this->remove($key);
+        }
+    }
+
+    public function clear(): void
+    {
+        $this->data = [];
+    }
+
+    public function serialize(): array
+    {
+        return [
+            'id' => (string) $this->id,
+            'created_at' => $this->createdAt->getTimestamp()->getSeconds(),
+            'last_active_at' => $this->lastActiveAt->getTimestamp()->getSeconds(),
+            'data' => $this->data,
+            'expired_keys' => $this->expiredKeys,
+        ];
+    }
+}
--- a/packages/http/src/Session/Managers/FileSessionManager.php
+++ b/packages/http/src/Session/Managers/FileSessionManager.php
@@ -6,6 +6,7 @@ namespace Tempest\Http\Session\Managers;
 
 use Tempest\Clock\Clock;
 use Tempest\Http\Session\Config\FileSessionConfig;
+use Tempest\Http\Session\GenericSession;
 use Tempest\Http\Session\Session;
 use Tempest\Http\Session\SessionCreated;
 use Tempest\Http\Session\SessionDeleted;
@@ -30,7 +31,7 @@ final readonly class FileSessionManager implements SessionManager
         $session = $this->load($id);
 
         if ($session === null) {
-            $session = new Session(
+            $session = new GenericSession(
                 id: $id,
                 createdAt: $now,
                 lastActiveAt: $now,
@@ -48,7 +49,7 @@ final readonly class FileSessionManager implements SessionManager
 
         Filesystem\write_file(
             filename: $this->getPath($session->id),
-            content: serialize($session),
+            content: serialize($session->serialize()),
             flags: LOCK_EX,
         );
     }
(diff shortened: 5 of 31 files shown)


aidan-casey:
Nice! Thanks for taking this on.

public/index.php
```diff
@@ -4,9 +4,21 @@
 
 use Tempest\Discovery\DiscoveryLocation;
 use Tempest\Router\HttpApplication;
+use Tempest\Router\WorkerApplication;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
+if (function_exists('frankenphp_handle_request')) {
```
xHeaven: This is not a good idea, see https://github.com/tempestphp/tempest-framework/pull/1792#discussion_r2614363129

src/Tempest/Framework/Installers/index.php
```diff
@@ -3,7 +3,17 @@
 declare(strict_types=1);
 
 use Tempest\Router\HttpApplication;
+use Tempest\Router\WorkerApplication;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
+if (function_exists('frankenphp_handle_request')) {
```
xHeaven: This is not a good idea, see https://github.com/tempestphp/tempest-framework/pull/1792#discussion_r2614363129

xHeaven (changes requested):

src/Tempest/Framework/Installers/index.php
aazsamir: yes, I fully agree. I guess in the final form of this PR I will just remove it, because as far as I understand from reading discord discussion, preparing tempest dockerfiles is another beast to deal with, and probably shouldn't even be handled in this PR. I made that draft-PR to mark/communicate that I'm willing to try work on worker mode support.

xHeaven: Sounds great, we're looking forward to your contribution! In this case, [this comment](https://github.com/tempestphp/tempest-framework/issues/1834#issuecomment-3719257223) might also be helpful, this is one of the known limitations.

aazsamir:
Okay, I'm building a benchmark test suite with different servers
- `server` - frankenphp normal php server
- `worker` - frankenphp using worker mode
- `local` - php local server (`php -S 0.0.0.0:8000 -t public`)
- `fpm` - nginx + fpm stack

Tests:
- `simple` - call simple endpoint returning OK HTTP 200, 1000 times sequentially
- `parallel` - call the same endpoint, but 1000 times in parallel using CURL MULTIEXEC
- `users` - call endpoint that fetches 1000 records from sqlite database, 1000 times sequentially
- `users parallel` - call the same endpoint, but 1000 times in parallel using CURL MULTIEXEC
- `echo` - call endpoint that echoes back the request body, 1000 times sequentially
- `echo parallel` - call the same endpoint, but 1000 times in parallel using CURL MULTIEXEC

Results:
```
Test: simple
  Host: worker (http://localhost:8022), RPS: 1055.1671012983, Performance: 100.00%
  Host: local (http://localhost:8023), RPS: 96.86567787388, Performance: 9.18%
  Host: fpm (http://localhost:8024), RPS: 92.100936451952, Performance: 8.73%
  Host: server (http://localhost:8021), RPS: 88.357682038286, Performance: 8.37%
Test: parallel
  Host: worker (http://localhost:8022), RPS: 4427.0892045539, Performance: 100.00%
  Host: fpm (http://localhost:8024), RPS: 512.60546514658, Performance: 11.58%
  Host: server (http://localhost:8021), RPS: 442.3487077214, Performance: 9.99%
  Host: local (http://localhost:8023), RPS: 100.63471332618, Performance: 2.27%
Test: users
  Host: worker (http://localhost:8022), RPS: 47.536233040406, Performance: 100.00%
  Host: local (http://localhost:8023), RPS: 36.347367183746, Performance: 76.46%
  Host: fpm (http://localhost:8024), RPS: 34.702199761434, Performance: 73.00%
  Host: server (http://localhost:8021), RPS: 31.737028577357, Performance: 66.76%
Test: users parallel
  Host: worker (http://localhost:8022), RPS: 234.80153710525, Performance: 100.00%
  Host: fpm (http://localhost:8024), RPS: 201.1042189988, Performance: 85.65%
  Host: server (http://localhost:8021), RPS: 169.46323360818, Performance: 72.17%
  Host: local (http://localhost:8023), RPS: 37.036573614678, Performance: 15.77%
Test: echo
  Host: worker (http://localhost:8022), RPS: 1015.9844876178, Performance: 100.00%
  Host: local (http://localhost:8023), RPS: 97.714257811262, Performance: 9.62%
  Host: fpm (http://localhost:8024), RPS: 93.914587716182, Performance: 9.24%
  Host: server (http://localhost:8021), RPS: 87.424431756847, Performance: 8.60%
Test: echo parallel
  Host: worker (http://localhost:8022), RPS: 4352.4999636799, Performance: 100.00%
  Host: fpm (http://localhost:8024), RPS: 507.39995364293, Performance: 11.66%
  Host: server (http://localhost:8021), RPS: 469.95287142354, Performance: 10.80%
  Host: local (http://localhost:8023), RPS: 99.925056874416, Performance: 2.30%
```

benchmark code is available here https://github.com/aazsamir/tempest-worker-test

All the tests are stateless, as it seems for me that there is some buggy behaviour about sessions. The more sessions there is, the slower framework is (I think that discovery mechanism tries to scan all the session files). Due to that, I had to change priority of a few middlewares, otherwise they weren't skipped (they were ran before route ignoring them was matched).

I am still not marking PR as ready for review, however if someone wants to tell me that the way I'm doing it is stupid, I'm open to hear it!

brendt:
Sorry for leaving this hanging so long, I'm diving into this PR this week!

brendt:
> I believe session lifecycle should be refactored, but I feel that this Pull Request is already too big, and decided to revert any changes related to that. It should be done in new PR.

Agree that it should be another PR — if needed

> WorkerRouter matches route at the beginning of request lifecycle, it breaks a few assumptions about it being matched later. I don't think it is a must-have to resolve it now, and this can be done in new PR.

What's the reason for having to change this order?

> I'm not sure that there should be two distinct routers - WorkerRouter and GenericRouter, especially that they are almost the same. However, it's a bit simpler that way and maybe it will be easier to "stabilize" worker mode if they will be kept separate.

For now it's best to keep them separate IMO, it's indeed easier to finetune worker mode support

> DeferredTasks and KernelEvent::SHUTDOWN - they are currently being ran on worker exit, it probably needs a bit more thought put into how it should behave, I don't think it is a must-have for first take.

Agree

packages/core/src/GenericResetHandler.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+namespace Tempest\Core;
+
+use Tempest\Container\Container;
+
+final readonly class GenericResetHandler implements ResetHandler
+{
+    public function __construct(
+        private ResetableContainer $resetableContainer,
```
brendt, line 12: I find `ResetableContainer` a bit of a confusing name, because I assumed it was a new container implementation, which it is not. Maybe something like `ResetableConfig` to keep in line with the rest of the framework would work better?

brendt:
Reading through the PR, I want to propose two alternative approaches.

1. `ResetWith` attribute:

```php
#[ResetWith(CookieManagerResetter::class)]
final class CookieManager
{

}

final class CookieManagerResetter implements Resetter
{
    public function reset(Container $container): void
    {
        CookieManager::$staticStuff = [];
        $container->get(CookieManager::class)->stuff = [];
    }
}
```

The benefits of this approach:

- Resettable dependencies themselves don't have to implement new interfaces with one or two method to reset them
- There are dedicated classes for resetting dependencies
- These reset classes can themselves decide what needs to be reset statically and non-statically, making the implementation a lot easier

2. No attribute at all

Writing out option 1, I wondered: why do we need coupling between resetters and their resettable dependencies? With discovery we can build a list of all classes implementing `Resetter`, and that's all we'd need. The attribute isn't required at all.


```php
final class CookieManagerResetter implements Resetter
{
    public function reset(Container $container): void
    {
        CookieManager::$staticStuff = [];
        $container->get(CookieManager::class)->stuff = [];
    }
}

final class ModelInspectorResetter implements Resetter
{
    public function reset(Container $container): void
    {
        ModelInspector::$inspectors = [];
        $container->get(ModelInspector::class)->resetMemoization();
    }
}

// …
```

Am I missing something with option 2?

packages/http/src/Session/Managers/FileSessionManager.php
```diff
@@ -48,7 +49,7 @@ public function save(Session $session): void
 
         Filesystem\write_file(
             filename: $this->getPath($session->id),
-            content: serialize($session),
+            content: serialize($session->serialize()),
```
brendt, line 52: Why was this change needed?

packages/http/src/Session/GenericSession.php
```diff
@@ -0,0 +1,118 @@
...
+            $this->remove($key);
+        }
+    }
+
+    public function clear(): void
+    {
+        $this->data = [];
+    }
+
+    public function serialize(): array
```
brendt, line 108: I prefer this method to be called `toArray` instead. `serialize` means the result should be a string, which this is not. This is making an array representation which can be used for serialization

packages/http/src/Session/OpaqueSessionInitializer.php
```diff
@@ -0,0 +1,26 @@
...
+use Tempest\Container\Initializer;
+use Tempest\Http\Session\Resolvers\CookieSessionIdResolver;
+
+final readonly class OpaqueSessionInitializer implements Initializer
+{
+    public function initialize(Container $container): OpaqueSession
+    {
+        $sessionIdResolverReflector = new ReflectionClass(CookieSessionIdResolver::class);
+        // sessionIdResolver depends on Request, which is not available at the time of initialization, so we create a lazy proxy for it
+        // TODO: maybe it should take request as a parameter, not a dependency
```
brendt, line 18: Usually that's indeed a good reason to move this dependency to the call-site

packages/http/src/RequestHolder.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Http;
+
+use Tempest\Container\Singleton;
+use Tempest\Core\Resetable;
+
+#[Singleton]
+final class RequestHolder implements Resetable
```
brendt, line 11: What's the reason why we can't reset request and session directly and need these opaque objects in between?

packages/mapper/src/CasterFactory.php
```diff
@@ -27,6 +28,11 @@ public function __construct(
         private readonly Container $container,
     ) {}
 
+    public function reset(): void
```
brendt, line 31: We could probably move the resetMemoization stuff to `HasMemoization`

brendt (changes requested):

brendt:
> All the tests are stateless, as it seems for me that there is some buggy behaviour about sessions. The more sessions there is, the slower framework is (I think that discovery mechanism tries to scan all the session files). Due to that, I had to change priority of a few middlewares, otherwise they weren't skipped (they were ran before route ignoring them was matched).

Could be, we can look into that

packages/http/src/Session/Managers/FileSessionManager.php
innocenzi: Because the session object no longer has a `__serialize`. But I disagree with this change, it should still have it (see other comment)

packages/http/src/Session/GenericSession.php
innocenzi: I think `toArray` would be super confusing. It would look like a method that returns all the key/values of the session.

I would suggest not having this method on the `Session` interface at all, it's an internal concern that shouldn't be in the public API.

I think that the session object should still have `__serialize`, as before.

packages/core/src/ResetableStatic.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Core;
+
+interface ResetableStatic
```
innocenzi, line 7: I'm wondering why we need a separate class for resetting static stuff?

packages/http/src/Session/GenericSession.php
brendt: Got it, makes sense

brendt:
FYI I'm exploring some related ideas in this PR: https://github.com/tempestphp/tempest-framework/pull/2049

I think it's likely that we'll make incremental changes with several smaller PRs before adding FrankenPHP support. I'll keep this PR open for now though because @aazsamir has done a lot of useful work and we'll probably end up cherry-picking parts of this PR.

PR #1997, opened 2026-02-20
feat(idempotency): add idempotency package

xHeaven:
## Overview

This PR adds idempotency support for HTTP routes and command bus commands. It prevents duplicate side effects by storing the result of the first execution and replaying it for subsequent requests with the same idempotency key.

Payment processing, order creation, resource provisioning and any operation where retrying the same request should not produce duplicate side effects. Timeouts, retries by clients, and accidental double clicks all cause the same problem: the server can't distinguish a retry from a new request.

### Routes

Add `#[Idempotent]` to a controller method. Clients send an `Idempotency-Key` header. The first request executes normally and caches the response. Subsequent requests with the same key replay the cached response.

Example:
```php
use Tempest\Idempotency\Attributes\Idempotent;

final readonly class OrderController
{
    #[Post('/orders')]
    #[Idempotent]
    public function create(CreateOrderRequest $request): Response
    {
        // Only executes once per idempotency key.
        // Retries replay the original response.
        $order = $this->orderService->create($request);

        return new GenericResponse(
            status: Status::CREATED,
            body: ['id' => $order->id],
        );
    }
}
```

### Commands
Mark a command with `#[IdempotentCommand]`. Duplicate dispatches with the same payload are silently skipped.

Example:
```php
use Tempest\Idempotency\Attributes\IdempotentCommand;

#[IdempotentCommand]
final readonly class ImportInvoicesCommand
{
    public function __construct(
        public string $vendorId,
        public string $month,
    ) {}
}
```

Commands can also provide explicit keys via `HasIdempotencyKey`.

Example:
```php
use Tempest\Idempotency\Attributes\IdempotentCommand;
use Tempest\Idempotency\Contracts\HasIdempotencyKey;

#[IdempotentCommand]
final readonly class ProcessPaymentCommand implements HasIdempotencyKey
{
    public function __construct(
        public string $paymentId,
        public int $amount,
    ) {}

    public function getIdempotencyKey(): string
    {
        return $this->paymentId;
    }
}
```

### Known limitations at the opening of PR
- [x] ~~No differentiation between in-progress/same key, different payload cases~~
- [x] ~~Scoping is weak, we need an additional component (user-provided resolver)~~
- [x] ~~Heartbeat renewal is not implemented yet~~
- [x] ~~We need to target only `POST` and `PATCH` methods, disallow others~~
- [x] Possibly other things I didn't think of

### Wonftixes (design decision)
- Stored responses must be JsonSerializable (more of a design decision than an issue)
- Windows incompatibility due to `posix` and `pcntl` usage


aidan-casey:
Pretty slick, man!

xHeaven:
Should be fine for review, can't seem to add/change much.

packages/idempotency/src/Attributes/IdempotentCommand.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Idempotency\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class IdempotentCommand
```
brendt: I dislike that we have both `Idempotent` and `IdempotentCommand`. Would there be a way to have only one attribute?

packages/idempotency/src/Exceptions/UnsupportedIdempotencyPlatform.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Idempotency\Exceptions;
+
+use RuntimeException;
+
+final class UnsupportedIdempotencyPlatform extends RuntimeException
```
brendt: We have different naming conventions for exceptions: https://tempestphp.com/3.x/extra-topics/contributing#exception-classes

packages/idempotency/src/Attributes/Idempotent.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Tempest\Idempotency\Attributes;
+
+use Attribute;
+use Tempest\Idempotency\Middleware\IdempotencyMiddleware;
+use Tempest\Router\Route;
+use Tempest\Router\RouteDecorator;
+
+#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS)]
+final readonly class Idempotent implements RouteDecorator
```
brendt, line 13: Just wanted to say: I love route decorators :p

packages/idempotency/src/Middleware/IdempotentCommandMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+use Tempest\Idempotency\Store\IdempotencyStore;
+use Tempest\Idempotency\Support\HeartbeatRenewer;
+use Tempest\Idempotency\Support\IdempotencyKeyResolver;
+use Tempest\Idempotency\Support\ProcessingOwner;
+use Tempest\Idempotency\Support\ProcessingOwnerLiveness;
+use Tempest\Reflection\ClassReflector;
+use Throwable;
+
+#[Priority(Priority::FRAMEWORK - 1)]
+final readonly class IdempotentCommandMiddleware implements CommandBusMiddleware
```
brendt, line 30: Why is this one discovered, but the Http version not?

brendt (changes requested):
Don't forget about adding the docs :) My suggestion would be to have a section both in the routing and command bus docs

Also: would there be any impact if we worked on FrankenPHP worker mode? Somethings we already need to account for?

packages/idempotency/src/Support/HeartbeatRenewer.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Tempest\Idempotency\Support;
+
+use Tempest\Idempotency\Store\IdempotencyStore;
+use Throwable;
+
+use const SIG_DFL;
+use const SIGALRM;
+
+final class HeartbeatRenewer
```
brendt, line 16: Can you add a comment here explaining what this class does?

brendt (changes requested):

xHeaven:
@brendt 

> Don't forget about adding the docs :) My suggestion would be to have a section both in the routing and command bus docs

Definitely, just wanted to gather some feedback first and iterate with things, to make sure there is no stale documentation anywhere when things get into their final form.

> Also: would there be any impact if we worked on FrankenPHP worker mode? Somethings we already need to account for?

Yes, we'd potentially have to revisit `HeartbeatRenewer` due to `pcntl_signal` and `pcntl_alarm` usage. Not sure how I'd handle this in worker mode just yet.

packages/idempotency/src/Attributes/IdempotentCommand.php
xHeaven: I agree, the only awkward bit is that `Idempotent` currently has HTTP-specific options, like `header` and `requireKey`, which don't make too much sense for commands.
Also, minor thing, but we'd lose a bit of strictness as command attributes can only be used on classes right now, whereas the HTTP attribute can be used on classes and methods too. Not a blocker, though.

packages/idempotency/src/Exceptions/UnsupportedIdempotencyPlatform.php
xHeaven: What do you think about this?
- `UnsupportedIdempotencyMethod` -> `IdempotencyMethodWasNotSupported`
- `UnsupportedIdempotencyPlatform` -> `IdempotencyPlatformWasNotSupported`

Also, would you prefer to keep extending `RuntimeException` or would you rather extend `Exception`?

packages/idempotency/src/Attributes/Idempotent.php
xHeaven: Hard agree!

packages/idempotency/src/Attributes/IdempotentCommand.php
brendt: Just thinking out loud here: what if only the command bus had idempotency? If you'd want an idempotent HTTP request, you could make a command for it. IDK if that's better, just a thought.

packages/idempotency/src/Middleware/IdempotentCommandMiddleware.php
xHeaven: The command bus only has a global middleware pipeline, so `IdempotentCommandMiddleware` needs to be discovered globally and then no-op for commands without the attribute.

HTTP already supports route-specific middleware, so `#[Idempotent]` adds `IdempotencyMiddleware` directly to the matching route. If the HTTP middleware were discovered as well, it would end up in the global HTTP stack and run for every request, which is not what we want.

packages/idempotency/src/Support/HeartbeatRenewer.php
xHeaven: Sure thing, will do.

packages/idempotency/src/Attributes/Idempotent.php
brendt: Outside the scope of this PR, but we could have the same for all other middleware busses (CLI, commandbus and eventbus)

packages/idempotency/src/Middleware/IdempotentCommandMiddleware.php
brendt: Gotcha, it's something we should improve in the future, but outside the scope of this PR

packages/idempotency/src/Exceptions/UnsupportedIdempotencyPlatform.php
brendt: Perfect, and yes, extending `Exception` is good enough. Also add an `IdempotencyException` which all specific exceptions extend from (and then `Exception`), so you could catch all idempotency-related exceptions.

packages/idempotency/src/Attributes/IdempotentCommand.php
xHeaven: That way, the end user would have to create a new command class for each idempotent endpoint, extracting the key header manually, dispatching the command, and handling the response.
We'd also lose IETF compliance. Correct status codes, `idempotency-replayed` header, `Idempotency-Key` reading would be gone, and the users would have to implement these semantics themselves.
Scoping would also be a bit broken as commands scope by class name, HTTP scopes by user identity.
Snapshotting/replaying would be gone on HTTP, too, as it's an HTTP concept, commands are fire-and-forget concepts.

While I also don't necessarily like having different attributes for essentially the same(ish) concept, I think removing HTTP idempotency would add too much friction for too little benefit.

brendt: Good points, we need to keep it :p

I don't think we'll find a better solution, so let's keep the two attributes for now

packages/idempotency/src/Contracts/HasIdempotencyKey.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Idempotency\Contracts;
```
brendt: Also one important thing: we never use a `Contracts` namespace. In general, all interfaces should go in the root namespace of the package, so `Tempest\Idempotency\HasIdempotencyKey`

docs/2-features/19-idempotency.md
```diff
@@ -0,0 +1,272 @@
...
+    #[Idempotent(ttlInSeconds: 172_800, requireKey: true)]
+    public function charge(ChargeRequest $request): Response
+    {
+        // Cached response persists for 48 hours instead of the default 24
+    }
+}
+```
+
+| Parameter | Type | Description |
+|---|---|---|
```
brendt, line 107: I want to give it one more try to make `Idempotent` and `IdempotentCommand` more streamlined. What if:

- We have an `#[Idempotent]` attribute that takes, `ttlInSeconds` and `pendingTtlInSeconds`. This one can be used for both controller actions and commands
- We also have an `#[IdempotentRoute]` with `requireKey` and `header` as paramaters. It can be used to add extra data specifically for routes:

```php
#[Post('/payments'), Idempotent, IdempotentRoute(requireKey: '…', header: '…')]
public function charge(ChargeRequest $request): Response
{
    // Cached response persists for 48 hours instead of the default 24
}
```

WDYT?

brendt (changes requested):

brendt (approved):

PR #2002, opened 2026-02-20
perf(discovery): improve discovery performance

brendt:


xHeaven:
Managed to shave off ~25-30% from discovery, see the updated benchmark above.

brendt:
Beautiful 😁

Issue #1984, opened 2026-02-18
Create docker-compose.yml


aidan-casey:
We should create a docker-compose.yml that supports the default FrankenPHP image and adds a database + node watcher.

Issue #352, opened 2024-08-29
`tempest deploy` command


brendt:
As the name suggests, a standardized way to directly deploy a tempest app via SSH

innocenzi:
Maybe worth trying to add a recipe to deployer.org?

brendt:
Could be, I'm open to ideas :)

aidan-casey:
I'm not convinced we want to take this on ourselves pre-alpha. We'd either need prep scripts that get run on the target system, or an orchestration syntax in deployment files. Neither of those are a problem in the long run, but a little distracting in the short term.

brendt:
Good point, I'm fine removing it from the milestone

aidan-casey:
I've been thinking about this a bit and just coming back to say it would be a crime not to call this `tempest ship` when the time comes. 🙃

brendt:
That name deserves the perfect storm label.

aidan-casey:
Just noting down some thoughts I had on this...

Similar to how we are talking about "transports" for the message bus, it may be nice to have "deployers" that can integrate with various external platforms. Ideally, a deployment wizard hands you off the deployer you specify on the initial run and remembers this for future runs.

PR #2004, opened 2026-02-22
feat(ship): first-party official docker images and supporting utilities

### LINKED ISSUE #1983: Create FrankenPHP Dockerfile

iamdadmin:
Closes #1983 

# Introducing tempestphp/ship

## Initial draft PR includes

- Commands to support publishing files etc
- Custom Caddyfile based on FrankenPHP official with extra settings
- Custom distroless and rootless Dockerfile based on GoogleContainerTools Debian Trixie, with a 'debug' version containing BusyBox
- Version pinning (i.e. frankenphp release-to-php release "1.11.2-8.5.3") as well as "latest" images

## Things to be checked

- composer.json probably isn't going to be wired correctly for the new module

## Future tasks

- Clarify CI/CD pipeline (may just be a Docs task)
- Tests
- Are any additional ENVs a good idea?
- Where will the images be pushed to? Dockerhub is obvious option but ghcr and others, of course.

## Future roadmap

- When worker mode is ready, we'll need to iterate


iamdadmin:
- CADDY_ prefix missing on some ENV in Caddyfile
- Redundant COPY statement in Dockerfile removed
- Added initial .dockerignore, it's mostly just copy-pasted from various so will need more thought but it's a starting point
- moved ServeCommand back to packages/router/ because I see aloft perhaps not being something pushed by default otherwise, since the primary consumption will be from a public registry

iamdadmin:
While this is a working start, there's more to be done on making a fixed bill-of-materials, pinning and tracing versions in each build. At the moment we're just inheriting frankenphp and plopping it in a distroless image which works, but we can do better.

Alpine is out for this reason <https://frankenphp.dev/docs/performance/#dont-use-musl>

> # Don’t Use musl
> 
> The Alpine Linux variant of the official Docker images and the default binaries we provide are using the musl libc.
> 
> PHP is known to be slower when using this alternative C library instead of the traditional GNU library, especially when compiled in ZTS mode (thread-safe), which is required for FrankenPHP. The difference can be significant in a heavily threaded environment.
> 
> Also, some bugs only happen when using musl.
> 
> In production environments, we recommend using FrankenPHP linked against glibc, compiled with an appropriate optimization level.
> 
> This can be achieved by using the Debian Docker images, using our maintainers .deb, .rpm, or .apk packages, or by compiling FrankenPHP from sources.
> 
> For leaner or more secure containers, you may want to consider a hardened Debian image rather than Alpine.

So that brings us to options.

1. Fork googlecontainertools distroless and maintain a PHP layer on top, there's an existing fork which does this but it's 4 years behind commits and also limited to PHP8.3 at the moment.

Downsides - php-zts packages aren't in Sury repo and will have to be built from source, Bazel really wants to deploy from repos, and it's a fairly big fork to maintain

2. Custom mmdebstrap/debootstrap wrapper with declarative package manifests.

Downsides - The tool is lower-level but can make fixed-file systems from package lists. We'll have to do some work to wrap a file-pinning BOM around it, and maintain that. Again, Sury repo doesn't do php-zts so we'd have to build PHP from source. We'd have to make and maintain it, breaking fairly fresh ground.

3. Use Ubuntu as base, using Chisel. 

Downsides - Adopting Canonical may not be the wisest due to their not entirely FOSS-friendly choices at times (i.e. insisting on Snap instead of using Flatpak) and Chisel *cannot* work with non-Ubuntu repos. So we'd have to build php-zts *and* frankenphp from source and manage all those deps, which is a bit less ideal. That means no single BOM.

4. NixOS container build. They already have php-zts and frankenphp, and it's updated fairly promptly. Builds are driven by a BOM so it is pinned and documented inherently.

Downsides - relying on community maintainers to bump the php-zts and frankenphp packages, although granted, we could potentially submit PRs if we needed to. Learning curve, I'm willing to dive in, but it is it's own whole thing.

### My suggestion

Given all of the above, my feeling is that NixOS is probably actually the best way to go. We can inject our Caddyfile, and Nix can build an OCI-compatible image that we can just import to Docker or podman locally, or push to a registry.

We'd produce a flake file for each build, and then as with composer, a flake.lock file is generated with a fixed package list. If we get a security advisory in dependencies, we can bump individual files and re-build.

When frankenphp and/or php gets a bump, aka our 'primary' packages, we could either manage with git tags so we can inspect point in time, or we can literally duplicate the original flake file side-by-side in another folder or something.

iamdadmin:
Okay, I've done a test with Nix, and there's some constraints that I think rule it out.

1. it's comparatively much slower than the build I've already done which translates into real wasted CI/CD minutes.
2. There's complications with Mac, under which it doesn't run and has to be QEMU'd or cross-compiled with MUSL with which are trying to avoid.
3. Building a single-stage multi-arch image means cross-compiling under emulator and appears to be quite slow, so you have to build separate arch and combine them after.
4. The images are nowhere near as small, which does defeat the point rather.

On to the next one!

packages/router/src/Commands/ServeCommand.php
aidan-casey, line 1: I think I'd scratch this file, personally. Let's assume the end user with Docker knows how to run this with Docker and provide a sensible docker-compose default.

packages/aloft/docker/Caddyfile.noworker
```diff
@@ -0,0 +1,56 @@
...
+	{$CADDY_GLOBAL_OPTIONS}
+
+	frankenphp {
+		{$FRANKENPHP_CONFIG}
+	}
+}
+
+{$CADDY_EXTRA_CONFIG}
+
+{$CADDY_SERVER_NAME:localhost}:{$CADDY_HTTP_PORT:8000}, {$CADDY_SERVER_NAME:localhost}:{$CADDY_HTTPS_PORT:8443}  {
```
aidan-casey: I _think_ you can leave off ports here because they are already defined globally above.

packages/aloft/docker/docker-bake.hcl
aidan-casey, line 1: My gut is that this adds a complexity that we're not after here and we either need to extract this out to a separate build process and publish to a registry or simplify.

packages/aloft/docker/Dockerfile
aidan-casey, line 1: I'm going back and forth on the distroless. I really like the idea in concept, but I have a few concerns:

1. I think the ideal world is that these are stubs we are publishing into the project for someone to use/customize. These would be a lot of stubs to publish and I'm guessing the average user is going to get overwhelmed and/or not understand what is going on here.
2. The resulting image is still pretty large. Knowing the issues with MUSL, I'd be curious if we could get the same/less size using RedHat UBI or something.
3. I am just imagining a situation where someone submits a bug report about something with the Docker image and we want logs or something.

packages/aloft/docker/usage-notes.md
aidan-casey, line 1: We probably need to track the env variables somewhere as there are quite a few at play here.

packages/aloft/docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ARG BASE_IMAGE=dunglas/frankenphp:${FRANKENPHP_VERSION}-php${PHP_VERSION}
+# Controls which distroless variant is used as the runner base.
+# Valid values mirror gcr.io/distroless/cc-debian13 tags: nonroot | debug-nonroot
+ARG DISTROLESS_VARIANT=nonroot
+
+FROM ${BASE_IMAGE} AS frankenphp
+
+RUN curl -sSLf \
+		-o /usr/local/bin/install-php-extensions \
+		https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions && \
```
aidan-casey: I concur with previous comments about `latest`. We should probably pin things like this and verify the hash.

packages/ship/stubs/Caddyfile
aidan-casey, line 1: We may [want to change the listening address for the admin API](https://caddyserver.com/docs/api).

packages/aloft/docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+COPY --from=frankenphp --chown=1002:1000 /etc/frankenphp /etc/frankenphp
+
+COPY Caddyfile.noworker /etc/frankenphp/Caddyfile
+
+WORKDIR /app
+
+EXPOSE 8000
+EXPOSE 8443
+EXPOSE 8443/udp
+EXPOSE 2019
```
aidan-casey: Any real reason to expose this?

aidan-casey (commented):
Hey, @iamdadmin! I apologize for only just now getting around to this, there's a lot to review here. 🙂 

First off, I appreciate the work you have put into this. Thank you!

Generally, as I look at this PR, I'm less concerned with the substance of it than I am of the complexity of it.

For example, let's envision the ideal user experience as something like:

```bash
# Publishes the Docker stubs to the root of our project.
tempest aloft:install

# Starts a dev environment.
docker compose up -d

# Builds a production image.
docker build .
```

If we are publishing the stubs to the root of our project, is the average user going to be able to figure out how to add a PHP extension easily? Will they be able to debug a distroless container with little to no logs?

I'm _thinking_ we might need to simplify this further, which might mean publishing a base build to the Docker registry and pulling from that for our stub. That would create an issue with distroless, however.

Dunno, I'm open to discussion on all this, just spitballing as I think about the dev experience.

packages/aloft/docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
+# Build-time arguments — set by bake.hcl, can be overridden individually
+ARG FRANKENPHP_VERSION=1.11.2
+ARG PHP_VERSION=8.5.3
+ARG BASE_IMAGE=dunglas/frankenphp:${FRANKENPHP_VERSION}-php${PHP_VERSION}
```
aidan-casey: I'd be half tempted to pin the version and reduce flexibility here.

aidan-casey:
> Okay, I've done a test with Nix, and there's some constraints that I think rule it out.
> 
> 1. it's comparatively much slower than the build I've already done which translates into real wasted CI/CD minutes.
> 2. There's complications with Mac, under which it doesn't run and has to be QEMU'd or cross-compiled with MUSL with which are trying to avoid.
> 3. Building a single-stage multi-arch image means cross-compiling under emulator and appears to be quite slow, so you have to build separate arch and combine them after.
> 4. The images are nowhere near as small, which does defeat the point rather.
> 
> On to the next one!

How about RedHat's UBI?

iamdadmin:
> Hey, @iamdadmin! I apologize for only just now getting around to this, there's a lot to review here. 🙂
> 
> First off, I appreciate the work you have put into this. Thank you!

No worries at all, it helps me as much as it helps the community as I intend to use it myself. Equally, if I am not going in the direction wanted for Tempest, that's also fine, I'm not going to try to force it on you or anything and if what I want doesn't match I will just make my own package and use it :) So please, don't hold anything back or worry about wording, I won't be offended.

> Generally, as I look at this PR, I'm less concerned with the substance of it than I am of the complexity of it.

I agree, this implementation is more fragile than I was aiming for.

> For example, let's envision the ideal user experience as something like:
> 
> ```shell
> # Publishes the Docker stubs to the root of our project.
> tempest aloft:install
> 
> # Starts a dev environment.
> docker compose up -d
> 
> # Builds a production image.
> docker build .
> ```
> 
> If we are publishing the stubs to the root of our project, is the average user going to be able to figure out how to add a PHP extension easily? Will they be able to debug a distroless container with little to no logs?

You're right that the PHP extensions issue is big. On the distroless side, the debug container has a shell, and we can figure out tweaks like logging configs, or even make sure all the logs go into /var/logs and we make that a container path so they're literally written in /logs/aloft/, or they can be viewed on the shell of course. We already set environment in .env so that can be utilised, or we can just turn all logs on in debug with a 7 day rotation/retention and all off in the distroless PROD image.

I think the goal to aim for is that both images are in the registry for those who want no changes, or extendable locally. 

> I'm _thinking_ we might need to simplify this further, which might mean publishing a base build to the Docker registry and pulling from that for our stub. That would create an issue with distroless, however.
> 
> Dunno, I'm open to discussion on all this, just spitballing as I think about the dev experience.

I agree that this won't work in it's current iteration. My next test is:

1. Writing a DTO class around a json package file, not entirely unlike composer.json in format since we're all used to it.
2. Using the Dagger.io PHP-SDK to programmatically construct an OCI image based on these packages dependencies [example](https://docs.dagger.io/extending/custom-applications/php/).

What this means is that one will be able to do 

```php
./tempest aloft:build --with-phpExtensions "yml,other-extension,another-one,whatever" --compose
```

Internally the --compose flag will just run ./tempest aloft:compose which can be used independently with the registry images.

What this will do is automate the docker-build, locally, and with the --compose flag also build a docker-compose for those who need the full stack.

For the end user? One or two simple commands.

For us? We can use it in CI/CD to make SBOM-traceable images for each version pairing of php/frankenphp. We should probably envision some kind of version strategy for example for when Tempest itself moves past PHP8.5 to PHP8.6 - at this point maybe we archive all older versions from the registry itself, excepting the very last PHP8.5.x and frankenphp image, while retaining the SBOMs in the github repo in case people need to build them locally for their own needs, to keep storage costs managed etc.

Looking forward, we should probably also tie in a package security check and actively remove images with known defects, I'm not sure on that mechanism yet and suspect best to put that in a second release as this commit will just get huge otherwise.

I'm working on the Dagger workflow at the moment, given I have a working Dockerfile it hopefully won't take too much effort to convert it into Dagger commands, and test. Will keep you posted.

packages/aloft/docker/docker-bake.hcl
iamdadmin: Removed docker-bake, we can script this in CI/CD and combine without bake.

packages/aloft/docker/Dockerfile
iamdadmin: Answering in turn

1. We now have aloft:publish or they can be built from aloft:build and optionally pass —with-php-extensions to add more extensions. It’s still distroless. We just now have some vanity commands wrapped around it and it’s significantly simpler. I really went overboard on the docs too.
2.  I’ve gotten it down to 263MB versus FrankenPHP’s Alpine at ~180MB (which doesn’t have any of our extensions which takes up a chunk of that) versus FrankenPHP’s Debian at 622MB.
3. There’s a debug version with busybox. We’d be able to provide instructions to people asking them to volume map specific configs or log files whether or not they use debug in theory. This is a bit of a ‘how long is a piece of string’ but we have options. I can create some internal troubleshooting guides that can live in a GitHub-only readme file if needed too.

packages/aloft/docker/Dockerfile
iamdadmin: Now outdated, different approach taken.

packages/ship/stubs/Caddyfile
iamdadmin: FrankenPHP would have to be built from source for this I think? It’s above 1024 so best to handle this in port mapping anyway.

packages/aloft/docker/Dockerfile
iamdadmin: Yes, if you want to route to the container from another within the docker network, in my case for example I have the docker image on a DMZ docker network with cloudflare tunnel, so I can reverse proxy over that, then it’s required.

When running as host with -p, then it won’t hurt to have it as anything on the network could in theory connect via host anyway. But we could in theory hide these behind ENV or ARG in some way just so they’re off unless needed on, if you have that preference?

if it helps, FrankenPHP exposes them by default also, so we're not breaking new ground.

packages/aloft/docker/Dockerfile
iamdadmin: They exist as ARGs with a default so we can use it in CI/CD. It also gives our users the ability to choose their own versions and we can update the stubs from time to time also. Possibly have the CI/CD write a change to the file to keep the defaults updated when a new version is published, dependabot style but doing a replace in file from a bash script.

iamdadmin:
> How about RedHat's UBI?

I am not sure that RedHat’s FOSS strategy works best for us, I used to use CentOS for example… while it still exists I’m out of the loop on its state and fitness for purpose. So it makes me a little wary that we base on that and they change their minds.

As I mentioned in another reply I managed to get file size down to within spitting distance of Alpine under Debian and it’s almost a default for docker containers to support Debian and Alpine (although not Alpine for us because of MUSL ofc).

It’s about ready for another check when you can Aidan, I will go back over the last comments above and be sure to reply explaining how I’ve dealt with them, some I haven’t fully addressed yet and will need another commit, and we will need to figure out CI/CD and a repo at some point too.

packages/router/src/Commands/ServeCommand.php
iamdadmin: I've removed it from the PR, it will remain in it's original location unchanged, as it still serves a purpose potentially.

packages/aloft/docker/Caddyfile.noworker
iamdadmin: I've tested removing it, it does indeed work fine, updated.

brendt:
I'm a total docker noob so I'll follow @aidan-casey 's opinion on anything.

One thing I do have an opinion on: we generally don't use _names_ for our packages. Just use `tempest/docker` or something generic. The problem with names is that eventually no one understands what it means anymore without reading through docs.

iamdadmin:
> I'm a total docker noob so I'll follow @aidan-casey 's opinion on anything.
> 
> One thing I do have an opinion on: we generally don't use _names_ for our packages. Just use `tempest/docker` or something generic. The problem with names is that eventually no one understands what it means anymore without reading through docs.

No worries, I would suggest we avoid `tempest/docker` though; while we’re using docker to build, containers are a generic technology and so this image should work in podman, k8s, could be imported as an LXC image, et al. And a tempest/container would confuse with `Container`. Therefore it may make more sense to have the module the beginnings of tempest/ship or tempest/deploy, where all techniques and utilities for shipping and deploying a tempest app eventually wind up? Happy to refactor to whatever though.

iamdadmin:
Refactored to use tempest/ship referencing #352 and will nudge it into Ready for review. Still perhaps some questions to answer on CI/CD and public images, but this should provide a fully local deployment as it stands, and we could consider public image repos etc separately.

packages/aloft/docker/usage-notes.md
iamdadmin: Updated docs to cover off the ENV vars :)

frank-laemmer:
Just skimming over this, I wonder if DDEV would be an alternative option, at least for containerized local dev?

brendt:
@aidan-casey all yours to review

iamdadmin:
> Just skimming over this, I wonder if DDEV would be an alternative option, at least for containerized local dev?

Just wanted to add, that I looked into this, and while it should be an option, it's a layer of complexity beyond what would be needed at this time. All we really need is an image with the apt-get packages installed, which is small, reproducible, and easy to maintain.

brendt:
@aidan-casey @iamdadmin 

Do I get it right that after this is merged, I'll have access to a working docker container with frankenphp? I really want to set up proper framework support, but I'm getting stuck on setting up the right environment. I was hoping this would solve my issue so that I could look into the framework side.

iamdadmin:
> @aidan-casey @iamdadmin
> 
> Do I get it right that after this is merged, I'll have access to a working docker container with frankenphp? I really want to set up proper framework support, but I'm getting stuck on setting up the right environment. I was hoping this would solve my issue so that I could look into the framework side.

Yep absolutely. I’m using the image right now for local dev and to host a prod site, have been since just before I marked it ready for review. Perhaps easier to clone the branch into a separate folder and do the build commands, so the image exists locally, then just manually launch the docker run from wherever you need, until Aidan has finished up his review. I will try to remember to rebase the branch today just to make that cleaner for you.

iamdadmin:
Needs stale label removing if you can spare a mo @aidan-casey @xHeaven @innocenzi @brendt many thanks o7

brendt:
@iamdadmin I'm sorry for being so late on this, but I'm still really interested in getting to merge this PR. Could you maybe add a docs entry when you have the time to explain how to use this? I'm ok if this is a whole new page under "features".

It would be helpful to me to try it out, following the actual docs others would follow as well

docs/0-getting-started/03-docker.md
```diff
@@ -0,0 +1,273 @@
...
+We monitor our three upstream providers, GoogleContainerTools, PHP, and FrankenPHP, for security defect announcements.
+
+- We will actively replace pinned-version images utilising an instance of the distroless base image found to have any security defects.
+- We will actively retire pinned-version images utilising an instance of PHP or FrankenPHP releases found to have any security defects.
+
+:::info
+We won't automatically retire 'patch' version releases i.e. PHP8.5.3 > PHP8.5.4, unless subject to security defects specifically, in which case we'll re-build the image and re-publish. We encourage you to monitor the upstreams and update regularly, or use the `:debug` or `:latest` releases where possible.
+:::
+
+## Developing your application with Docker
```
iamdadmin, line 53: @brendt This section and the one below should cover off how to actually use the docker images. I can move the information above to the end as a footnotes/appendix if it makes it clearer? Just let me know what helps.

`./tempest ship:serve` will run it, first from the public repo which doesn’t exist yet, or from a local image - so run `./tempest ship:build` to build the debug image from defaults and then run the serve command once more.

The dockerfiles will be for older-than-latest versions due to some drift but I can do a push for that before merge once review is done and approach is all okay?

aidan-casey:
@iamdadmin - Likewise I am sorry to just be responding on this PR now. It's been open a while and I appreciate your efforts here. Just a few comments on what I'd love to see here before we get this merged.

1. To me, I think the `tempest ship` commands are somewhat useless. The benefit of Docker is the ability for an end-user to completely mirror their production environment without installing any outside dependencies. Because of this, I'd rather not mask the functionality like Laravel chose to, but help teach the user about Docker through our documentation.

2. That said, I think it is still useful to make the environment easy to setup. Because of this, I think we would be wise to introduce a `docker-compose.yml` that gives them a workbench + web server + database server.

3. I think I'd rather avoid a separate debug Dockerfile in the spirit of mirroring production. Was there a specific reason this had to be introduced?

I haven't had enough time to completely review the Dockerfile itself, so I will plan to do so over the weekend.

Thanks again for continuing to move this forward!

iamdadmin:
> @iamdadmin - Likewise I am sorry to just be responding on this PR now. It's been open a while and I appreciate your efforts here. Just a few comments on what I'd love to see here before we get this merged.

No worries.

> 1. To me, I think the `tempest ship` commands are somewhat useless. The benefit of Docker is the ability for an end-user to completely mirror their production environment without installing any outside dependencies. Because of this, I'd rather not mask the functionality like Laravel chose to, but help teach the user about Docker through our documentation.

Fair enough, the publish command is still relevant of course, and the reason I included the serve command is as a docker-based alternative to the existing serve command using the PHP built-in. Build could easily be replaced with documentation if you prefer of course.

> 2. That said, I think it is still useful to make the environment easy to setup. Because of this, I think we would be wise to introduce a `docker-compose.yml` that gives them a workbench + web server + database server.

I very specifically did not do that, because there’s a separate #1984 to cover that. Doing both in this PR I think may have covered too much ground, and it’s already been difficult to constrain the scope to something you all can review as evidenced by the fact it’s been hard to review :)

> 3. I think I'd rather avoid a separate debug Dockerfile in the spirit of mirroring production. Was there a specific reason this had to be introduced?

I made a choice to go with a distroless, rootless image for the ‘prod’ image, and given the principles of distroless, that means no shell. The debug image is identical, but adds busybox. I think there’s merit in both. The prod image follows best practice and has no shell, the dev image is identical except layering on a shell, so it’s representative in dev but allows access to the shell when needed.

We could avoid the shell, and map more log files via volumes and such so things can simply be interrogated in m the host, but again this is already a fairly big PR, review time, etc. There will also be some people going ‘where’s the shell’ so this way they can simply swap the image with the same compose or run command and hop into the shell, do what’s needed, and swap back after.

> I haven't had enough time to completely review the Dockerfile itself, so I will plan to do so over the weekend.
> 
> Thanks again for continuing to move this forward!

No worries. I had to make a lot of choices here to bring this to a starting point, perhaps it merits being merged to a feature branch with several smaller PRs considered individually as smaller pieces towards a whole - and then the whole finished branch merged to main when it’s ready. Still a lot of direction and input needed from you all so might make more sense.

packages/ship/stubs/Dockerfile.debug
```diff
@@ -0,0 +1,108 @@
...
+        echo "Pin: version ${FRANKENPHP_VERSION}+php85*"; \
+        echo 'Pin-Priority: 1001'; \
+        echo ''; \
+        echo 'Package: php*'; \
+        echo 'Pin: release o=Debian'; \
+        echo 'Pin-Priority: -1'; \
+    } > /etc/apt/preferences.d/static-php85; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends busybox ca-certificates curl mailcap xz-utils; \
+    curl https://pkg.henderkes.com/api/packages/85/debian/repository.key -o /etc/apt/keyrings/static-php85.asc; \
```
osbre, line 28: Do we trust this repo enough to use it over the widely known one by Ondřej Surý? Never seen it before.

iamdadmin: It’s from FrankenPHP docs https://frankenphp.dev/docs/#deb-packages - FrankenPHP and PHP-ZTS both come from the same repo.

It’s maintained by one of the FrankenPHP maintainers - https://github.com/henderkes

I have no idea if provenance has been published or anything.

Sury repo doesn’t provide 8.5 (or 8.4 for that fact?) or zts builds.

I did try a build from source for all, using the git repos, but that became dependency hell. I did not try static builds, which might’ve been a good step to avoid dependency issues, but give this is the formal repo for FrankenPHP I didn’t think making Tempest a maintainer of a set of packages added value.

Open to further comments if you have any for either argument or another alternative. I want this to be right as people could end up serving real websites on it or a future iteration so good fundamentals is important.

PR #2009, opened 2026-02-25
feat(database): on database support for isdatabasemodel

### LINKED ISSUE #1660: Support per-operation database selection using `IsDatabaseModel`
### Description

When using multiple databases it's not possible to specifically set which database queries built by IsDatabaseModel should target

On My test project I forked that trait and added $database param to methods but maybe you will think of something better
Or maybe add handling OnDatabase trait

### Benefits

allow for multi database setup when using IsDatabaseModel

laylatichy:
Closes #1660


brendt:
Thank you!

PR #2012, opened 2026-02-26
feat(http): add overridable `getBody()` hook to `IsResponse`

Maarten-Dekker:
Adds a protected `getBody()` method to the `IsResponse` trait that can be overridden in classes that need lazy body resolution. This is useful when building response objects that support method chaining (e.g. `->with()`, `->flash()`, `->cache()`) where the body should only be resolved once all mutations have been applied, rather than eagerly in the constructor.

### DIFF
--- a/packages/http/src/IsResponse.php
+++ b/packages/http/src/IsResponse.php
@@ -19,7 +19,12 @@ trait IsResponse
 {
     private(set) Status $status = Status::OK;
 
-    private(set) View|string|array|Generator|JsonSerializable|null $body = null;
+    private View|string|array|Generator|JsonSerializable|null $bodyValue = null;
+
+    private(set) View|string|array|Generator|JsonSerializable|null $body {
+        get => $this->getBody();
+        set => $this->bodyValue = $value;
+    }
 
     /** @var \Tempest\Http\Header[] */
     private(set) array $headers = [];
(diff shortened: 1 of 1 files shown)


packages/http/src/IsResponse.php
```diff
@@ -123,6 +128,11 @@ public function setStatus(Status $status): self
         return $this;
     }
 
+    protected function getBody(): View|string|array|Generator|JsonSerializable|null
+    {
+        return $this->bodyValue;
+    }
+
     public function setBody(View|string|array|Generator|null $body): self
```
Maarten-Dekker, line 136: Note: this should accept `JsonSerializable` to match the $body property type,

aidan-casey:
Hmmm... My leaning is that generally those use cases deserve their response factory and that should be done there. At the same time, we do have some builder methods on the response trait.

innocenzi:
That seems like a workaround for a deeper problem honestly :/ I agree with Aidan here that we should have a proper factory, even if it's just an alternative to classes using `IsResponse` for more advanced cases

Maarten-Dekker:
I agree with the separation of concerns argument and understand the hesitation. I just felt that since `IsResponse` already has builder methods that mutate state after construction, a hook like this would fit naturally alongside them. But I understand the reasoning.

aidan-casey:
Thanks for your work here, regardless, @Maarten-Dekker!

Issue #2016, opened 2026-03-03
datetime() default expression is quoted (NOW() becomes string literal)


xHeaven:
## Overview

When using a SQL expression as default in a migration, Tempest wraps it in quotes.

Example:
```php
->datetime('created_at', default: 'NOW()')
```

This compiles to `DEFAULT 'NOW()'` instead of `DEFAULT NOW()`.

## Minimal reproduction

```php
use Tempest\Database\Config\DatabaseDialect;
use Tempest\Database\QueryStatements\CreateTableStatement;

$sql = new CreateTableStatement('users')
    ->datetime('created_at', default: 'NOW()')
    ->compile(DatabaseDialect::MYSQL);

echo $sql;

// CREATE TABLE `users` (
//     `created_at` DATETIME DEFAULT 'NOW()' NOT NULL
// );
```

aidan-casey:
Presumably this would be applicable to Boolean defaults as well.

Not sure the most elegant way to solve this one. A VO comes to mind, but not sure that's the best DX.

innocenzi:
Value objects are definitely the right solution here, but we could add a convenient parameter to `datetime` as well

innocenzi:
Or another idea: do not quote values here, users can quote them themselves. That removes the need for a value object. It kind of makes sense but it's not very user friendly

xHeaven:
Perhaps the #2017 modification could be added to `date` too?

PR #2017, opened 2026-03-03
feat(database): support datetime columns with `CURRENT_TIMESTAMP` as the default value

innocenzi:
This PR adds support for creating a datetime column with its default value set to the current timestamp.

```php
new CreateTableStatement('users')
    ->datetime('created_at', current: true);
```

Related to #2016


packages/database/src/QueryStatements/CreateTableStatement.php
xHeaven, line 249: Perhaps this could be added to `date` too?

innocenzi: Absolutely

PR #2027, opened 2026-03-07
feat(view): add optional scopedVariables parent-child inheritance for view components

### LINKED ISSUE #2026: Enable optional scopedVariables inheritance for view components
### Description

Consider this ViewComponent, called `<x-header>`

```php
<header>
    
    <x-container>

        <x-slot name="left">
            <x-header-left
                :siteName="$siteName ?? 'My Title'" />
        </x-slot>

    </x-container>

</header>
```

The contents of `<x-container>` are simply a div with Tailwind container utilities, and a default slot.

In my <x-base-page> I call it with the following (assuming I am also passing it in from elsewhere, of course):

```php
<x-header :siteName="$siteName ?? null" />
```

As it stands today, in order for `$siteName` to be passed down to `<x-header-left>` I'd have to also declare it on `<x-container>` also, like so:

```php
<header>
    
    <x-container
        :siteName="$siteName ?? 'My Title'">

        <x-slot name="left">
            <x-header-left
                :siteName="$siteName ?? 'My Title'"
            />
        </x-slot>

    </x-container>

</header>
```

This is error-prone, and repetitive. The null-coalescing default need only be declared on `<x-container>` in this example, however doing so is further error prone; if it's not declared in the right sequence, or too late, you're still going to end up with variable not found errors. Consistent declaration avoids this, but on the flip-side, it's another dataExpression which requires eval each time.

## How I suggest to fix it

I propose to modify ViewConfig to hold an optional, default-false boolean, to retain current default behaviour. When this is set to true, it will enable the scopedVariables of the parent to be inherited by child components, automatically. Siblings components will receive a copy of the parent scopedVariables, but would not receive anything from siblings. I would also propose ON_EXCEPT and OFF_EXCEPT functionality, should there be some reason to be selective.

Why in the viewconfig? Because the scopedVariables are compiled before the actual content of an individual view, so placing it on a file-by-file basis would be out of order. And also a lot of repetitious variable declarations in a large component kit. So, viewconfig, and support for a pattern-matched whitelist/blacklist.

## Benefits

- Reduce errors.
- Reduce boilerplate.
- Reduce unnecessary repetition and apply DRY principles.
- Default off means it's a non-breaking change, but available for those who need it.

iamdadmin:
Closes #2026

- add ParentScopedVariablesMode and Patterns to ViewConfig
- default off, options OFF, ON, OFF_EXCEPT (whitelist), ON_EXCEPT (blacklist)
- patterns optional with whitelist/blacklist
- add command to publish a views.config.php from stub
- add various tests to check behaviour is as expected

### DIFF
--- a/docs/5-extra-topics/04-contributing.md
+++ b/docs/5-extra-topics/04-contributing.md
@@ -321,3 +321,12 @@ Finally, as we close in on tagging `next major`, features that would usually go 
 ### Milestone deadlines
 
 Even though we release on a non-fixed schedule, we do assign deadlines to the `next major` version. This gives all contributors a clear goal to work towards, and helps us stay on track. The dealine for `next major` also determines the end date of `current minor`
+
+## Brand Guidelines
+
+### Colors
+
+<span class="swatch" style="--color: #1b1429">#1b1429</span>
+<span class="swatch" style="--color: #29abe2">#29abe2</span>
+<span class="swatch" style="--color: #00e7ff">#00e7ff</span>
+<span class="swatch" style="--color: #0071bc">#0071bc</span>


iamdadmin:
I brought this forward in order to reduce repetition in the view components, when you know you want to pass parent-child inheritance automatically. The scopedVariables mechanism is already there, so when the config is set accordingly, it just passes it down. When not required, it does not. Ultimately I would like to evolve this into a system more like Vue, where the props are set, in a computed array, and pushed into the scope. That has challenges of course as it relies on generics, which as we well-know, PHP does not really do. Even with that implemented in some sensible and elegant way, I feel there may still be a good case for a default 'always-push' mechanism. In a complex UI kit it will require attention to detail of a different kind, that of making sure there are no unintended clashes in prop-naming, but I feel that's a simpler thing to build and manage for a developer than the repetitive manual declaration of props.

iamdadmin:
Actually … I’ve had some inspiration and will set this back to draft for more work!

PR #2029, opened 2026-03-09
ci: improve action performance

brendt:
For this one I used Claude's help instead of Aidan's :p

### DIFF
--- a/.github/workflows/coding-conventions.yml
+++ b/.github/workflows/coding-conventions.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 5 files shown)


brendt:
Ok the integration test run isn't shown, so I'll link it manually: https://github.com/tempestphp/tempest-framework/actions/runs/22844937934

My reasoning with these changes is:

- Concurrency just makes a lot of sense
- Reducing the matrix: I know _in theory_ we're missing out on specific cases now, but I want to give this a try anyway. I don't think there ever was a PR where one specific matrix-entry failed; if something fails, it's always a bunch of entries failing together. So my suggestion is to try this out, and see whether we actually run into issues or not. If we do, we can always revert.

PR #2035, opened 2026-03-10
fix(auth): handle OIDC provider identifier keys

luggesexe:
Hello,

after talking on the Discord we figured, that Keycloak (and supposedly other Providers) pass "sub" instead of "id, uuid..." as identificator. Thus adding to the mapping with this PR.

Also: If a property is readonly, we should not try to unset it.

Credits go to @xHeaven


xHeaven (approved):

brendt:
Tagged: https://github.com/tempestphp/tempest-framework/releases/tag/v3.3.1

PR #2039, opened 2026-03-11
chore: improve generic typing for IDE autocompletion

### LINKED ISSUE #1644: Generics are incorrect
### Tempest version

1.0-alpha.6

### PHP version

8.4

### Operating system

macOS

### Description

Calling ->first() should return type `TModel|null`, but it lost generic during the chaining

Image

### Steps to reproduce

```php
        $queryBuilder = query(Book::class);
        $selectQueryBuilder = $queryBuilder->select();
        $where = $selectQueryBuilder->where('title', 'LOTR 5');
        $result = $where->first();
        $result->auth<code completion>
```

brendt:
Closes #1644


brendt:
Unfortunately, making PHPStan and PhpStorm play nice together, while keeping autocompletion in the places we want, seems to be impossible :( In order to make PHPStan happy, I have to add stuff to the template definitions, which causes PhpStorm to not recognize them anymore.

With the changes in this PR though, we get autocompletion on all querybuilder methods, which is far more important than PHPStan having all insights it needs. 

I hope this doesn't break anything.

Once again… a proper spec for generics that everyone had to follow would be the solution here.

PR #2045, opened 2026-03-13
feat(discovery): add support for dynamic path exlusion

innocenzi:
This pull request adds support for dynamically excluding paths from being discovered:

```php
new DiscoveryConfig(
    locations: [
        new DiscoveryLocation('Module\\', base_path('src/Modules')),
        new DiscoveryLocation('Infrastructure\\', base_path('src/Infrastructure')),
        new DiscoveryLocation('Integrations\\', base_path('src/Integrations')),
    ],
)->skipUsing(static function (string $input) {
    if (str_ends_with($input, needle: 'Test.php')) {
        return true;
    }

    if (str_ends_with($input, needle: 'Pest.php')) {
        return true;
    }

    return false;
});
```

It also includes another fix for when Tempest scans a Pest test file:

CleanShot 2026-03-13 at 16 19 47

When encountering a test or `Pest.php` file, `class_exists` or `new ClassReflector` throws because it autoloads the file.

### DIFF
--- a/packages/discovery/src/DiscoveryConfig.php
+++ b/packages/discovery/src/DiscoveryConfig.php
@@ -2,11 +2,15 @@
 
 namespace Tempest\Discovery;
 
+use Closure;
 use Tempest\Support\Filesystem;
 
 final class DiscoveryConfig
 {
     private array $skipDiscovery = [];
+
+    /** @var array<Closure(string): bool> */
+    private array $skipUsing = [];
 
     /** @var array<array-key, class-string<\Tempest\Discovery\Discovery>> The loaded discovery classes that will be used during discovery */
     public array $classes = [];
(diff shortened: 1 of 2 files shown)


aidan-casey:
@innocenzi - why don't we just add exclusion rules and include a couple smart defaults?

innocenzi:
The discovery config already has something similar, but nothing dynamic, I assume for performance. 

I think adding static closure support would be fine too, but in this specific case, we already have a workaround for Pest _within_ Tempest, and I'm extending it for outside of Tempest, I don't think all Pest users should manually have to debug that weird error

innocenzi:
@aidan-casey I added the functionality anyway because it's useful!

packages/discovery/src/DiscoveryConfig.php
```diff
@@ -25,7 +29,25 @@ public static function autoload(string $path): self
 
     public function shouldSkip(string $input): bool
     {
-        return $this->skipDiscovery[$input] ?? false;
+        if (array_key_exists($input, $this->skipDiscovery)) {
+            return true;
+        }
+
+        foreach ($this->skipUsing as $closure) {
```
brendt, line 36: What worries me here is performance, we're running this check for every file during discovery scanning. Luckily most is done before discovery cache, so the only place this might have a negative effect is when there are many non-vendor based closures. I think it's fine to add for now, but maybe this will cause issues at one point.

brendt (approved):

brendt:
Released: https://github.com/tempestphp/tempest-framework/releases/tag/v3.4.1

PR #2049, opened 2026-03-18
feat(container): resettable container

brendt:
This is an experiment in preparation of FrankenPHP support, for which we need a way to reset the application's state. 

I think the right place to handle this reset flow is within the container, since everything within Tempest goes through there. The `Resettable` interface can be implemented by anything, but I envision initializer classes to implement it, keeping all logic regarding a certain dependency in one place. 

Here's one example:

```php
final readonly class SessionInitializer implements Initializer, Resettable
{
    #[Singleton]
    public function initialize(Container $container): Session
    {
        $sessionManager = $container->get(SessionManager::class);
        $sessionIdResolver = $container->get(SessionIdResolver::class);

        return $sessionManager->getOrCreate($sessionIdResolver->resolve());
    }

    public function reset(): void
    {
        $sessionManager = $this->container->get(SessionManager::class);
        
        // Do whatever needs to happen to reset the session
    }
}
```


packages/container/src/GenericContainer.php
```diff
@@ -678,4 +682,27 @@ private function resolveDecorator(string $className, mixed $instance, null|strin
...
+        return $this;
+    }
+
+    public function reset(): self
+    {
+        foreach ($this->resettables as $resettableClass) {
+            /** @var Resettable $resettable */
+            $resettable = $this->get($resettableClass);
+
+            $resettable->reset();
```
xHeaven, line 727: Ideally, `$singletons` would also be evicted here, otherwise we can receive stale singletons after `$container->reset()`.

packages/container/src/GenericContainer.php
```diff
@@ -41,6 +41,10 @@ public function __construct(
 
         /** @var ArrayIterator<array-key, class-string[]> $decorators */
         private(set) ArrayIterator $decorators = new ArrayIterator(),
+
+        /** @var ArrayIterator<array-key, class-string<\Tempest\Container\Resettable>> $resettables */
+        private(set) ArrayIterator $resettables = new ArrayIterator(),
```
xHeaven, line 51: We don't have `getResettables()`/`setResettables()` for this collection. Other collections have these methods.

packages/container/src/GenericContainer.php
```diff
@@ -678,4 +682,27 @@ private function resolveDecorator(string $className, mixed $instance, null|strin
...
         return $instance;
     }
+
+    public function addResettable(string|ClassReflector $resettableClass): Container
+    {
+        if ($resettableClass instanceof ClassReflector) {
+            $resettableClass = $resettableClass->getName();
+        }
+
+        $this->resettables[] = $resettableClass;
```
xHeaven: Perhaps it'd be worth hardening this with either keying or deduplication.

Issue #1480, opened 2025-08-12
Missing M:N relations in ORM/Query builder


asakurayoh:
### Description

The is N:1, 1:N and 1:1 relation, but no N:M relation. 
Possibility to add re condition to retrieve only some of the relation (ex, type of the relation) and retrieve data saved on the middle table.
Here my example:

**movie**
- id
- title
- actors
- screenwriters

**people**
- id
- name

**movie_person**
- id
- movie_id
- person_id
- type (ex: actor, sreenwriter)
- role

So the movie can have many person that are of type "actor", with a role information (can have many of the same actor on the same movie, with different role), and many person of type "screenwriter" without role.

Is there already a way to do it (can't find it). 

Or what would be a good way to do it without much db queries. (the mapper is still a bit of a mystery, the doc is really light on how it work).

Thanks.

### Benefits

Possibility to use M:N relations easily.

brendt:
For now, you must manually create and query the relation model. We'll keep this issue around and I'll work on it after #1458 is merged

PR #2050, opened 2026-03-18
feat(database): add support for hasonethrough relation

laylatichy:
I saw some open issues about missing relation types and as Im building my project I run into a lot of N+1 because of mostly missing belongs to many but then I saw has one tru and has many tru are missing as well so added 1st pr, will come with the other 2 separately stacked

linked issues
#1480

### DIFF
--- a/docs/1-essentials/03-database.md
+++ b/docs/1-essentials/03-database.md
@@ -250,6 +250,49 @@ final class Book
 }
 ```
 
+### Has one through
+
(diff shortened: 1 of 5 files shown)


laylatichy:
mago fails to unrelated changes to my pr
image
I can open second pr with fixes to this but not sure if there was change to some mago configs and you happy with these or not or you want to do them yourself

brendt (approved):
Looks clean! @innocenzi anything to add?

brendt:
Gonna merge it, looks good :)

PR #2056, opened 2026-03-19
fix(generics): add return type annotations to array methods

laylatichy:
added missing generics to `ManipulatesArray` because phpstan is crying a lot ;)

### DIFF
--- a/packages/support/src/Arr/ManipulatesArray.php
+++ b/packages/support/src/Arr/ManipulatesArray.php
@@ -112,6 +112,8 @@ trait ManipulatesArray
 
     /**
      * Shuffles the array.
+     *
+     * @return static<TKey, TValue>
      */
     public function shuffle(): self
     {
@@ -122,6 +124,8 @@ trait ManipulatesArray
      * Removes the specified keys and their values from the array.
      *
      * @param array-key|array<array-key> $keys The keys of the items to remove.
+     *
+     * @return static<TKey, TValue>
      */
     public function removeKeys(string|int|array $keys): self
     {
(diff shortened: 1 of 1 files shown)


innocenzi:
cc @xHeaven @brendt since you guys touched that recently. I think we wanna prioritize IDE support over phpstan here, not sure if we can have both

xHeaven:
> cc @xHeaven @brendt since you guys touched that recently. I think we wanna prioritize IDE support over phpstan here, not sure if we can have both

We can have both in certain cases, but not always. Unfortunately PhpStorm seems to parse phpstan-specific (prefixed) phpdocs too and it messes with certain things in the IDE, like autocomplete.

Also one thing to note: PhpStorm can behave very weirdly with traits and static/self stuff. Afaik the PhpStorm team is aware of this and probably working on it.

laylatichy:
Maybe im in minority here but If i had to choose between these 2 I rather be yelled at by analysis
sure IDE support is nice to have, but imho phpstan is must have, one is a nice qol change one prevents shipping bugs and right now i think only ->map() resolves correct type
and unless its nicely resolved by LSP you will be left trying to get it to work in other IDEs too

xHeaven:
> Maybe im in minority here but If i had to choose between these 2 I rather be yelled at by analysis sure IDE support is nice to have, but imho phpstan is must have, one is a nice qol change one prevents shipping bugs and right now i think only ->map() resolves correct type and unless its nicely resolved by LSP you will be left trying to get it to work in other IDEs too

I think I agree with both you and Enzo. IDE support is pretty much a must-have, but it's known that PhpStorm has issues with traits right now, which hopefully will be fixed soon.

PHPStan is mostly right about things, though, so I definitely like it yelling at me when I'm doing something that is not too smart.

brendt (approved):

brendt:
Thanks!

PR #2058, opened 2026-03-20
fix(database): auto resolve queries for new relations on insert and update

laylatichy:
@brendt shipped a bug yesterday, didn't. notice queries handle inserting/updating relations separately


brendt (approved):

brendt:
Thanks!

PR #2063, opened 2026-03-21
fix(database): fix is relation loaded and correctly unset properties

laylatichy:
unset property so lazy loading works correctly for belongstoman, isRelationLoaded returns corect value and empty results are correctly set

### DIFF
--- a/packages/database/src/IsDatabaseModel.php
+++ b/packages/database/src/IsDatabaseModel.php
@@ -347,13 +347,11 @@ trait IsDatabaseModel
             return $property->getValue($this);
         }
 
-        $type = $property->getType();
-
-        if ($type->isRelation()) {
+        if (inspect(model: $this)->isRelation(name: $name)) {
             throw new RelationWasMissing($this, $name);
         }
 
-        if ($type->isBuiltIn()) {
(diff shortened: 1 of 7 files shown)


laylatichy:
@brendt some fixes so isRelationLoaded works correctly

but, it might have breaking behavior, kinda I guess if somebody relied on broken behavior in his app to work

now
```php
class Author {
    public ?Publisher $publisher = null
}
```
isrelationloaded false, author->publisher throws relationmissing if we dont ->load('publisher') before

```php
class Author {
    #[Lazy]
    public ?Publisher $publisher = null
}
```
isrelationloaded false, author->publisher loads data

```php
class Author {
    #[Eager]
    public ?Publisher $publisher = null
}
```
isrelationloaded true, author->publisher shows data loaded at the time of author fetch

I guess thats how it should work?

brendt:
```php
class Author {
    public ?Publisher $publisher = null
}
```

This was kind of intentional that it returned true, because if has a "default value". However, I agree that it's weird, and I'm fine changing it. We'll see if it breaks anything

brendt (approved):

PR #2064, opened 2026-03-21
fix(database): relations joins overshadowing if property name conflict in chain

laylatichy:
in a case the parent model has the same property name as child model one is overshadowed leading to invalid query

Given models like this:

```php
final class User
{
    public Role $role;

    #[BelongsToMany]
    public array $permissions;
}

final class Role
{
    #[BelongsToMany]
    public array $permissions;
}

final class Permission
{
    public string $label;
}
```

User::select()->with('role', 'role.permissions', 'permissions')->all();

would lead to user.role.permissions being overshadowed by user.permissions and dropped


brendt (approved):

brendt:
Thanks!

PR #2066, opened 2026-03-23
fix(http): remove session middleware while running static:generate

brendt:


brendt:
This makes for a massive improvement while rendering static pages

PR #2068, opened 2026-03-23
feat(http): introduce session cleanup strategies

### LINKED ISSUE #2061: Improve performance of session cleanup
### Tempest version

3.6

### PHP version

8.5

### Operating system

macOS

### Description

Im doing some optimizations on my app for queries as I added debug bar and noticed that session cleanup runs on every request

https://github.com/tempestphp/tempest-framework/blob/3.x/packages/http/src/Session/ManageSessionMiddleware.php

`$this->sessionManager->deleteExpiredSessions()`

there is already command that cleans this up every minute

packages/http/src/Session/CleanupSessionsCommand.php

@innocenzi was there a specific reason for this? Im fine submitting pr but maybe I'm missing osmething

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/2061

This pull request refactors session clean up to allow the choice between cleaning up sessions after every request (the current behavior), cleaning up session after random requests (2/100 chance), or disabling session cleanup in favor of a schedule-based clean up.

The database session installer has been removed in favor of a session installer, which allows choosing a driver (file, redis, database), and allows specifying the clean up strategy (except if Redis is chosen). When the session cleanup is disabled, the installer suggests installing the scheduled cleaning up command.

CleanShot 2026-03-23 at 19 50 48


packages/http/src/Session/ManageSessionMiddleware.php
```diff
@@ -26,7 +29,27 @@ public function __invoke(Request $request, HttpMiddlewareCallable $next): Respon
...
+    {
+        $this->deferredTasks->add(
+            task: $this->sessionManager->deleteExpiredSessions(...),
+            name: 'tempest:session-cleanup',
+        );
+    }
+
+    private function maybeScheduleSessionCleanup(): void
+    {
+        if (random_int(min: 2, max: 100) === 2) {
```
brendt: FYI this isn't a 2 in 100 chance, it's 1 in 99. 2/100 would be `random_int(1, 100) <= 2`

brendt (approved):
Looking great. Feel free to fix that one mistake and then merge

packages/http/src/Session/ManageSessionMiddleware.php
innocenzi: Woops, thanks for fixing!

PR #2070, opened 2026-03-24
feat(discovery): skip discovery with closure

### LINKED ISSUE #1657: SkipDiscovery based on closure

brendt:
Closes #1657

### DIFF
--- a/docs/1-essentials/05-discovery.md
+++ b/docs/1-essentials/05-discovery.md
@@ -306,6 +306,44 @@ new BootDiscovery(
 )();
 ```
 
+You can also mark classes themselves to be skipped entirely by discovery:
+
+```php
+use Tempest\Discovery\SkipDiscovery;
+
+#[SkipDiscovery]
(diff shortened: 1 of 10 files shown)


brendt (commented):
We really need to check why Mago is making random changes… but out of scope for this PR

innocenzi:
> We really need to check why Mago is making random changes… but out of scope for this PR

Yeah I got those when I PR'd yesterday. Those changes are actually legit, the weird thing is that they're not reported in CI

PR #2073, opened 2026-03-25
refactor(core)!: migrate installers from interfaces to attributes

innocenzi:
This pull request changes the API for installers from an interface to an attribute.

This makes it easier to provide metadata (such as a friendly name) and makes them more flexible as well, as we can now have multiple installers in the same class.

I also made it so arguments can be typed in the signature, just like console commands, instead of being dynamic:

```php
#[Installer('Vite', alias: 'vite')]
public function install(
    #[InstallerArgument(prompt: 'Install Tailwind CSS as well?')]
    bool $tailwindcss,
): void {}
```

When an argument is mandatory, it will be prompted by the `install` command, and the prompt can be configured.

BREAKING CHANGE: `Installer` is no longer an interface but an attribute, any custom installer will need to be adapted.

### DIFF
--- a/packages/console/src/Actions/PromptForConsoleInput.php
+++ b/packages/console/src/Actions/PromptForConsoleInput.php
@@ -9,7 +9,6 @@ use Tempest\Console\Console;
 use Tempest\Console\Input\ConsoleArgumentBag;
 use Tempest\Console\Input\ConsoleArgumentDefinition;
 use Tempest\Console\Input\ConsoleInputArgument;
-use Tempest\Support\Str;
 use Tempest\Validation\Rules\IsBoolean;
 use Tempest\Validation\Rules\IsEnum;
 use Tempest\Validation\Rules\IsNotEmptyString;
@@ -26,15 +25,14 @@ final readonly class PromptForConsoleInput
         ConsoleArgumentDefinition $argumentDefinition,
     ): void {
         $isEnum = is_a($argumentDefinition->type, BackedEnum::class, allow_string: true);
-        $question = Str\to_sentence_case($argumentDefinition->name);
 
         $value = match ($argumentDefinition->type) {
             'bool' => $this->console->confirm(
-                question: $question,
+                question: $argumentDefinition->prompt ?? $argumentDefinition->name,
                 default: $argumentDefinition->default ?? false,
             ),
             default => $this->console->ask(
-                question: $question,
+                question: $argumentDefinition->prompt ?? $argumentDefinition->name,
                 options: match (true) {
                     $isEnum => $argumentDefinition->type::cases(),
                     default => null,
--- a/packages/console/src/Actions/PromptForConsoleInput.php
+++ b/packages/console/src/Actions/PromptForConsoleInput.php
@@ -14,6 +14,7 @@ use Tempest\Validation\Rules\IsEnum;
 use Tempest\Validation\Rules\IsNotEmptyString;
 use Tempest\Validation\Rules\IsNumeric;
 
+/** @internal */
 final readonly class PromptForConsoleInput
 {
     public function __construct(
--- a/packages/console/src/Actions/PromptForConsoleInput.php
+++ b/packages/console/src/Actions/PromptForConsoleInput.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tempest\Console\Actions;
+
+use BackedEnum;
+use Tempest\Console\Console;
+use Tempest\Console\Input\ConsoleArgumentBag;
+use Tempest\Console\Input\ConsoleArgumentDefinition;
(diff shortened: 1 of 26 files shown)


packages/console/src/Actions/PromptForConsoleInput.php
```diff
@@ -0,0 +1,61 @@
...
+use Tempest\Console\Console;
+use Tempest\Console\Input\ConsoleArgumentBag;
+use Tempest\Console\Input\ConsoleArgumentDefinition;
+use Tempest\Console\Input\ConsoleInputArgument;
+use Tempest\Validation\Rules\IsBoolean;
+use Tempest\Validation\Rules\IsEnum;
+use Tempest\Validation\Rules\IsNotEmptyString;
+use Tempest\Validation\Rules\IsNumeric;
+
+final readonly class PromptForConsoleInput
```
innocenzi, line 18: Since the logic for prompting missing input for console command was basically the same, I preferred to extract it to an action.

Happy to change its name though

packages/console/src/Actions/ResolveConsoleInput.php
```diff
@@ -0,0 +1,53 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\Console\Actions;
+
+use Tempest\Console\Input\ConsoleArgumentBag;
+use Tempest\Console\Input\ConsoleArgumentDefinition;
+use Tempest\Console\Input\ConsoleInputArgument;
+
+final readonly class ResolveConsoleInput
```
innocenzi, line 12: Same thing, this is extracted from `ConsoleInputBuilder`. It was the easiest thing to do, but I'm wondering if we should really hijack "console arguments" when they aren't really, they're installer argument and are not exactly the same

But that would probably be a lot more abstraction work for no big benefit

packages/console/src/Exceptions/InvalidCommandException.php
```diff
@@ -10,10 +10,11 @@
 
 final class InvalidCommandException extends ConsoleException
 {
+    /**
+     * @param ConsoleArgumentDefinition[] $invalidArguments
+     */
     public function __construct(
         public readonly ConsoleCommand $consoleCommand,
-
-        /** @var \Tempest\Console\Input\ConsoleArgumentDefinition[] $invalidArguments */
```
innocenzi, line 16: `@var` in constructor parameters doesn't work on VS Code, not sure if that's the case for PhpStorm. I'm assuming it works? But `@param` on the constructor should work as well, even though it looks less nice

packages/console/src/GenericConsole.php
```diff
@@ -224,7 +223,7 @@ public function ask(
             $options = wrap($options);
         }
 
-        if (is_string($options) && is_a($options, BackedEnum::class, allow_string: true)) {
+        if (is_string($options) && is_a($options, UnitEnum::class, allow_string: true)) {
```
innocenzi, line 226: This is an unrelated bug that I stumbled upon, there's no need to check specifically for backed enums, all components accept unit ones

packages/core/src/Commands/InstallCommand.php
```diff
@@ -21,38 +28,148 @@
...
+                return null;
+            }
+
+            return Arr\first(
+                array: $this->installerConfig->installers,
+                filter: fn (InstallerDefinition $installer) => in_array($search, $installer->aliases, strict: true) || $installer->id === $search,
+            );
+        }
+
+        private function resolveInstallerArguments(InstallerDefinition $installer): ?array
```
innocenzi, line 86: So in order to support automatically prompting for mandatory arguments, this was required (this and the methods below). It bothers me a bit, but I'm unsure where else to put this.

packages/console/src/Actions/PromptForConsoleInput.php
brendt: Is fine; let's mark it as internal so that we don't have to worry about making breaking changes here

packages/console/src/Actions/ResolveConsoleInput.php
brendt: I think this is fine for now. Can you mark this one as internal as well?

packages/console/src/Exceptions/InvalidCommandException.php
brendt: Feel free to change to @param, but yes indeed @var works with PhpStorm

packages/core/src/Commands/InstallCommand.php
brendt: I think it's fine for now

brendt (dismissed):

brendt:
Can you also add the breaking change in the PR's description? Just so that if people look it up, they immediately see what's breaking.

I'm fine merging this in a minor, as installers are an internal feature, but we do need to document the break it in this PR

PR #2074, opened 2026-03-26
fix(idempotency)!: move attributes to right namespace

brendt:


brendt:
I reverted this because it's a bit too sketchy if someone actually relies on this… We'll have to do it in a major :( https://github.com/tempestphp/tempest-framework/issues/2075

PR #2076, opened 2026-03-26
fix(view): improve slot rendering

### LINKED ISSUE #2071: Incorrect Slot Rendering in Nested Components
### Tempest version

3.6

### PHP version

8.5

### Operating system

macOS

### Description

When using nested custom components, the templating engine incorrectly handles the passing of named slots from a parent component to a child component. Specifically, content intended for a named slot in the child component is not rendered in its designated slot; instead, it is concatenated with the default slot content and rendered within the child's default slot.

**Steps to Reproduce:**

1. Create a new project.
2. Add the following three files to the project:

-   x-index.view.php
-   x-panel.view.php
-   home.view.php

4. Render the home.view.php template.

**Expected Output:**
```html
<div style="display: flex; flex-direction: row; gap: 150px;">
    <div>
        <p>Left</p>
    </div>
    <div>
        <p>main</p>
    </div>
</div>
```

In this expected output, the content <p>Left</p> is correctly rendered within the left named slot of the x-index component, and <p>main</p> is rendered within its default slot.

**Actual Output:**

```html
<div style="display: flex; flex-direction: row; gap: 150px;">
    <div>
        <!--<x-slot name="left"/>-->
    </div>
    <div>
        <p>Left</p>
        <p>main</p>
    </div>
</div>
```

The actual output shows that the left named slot of x-index remains commented out, while the content `<p>Left</p>` (which was intended for the left slot) is incorrectly rendered alongside `<p>main</p>` within the default slot of x-index. This indicates that the named slot content from x-panel is being treated as part of the default content for x-index rather than being assigned to its specific named slot.

---

`x-index.view.php`
```html
<!doctype html>
<html lang="en">
<head>
    <title>Document</title>
</head>
<body>
<div style="display: flex; flex-direction: row; gap: 150px;">
    <div>
        <x-slot name="left"/>
    </div>

    <div>
        <x-slot/>
    </div>
</div>
</body>
</html>
```

`x-panel.view.php`
```html
<x-index>
    <x-slot name="left">
        <p>Left</p>
    </x-slot>
    <x-slot/>
</x-index>
```

`home.view.php`
```html
<x-panel>
    <p>main</p>
</x-panel>

```

brendt:
This PR moves from regex based slot rendering to to AST, and then does some cleanup so that we can combine two AST-parse steps into one.

Because of AST-based slot rendering, we can now detect wether a named slot lives inside a component, thus meaning it's a named slot for that component and should be compiled differently. This closes #2071

### DIFF
--- a/packages/view/src/Elements/ViewComponentElement.php
+++ b/packages/view/src/Elements/ViewComponentElement.php
@@ -41,8 +41,7 @@ final class ViewComponentElement implements Element, WithToken
         private readonly ViewCache $viewCache,
         private readonly ViewComponent $viewComponent,
         array $attributes,
-    )
-    {
+    ) {
         $this->attributes = $attributes;
 
         $this->viewComponentAttributes = arr($attributes)
@@ -100,7 +99,8 @@ final class ViewComponentElement implements Element, WithToken
     {
         $slots = $this->getSlots();
 
-        $compiled = $this->compileComponent()
+        $compiled = $this
+            ->compileComponent()
             ->prepend(
                 sprintf(
                     '<?php return function ($attributes, $slots, $scopedVariables %s %s %s) { extract($scopedVariables, EXTR_SKIP); ?>',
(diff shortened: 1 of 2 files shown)


packages/view/src/Elements/ViewComponentElement.php
```diff
@@ -185,6 +154,157 @@ public function compile(): string
...
+    private function compileSlotTokens(
+        iterable $tokens,
+        ?Token $parentToken = null,
+    ): string {
+        $buffer = '';
+        $isNestedComponentToken = $parentToken !== null && $this->isNestedComponentToken($parentToken);
+
+        foreach ($tokens as $token) {
+            if ($token->tag === 'x-slot') {
+                if ($isNestedComponentToken && $token->getAttribute('name') !== null) {
```
brendt: This is basically the part that fixes the original bug. If this slot is within another component, it's compiled differently

PR #2080, opened 2026-03-27
fix(database): support multiple model properties joining the same table

### LINKED ISSUE #2079: Multiple `#[Eager]` `#[BelongsTo]` relations to same table produce duplicate JOINs with wrong aliases
### Tempest version

3.7

### PHP version

8.5

### Operating system

Linux, Windows

### Description

When a model has multiple `#[Eager]` `#[BelongsTo]` relations referencing the same table, the query builder generates invalid SQL with duplicate JOINs using the same alias.

### Steps to Reproduce

Create a model with multiple `#[BelongsTo]` relations to the same table, both marked with `#[Eager]`, then query for instances:

```php
// Model definition
#[Table('roles')]
class RoleBrowse
{
    public PrimaryKey $id;

    public string $code;

    #[Eager]
    #[BelongsTo('roles.created_by', 'users.id')]
    public ?UserLookup $created_by = null;

    #[Eager]
    #[BelongsTo('roles.updated_by', 'users.id')]
    public ?UserLookup $updated_by = null;
}

// Query
$roles = query(RoleBrowse::class)->select()->whereIn('code', array_column(SystemRole::cases(), 'value'))->all();
```

### Expected Output

```sql
SELECT roles.id AS `roles.id`, ...,
    created_by_user.id AS `created_by.id`, created_by_user.name AS `created_by.name`,
    updated_by_user.id AS `updated_by.id`, updated_by_user.name AS `updated_by.name`
FROM `roles`
LEFT JOIN users AS created_by_user ON roles.created_by = created_by_user.id
LEFT JOIN users AS updated_by_user ON roles.updated_by = updated_by_user.id
WHERE `roles`.`code` IN ('admin','player')
```

### Actual Output

```sql
SELECT roles.id AS `roles.id`, ...,
    users.id AS `created_by.id`, users.name AS `created_by.name`,
    users.id AS `updated_by.id`, users.name AS `updated_by.name`
FROM `roles`
LEFT JOIN users ON roles.created_by = users.id  -- Same alias "users" used twice
LEFT JOIN users ON roles.updated_by = users.id  -- Duplicate JOIN without unique alias
WHERE `roles`.`code` IN ('admin','player')
```

laylatichy:
fixes #2079 

tldr we need to alias using property name instead of table name if duplicate relations are present


packages/database/src/Builder/QueryBuilders/SelectQueryBuilder.php
```diff
@@ -435,6 +436,27 @@ public function getResolvedRelations(): array
...
 
+        return $this->shouldUsePropertyNameAlias($relations);
+    }
+
+    /**
+     * @param Relation[] $relations
+     *
+     * @return Relation[]
+     */
+    private function shouldUsePropertyNameAlias(array $relations): array
```
brendt, line 511: If we have an implementation that works with the property name, why not always use that instead of making a distinction?

brendt (changes requested):

packages/database/src/Builder/QueryBuilders/SelectQueryBuilder.php
laylatichy: @brendt Because if we always alias books as book, then whereRaw('books.title = ?') breaks — books no longer exists in the query scope, only book does. That's a breaking change for every user who references table names in raw queries.

  By only aliasing when there are actual duplicates, existing queries keep working. The alias only kicks in when two relations point to the same table and the SQL would otherwise be ambiguous.

brendt (approved):
Can you fix the merge conflicts? After that I'm fine to merge

laylatichy:
@brendt resolved

PR #2083, opened 2026-03-27
feat(database): add `min`, `max`, `avg` and `sum` to query builders

laylatichy:
adds missing aggregates

not sure why mago changes
```
$result = SelectQueryBuilder::fromQueryBuilder(
    source: $this,
    fields: arr(input: [$field]),
)
```
into
```
$result =
      SelectQueryBuilder::fromQueryBuilder(
          source: $this,
          fields: arr(input: [$field]),
      )
``` 

swapped int/float detection logic to str_contains because mago doesnt allow string + 0 which if by far the fastest  
https://github.com/carthage-software/mago/issues/1525

str_contains is about 10% slower but mago is not crying


brendt:
🔥

PR #2087, opened 2026-03-29
feat(database)!: allow setting char size

regnerisch:
Until now it was not possible to set a specific CHAR size inside the migration: 

```php
->char('uuid')
```

With this PR, users are now forced to set a CHAR size: 

```php
->char('uuid', 36)
```

As Postgres and MySQL consider the previous `foo CHAR` as `foo CHAR(1)`, this seems to be a breaking change. So maybe changing `private int $size = 1` would be better? 

Postgres: 
> If character (or char) lacks a specifier, it is equivalent to character(1)
(https://www.postgresql.org/docs/current/datatype-character.html)

MySQL: 
> If M is omitted, the length is 1
(https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html)

### DIFF
--- a/packages/database/src/QueryStatements/CreateTableStatement.php
+++ b/packages/database/src/QueryStatements/CreateTableStatement.php
@@ -186,7 +186,7 @@ final class CreateTableStatement implements QueryStatement, HasTrailingStatement
     /**
      * Adds a `CHAR` column to the table.
      */
-    public function char(string $name, int $size, bool $nullable = false, ?string $default = null): self
+    public function char(string $name, bool $nullable = false, ?string $default = null, int $size = 1): self
     {
         $this->statements[] = new CharStatement(
             name: $name,
--- a/packages/database/src/QueryStatements/CreateTableStatement.php
+++ b/packages/database/src/QueryStatements/CreateTableStatement.php
@@ -186,10 +186,11 @@ final class CreateTableStatement implements QueryStatement, HasTrailingStatement
     /**
      * Adds a `CHAR` column to the table.
      */
-    public function char(string $name, bool $nullable = false, ?string $default = null): self
+    public function char(string $name, int $size, bool $nullable = false, ?string $default = null): self
     {
         $this->statements[] = new CharStatement(
             name: $name,
+            size: $size,
             nullable: $nullable,
             default: $default,
         );
--- a/packages/database/tests/QueryStatements/CharStatementTest.php
+++ b/packages/database/tests/QueryStatements/CharStatementTest.php
@@ -20,7 +20,9 @@ final class CharStatementTest extends TestCase
         );
 
         $expectedMysql = '`foo` CHAR(36) DEFAULT \'019d38a9-5504-7a16-ab9d-520bbc289ecc\' NOT NULL';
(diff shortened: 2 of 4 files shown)


packages/database/tests/QueryStatements/CharStatementTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Tempest\Database\QueryStatements\JoinStatement;
+use Tempest\Database\QueryStatements\OrderByStatement;
+use Tempest\Database\QueryStatements\SelectStatement;
+use Tempest\Database\QueryStatements\WhereStatement;
+
+use function Tempest\Support\arr;
+
+final class CharStatementTest extends TestCase
+{
+    public function test_char(): void
```
innocenzi, line 13: Can you use the `#[Test]` attribute?

innocenzi (dismissed):
While this is definitely a breaking change, I think that's something we want here, be more explicit 👍 

If you use `CHAR`, it's generally because you expect a specific length, so it doesn't make sense to have an implicit `1` as a default

packages/database/tests/QueryStatements/CharStatementTest.php
regnerisch: Done

packages/database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -187,10 +187,11 @@ public function string(string $name, int $length = 255, bool $nullable = false,
     /**
      * Adds a `CHAR` column to the table.
      */
-    public function char(string $name, bool $nullable = false, ?string $default = null): self
+    public function char(string $name, int $size, bool $nullable = false, ?string $default = null): self
```
brendt: Let's set the default to `1`, that way it's not a breaking change and I think it makes sense for `1` to be default.

brendt (dismissed):

regnerisch:
Unsure if it was the right way to rebase 🤔 If it was wrong, maybe someone can help me fix the mess. I'm sorry.

iamdadmin:
> Unsure if it was the right way to rebase 🤔 If it was wrong, maybe someone can help me fix the mess. I'm sorry.

When you rebase you need to `git push --force`, on the command line, don't use the 'Sync Now' button in the VS Code plugin (if you're using vscode), so it doesn't replay all the commits on your branch too (I've done this before as well!)

You will need to make a copy of or back up the changes you made in commit 51460d2, do a hard reset back to 095abe6, then again rebase but this time do `git push --force`. You can then replay 51460d2.

To get the replay, you can do `git diff 2ae34c2 > ../(outside the git repo)/51460d2.patch` - if you don't put it outside the repo you'll have to stash it or it might get wiped ... just put it outside the git repo! You can then re-apply it with `git apply ../51460d2.patch`.

laylatichy:
@regnerisch i can try to untangle it but not guarantee ;p

laylatichy:
@regnerisch 

git remote add reshepe https://github.com/reshepe/tempest-framework.git
git fetch reshepe fix/char-size


git push origin reshepe/fix/char-size:3.x --force
  
but make backup first please ;p

regnerisch:
Thank you guys!

brendt:
Mind checking the failing CI?

brendt:
One nice-to-have (in another PR): determine the size if none is set based on the default value if that's provided.

brendt:
Thanks!

Issue #2086, opened 2026-03-29
#[BelongsTo] attribute ignored when property name contains "metadata" (or other Latin words)


laylatichy:
### Tempest version

3.7

### PHP version

8.5

### Operating system

macOS

### Description

`ModelInspector::getBelongsTo()` singularizes the property name before checking for
`#[BelongsTo]`. So `$product_metadata` becomes `product_metadatum`, no property found,
attribute never seen. The FK isn't extracted — the whole model gets serialized into the
column instead.

```php
class ProductMetadataLink implements Model {
    use IsDatabaseModel;

    public function __construct(
        public Product $product,
        #[BelongsTo]
        public ProductMetadata $product_metadata, // "metadata" singularizes to "metadatum"
        public string $value,
    ) {}
}
```
```php
(new ProductMetadataLink(...))->save(); // FK not resolved
```

Same bug in getHasOne() and getHasOneThrough() — all three recurse with the
singularized name without checking the original property first. getHasMany() and
getBelongsToMany() don't singularize so they're fine.

Fix — check original property for explicit attribute before recursing in all three methods:
```php
// getBelongsTo() ~line 197
if (! $singularizedName->equals($name)) {
    if ($this->reflector->hasProperty($name)) {
        if ($this->reflector->hasProperty($name) && $this->reflector->getProperty($name)->hasAttribute(BelongsTo::class) {
             return $this->reflector->getProperty($name)->getAttribute(BelongsTo::class);
         }
    }
    return $this->getBelongsTo($singularizedName);
}
```

// Same pattern for getHasOne() (~line 265, check HasOne::class)
// Same pattern for getHasOneThrough() (~line 344, check HasOneThrough::class)

I'll handle the fix and open a PR when i wake up just so I remember about it, unless somebody else finds time.

innocenzi:
Thanks @laylatichy, your contributions are appreciated 🙏🏼

PR #2089, opened 2026-03-29
fix(intl): handle snake case when singularizing/pluralizing last words

laylatichy:
first half fix for #2086 

made regex more robust to handle things like
some__sentence
someSENTENCE
someSentence__

### DIFF
--- a/packages/intl/src/Pluralizer/InflectorPluralizer.php
+++ b/packages/intl/src/Pluralizer/InflectorPluralizer.php
@@ -9,6 +9,8 @@ use Doctrine\Inflector\Inflector;
 use Doctrine\Inflector\InflectorFactory;
 use Stringable;
 
+use function Tempest\Support\str;
+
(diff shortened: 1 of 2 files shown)


innocenzi:
Do you think the helpers methods you created here would make sense as string utils? How does your `splitWords` compare to [`to_words`](https://github.com/tempestphp/tempest-framework/blob/3.x/packages/support/src/Str/functions.php#L82)?

laylatichy:
@innocenzi I can throw splitWords into str helper, but to_words serves a different purpose, from looking  at regex it splits on punctuation, while here we use that one for mostly code identifiers, classes, properties, to_words doesnt need to split things like bookAuthor while mine helper do.

I'm happy to throw that helper into str functions, just cant think of suitable name

innocenzi:
Alright I think it's fine to keep it there then

PR #2090, opened 2026-03-29
fix(database): check explicit relation attributes after singularizing property names

laylatichy:
second half fix for https://github.com/tempestphp/tempest-framework/issues/2086

check if singularized property exists


brendt:
Perfect :)

PR #2092, opened 2026-03-30
fix(view)!: fix inconsistent fallthrough attributes and introduce apply attribute for full control

### LINKED ISSUE #2040: applyFallthroughAttributes doesn't apply when a <?php preamble is present
### Tempest version

3.3.1 dev-main

### PHP version

8.5

### Operating system

Linux, macOS

### Description

applyFallthroughAttributes uses `/^<(?<tag>[\w-]+)/ ` without the m (multiline) flag which anchors to the absolute start of the viewComponent.

For `x-form-view.php`, `x-input-view.php`, `x-submit`view.php` and any of the other built-ins or self-created components with a `<?php` preamble  the regex never matches, and fallthrough attributes silently don't apply.

I've extended the Fixtures and test to demonstrate this - do I have a shot at Perfect Storm?

[fallthrough-preamble.view.php](https://github.com/user-attachments/files/25894349/fallthrough-preamble.view.php)
[x-fallthrough-preamble-dynamic-test.view.php](https://github.com/user-attachments/files/25894347/x-fallthrough-preamble-dynamic-test.view.php)
[x-fallthrough-preamble-test.view.php](https://github.com/user-attachments/files/25894348/x-fallthrough-preamble-test.view.php)

[FallthroughAttributesTest.php](https://github.com/user-attachments/files/25894355/FallthroughAttributesTest.php)

iamdadmin:
Closes #2040.

This is likely to be a breaking change, I've taken the liberty of writing a blog post in Markdown explaining it, which you're welcome to post, adapt, or not use as suits your needs.

If changes are needed, I can adjust as well. Here's it pasted in for all to read.

# Updated approach for fallthrough attributes in Tempest View

Tempest's view components provide a convenient automatic fallthrough of `class`, `style`, and `id` from the call site onto the root element of a component template. Pass `class="mt-4"` to `<x-button />` and it appears on the `<button>` inside — no boilerplate needed.

However, a simple bug in the code meant that this didn't happen consistently, and even then there was no method to control the behaviour if it was not required. The inconsistent nature of the bug - anchored on whether or not there was a PHP preamble or other content before the 'first' HTML token in the component - means that a direct fix would be a breaking change for many applications.

## The Bug

The original implementation used a single regular expression anchored to the start of the compiled template string:

```php
$compiled->replaceRegex(
    regex: '/^<(?<tag>[\w-]+)(.*?["\s])?>/',
    ...
);
```

The `^` anchor means: match only if the very first character of the file is `<`. For a component like this, it worked fine:

```html x-button.view.php
<button class="rounded-md px-2.5 py-1.5 text-sm">
    <x-slot />
</button>
```

But the moment a PHP preamble appeared before the first tag — even a `declare` statement or a comment — the regex found nothing, silently produced no output, and the fallthrough was lost entirely:

```php x-button.view.php
<?php
$variant = $variant ?? 'primary';
?>
<button class="rounded-md px-2.5 py-1.5 text-sm">
    <x-slot />
</button>
```

No error. No warning. The `class` you passed from the parent simply never arrived.

Tempest version 3.8.0 brought changes to the way a view component was compiled, but the bug still asserted itself. The regex was removed and replaced with a proper AST walk using `TempestViewParser::ast()` - a big step forward - but the fallthrough check was anchored to the token index instead of the token type:
 
```php
$shouldApplyFallthrough = $i === 0 && $token->type === TokenType::OPEN_TAG_START && $token->tag !== 'x-slot';
```
 
`$i === 0` means: only attempt fallthrough if the first token in the parsed AST is the opening tag. However, a PHP preamble is its own token. So when one was present, the HTML element appeared at index `1` or later, `$i === 0` was false, `$shouldApplyFallthrough` was false, and the fallthrough was silently dropped — exactly as before, just through different code. The same tests that failed before continued to fail locally.

### Why wasn't it spotted before?

There are tests for this - several in fact. However, they all used minimal fixtures, and none of which had a PHP preamble.

```php fallthrough.view.php
<?php
$componentClass = 'component-class';
$componentStyle = 'display: block;';
?><x-fallthrough-test class="component-class" />
<x-fallthrough-test :class="$componentClass" />
<x-fallthrough-dynamic-test c="component-class" s="display: block;" />
<x-fallthrough-dynamic-test :c="$componentClass" :s="$componentStyle"/>
```

```html x-fallthrough-test.view.php
<div class="in-component"></div>
```

```html x-fallthrough-dynamic-test.view.php
<div :class="$attributes->get('c')" :style="$attributes->get('s')"></div>
```

This update also brings updated tests, all of which include a PHP preamble, with simply a comment to hold it in the file, in order to ensure that the issue is tested for going forwards.

## No developer control over fallthrough attributes

Beyond the bug, the original implementation was all-or-nothing. It always attempted to merge `class`, `style`, and `id` onto the first element, with no way to opt out, no way to target a specific element, and no way to know it had run. If your component defined its own `class` on the root element, the incoming value was appended on top of it regardless.

### Finding the first tag properly

The regex was replaced with a proper AST walk. `TempestViewParser::ast()` already tokenises the template, so instead of guessing with a pattern, we iterate tokens and find the first real `OPEN_TAG_START` or `SELF_CLOSING_TAG` — skipping past PHP blocks, whitespace, comments, and doctypes entirely:

```php
foreach ($tokens as $token) {
    if (in_array($token->type, [TokenType::OPEN_TAG_START, TokenType::SELF_CLOSING_TAG], true)
        && $token->tag !== 'x-slot') {
        $firstToken = $token;
        break;
    }
}
```

The first valid HTML token is found regardless of what precedes it in the file. PHP preambles, `declare(strict_types=1)`, comment blocks — none of them interfere.

### Respecting what the component already declares

The previous behaviour unconditionally merged incoming values onto whatever the root element had. The new behaviour checks first: if the root element already declares `class`, `:class`, `style`, `:style`, `id`, or `:id`, the fallthrough for that attribute is now *skipped* entirely.

```php
if (array_key_exists($name, $firstToken->htmlAttributes)
    || array_key_exists(':' . $name, $firstToken->htmlAttributes)) {
    continue;
}
```

For applications previously relying on the mandatory merge of these attributes, this is a breaking change, meaning you'll need to adjust your views to manually implement the merge. This was a necessary change however, in order to add control.

This now means a component can take ownership of any of these attributes simply by declaring them. Pass whatever you like from the call site — if the component has already configured it, the component wins:

```html x-button.view.php
<button :class="$class ?? 'rounded-md px-2.5 py-1.5 text-sm'">
    <x-slot />
</button>
```

With the above, `class` is owned by the component. Pass `class="something-else"` from the call site and the component's expression takes precedence. `id` and `style`, having no declaration on the root, still fall through as before.

### The new `:apply` attribute

For cases where you want full manual control — spreading additional attributes, filtering specific ones, or building the attribute set dynamically — the new `:apply` attribute hands you that control explicitly.

Placing `:apply` anywhere in a component template disables automatic fallthrough for that component entirely. The `$attributes` variable, an `ImmutableArray` of everything passed at the call site, is always available:

```html x-button.view.php
<button :apply="$attributes">
    <x-slot />
</button>
```

Because `$attributes` is an `ImmutableArray`, you can filter it before spreading. To exclude specific attributes:

```html
<button :apply="$attributes->removeKeys['id', 'style'])">
    <x-slot />
</button>
```

To include only specific attributes:

```html
<button :apply="$attributes->removeKeysExcept(['class'])">
    <x-slot />
</button>
```

You can also build the array entirely yourself in a PHP preamble and pass it in, which pairs naturally with the `:as` attribute for components that conditionally render as different elements:

```php x-button.view.php
<?php
$apply = [
    'class' => $class ?? null,
    'href'  => $href ?? '',
    'target' => (isset($href) && str_contains($href, 'http')) ? '_blank' : null,
];
?>
<button :as="$apply['href'] !== '' ? 'a' : 'button'" :apply="$apply">
    {{ $label ?? '' }}
</button>
```

`:apply` stringifies attribute values according to the following rules: `true` emits a bare attribute name; `false`, `null`, and empty string are omitted; everything else is rendered as `name="value"`.


iamdadmin:
Weird, I can't replicate the run style check's unused view import. Locally mago on the same version doesn't find any unused view imports, and the logs don't show where. I'll do some digging.

Edit to clarify: I did find this, it was in one of the tests.

docs/1-essentials/02-views.md
```diff
@@ -341,21 +341,128 @@ The idiomatic way of using attributes is to always use `{txt}kebab-case`.
...
+#### Excluding specific fallthrough attributes
+
+To exclude specific attributes from falling through, configure your `button` view component like this:
+```html x-button.view.php
+<button :apply="$attributes->diffKeys(array_flip(['id', 'style']))">
+	<x-slot />
 </button>
 ```
+:::info
+Why array_flip? In `$attributes` the keys are the attributes, in the array in the example above, the values are the attributes, you could also pass `['class' => 0, 'width' => 1, etc]` without a flip.
```
brendt: This is cumbersome. Isn't there a better way of doing this?

brendt: FYI, I'm fine with adding a dedicated class for attributes that has dedicated methods to handle this.

brendt: Or add a method on Collection like `$attributes->with('class', 'id')` and `$attributes->without('style', 'id')` — I'm just thinking out loud

packages/view/src/Elements/ViewComponentElement.php
```diff
@@ -156,34 +186,46 @@ public function compile(): string
 
     private function compileComponent(): ImmutableString
     {
+        // If the component template itself uses :apply= anywhere, the developer is controlling
+        // attribute spreading manually. If :apply was set at the call site, all attributes are
+        // forwarded explicitly via the merged $attributes array. Either way, skip auto-fallthrough.
+        $skipFallthrough = str_contains($this->viewComponent->contents, ':apply=') || $this->applyExpression !== null;
```
brendt: Seems very brittle, no?

brendt (changes requested):

docs/1-essentials/02-views.md
iamdadmin: I was thinking that a with or without method would be more elegant, I will do that as a separate PR.

packages/view/src/Elements/ViewComponentElement.php
iamdadmin: It was definitely a concern - but on some consideration I figured that the code itself would be checking an un-hydrated view component, and it only needs to find the string once. Coupled with the semicolon, I thought it likely to be unique enough.

I will look into retrieving it from the AST instead, though. I’m not really wanting to introduce tech debt if I can help it!

docs/1-essentials/02-views.md
iamdadmin: Submitted #2094 and a commit updating the docs.

packages/view/src/Elements/ViewComponentElement.php
iamdadmin: Commit [5b85298](https://github.com/tempestphp/tempest-framework/pull/2092/commits/5b85298f0278d8eebbe8082fe684bb2c62cf856a) refactors to search the AST for it instead, let me know if all okay or if further needed :)

iamdadmin:
Blocked by #2094

iamdadmin:
Will update for #2094 changes and ready it up again after.

brendt:
As discussed in #2097, we'll need a dedicated attributes value object with dedicated methods (I prefer `with` and `without`)

iamdadmin:
> As discussed in #2097, we'll need a dedicated attributes value object with dedicated methods (I prefer `with` and `without`)

There's several instances of `$attributes` spread across `tempest/view`, some of them are array[] and some are ImmutableArray via arr(), each in different spaces and generally contained within itself or between a couple of classes.

Anyone that's already typing $attributes out in their views is going to find this to be a breaking change to their typing at a minimum; in order to maintain compatibility and prevent it breaking anything else people could already be doing to it, and with usage of methods already within `tempest/views`, $attributes effectively needs to just be a copy of ImmutableArray with `with` and `without` declared in the `final class`, because it needs to expose the same methods.

Does this really add value? It seems like we're layering an abstraction just to have aliases for `with` -> `removeKeys` and `without` -> `removeKeysExcept`.

brendt:
Ok the breaking change is a good point…

I guess we can then just use `removeKeys` and `removeKeysExcept` for now.

I want the record to state I don't like it, but I also get why this is the best approach…

iamdadmin:
> Ok the breaking change is a good point…
> 
> I guess we can then just use `removeKeys` and `removeKeysExcept` for now.
> 
> I want the record to state I don't like it, but I also get why this is the best approach…

I'll raise a separate FR to handle the conversion for $attributes as it's own piece of work. Perhaps it can go into 4.0 as it'd potentially be possible to use the major bump rector to re-type it anywhere it's found in developer applications?

In the meantime this PR should be ready to go again, and I will edit this with the ref to the FR in a few.

Edit: #2103 for the $attributes value object class.

brendt (approved):

brendt:
Looks good, thanks!

PR #2095, opened 2026-03-31
fix(database): target `whereHas` callback to the target table instead of the pivot table

laylatichy:
needed to add missing join for belongs to many, has one through and has many through because callback from where has helpers was targeting pivot table instead of target table


brendt:
Thanks!

PR #2097, opened 2026-04-01
refactor(support): remove `withoutKeys` and rename `withKeys` to `removeKeysExcept`

innocenzi:
Follow-up of https://github.com/tempestphp/tempest-framework/pull/2094:

- `withKeys` implied to me that we're adding keys to the array, which was not the case
- `withoutKeys` is almost the same as `removeKeys`

I initially thought it would be better to have different names based on the mutability of the array, but I realized we already had this issue with `removeKeys` for instance, and other existing methods as well. So that will be for another time.

Instead, I made these changes:
- Removed `withoutKeys`, which had almost the same functionality as `removeKeys`
- Removed `withKeys` in favor of `removeKeysExcept`, which has almost the same functionality as `withKeys` except it has the same API as `removeKeys`.

### DIFF
--- a/packages/support/src/Arr/ManipulatesArray.php
+++ b/packages/support/src/Arr/ManipulatesArray.php
@@ -129,7 +129,19 @@ trait ManipulatesArray
      */
     public function removeKeys(string|int|array $keys): self
     {
-        return $this->createOrModify(remove_keys($this->value, $keys));
+        return $this->createOrModify(namespace\remove_keys($this->value, $keys));
+    }
+
+    /**
+     * Removes all keys and their values from the array, except the ones provided by `$keys`.
+     *
+     * @param array-key|array<array-key> $keys The keys of the items to preserve.
+     *
+     * @return static<TKey, TValue>
+     */
+    public function removeKeysExcept(string|int|array $keys): self
+    {
+        return $this->createOrModify(namespace\remove_keys_except($this->value, $keys));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


iamdadmin:
For a mutable array, why not extend removeKeys to accept string|array[]?

innocenzi:
I did suggest that yeah—any idea for `keepKeys` though?

iamdadmin:
> I did suggest that yeah—any idea for `keepKeys` though?

Ah yes sorry I see that now. Maybe ‘onlyKeys’ or ‘reduceKeys’ as it explains that it’ll be mutated to the end goal?

iamdadmin:
How about `retainKeys` and `releaseKeys`?

innocenzi:
@iamdadmin how important is it to you that the values are taken into account? Adapting `removeKeys` would be a confusing breaking change here.

I went ahead and did that:
- Removed `withoutKeys`, which had almost the same functionality as `removeKeys`,
- Removed `withKeys` in favor of `removeKeysExcept`, which has almost the same functionality as `withKeys` except it has the same API as `removeKeys`.

From the examples in your PR, it doesn't seem like you needed to check against the values. I would agree that it would be a nice feature, but I don't think it warrants having confusing new methods that seemingly overlap with existing ones...

Would that be okay for you?

iamdadmin:
> @iamdadmin how important is it to you that the values are taken into account? Adapting `removeKeys` would be a confusing breaking change here.
> 
> I went ahead and did that:
> 
> * Removed `withoutKeys`, which had almost the same functionality as `removeKeys`,
> * Removed `withKeys` in favor of `removeKeysExcept`, which has almost the same functionality as `withKeys` except it has the same API as `removeKeys`.
> 
> From the examples in your PR, it doesn't seem like you needed to check against the values. I would agree that it would be a nice feature, but I don't think it warrants having confusing new methods that seemingly overlap with existing ones...
> 
> Would that be okay for you?

the requirement I have is just:

diffKeys(array_flip([‘key’, ‘anotherkey’]))

Or intersectKeys. Both return a pruned array, one way or the other. Hope that helps!

innocenzi:
Well, the implementation is slightly different but the result is the same, so I'm taking that as a yes!

brendt:
No but… the whole point here was to not have a double negation 😖 `removeExcept` is so much more complex to understand compare to "keep only these keys".

iamdadmin:
Okay how about removeKeys and retainKeys? It clearly shows you’re removing them, or keeping them, and are grammatic sugar.

brendt:
Where did "retain" come from? It's a more difficult word than "keep" and means the same  😅 `removeKeys` and `keepKeys` are fine. Although from the attribute's point of view they still feel… technical. 

`$attributes->keepKeys('a', 'b', …)`

I get that it's technically correct, but I much rather write `$attributes->with` and `$attributes->without`.

I also suggested `$attributes` should be a dedicated class, separate from these array implementations. Maybe that's the better approach after all.

innocenzi:
> `removeExcept` is so much more complex to understand compare to "keep only these keys".

While I wholeheartedly disagree (you removed a word from the name, that makes a difference), I'm totally open to a better (explicit) name

> I also suggested `$attributes` should be a dedicated class, separate from these array implementations. Maybe that's the better approach after all.

That I agree with. That would make things less confusing regarding the naming (with/without could maybe make sense here).

Want me to remove `removeKeysExcept` entirely, and we let @iamdadmin PR an attribute class?

iamdadmin:
How about this:

- remove withKeys/withoutKeys
- forget which is based on removeKeys and therefore both, extended to also accept an array of keys to remove
- get (and underlying get_by_key) is extended to accept an array of keys to fetch

Whether or not $attributes deserves it's own value object class can be a separate question, but we surgically add the ability to push more than one key into these existing, well-named, methods.

For a mutablearray, it'll adjust the self, for a mutablearray, it returns a copy with the mutations, in line with what these commands already do for singular keys, and we avoid cluttering the namespace with several very similar commands. Happy to do the leg work.

brendt:
Look, I'm fine merging this PR, even though I disagree on the name. However, for #2092 I think that attributes value class will be a requirement now.

PR #2099, opened 2026-04-02
feat(database): autodetect char size by default value

regnerisch:
The default value will now be considered as char size if no char size ist specified.

### DIFF
--- a/packages/database/src/QueryStatements/CharStatement.php
+++ b/packages/database/src/QueryStatements/CharStatement.php
@@ -19,11 +19,11 @@ final readonly class CharStatement implements QueryStatement
     public function compile(DatabaseDialect $dialect): string
     {
         return sprintf(
-            '%s CHAR(%s) %s %s',
+            '%s CHAR(%s)%s%s',
             $dialect->quoteIdentifier($this->name),
-            $this->size,
-            $this->default !== null ? "DEFAULT '{$this->default}'" : '',
-            $this->nullable ? '' : 'NOT NULL',
+            $this->size === 1 && $this->default !== null ? mb_strlen($this->default) : $this->size,
+            $this->default !== null ? " DEFAULT '{$this->default}'" : '',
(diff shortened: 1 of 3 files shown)


brendt:
Ah I think we need to change it up a bit: make `$size` nullable, and set the default value to null. Then:

- Is $size is provided, do nothing
- If $size is null and default value is provided: $size = mb_str_len($default)
- If $size is null and no default value is provided: $size = 1

The reason for this is because we want to distinguish between a user-provided `1`, and nothing provided at all.

regnerisch:
Some cases: 

```php
// $size = 1
->char('char') // CHAR(1)
->char('char', default: 'char') // CHAR(4)
->char('char', size: 10, default: 'char') // CHAR(10)
->char('char', size: 1, default: 'char') // CHAR(4)

// $size = null
->char('char') // CHAR(1)
->char('char', default: 'char') // CHAR(4)
->char('char', size: 10, default: 'char') // CHAR(10)
->char('char', size: null, default: 'char') // CHAR(4)
->char('char', size: 1, default: 'char') // CHAR(1)
```

With `$size = 1`, it will always use the default length (if given), which could leed to unexpected behaviour. But same with `$size = null`: Setting `size: 1, default: 'char'` could produce invalid SQL, or unexpected database behaviour. 

I don't know which one is less problematic. I still prefer not setting `$size` at all, so it's more explicit, and the user exactly knows which char length he will get. But of course, this solution does not prevent someone from using `->char('char', size: 1, default: 'invalid')`.

innocenzi:
We should just throw an exception if the user provides a `size` and a `default`, and both don't match

brendt:
> We should just throw an exception if the user provides a size and a default, and both don't match

I'm fine with that as well. Although small change: when the default's length is bigger than the size. 

Actually, I believe that will already happen when running the migration, although a dedicated exception is better than a PDO exception

brendt:
I like it :)

PR #2100, opened 2026-04-03
feat(database): add current to date

regnerisch:
Adds `current` to the `date` function. Extends #2016 

Will only work with MySQL >= 8.0.13 as default expressions for e.g. CURRENT_DATE were added then.


brendt:
Cool! Appreciate it :)

PR #2101, opened 2026-04-04
fix(database): fix inserting multiple properties pointing to the same table

laylatichy:
related to https://github.com/tempestphp/tempest-framework/issues/2079 but on the insert side of things when no attribute is used


brendt:
I like it when a PR with 64 additions is 63 lines of test code, and 1 line of actual fix 👍

PR #2104, opened 2026-04-04
feat(view): add x-slot define attribute to declare slot ownership in nested components

iamdadmin:
Consider this `x-header` component, which uses an `x-container` to position it's content. `x-container` itself is just a `<div>` with some classes for formatting, it has only a default slot.

```php x-header.view.php
<header>
    <x-slot name="top" />
    <x-container>
        <x-slot name="left"> // this slot is assumed to exist within x-container
            <x-header-left />
        </x-slot>
        <div>
            I am in the center
        </div> 
        <x-slot name="right">  // this slot is assumed to exist within x-container
            <x-header-right />
        </x-slot>
    </x-container>
    <x-slot name="bottom" />
</header>
```
At the callsite, if you want to override the default content of the "left" or "right" slot, you can't, because tempest/view assumes these are slots of `x-container`.

This, for example, doesn't match because `x-header` itself doesn't have a "left" slot.
```php callsite.view.php
<x-header>
    <x-slot name="left">Some content I want to insert</x-slot>
</x-header>
```
This PR adds the ability to `define` a slot, as owned in the ViewComponent it's `define`d in. The `x-header` becomes:
```php x-header.view.php
<header>
    <x-slot name="top" />
    <x-container>
        <x-slot define="left"> // define this slot as a slot of x-header, compiled and passed into x-container's default slot
            <x-header-left />
        </x-slot>
        <div>
            I am in the center
        </div> 
        <x-slot define="right">  // define this slot as a slot of x-header, compiled and passed into x-container's default slot
            <x-header-right />
        </x-slot>
    </x-container>
    <x-slot name="bottom" />
</header>
```
At the callsite, this now works
```php callsite.view.php
<x-header>
    <x-slot name="left">Some content I want to insert</x-slot>
</x-header>
```
You can also push content into a child's named slot this way:
```php x-outer.view.php
        <div class="outer">
            <x-inner>
                <x-slot name="left">
                    <x-slot define="left"><x-header-left /></x-slot>
                </x-slot>
            </x-inner>
        </div>
```
At the callsite:
```php outercallsite.view.php
<x-outer>
    <x-slot name="left">My override</x-slot>
</x-outer>
```

### DIFF
--- a/docs/1-essentials/02-views.md
+++ b/docs/1-essentials/02-views.md
@@ -579,11 +579,17 @@ For instance, the snippet below implements a tab component that accepts any numb
 
 ### Define slot ownership in nested view components
 
+You can use `<x-slot name="mySlot" />` interchangeably to both *define* a slot with optional default content, or to provide content to *populate* the slot. Tempest will consider the hierarchy of the components from the AST to automatically detect your intent, however in more complex, especially nested, view components this can result in unexpected behaviour.
+
+To override this behaviour and manually control in which view component your slots are considered to be *defined*, you can use `<x-slot define="mySlot" />` syntax instead. This causes the slot to be registered against the view component in which the keyword 'define' is used, instead of where the `slot` itself appears in the AST.
+
+#### Extendable view component example using `define`
+
 Let us assume you have an `x-container` view component, which is a `<div>` with formatting to act as a flex container for responsive sizing. You use this component repeatedly across your project, and it's effectively a macro to open and close the `<div>`; it doesn't have any slots or do anything special itself otherwise, with only a default `<x-slot/>` to render whatever it is given.
 ```html x-container.view.php
 <div class="w-full max-w-7xl mx-auto px-4 sm:px-6 lg:px-8"><x-slot/></div>
 ```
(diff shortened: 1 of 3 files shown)


iamdadmin:
This isn't a breaking change, happily; it doesn't change the behaviour of `<x-slot name="">` - and is actually a very simple implementation. It's also optional, meaning no need to change anything at all, unless you specifically need this type of slot.

Writing up some docs now.

docs/1-essentials/02-views.md
```diff
@@ -470,6 +470,53 @@ For instance, the snippet below implements a tab component that accepts any numb
 </x-tabs>
 ```
 
+### Define slot ownership in nested view components
+
+Let us assume you have an `x-container` view component, which is a `<div>` with formatting to act as a flex container for responsive sizing. You use this component repeatedly across your project, and it's effectively a macro to open and close the `<div>`; it doesn't have any slots or do anything special itself otherwise, with only a default `<x-slot/>` to render whatever it is given.
```
brendt, line 588: I have a hard time understand what this actually does without re-reading the docs three times.

I think it's a good feature, but I wonder if we can write the docs differently so that it's more clear what it does and how it works?

I'd start with one or two sentences just explaining what the feature is about, and only then move to examples.

brendt (changes requested):

docs/1-essentials/02-views.md
iamdadmin: Yep no worries will give it some more work!

iamdadmin:
Added more detail, ready for another check (and of course, happy to work on it more if it's still unclear).

iamdadmin:
Rebased with latest 3.x commits.

brendt (approved):

brendt:
Nice!

PR #2105, opened 2026-04-04
feat(class-variance): class-variance authority with slot support, and conflict-aware merging

iamdadmin:
`tempest/class-variance` - a PHP package for building component CSS class strings with variants, compound variants, slot support, and conflict-aware merging. Heavily inspired by [CVA](https://cva.style) and [Tailwind-Variants](https://www.tailwind-variants.org/), but with the ability to extend and support other CSS frameworks also.

## Two flavours:

- **`cv()`** — separator-based merging, framework-agnostic CSS, somewhat limited by default but can be extended to support other frameworks.
- **`tv()`** — builds on `cv()` with Tailwind-aware merging, aiming to provide the same outcomes as [tailwind-merge](https://github.com/dcastil/tailwind-merge).

## Main features

- Implements syntax similar to that of CVA and Tailwind-Variants, supporting `slot`, `variants`, `compoundVariants` and `defaultVariants`,
- Note: internally, it assumes 'base' to be the default slot, as per Tailwind-Variants,
- Can be extended to support additional frameworks in the future (contribs welcome),
- Can be installed standalone, although recommended to at least use with `tempest/discovery`,
- initialize once, use many, the tailwind config is quite large and other frameworks likely will be also, so this way it's not initialized many, many times throughout the application,
- Supports custom configs for Tailwind again inspired by Tailwind-Variants, supporting prefixes, separators, and you can both merge and overwrite the Tailwind config through custom configs.

## Why this? / Notes

This is the biggest bit of work I've done; it's based on a fork I put together [over here](https://github.com/iamdadmin/vewe-classvariance) but attempting to fully adopt Tempest paradigms for integration.

Previously discussed with innocenzi in the discord about whether this would be welcome as a contrib ... so I hope you like it and I'm happy to work on any and all feedback and suggestions.

## Acknowledgements / license

All code that went into this from my original fork was released under the MIT license (not much in the end, but it was a rough basis). CVA and Tailwind-Variants are both JS-sphere packages and no actual code was taken from these, not even as a pseudo-conversion to PHP. It's a ground-up implementation of similar syntax, which delivers similar outcomes, and is separate code entirely, however this package could not exist without their work, and the work of their inspirations. There's also an import from a MIT-licensed tailwind-merge package, with acknowledgements in the comments therein.

(If suited, I can put together a detailed ACKNOWLEDGEMENTS.md for this package specifically?)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -97,6 +97,7 @@
     "replace": {
         "tempest/auth": "self.version",
         "tempest/cache": "self.version",
+        "tempest/class-variance": "self.version",
         "tempest/clock": "self.version",
         "tempest/command-bus": "self.version",
         "tempest/console": "self.version",
@@ -138,6 +139,7 @@
         "psr-4": {
             "Tempest\\Auth\\": "packages/auth/src",
             "Tempest\\Cache\\": "packages/cache/src",
+            "Tempest\\ClassVariance\\": "packages/class-variance/src",
             "Tempest\\Clock\\": "packages/clock/src",
             "Tempest\\CommandBus\\": "packages/command-bus/src",
             "Tempest\\Console\\": "packages/console/src",
@@ -171,6 +173,7 @@
             "Tempest\\Vite\\": "packages/vite/src"
         },
         "files": [
+            "packages/class-variance/src/functions.php",
             "packages/clock/src/functions.php",
             "packages/command-bus/src/functions.php",
             "packages/container/src/functions.php",
--- a/docs/3-packages/03-class-variance.md
+++ b/docs/3-packages/03-class-variance.md
@@ -4,14 +4,12 @@ description: "The Class Variance package can be used standalone or with Tempest 
 ---
 ## Introduction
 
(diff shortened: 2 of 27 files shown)


iamdadmin:
The failing test is a CI issue, it couldn't pull redis image so didn't complete. If someone could re-run just that single one that'd be great.

iamdadmin:
I should add, I likely didn’t include everything a new package would need because I’m actually not sure what’s needed. I obviously don’t have GitHub permissions for setting up the split repo etc. I am happy to do the leg work on the parts I can do, with guidance on what’s needed!

brendt:
Sorry for the late reply, I've been busy 😬

TBH I'm not sure whether this should be a core package. I think it's a good candidate for a third party package (which I'm happy to endorse)

brendt:
I've closed for now, but feel free to continue the discussion here :)

iamdadmin:
> I've closed for now, but feel free to continue the discussion here :)

No worries, Enzo expressed interest so I popped it in to be given a review to prompt discussion and decision, so it having done that is fine.

If you have any further thoughts on it and ultimately you think it adds value to the framework after all, but don’t want it to be a standalone package, I can refactor it into `tempest/support` instead? Or wherever. I wasn’t sure where it might land, so I chose to PR it separate so there would be less merge issues etc.

PR #2109, opened 2026-04-05
fix(view): data attribute handling for x-component

### LINKED ISSUE #2000: Dynamic components assume attributes to be string
### Tempest version

3.0

### PHP version

8.5

### Operating system

macOS

### Description

When I try to pass a non-string to a dynamic component like so:

```html
<div :foreach="$config->fields as $field">
    <x-component :is="$field->component" :field="$field" />
</div>
```
I get a `Tempest\View\Exceptions\ViewCompilationFailed` error. The `$field` variable, which is an object, seems to be the culprit. I think this has to do with the following line: https://github.com/tempestphp/tempest-framework/blob/092791b48782454545faf396cb9433b8afb22c8e/packages/view/src/Components/x-component.view.php#L14
It assumes all attributes are strings, but they can be anything. In this case, an object, but when I try to pass an integer or float, I get the same error. Any non-string, basically.

I guess it would be better to use `mixed` here?

iamdadmin:
Closes #2000 

- Extended `x-component` to ensure it handles data, boolean and stringable attribute spread.
- Added tests.

### DIFF
--- a/packages/view/src/Components/x-component.view.php
+++ b/packages/view/src/Components/x-component.view.php
@@ -10,9 +10,20 @@ use Tempest\View\Slot;
 use function Tempest\Container\get;
 use function Tempest\View\view;
 
-$attributeString = $attributes
-    ->map(fn (string $value, string $key) => "{$key}=\"{$value}\"")
-    ->implode(' ');
+$htmlAttributes = [];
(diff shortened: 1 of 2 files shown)


brendt:
Looking good :)

PR #2113, opened 2026-04-07
fix(framework): add replace entries to `composer.json` to prevent polyfills

brendt:
Read the reasoning here: 

- https://stitcher.io/blog/dependency-hygiene
- https://stitcher.io/blog/more-depedency-considerations


brendt:
Open question: we should probably also add these to composer.json of each package?

innocenzi:
> Open question: we should probably also add these to composer.json of each package?

I think so yes, at least for those we want to be available standalone

innocenzi:
@brendt what's the reason for closing? 👀

brendt:
I'm afraid it will lead to situations where people who do want to install any of these packages for _any_ reason alongside Tempest will get stuck.

As unfortunate as it is, I think the only real solution with the tools we currently have is to manage this on the project level :/

PR #2114, opened 2026-04-08
feat(discovery): improve autoloading of discovery paths

innocenzi:
This pull requests slightly improves the performance of the building of discovery by only reading `installed.json` once, and looping on the installed packages once instead of three times.

Additionally:
- I made it so all `tempest/` packages were included in Discovery, not just core/framework/discovery
- I added the ability for a package to define a `extra.tempest.can-discover` property to its `composer.json`, so it can opt in for Discovery without requiring a Tempest package
- I also added the ability for a package that has opted in for Discovery to ignore specific files through `extra.tempest.ignore`, mostly to prevent issues with optional dependencies

### DIFF
--- a/docs/3-packages/02-console.md
+++ b/docs/3-packages/02-console.md
@@ -35,7 +35,7 @@ You may read more about building commands in the [dedicated documentation](../1-
 Tempest will automatically discover all console commands from multiple sources:
 
 1. **Core Tempest packages** — Built-in commands from Tempest itself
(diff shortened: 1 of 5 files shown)


brendt (changes requested):
Very nice, before merging, can you add docs for the new composer config values? I think that's a really great idea!

innocenzi:
Good catch, should definitely be documented!

brendt (approved):

brendt:
Perfect 👌

PR #2116, opened 2026-04-13
refactor(core)!: move `#[Priority]` to `tempest/support`

innocenzi:
This pull request moves `Tempest\Core\Priority` to `tempest/support`.

The reasoning is that it's a class used by multiple packages, which then depend on `tempest/core` and are harder to use standalone without Tempest.

This is a breaking change, as this attribute might be used in userland, so I added a rector rule for it.


eshimischi:
This is a useful feature for developing libraries dependent on this framework. We ask you to consider adopting these changes soon.

brendt:
I'm ok with this, but in order to do it non-breaking, did you consider using `class_alias`? We don't use it atm, but maybe it's not a bad idea?

brendt:
Discussed the class_alias stuff on Discord. We'll not do it. Rather teach our community that we allow automated breaking changes in minor releases. I'll blog about it though and make the docs a bit more clear about it as well

PR #2117, opened 2026-04-13
refactor(generation): remove `tempest/container` dependency

innocenzi:
Depends on https://github.com/tempestphp/tempest-framework/pull/2116, will update this PR when merged

## What

This pull request removes the dependency of `tempest/generation` on `tempest/container` by using PSR-11 instead. This makes the package usable outside of Tempest in applications that user other containers.

## Why

With `tempest/container` as a dependency, Discovery will find many discoveries depending on Tempest's own container, which is obviously not registered (and should not be alongside another container), so it will fail at the `BootDiscovery` step.

I'm thinking that for all packages that we want to be usable standalone, we probably want to do the same.

### DIFF
--- a/packages/generation/composer.json
+++ b/packages/generation/composer.json
@@ -11,6 +11,11 @@
         "tempest/support": "3.x-dev",
         "psr/container": "^2.0"
     },
+    "require-dev": {
+        "spatie/phpunit-snapshot-assertions": "^5.1.8",
+        "phpunit/phpunit": "^12.5.22",
+        "tempest/container": "3.x-dev"
+    },
     "autoload": {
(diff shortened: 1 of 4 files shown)


eshimischi:
#2116 merged

eshimischi:
@brendt would love to see this merged, thanks!

brendt:
Sorry for being late!

brendt:
It's now merged, tagging later today 👍

eshimischi:
> It's now merged, tagging later today 👍

thanks!

brendt:
Tagged in 3.11.0 👍

PR #2122, opened 2026-04-18
chore(deps): update phpunit to 12.5.22

laylatichy:
update phpunit to not affected version

Problem 1
    - Root composer.json requires phpunit/phpunit 12.5.8 (exact version match: 12.5.8 or 12.5.8.0), found phpunit/phpunit[12.5.8] but these were not loaded, because they are affected by security advisories ("PKSA-5jz8-6tcw-pbk4"). Go to https://packagist.org/security-advisories/ to find advisory details. To ignore the advisories, add them to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.
    
https://packagist.org/packages/phpunit/phpunit/advisories?version=9715549

### DIFF
--- a/.github/workflows/isolated-tests.yml
+++ b/.github/workflows/isolated-tests.yml
@@ -70,7 +70,7 @@ jobs:
         uses: supercharge/redis-github-action@1.8.1
 
       - name: Install PHPUnit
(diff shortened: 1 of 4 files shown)


brendt:
Yes, thanks!

PR #2126, opened 2026-04-22
fix(core): report deprecations without throwing

### LINKED ISSUE #2124: Warnings and deprecations get converted to uncatchable ErrorExceptions
So for warnings and deprecations, we convert them to ErrorException:

```
        ini_set('display_errors', 'Off'); // @mago-expect lint:no-ini-set
        set_exception_handler($handler->handle(...));
        set_error_handler(function (int $code, string $message, string $filename, int $line) use ($handler): bool {
            $handler->handle(new ErrorException(
                message: $message,
                code: $code,
                filename: $filename,
                line: $line,
            ));

            return true;
        });
```

So they get reported by the error handler. However, these exceptions aren't actually thrown, so it's super confusion to get a exception page, but then not being able to catch it.

innocenzi:
Closes https://github.com/tempestphp/tempest-framework/issues/2124

This pull request improves our error handling to:
- report notices/deprecations/warnings without throwing
- ignore suppressed errors (with `@`)
- add some tests regarding all of that

That should be a more natural behavior, but we might need to iterate on that (make it configurable?)

### DIFF
--- a/packages/core/src/FrameworkKernel.php
+++ b/packages/core/src/FrameworkKernel.php
@@ -9,6 +9,7 @@ use ErrorException;
 use RuntimeException;
 use Tempest\Container\Container;
 use Tempest\Container\GenericContainer;
+use Tempest\Core\Exceptions\ExceptionProcessor;
 use Tempest\Core\Kernel\FinishDeferredTasks;
 use Tempest\Core\Kernel\LoadConfig;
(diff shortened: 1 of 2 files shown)


brendt:
> That should be a more natural behavior, but we might need to iterate on that (make it configurable?)

Sounds like a good thing to have, although outside the scope of this PR

brendt:
Thanks for fixing this so quickly btw!

PR #2130, opened 2026-05-16
fix(validator): improve validateObject and validateValueForProperty methods

### LINKED ISSUE #2131: Validator throw a `ValidationFailed` exception when `validateObject` is used even if object is valid
### Tempest version

3.10

### PHP version

8.5

### Operating system

Linux

### Description

Hi!
The method `validateObject` defined here
https://github.com/tempestphp/tempest-framework/blob/3.x/packages/validation/src/Validator.php#L39-L58, will always throw an exception because this condition https://github.com/tempestphp/tempest-framework/blob/3.x/packages/validation/src/Validator.php#L55 is always **true**.

This test simple test fails:
```php
    public function test_validate_passes_with_valid_object(): void
    {
        $this->validator->validateObject(new ObjectToBeValidated(name: 'ab'));

        $this->expectNotToPerformAssertions();
    }
```

In addition, when an exception is thrown, the property `ValidationFailed->$targetClass` is not set and for each rule which validation failed (`$failingRules`),  `FaillingRule->$field` is null.

I know it's a not document method `validateObject`, but I was trying to use anyway, and I found the bug. If it's something internal, please ignore this Bug Report.

I'll try to open also a PR.

tux-eithel:
Fix #2131.

I added several assertions for the exception. 
Let me know if they're okay, or should be moved in a specific test inside `tests/Integration/Validator`.

### DIFF
--- a/packages/validation/src/Validator.php
+++ b/packages/validation/src/Validator.php
@@ -49,11 +49,15 @@ final readonly class Validator
 
             $value = $property->getValue($object);
 
-            $failingRules[$property->getName()] = $this->validateValueForProperty($property, $value);
+            $failingRulesForProperty = $this->validateValueForProperty($property, $value);
+
+            if ($failingRulesForProperty !== []) {
+                $failingRules[$property->getName()] = $failingRulesForProperty;
+            }
         }
 
         if ($failingRules !== []) {
-            throw $this->createValidationFailureException($failingRules, $object);
+            throw $this->createValidationFailureException($failingRules, $object, $object::class);
         }
     }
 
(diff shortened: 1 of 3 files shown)


brendt:
Nice, I'm happy to merge this :) And sorry for the late reply!

Can you take a look at that one failing test?

tux-eithel:
Hi @brendt !

Done!

tux-eithel:
Do you think is worth adding documentation for `validateObject` method? I don't see it's mentioned in https://tempestphp.com/3.x/features/validation.

If the feature can be made public,  I can open another PR, with documentation improvements.

Let me know!

brendt:
Definitely worth documenting! Feel free to send a followup PR

brendt:
Thanks! Will tag in a second

PR #2165, opened 2026-06-07
fix(cryptography): support php builds without argon2

### LINKED ISSUE #2147: tempest/cryptography has undocumented hard dependency on Argon2
### Tempest version

3.0

### PHP version

8.5

### Operating system

Linux

### Description

Trying to use Tempest on a PHP version that does not include Argon2 will fail:

```
  Fatal error: Uncaught Error: Undefined constant "Tempest\Cryptography\Password\PASSWORD_ARGON2ID" in vendor/tempest/framework/packages/cryptography/src/Password/HashingAlgorithm.php:10
  Stack trace:
  #0 vendor/tempest/framework/packages/cryptography/src/Password/hashing.config.php(5): [constant expression]()
  #1 vendor/tempest/framework/packages/core/src/Kernel/LoadConfig.php(33): require()
  #2 vendor/tempest/framework/packages/core/src/FrameworkKernel.php(191): Tempest\Core\Kernel\LoadConfig->__invoke()
  #3 vendor/tempest/framework/packages/core/src/FrameworkKernel.php(73): Tempest\Core\FrameworkKernel->loadConfig()
  #4 vendor/tempest/framework/packages/core/src/Tempest.php(14): Tempest\Core\FrameworkKernel::boot('/home/runner/wo...', Array, NULL, NULL)
  #5 vendor/tempest/framework/packages/console/src/ConsoleApplication.php(42): Tempest\Core\Tempest::boot(NULL, Array, NULL)
  #6 vendor/tempest/framework/packages/console/bin/tempest(8): Tempest\Console\ConsoleApplication::boot()
  #7 vendor/bin/tempest(119): include('/home/runner/wo...')
  #8 {main}
```

This will happen even when overriding the password hashing algorithm to BCrypt in `app/hashing.config.php`:

```
<?php

use Tempest\Cryptography\Password\BcryptConfig;

return new BcryptConfig();
```

WarLikeLaux:
`HashingAlgorithm` took its case values from `PASSWORD_ARGON2ID` / `PASSWORD_BCRYPT`, but `PASSWORD_ARGON2ID` only exists on PHP builds compiled with Argon2. Without it the enum can't load, which also kills the bcrypt path (`BcryptConfig` references `HashingAlgorithm::BCRYPT`) and the default `hashing.config.php` (`new ArgonConfig()`) hits a fatal on boot before any bcrypt override applies. The enum now uses the literal strings behind those constants (`'argon2id'`, `'2y'`) - the same values `password_hash()` accepts and `password_get_info()` returns.

The default config falls back to bcrypt only when Argon2 isn't available, so Argon2id stays the default on every build that supports it and the framework boots on the ones that don't.

Fixes #2147


innocenzi:
@brendt would we be fine with a breaking change that would use bcrypt by default?

We can't automate config changes unfortunately, and this fallback implementation is not really ideal.

As far as I know, Argon2 support is in most default PHP installations, so we can also just specify in the docs that it's required for Tempest to run by default

aidan-casey:
@innocenzi whats the reason to use Argon2 over bcrypt?

Side note to keep in mind: in theory, this would have been less of an issue if we published configurations. By not publishing configurations these things become breaking changes.

I'd suggest we just slot this for next major and document for the time being.

innocenzi:
> @aidan-casey
> whats the reason to use Argon2 over bcrypt?

Mainly because it would be a breaking change otherwise. Also, my understanding is that argon2 underperforms compared to bcrypt only if it's not properly configured, it's not worse than bcrypt per-se.

> Side note to keep in mind: in theory, this would have been less of an issue if we published configurations. 

I'm confused here—do you mean that Tempest shouldn't have its internal config files? Would be really bad UX if we force-published them in the default install imo.

> I'd suggest we just slot this for next major and document for the time being.

Fine by me, but I really don't think we _have_ to replace argon2 by bcrypt

aidan-casey:
> Also, my understanding is that argon2 underperforms compared to bcrypt only if it's not properly configured

The [claim](https://x.com/TerahashCorp/status/1155129705034653698), I believe, is that Argon2 is weaker for faster runtimes, which I think we probably are.

> do you mean that Tempest shouldn't have its internal config files?

I'm conflicted. I get the DX side of the argument, but think about it... one change to a config file setting could throw your whole app off if you aren't overriding.

innocenzi:
I read that claim, but my understanding is still that it's a configuration issue, and that it's not easy to configure / there is no good default to recommend.

But granted, I'm not very knowledgeable in this area. If this wasn't for the breaking change, I'd be fine reverting to bcrypt anyway.

> one change to a config file setting could throw your whole app off if you aren't overriding.

I don't think it's that big of a deal, I'd take that tradeoff for the benefits it brings.

Now that I think about it, the breaking change could be mitigated by having our upgrade command publish a config with argon2, so the new default config could use bcrypt without it being breaking?

brendt:
Sorry for being late to this one folks.

So, the breaking change is for projects who use the default config file, but don't have argon available, right? Because if you'd have argon available the current PR will still favor it over bcrypt. Or is the breaking change with the enum value change?

I'm just trying to understand how this would be breaking? If the default config works, it means it'll continue to work with this PR. If it didn't work, people's installation would crash, and they would have to overwrite the config with bcrypt anyway.

Either scenarios won't lead to a breaking change if we merge this PR?

Please tell me where I'm wrong 😁

innocenzi:
@brendt the PR is not breaking currently.

I asked about whether it would be fine because I don't like the condition in the default config, I would prefer a simpler, straightforward default

brendt:
Aaah I see! In that case we'll merge this as is, and then we can consider refactor it for 4.x to make it more clean :)

PR #2166, opened 2026-06-10
docs(database): add missing query function import

tobz-nz:


innocenzi:
Thanks, but FYI we have many snippets with missing imports on purpose to not bloat the snippet with too much information. I think this case is fine to add the import though

tobz-nz:
Ok, Will keep that in mind.  In this case was because it took me ages to find where to get the query function from & I thought it might help someone else adding it.

innocenzi:
Thank you!

PR #2169, opened 2026-06-12
fix: preserve numeric zero values in view expression attributes

### LINKED ISSUE #2167: View attribute binding drops numeric zero values
### Tempest version

3.11.6

### PHP version

8.5

### Operating system

macOS

### Description

Tempest view attribute bindings appear to omit an attribute when the bound value is numeric 0 or float 0.0. This is surprising for non-boolean attributes such as data-*, where "0" is a meaningful value.

We hit this while rendering a proportional conference schedule. A talk starting exactly at the beginning of the timeline had start_percent = 0.0, which we render as:

```
<article
    data-talk-lineup-talk
    :data-start-percent="$lineupTalk['start_percent']"
    :data-width-percent="$lineupTalk['width_percent']"
>
```

Expected output:

```
<article data-talk-lineup-talk data-start-percent="0" data-width-percent="10">
```

Actual output:

```
<article data-talk-lineup-talk data-width-percent="10">
```

This is (to me, at least) unexpected behavior. It looks like the intent was to drop `:empty-attribute=""` - which is a defensible choice given HTML's semantics, but it's a bit too wide. On a data-level, the value '0' is semantically relevant - it shouldn't be a view layer's decision to completely omit the attribute in this case. I expect this to be a bug due to a non-strict `!empty()`-check.

sawirricardo:
## Problem

View expression attributes (`:data-xxx="$value"`) silently drop numeric zero values (`0`, `0.0`). Reported in #2167.

```html
<div :data-start-percent="$start" :data-width-percent="$width">
```

With `start = 0.0, width = 10`, the `data-start-percent` attribute was omitted entirely.

## Root Cause

`ExpressionAttribute::render()` used a loose falsy check (`! $value`) which treats PHP `0` and `0.0` as falsy.

## Fix

One-line change in `ExpressionAttribute.php`:

```php
// Before
if (! $value) {

// After
if ($value === false || $value === null || $value === ) {
```

Strict checks preserve `false`/`null`/`""` omission while rendering numeric zero as a proper attribute (`data-start-percent="0"`).

Fixes #2167


sawirricardo:
Closing to recreate with minimal diff.

PR #2172, opened 2026-06-17
feat(core)!: worker mode support

brendt:
## Breaking changes

- Change `Tempest\Core\Kernel::shutdown()` signature
- Add `Tempest\Database\Connection::inTransaction(): bool`
- Add `Tempest\Database\Connection::ping(): bool`
- Add `Tempest\Database\Connection::reconnect(): void`

## Upgrade guide

Start by installing Rector if you haven't yet:

```
composer require rector/rector --dev 
vendor/bin/rector
```

Next, update Tempest; it's important to add the `--no-scripts` flag to prevent any errors from being thrown during the update.

```sh
composer require tempest/framework:^3.14 --no-scripts
```

Then configure Rector to upgrade to Tempest 3.14:

```php
// rector.php

use \Tempest\Upgrade\Set\TempestSetList;

return RectorConfig::configure()
    // …
    ->withSets([TempestSetList::TEMPEST_314]);
```

Next, run Rector:

```
vendor/bin/rector
```

Finally: clear config and discovery caches, and regenerate discovery:

```
rm -r .tempest/cache/config
rm -r .tempest/cache/discovery
./tempest discovery:generate
```


packages/database/src/DatabaseInitializer.php
```diff
@@ -26,26 +29,43 @@ public function canInitialize(ClassReflector $class, null|string|UnitEnum $tag):
...
 
-                $connection = new PDOConnection($config);
-                $connection->connect();
+        $connection = $config->usePersistentConnection
+            ? self::$connections[$connectionKey] ?? null
+            : null;
 
-                return $connection;
-            },
+        if (! $connection) {
```
brendt, line 39: We don't use the default connection from `ConnectionInitializer` because within the context of "database" there might be different connections based on the database tag. 

We might make this more consistent in the future by refactoring `ConnectionInitializer`, but that's outside the scope of this PR

packages/database/src/DatabaseInitializer.php
```diff
@@ -26,26 +29,43 @@ public function canInitialize(ClassReflector $class, null|string|UnitEnum $tag):
...
+            $connection = new PDOConnection($config);
+            $connection->connect();
+            self::$connections[$connectionKey] = $connection;
+        }
+
+        if ($connection instanceof PDOConnection && $connection->ping() === false) {
```
brendt: In 4.0 we'll add `ping` and `reconnect` to the `Connection` interface: #2173

packages/auth/src/Authentication/SessionAuthenticatorReset.php
```diff
@@ -0,0 +1,17 @@
...
+
+final readonly class SessionAuthenticatorReset implements Resettable
+{
+    public function __construct(
+        private SessionAuthenticator $sessionAuthenticator,
+    ) {}
+
+    public function reset(): void
+    {
+        $this->sessionAuthenticator->clearCurrent();
```
brendt, line 15: I think I prefer this pattern where a dedicated class is responsible for resetting rather then the subject itself. That way it's also easier to hook into eg. static class resets (if ever needed) rather than eg. `StaticResettable` like the previous PR did.

brendt:
@laylatichy it took a little longer, but here's my draft PR to implement the resettable interface and provide persistent database connections. Would love if you could share your input and possibly tell me what I missed :)

brendt:
To be clear: the goal of this PR isn't to provide the worker application, I'll work on that in a separate PR. I first want to make sure I haven't missed anything obvious in the base implementation

brendt:
Also something that's missing compared to #1996 is the memoization reset. I don't think they need to be reset, since they only contain reflection data that never changes. In fact, I think it's better not to reset them, since resetting would decrease performance.

packages/database/src/DatabaseReset.php
```diff
@@ -0,0 +1,18 @@
...
+
+final readonly class DatabaseReset implements Resettable
+{
+    public function __construct(
+        private Container $container,
+    ) {}
+
+    public function reset(): void
+    {
+        $this->container->unregister(Database::class);
```
laylatichy: shouldn't tagged be cleared too?

packages/database/src/DatabaseInitializer.php
```diff
@@ -26,26 +29,43 @@ public function canInitialize(ClassReflector $class, null|string|UnitEnum $tag):
...
         );
     }
+
+    private function getConnectionKey(DatabaseConfig $config): string
+    {
+        return hash('xxh128', serialize([
+            $config->dsn,
+            $config->username,
+            $config->options,
+        ]));
```
laylatichy, line 69: would probably add password to hash, in case of password rotation

packages/database/src/Connection/ConnectionReset.php
```diff
@@ -0,0 +1,18 @@
...
+
+final readonly class ConnectionReset implements Resettable
+{
+    public function __construct(
+        private Container $container,
+    ) {}
+
+    public function reset(): void
+    {
+        $this->container->unregister(Connection::class);
```
laylatichy: same about tagged

packages/database/src/DatabaseInitializer.php
```diff
@@ -26,26 +29,43 @@ public function canInitialize(ClassReflector $class, null|string|UnitEnum $tag):
...
-            },
+        if (! $connection) {
+            $connection = new PDOConnection($config);
+            $connection->connect();
+            self::$connections[$connectionKey] = $connection;
+        }
+
+        if ($connection instanceof PDOConnection && $connection->ping() === false) {
+            $connection->reconnect();
+        }
```
laylatichy, line 45: think you will want to change it to elseif, to avoid pinging just created connection from the if above

packages/database/src/Connection/ConnectionInitializer.php
```diff
@@ -11,13 +11,24 @@
...
+    private static ?Connection $connection = null;
+
     #[Singleton]
     public function initialize(Container $container): Connection
     {
         $databaseConfig = $container->get(DatabaseConfig::class);
 
-        $connection = new PDOConnection($databaseConfig);
-        $connection->connect();
+        $connection = self::$connection;
```
laylatichy: maybe could do with $databaseConfig->usePersistentConnection check similar to DatabaseInitializer

laylatichy:
worth noting DeferredTasks, right now you finish them in shutdown, resetting them is kinda tricky, might require different approach 

and db connection reset is a bit tricky with transactions, on db connection from request crashes mid transaction ping() i think returns fine, next requests reuses that transaction, but might need validating, a lot of time passed since i digged into it 

on the memo yeah it should stay no point in resting it with how it's used now

aside from that looks good

brendt:
> and db connection reset is a bit tricky with transactions, on db connection from request crashes mid transaction ping() i think returns fine, next requests reuses that transaction, but might need validating, a lot of time passed since i digged into it

Wouldn't the right approach be that all transactions should be finished before ending the request? My first idea is to have a `TransactionReset`, which will simply throw an exception when there's still an open transaction. Is that too naive?

> worth noting DeferredTasks, right now you finish them in shutdown, resetting them is kinda tricky, might require different approach

Good point. Should we support it though? I think an easier "solution" is to not allow deferred tasks in worker mode, and force users to use async commands instead. Deferred tasks are a lazy solution anyway for those who don't want to set up async commands, but I think it's fair to assume people who chose worker mode won't mind using the command bus. WDYT?

brendt:
Edit: having said that, I realize we actually use deferred tasks to clean up sessions 😅 Ok I'll think about it some more

brendt:
Oh, actually: defer will work as expected:

> Defer uses fastcgi_finish_request() under the hood. FrankenPHP implements this function (even if we don't rely on FastCGI/FPM at all, we can, and do emulate this function).

[source](https://github.com/php/frankenphp/issues/1046)

That's good, it means I just have to refactor the shutdown function. 👍

brendt:
Ok @laylatichy a couple more changes:

- Applied all your review comments
- Added a check for open transactions when resetting. I'm currently throwing an exception. Does that make sense? Would auto-commit be better? It feels weird to do that, especially since everything that's using `Database` instead of `Connection` or `TransactionManager` should auto-commit or close, so if you end up with an open transaction, something lower-level weird is going on and I'd say likely a bug
- Added `Tempest\Router\WorkerApplication` that should be used in worker mode. This one never exits but instead resets the container after every run. 
- Deferred tasks are handled before reset, which FrankenPHP apparently has built-in support for, so I _guess_ they should be fine?
- Also added a couple more kernel events: `SHUTTING_DOWN`, `RESETTING`, and `RESET`; we don't use them, but I reckon they might be convenient for debugging. The full flow in worker mode is as follows: `shutting_down > shutdown > resetting > reset`.

packages/router/src/Exceptions/HttpExceptionHandler.php
```diff
@@ -34,6 +34,8 @@ public function handle(Throwable $throwable): void
             $this->responseSender->send($this->renderResponse($request, $throwable));
         } finally {
             $this->kernel->shutdown();
+
+            exit();
```
xHeaven: This will cause 4xx/5xx/error responses to kill worker mode. We might need a separate pipeline for worker mode that allows rendering but doesn't kill the app.

packages/database/src/DatabaseInitializer.php
```diff
@@ -26,26 +29,42 @@ public function canInitialize(ClassReflector $class, null|string|UnitEnum $tag):
...
     }
+
+    private function getConnectionKey(DatabaseConfig $config): string
+    {
+        return hash('xxh128', serialize([
+            $config->dsn,
+            $config->username,
+            $config->options,
+            $config->password,
+        ]));
```
xHeaven, line 69: We should include the `tag` as well, otherwise, 2 connections with the same setup but different tags will collide.

packages/database/src/Connection/ConnectionInitializer.php
```diff
@@ -11,13 +11,24 @@
 
 final class ConnectionInitializer implements Initializer
 {
+    private static ?Connection $connection = null;
+
     #[Singleton]
     public function initialize(Container $container): Connection
```
xHeaven, line 18: Maybe it'd be wise to use a keyed cache for connections here as well, like in `DatabaseInitializer`. Resolving a stale `Connection` after config changes might cause connecting to a stale database.

xHeaven (changes requested):

xHeaven:
- If a singleton definition is not an instance (it doesn't have to be) but a closure, `ConnectionReset::reset()` might crash the app.

packages/upgrade/src/Tempest314/UpdateConnectionImplementationsRector.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        if (! $this->implementsConnection($node)) {
+            return null;
+        }
+
+        $hasChanged = false;
+
+        foreach (self::METHODS as $methodName => $returnType) {
+            if ($node->getMethod($methodName) instanceof ClassMethod) {
```
xHeaven, line 44: We should check for inheritance here as well with reflection. We might override inherited methods if we don't check them.

xHeaven (dismissed):

packages/router/src/Exceptions/HttpExceptionHandler.php
brendt: Good point. I've refactored it so that resetting and exiting are optionally handled withinKernel::shutdown 👍

packages/upgrade/src/Tempest314/UpdateConnectionImplementationsRector.php
brendt: Since we actively promote against inheritance from a technical point of view, I don't think we should add it.

brendt:
I suspect the failing job having something to do with PHPUnit rather than this PR. Not going down that rabbit hole in this PR though

PR #2178, opened 2026-06-22
fix(database): use correct return type for IsDatabaseModel::resolve

brendt:


brendt:
PhpStorm complains of `self` is used in a trait while the interface defines `static`. Making this change makes PHPStan complain, which I then suppresed. I think it's more important to have accurate realtime IDE support

PR #2182, opened 2026-06-23
feat(testing): add model factories

brendt:


brendt:
When it comes to generating random values, there's much that can be improved. However this minimal version is a good starting point and I'd like to invite others to add whatever they can think of :)

rexiliano89 (approved):

PR #2183, opened 2026-06-23
fix(database): model inspector `getSelectFields` returns duplicates

NeoIsRecursive:
This causes the mapper to fail "silently".

Example:

```php
// the credential here belongs to the user with id = 2
$passkey = query(Passkey::class)->find(credential_id: $publicKeyCredential->rawId)->first();

// validation logic

// here the user_id would always be 1, signing in the wrong user
$user = query(User::class)->findById($passkey->user_id);
```

After some digging I found out that the user_id returned from the database as an array [1,1] or [2,2] and so on, which the mapper always map to 1?

And the reason for this is that my model has both a relation and a relation_id field.

```php
final class Passkey
{
    public PrimaryKey $id;

    public User $user;

    public function __construct(
        ...
        public int $user_id,
        ...
    ){}
}
```

And since the model inspector appends relation ids to the select fields, there came duplicates which made pdo/sqlite(?) return a list for the user_id key.

Anyhow, very simple fix (unless Im missing something) `->unique()` on the fields.

### DIFF
--- a/packages/database/src/Builder/ModelInspector.php
+++ b/packages/database/src/Builder/ModelInspector.php
@@ -582,7 +582,7 @@ final class ModelInspector
             }
         }
 
-        return $selectFields;
+        return $selectFields->unique();
     }
 
     public function resolveRelations(string $relationString, string $parent = '', array $visitedPaths = []): array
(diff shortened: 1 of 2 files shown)


brendt:
Perfect! Thanks :D

brendt:
Tagged as v3.12.0

PR #2184, opened 2026-06-24
fix(testing): allow to throw exceptions

brendt:


brendt:
Some mago updates as well. I'll just keep it in the same PR

innocenzi:
Curious what was the use case here?

PR #2187, opened 2026-06-24
feat(auth): add middleware and route decorator to auth installer

brendt:


brendt:
As with everything auth, it's not possible to directly ship these with the framework because likely the user wants to make changes to the very simple basic implementation. Rather, it's a starting point for developers to build on

brendt:
@innocenzi wdyt?

innocenzi:
Besides the names, very good idea!

I suggest`AuthMiddleware` -> `MustBeAuthenticatedMiddleware` and `#[Auth]` -> `#[MustBeAuthenticated]`

I have an issue with `Auth` as a name, it's too generic and doesn't convey what it does

innocenzi (approved):

Issue #2194, opened 2026-06-25
Make `ValidationFailed` throwable without rules


NeoIsRecursive:
### Description

Perhaps a bad title, but I think this would be nice to have (standardizing "failure" responses in tempest apps).

For example, here, where the request validations passed, but we need to do additional "validation":

```php
#[Post('/webauthn/authenticate')]
public function authenticate(AuthenticatePasskeyRequest $request, Authenticator $authenticator)
{
   // some deserialization and other logic

    $passkey = query(Passkey::class)->find(credential_id: $publicKeyCredential->rawId)->first();

    if (!$passkey) {
        // here id like to throw a validation exception
    }
```

In laravel there is the `ValidationException::withMessages()` which is what I am missing I suppose. It is possible to do something like that with `ValidationFailed` but it requires a lot of boilerplate + a rule to make the message.

```php
throw new ValidationFailed([
    'answer' => [
        new FailingRule(rule: new MessageRule('Passkey not valid')),
    ],
]);

//

final readonly class MessageRule implements Rule, HasErrorMessage
{
    public function __construct(
        private string $message,
    ) {}

    public function isValid(mixed $value): bool
    {
        return false;
    }

    public function getErrorMessage(): string
    {
        return $this->message;
    }
}
```

Perhaps this is fine, and all that is needed is some GenericRule that takes a message?

(Another thing I found when testing this is that `Validator::getErrorMessage` doesn't check the `HasErrorMessage` if the rule is in a `FailingRule`)

Or is there a better way to handle these cases that I am missing?

PR #2195, opened 2026-06-25
feat(validator): add failing test for `Validator::getErrorMessage`

NeoIsRecursive:
Adds a failing test for the validation error message issue mentioned in #2194, and since I didnt find a more generic test for the translation keys, I added one for that too.

Can add a fix in here too, but im wondering if the `HasErrorMessage` interface is to be deprecated since it is not mentioned in the documentation?

Cheers! 😄

### DIFF
--- a/packages/validation/tests/ValidatorTest.php
+++ b/packages/validation/tests/ValidatorTest.php
@@ -7,7 +7,9 @@ namespace Tempest\Validation\Tests;
 use PHPUnit\Framework\TestCase;
 use Tempest\Reflection\ClassReflector;
 use Tempest\Validation\Exceptions\ValidationFailed;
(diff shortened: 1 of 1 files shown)


innocenzi:
> I'm wondering if the HasErrorMessage interface is to be deprecated since it is not mentioned in the documentation?

No it's not, it's used internally and useful for custom rules for which the translator isn't needed, we should document that

Should we close this PR in favor of Màrk's?

NeoIsRecursive:
> No it's not, it's used internally and useful for custom rules for which the translator isn't needed, we should document that

Ah, good! (mainly wanted to check before doing work).

> Should we close this PR in favor of Màrk's?

Yes! His contains fixes and everything, much better

Issue #210, opened 2024-03-10
Consider reworking flash sessions


brendt:
Right now, a flash session is removed as soon as it's retrieved from the session. Ideally it should be removed when the response is sent, so that you have the possibility to retrieve it more then once during the request.

PR #2198, opened 2026-07-05
fix(session): age flash values at the start of the request

### LINKED ISSUE #1734: Session::cleanup() is not called before shutdown

osbre:
Closes #1734. Addresses #210.

### Problem

A flash value is meant to live for one request and then go away.

The old logic removed a flash value once it was read. But the cleanup happened before the view was rendered - and views are often where the value is actually read (for example, the default form and input components). By then it was too late: the value was already saved back into the session, so it showed up again on the next request.

And if a flash value was never read, it was never removed at all.

### Fix

Instead of expiring flash values at the end of the request, expire them at the start:

- values flashed on the previous request are removed;
- values flashed since then are marked to go on the next request.

Reads no longer affect this, so a flash value stays readable for the whole request, including whilst the view renders, and is reliably gone on the next one.

The session is still saved right away. This is a bit different from the idea in #210 (removing once the response is sent): expiring at the start of the next request gives the same result without deferring any work, and doesn't change the public API.

### Notes

Three tests checked the old read-based timing (`flash()` then a single `cleanup()`); they now check the two-request lifecycle. A new test covers a flash value expiring even when it's never read.

### DIFF
--- a/packages/http/src/Session/ManageSessionMiddleware.php
+++ b/packages/http/src/Session/ManageSessionMiddleware.php
@@ -24,10 +24,14 @@ final readonly class ManageSessionMiddleware implements HttpMiddleware
 
     public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
     {
+        // Flash values are aged at the start of the request, before the response
+        // body is rendered, so a value flashed on the previous request stays
+        // readable this request and is expired on the next one.
+        $this->session->cleanup();
+
         try {
             return $next($request);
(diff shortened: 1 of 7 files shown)


innocenzi (approved):

brendt:
Thank you, much appreciated!

PR #2201, opened 2026-07-06
docs: fix typo on docs 1-essentials/03-database

davidtaim:
A simple typo that I notice while reading the docs.


brendt:
Thanks for taking the time, it's the small things that matter 💪

PR #2203, opened 2026-07-08
feat(validation): include error messages in ValidationFailed exception message

### LINKED ISSUE #2189: Add validation failures in ValidationFailed message
This is useful during testing so that you immediately see what failed from the message

mgalexandre:
**1. Implements `ProvidesContext`**, exposing the errors as structured context, so they show up in log context via `LoggingExceptionReporter` and on the development exception page, the same pattern as `QueryWasInvalid`, `EncryptionFailed`, and `ProcessExecutionWasForbidden`.

**2. Appends the failures to the message in a human-readable format:**

```
Validation failed for App\User.
- credential: Passkey not valid
- email: Email is already taken; Email domain is not allowed
```

The message stays untouched (`Validation failed.`) when there are no error messages.

Fixes #2189

### DIFF
--- a/packages/validation/src/Exceptions/ValidationFailed.php
+++ b/packages/validation/src/Exceptions/ValidationFailed.php
@@ -5,12 +5,11 @@ declare(strict_types=1);
 namespace Tempest\Validation\Exceptions;
 
 use Exception;
+use Tempest\Core\ProvidesContext;
 use Tempest\Validation\FailingRule;
 use Tempest\Validation\Internal\MessageRule;
 
-use function Tempest\Support\Json\encode;
-
-final class ValidationFailed extends Exception
+final class ValidationFailed extends Exception implements ProvidesContext
 {
     /**
      * @template TKey of array-key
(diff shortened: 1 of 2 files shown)


innocenzi:
I don't think this is the right thing to do, honestly. Appending the JSON-encoded message array seems rough around the edges.

In what situations did you need the `message` property of the exception to contain the details, and you could not reach for `errorMessages`?

mgalexandre:
I've reworked it to do two things:
  - ValidationFailed now implements ProvidesContext (as @NeoIsRecursive suggested on the
  issue), exposing the errors as structured context for logs and the exception page.
  - The message lists the failures in a readable format instead of JSON:
  Validation failed for App\User.
  - email: Email is already taken

innocenzi:
Right, didn't take a look at the original issue, thanks @mgalexandre 

@brendt, in what situations do you need the message to contain the actual errors? I do think the context, as suggested by Neo, is more suited here

brendt:
The problem I want solved is this one: you're testing HTTP endpoints with exceptions enabled (a common use case), and you run into a validation error. There's no way to see what's wrong here without making changes to the test code, which is very annoying.

Screenshot 2026-07-10 at 09 54 34

That's the only thing that needs improvement. I don't really care how, but I want to see immediately in the exception message what's wrong

innocenzi:
Thanks, yeah that's definitely and issue. Then I believe our exception renderer should be responsible for that instead of hijacking the exception message.

Maybe by showing the context for all exceptions, or by special-casing `ValidationFailed`?

brendt:
Right but I don't think this is using an exception renderer? It's just PHPUnit showing the exception output during testing — AFAIK

innocenzi:
Ah, my bad, somehow thought this was our console exception renderer... yeah, not much we can do here if you want to use `throwExceptions` instead of `dd` on the response (which is what I do and recommend, btw)

brendt:
@mgalexandre can you run `composer qa` locally and fix the failing Mago rule? After that I'll merge :)

brendt (approved):

mgalexandre:
sorry, missed it! it's done :)

brendt:
Thank you :)

PR #2204, opened 2026-07-08
feat(auth): support refreshing OAuth access tokens

### LINKED ISSUE #2037: Add the ability to refresh OAuth tokens
See: https://oauth2-client.thephpleague.com/usage/#refreshing-a-token

boydbloemsma:
Fixes #2037

### DIFF
--- a/packages/auth/src/OAuth/GenericOAuthClient.php
+++ b/packages/auth/src/OAuth/GenericOAuthClient.php
@@ -101,6 +101,32 @@ final class GenericOAuthClient implements OAuthClient
         }
     }
 
+    public function refreshAccessToken(string $refreshToken): AccessToken
+    {
+        if ($refreshToken === '') {
+            throw OAuthTokenCouldNotBeRetrieved::missingRefreshToken();
+        }
+
+        try {
+            $token = $this->provider->getAccessToken('refresh_token', [
+                'refresh_token' => $refreshToken,
+            ]);
+
+            if ($token instanceof AccessToken) {
+                return $token;
+            }
+
+            return new AccessToken([
+                ...$token->getValues(),
+                'access_token' => $token->getToken(),
+                'refresh_token' => $token->getRefreshToken(),
+                'expires' => $token->getExpires(),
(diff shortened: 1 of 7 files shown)


boydbloemsma:
I don't think the check failing is caused by my changes. Please let me know if I'm wrong.

packages/auth/src/OAuth/GenericOAuthClient.php
```diff
@@ -101,6 +101,32 @@ public function requestAccessToken(string $code): AccessToken
         }
     }
 
+    public function refreshAccessToken(string $refreshToken): AccessToken
```
brendt, line 104: Up for discussion:

- could we call the method `refresh`?
- Maybe it makes sense to allow `string|AccessToken`, and if an access token is provided, automatically resolve the refresh token?

Not sure either of these make sense. Putting it out there to hear your thoughts

brendt: Edit on the first point: just noticed that there's also the `requestAccessToken` method, so it's best to keep naming consistent 👍

brendt (commented):
Solid PR 👌

One small question where I'm curious about your thoughts.

brendt:
Oh and yes, that failing CI run wasn't your fault, it's fixed now :)

packages/auth/src/OAuth/GenericOAuthClient.php
boydbloemsma: I did think about this initially, but I'm not a big fan of using union types.
Also my idea was that most users will store the token as a string in a database and probably not the whole AccessToken object.

One question I do have though is that in this PR I didn't wrap the `League\OAuth2\Client\Token\AccessToken`, in a Tempest version, would you like to see that still? Or keep it as is and users will need to include the League version?

brendt: I'm ok keeping it as is for now :)

brendt:
Thank you, great PR! :)

PR #2206, opened 2026-07-12
feat(tests): allow test response helper to be container injectable

Opblaasmaatje:
I'd like the ability to add my own assertion methods onto the `TestResponseHelper`. 

I believe with container injection this it should be possible, without having to overwrite and maintain the `HttpRouterTester::sendRequest(...)` method.

If there's already a solution in place for this that I've missed, or if this needs any changes or has unexpected side effects, please let me know! 😅


Opblaasmaatje:
PHPStan is going off because I removed `final` from `TestResponseHelper` which is by design so it can be extended when replacing it in the container.

PR #2209, opened 2026-07-15
fix(view): escape resolved expression attribute values

brendt:


brendt:
Thanks to @edorian for flagging this

PR #2210, opened 2026-07-15
fix(router)!: block all cross-site unsafe requests when using `PreventCrossSiteRequestsMiddleware`

innocenzi:
I initially implemented this with the idea that Tempest defaults to using LAX cookies, so sessions are not carried during cross-site POST requests.

However, it's better to special-case the routes that need support for this instead of it being allowed by default, so this PR changes that.

Marking as breaking because technically it is, payment provider flows might be affected and related routes will need to get hand-rolled protection.


brendt (approved):

brendt (approved):

brendt:
Thanks to @edorian for flagging this

PR #2218, opened 2026-07-16
docs: fix syntax in mapper

MrYamous:
While reading Mapper doc I noticed in [this part](https://github.com/tempestphp/tempest-framework/blob/580b059a0b40dc6c1ac0cc9e26a99f7396ec1faf/packages/router/src/functions.php#L23) a malformed ref to point to attribute file on Github


brendt:
Thank you!

PR #2225, opened 2026-07-17
fix(view): lexer source mapping

brendt:
I broke it, then I fixed it. Well technically Mark fixed it and I copied his solution :p


xHeaven:
I've added coverage for newline-with-indentation as well.

xHeaven:
It's okay, Windows CI is at 30% after 16 minutes. A couple of hours and it'll finish. No pressure.

brendt:
Don't you dare push a new commit

PR #2226, opened 2026-07-17
feat(core): add config installer

### LINKED ISSUE #2224: Add `install config` command
A simple installer that lists all discovered vendor config files, and allows you to select any number that will be installed into your project.

Alongside this feature, we should also document that it's best practice to install any config file you're actively relying on (especially in production envs)

brendt:
Closes #2224

### DIFF
--- a/tests/Integration/Core/ConfigInstallerTest.php
+++ b/tests/Integration/Core/ConfigInstallerTest.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Tempest\Integration\Core;
+
+use Tempest\Core\Kernel\LoadConfig;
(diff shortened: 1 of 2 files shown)


tests/Integration/Core/ConfigInstallerTest.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    protected function tearDown(): void
+    {
+        $this->installer->clean();
+
+        parent::tearDown();
+    }
+
+    public function test_it_can_install(): void
```
innocenzi:
```suggestion
    #[Test]
    public function it_can_install(): void
```

innocenzi (approved):

PR #2227, opened 2026-07-17
feat(support)!: use PHP 8.5 native URI parser

### LINKED ISSUE #1641: Use new core URI implementation
See https://thephp.foundation/blog/2025/10/10/php-85-uri-extension/

boydbloemsma:
Fixes: #1641

### DIFF
--- a/packages/http/src/IsRequest.php
+++ b/packages/http/src/IsRequest.php
@@ -6,6 +6,7 @@ namespace Tempest\Http;
 
 use Tempest\Http\Cookie\Cookie;
 use Tempest\Http\Session\Session;
+use Tempest\Support\Uri;
 use Tempest\Validation\SkipValidation;
 
 use function Tempest\Container\get;
@@ -96,21 +97,12 @@ trait IsRequest
 
     private function resolvePath(): string
     {
-        $decodedUri = rawurldecode($this->uri);
-        $parsedUrl = parse_url($decodedUri);
-
-        return $parsedUrl['path'] ?? '';
+        return rawurldecode(Uri\get_path($this->uri) ?? '');
     }
 
     private function resolveQuery(): array
     {
-        $decodedUri = rawurldecode($this->uri);
-        $parsedUrl = parse_url($decodedUri);
-        $queryString = $parsedUrl['query'] ?? '';
-
-        parse_str($queryString, $query);
-
-        return $query;
+        return Uri\get_query($this->uri);
     }
 
     public function has(string $key): bool
--- a/packages/http/src/IsRequest.php
+++ b/packages/http/src/IsRequest.php
@@ -6,7 +6,7 @@ namespace Tempest\Http;
 
 use Tempest\Http\Cookie\Cookie;
 use Tempest\Http\Session\Session;
-use Tempest\Support\Uri;
+use Tempest\Support\Uri\Uri;
 use Tempest\Validation\SkipValidation;
 
 use function Tempest\Container\get;
@@ -65,8 +65,9 @@ trait IsRequest
         $this->files = $files;
         $this->raw = $raw;
 
-        $this->path ??= $this->resolvePath();
-        $this->query ??= $this->resolveQuery();
+        $uri = Uri::from($this->uri);
+        $this->path ??= rawurldecode($uri->path ?? '');
+        $this->query ??= $uri->query;
     }
 
     public function get(string $key, mixed $default = null): mixed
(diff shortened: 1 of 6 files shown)


packages/support/src/Uri/Uri.php
```diff
@@ -67,21 +66,47 @@ public function __construct(
...
+            return new self(
+                user: $url->getUsername(),
+                password: $url->getPassword() === '' ? null : $url->getPassword(),
+                host: $url->getAsciiHost(),
+                port: $url->getPort(),
+                path: $url->getPath() ?: null,
+                queryString: $url->getQuery(),
+                fragment: $url->getFragment(),
+            );
+        }
```
xHeaven: Using `http://localhost` as the base changes the semantics of scheme-relative URLs. For `//example.com:80/path`, WHATWG treats `80` as HTTP's default and `getPort()` returns `null`, so the URI serializes as `//example.com/path`. If a caller later applies `https`, it silently changes the destination from port 80 to 443.

Could we preserve the explicit port for scheme-relative inputs while still using the HTTP parse for WHATWG/IDNA host normalization? A regression test for `//example.com:80/path` (and ideally `:443` plus a Unicode host) would cover this.

packages/http/src/IsRequest.php
```diff
@@ -96,21 +97,12 @@ public function getCookie(string $name): ?Cookie
...
     private function resolveQuery(): array
     {
-        $decodedUri = rawurldecode($this->uri);
-        $parsedUrl = parse_url($decodedUri);
-        $queryString = $parsedUrl['query'] ?? '';
-
-        parse_str($queryString, $query);
-
-        return $query;
+        return Uri\get_query($this->uri);
```
xHeaven: `resolvePath()` and `resolveQuery()` each call a helper that constructs `Uri::from($this->uri)`. For normal request targets, that means the same URI is parsed twice; in this implementation each relative URI can also trigger a failed no-base parse followed by a base parse.

My local comparison confirms the PR's benchmark signal too: isolated route matching regresses 90–160%, while full router dispatch regresses 5–10%. Could we parse once in the constructor and derive both `path` and `query` from the same `Uri` instance? Skipping the guaranteed no-base parse for `/`, `?`, `#`, and `*` targets also restores most of the lost throughput.

xHeaven (changes requested):

packages/support/src/Uri/Uri.php
boydbloemsma: Thank you for the feedback, you are right that this was an oversight.

packages/http/src/IsRequest.php
boydbloemsma: Completely valid, my latest commit will have fixed the perf issues.

packages/http/src/IsRequest.php
```diff
@@ -64,8 +65,9 @@ public function __construct(
         $this->files = $files;
         $this->raw = $raw;
 
-        $this->path ??= $this->resolvePath();
-        $this->query ??= $this->resolveQuery();
+        $uri = Uri::from($this->uri);
```
xHeaven, line 75:
```suggestion
        if ($this->method === Method::CONNECT) {
            $this->path ??= '';
            $this->query ??= [];

            return;
        }

        $uri = Uri::from($this->uri);
```

`CONNECT` uses authority-form (`host:port`), so parsing it as a WHATWG URL produces incorrect and inconsistent paths (`example.com:443` → `443`, while `[::1]:443` remains unchanged). Could we skip URI parsing for `Method::CONNECT`, leave `path` empty and `query` unset, and add hostname/IPv6 regression tests?

xHeaven (changes requested):
Other than this, LGTM! The benchmark numbers are expected (and much better now), `parse_url` is more performant in general.

brendt:
I'm trying to decide whether this is breaking or not. We now follow the spec that "an empty path" still needs a `/`, and some of the existing tests were adjusted accordingly. 

Since we only use this URI implementation within the context of requests, my feeling is that this change doesn't have any breaking impact, because the browser will be sending the correct URIs anyway. 

Anything I'm missing?

xHeaven:
> I'm trying to decide whether this is breaking or not. We now follow the spec that "an empty path" still needs a `/`, and some of the existing tests were adjusted accordingly.
> 
> Since we only use this URI implementation within the context of requests, my feeling is that this change doesn't have any breaking impact, because the browser will be sending the correct URIs anyway.
> 
> Anything I'm missing?

From the `Request` standpoint, it is not breaking. From the `Uri` API's standpoint, it's an observable change:
```php
Uri::from('https://example.com')->path; // null → '/'
(string) Uri::from('https://example.com'); // 'https://example.com' → 'https://example.com/'
```
Callers may rely on the exact string for comparisons, cache keys, signatures, etc. The new parser also changes more than the empty path, because it normalizes hosts and IDNs, strips default ports, and resolves dot segments.

However, as we discussed before, the current behavior is more like a bug. I personally would be fine shipping this as a 3.x bugfix, but I'd mark it as breaking, just for communication's sake.

brendt:
> but I'd mark it as breaking, just for communication's sake.

Agree

xHeaven (approved):
LGTM!

brendt:
Merged! Thanks @boydbloemsma , really great PR :)

PR #2234, opened 2026-07-21
ci: fix PHPStan not using result cache

staabm:
~~CI debug first, why it doesn't work.. than we will see how a fix looks like~~

ok, findings are
- we had a bug in PHPStan result cache handling, when projects had composer.lock on gitignore, which we fixed with https://github.com/phpstan/phpstan/issues/14984
- CI should build the cache on the 3.x branch, so new PRs immediately benefit from a warm cache
- updating to latest PHPStan unlocks new performance level (cold cache takes 10 seconds on my local machine to scan ~2600 files; previosly PHPStan 2.1.40 took 20s)
- after updating PHPStan, I fixed upcoming new errors and cleaned up no longer necessary ignore rules

---

relevant changes are in 
- `.github/workflows/coding-conventions.yml`
- `composer.json`

everything else is handling of PHPStan errors to get the build green

### DIFF
--- a/.github/workflows/coding-conventions.yml
+++ b/.github/workflows/coding-conventions.yml
@@ -1,6 +1,8 @@
 name: Coding Conventions
 
 on:
+  push:
+    branches: [3.x]
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/coding-conventions.yml
+++ b/.github/workflows/coding-conventions.yml
@@ -63,7 +63,7 @@ jobs:
             phpstan-result-cache-
 
       - name: Run PHPStan
-        run: vendor/bin/phpstan --error-format=github
+        run: vendor/bin/phpstan --error-format=github -vv
 
       - name: Save PHPStan result cache
         uses: actions/cache/save@v4
--- a/.github/workflows/coding-conventions.yml
+++ b/.github/workflows/coding-conventions.yml
@@ -63,7 +63,7 @@ jobs:
             phpstan-result-cache-
 
       - name: Run PHPStan
-        run: vendor/bin/phpstan --error-format=github -vv
+        run: vendor/bin/phpstan --error-format=github -vv 
 
       - name: Save PHPStan result cache
         uses: actions/cache/save@v4
--- a/packages/container/src/GenericContainer.php
+++ b/packages/container/src/GenericContainer.php
@@ -234,7 +234,7 @@ final class GenericContainer implements Container
             return $this->invokeClosure($method, ...$params);
         }
 
-        if (is_array($method) && count($method) === 2) {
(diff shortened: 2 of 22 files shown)


.github/workflows/coding-conventions.yml
```diff
@@ -1,6 +1,8 @@
 name: Coding Conventions
 
 on:
+  push:
+    branches: [3.x]
```
staabm, line 5: build the cache on the target branch, so new PRs immediately benefit from a warm cache

packages/container/src/GenericContainer.php
```diff
@@ -234,7 +234,7 @@ public function invoke(ClassReflector|MethodReflector|FunctionReflector|callable
             return $this->invokeClosure($method, ...$params);
         }
 
-        if (is_array($method) && count($method) === 2) {
+        if (is_array($method)) {
```
staabm, line 237: since the native type is `callable`, PHPStan knows when it is a array, it must have a count of 2

brendt:
Thanks for looking into it!

So should I still merge this branch? I reckon I can since there are some code fixes in it as well?

staabm:
the branch is ready to merge, indeed.

I will watch PRs coming in after the merge to make sure it works now as expected

brendt:
@xHeaven could the failing tests here have something to do with your refactor in #2238 ?

staabm:
I think we also need to update rector for compat with the latest PHPStan release.

rector depends on non API internals of PHPStan, which usually means it easily breaks right after a PHPStan release, see https://github.com/rectorphp/rector-src/pull/8208

xHeaven:
> @xHeaven could the failing tests here have something to do with your refactor in #2238 ?

I don't think so, or at least I don't see anything related. It's likely what @staabm mentioned above.

@staabm If @brendt gives his blessing, I think you can upgrade Rector in this same PR, as it's sort of related.

rector.php
```diff
@@ -54,6 +55,7 @@
         ArrayKeyExistsNullToEmptyStringRector::class,
         StringClassNameToClassConstantRector::class,
         ReturnBinaryOrToEarlyReturnRector::class,
+        SafeDeclareStrictTypesRector::class,
```
staabm, line 58: added a new rector skip rule here, as it would affect a lot of files and create a lot unnecessary noise in this PR.

in case you want this "declare-strict" types, you can re-enable the rule in a followup PR and see whether you like the results.

staabm:
(fixed the rector update from my company laptop, therefore commited with a different github.com account)

brendt (approved):

brendt:
Thanks! Appreciate it :D

PR #2237, opened 2026-07-24
fix(console): interactive components taking terminal height into account

brendt:


brendt:
This was a test but didn't turn out to fix the issue

Issue #2241, opened 2026-07-27
Replace ´tempest/core´ with Mapper’s direct dependencies


steffans:
### Description

I noticed `tempest/mapper` does not use any Tempest\Core\* APIs. Its `tempest/core` requirement only provides access to Container, Discovery, Reflection and Support. Could the `tempest/core` dependency be removed and the packages required directly instead. This would also be great for the `tempest/database` package which requires `tempest/mapper` and therefore would be independent of the `tempest/core`. It also avoids installing Core's additional dependencies like `vlucas/phpdotenv`, `symfony/cache` and `filp/woops`. 

Thanks, for considering. I like the composability of Tempest, so i can choose what packages i really need.

brendt:
For sure! Feel free to send a PR :)

Bapawe:
@brendt issue can be closed, the PR has been merged.

PR #2242, opened 2026-07-28
refactor(mapper): replace tempest/core with direct package dependencies

steffans:
PR for #2241, im not sure about including the `tempest/discovery` package, since it's optional. The `tempest/container` provides Discovery, but also does not include it in its dependencies.


brendt:
Thanks, sorry that it took a bit longer

PR #2244, opened 2026-08-01
feat(http): client IP address behind trusted proxies

osbre:
Right now, getting the IP address a request came from means digging into `$_SERVER` yourself. The PSR request has it in `REMOTE_ADDR`, but the mapper to `GenericRequest` throws the server params away.

This PR adds `Request::$ip`:

```php
#[Get(uri: '/aircraft')]
public function index(Request $request): View
{
    $ip = $request->ip;
}
```

It's `null` when the server doesn't report an address. In tests, `fromIp` sets it:

```php
$this->http
    ->fromIp('203.0.113.9')
    ->get('/aircraft')
    ->assertOk();
```

## Trusted proxies

Behind a reverse proxy you get the proxy's address, and the client's real one sits in a header like `X-Forwarded-For`. Anyone can set that header, so it's only read if the request came through a proxy you've declared. Nothing is trusted by default:

```php app/trusted-proxies.config.php
use Tempest\Http\Ip\TrustedProxiesConfig;

return new TrustedProxiesConfig(
    proxies: ['10.0.0.0/8'], // or PRIVATE_RANGES, or ANY
);
```

Addresses and CIDR ranges both work, IPv4 and IPv6. In a chain of hops, the closest one that isn't a trusted proxy is the client. Header names are configurable too, for the likes of Cloudflare. Docs included, under routing.

## Notes

- `ip` is now a reserved request parameter name, since it's a property on the `Request` interface. If a payload has a field called `ip` and gets mapped onto a custom request object, it'll throw `RequestParametersIncludedReservedNames`, same as it already does for `path`, `query` and so on. Probably worth a line in the release notes.
- Went with `ip` rather than `clientIp` to keep it short, like Laravel's `$request->ip()`.
- The range matching lives in `Tempest\Support\Ip` rather than the HTTP package, since it seemed generally useful: `matches()`, `matches_any()`, `is_private()`.

### Included changes

- `Request::$ip` on the interface and in `IsRequest`
- `TrustedProxiesConfig` and `ClientIpResolver` in `Tempest\Http\Ip`, `Tempest\Support\Ip` functions and `PRIVATE_RANGES`
- `PsrRequestToGenericRequestMapper` resolves the address, `RequestToPsrRequestMapper` writes it back for the tester's dispatch, and `RequestToObjectMapper` carries it onto custom request objects
- `HttpRouterTester::fromIp()`, across all verb methods and `makePsrRequest`
- Docs and tests

### DIFF
--- a/docs/1-essentials/01-routing.md
+++ b/docs/1-essentials/01-routing.md
@@ -466,6 +466,59 @@ final readonly class AircraftController
 }
 ```
 
+### Client IP address
+
+The address a request came from is available as the `ip` property of {b`Tempest\Http\Request`}. It is `null` when the server does not report one, such as for requests that were not made over a network.
+
+```php app/AircraftController.php
+use Tempest\Router\Get;
+use Tempest\Http\Request;
+
+final readonly class AircraftController
+{
+    #[Get(uri: '/aircraft')]
+    public function index(Request $request): View
+    {
+        $ip = $request->ip;
+    }
+}
+```
+
+Within tests, the address can be specified using the `fromIp()` method.
+
+```php
+$this->http
+    ->fromIp('203.0.113.9')
+    ->get('/aircraft')
+    ->assertOk();
+```
+
+#### Trusted proxies
+
+Forwarding headers are only read when the request came from a trusted proxy. None are trusted by default; they can be declared by creating a `trusted-proxies.config.php` file [anywhere](../1-essentials/06-configuration.md#configuration-files).
+
+```php app/trusted-proxies.config.php
+use Tempest\Http\Ip\TrustedProxiesConfig;
+
+return new TrustedProxiesConfig(
+    proxies: ['10.0.0.0/8'],
+    headers: ['x-forwarded-for'],
+);
+```
+
+Addresses and CIDR ranges are accepted, in IPv4 as well as IPv6, along with two constants for proxies that have no stable address:
+
+- `TrustedProxiesConfig::PRIVATE_RANGES` trusts the application's own network, which covers container platforms such as Docker, Railway or Fly.
+- `TrustedProxiesConfig::ANY` trusts whichever address connects, which covers a CDN such as Cloudflare.
+
+Headers are read in order of preference, each as a comma-separated chain of hops, of which the nearest one that is not itself a trusted proxy is the client.
+
+Note that only the client address is derived from a trusted proxy—the scheme and the host are not affected by this configuration. Headers that describe hops differently, such as the `{txt}Forwarded` header defined by RFC 7239, are not supported.
+
+:::warning
+Trusting a proxy that is not actually in front of the application allows any client to choose its own address, defeating anything built on top of it, such as rate limiting or allow-listing.
+:::
+
 ## Form validation
 
 When users submit forms—like updating profile settings, or posting comments—the data needs validation before processing. Tempest automatically validates request objects using type hints and validation attributes, then provides errors back to users when something is wrong.
--- a/docs/1-essentials/01-routing.md
+++ b/docs/1-essentials/01-routing.md
@@ -468,7 +468,7 @@ final readonly class AircraftController
 
 ### Client IP address
 
-The address a request came from is available as the `ip` property of {b`Tempest\Http\Request`}. It is `null` when the server does not report one, such as for requests that were not made over a network.
+The address a request came from is available as the `ip` property of {b`Tempest\Http\Request`}, an instance of {b`Tempest\Support\Ip\IpAddress`}. It is `null` when the server does not report one, such as for requests that were not made over a network.
 
 ```php app/AircraftController.php
 use Tempest\Router\Get;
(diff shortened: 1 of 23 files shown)


xHeaven:
Perhaps we could make this a bit more exhaustive if we're adding it. ~~`SERVER_ADDR`~~`REMOTE_ADDR` doesn't always contain the client's real IP address, especially behind Cloudflare, Docker and other services. What do you think?

osbre:
@xHeaven You mean `REMOTE_ADDR`, yes. Updated to include trusted proxies and relevant IP helpers under `support` package, hopefully that's not too many changes for one PR.

docs/1-essentials/01-routing.md
```diff
@@ -466,6 +466,59 @@ final readonly class AircraftController
...
+
+#### Trusted proxies
+
+Forwarding headers are only read when the request came from a trusted proxy. None are trusted by default; they can be declared by creating a `trusted-proxies.config.php` file [anywhere](../1-essentials/06-configuration.md#configuration-files).
+
+```php app/trusted-proxies.config.php
+use Tempest\Http\Ip\TrustedProxiesConfig;
+
+return new TrustedProxiesConfig(
+    proxies: ['10.0.0.0/8'],
```
brendt, line 510: Ok so, one thought I have it whether we could introduce an `IpAddress` value object that's used for the request's property and (optionally) in this config as well. 

I believe comparing IPs is more involved than a simple `===` operation on strings, so maybe it could have an `$request->ip->equals($ip)` method?

packages/support/src/Ip/functions.php
```diff
@@ -0,0 +1,114 @@
...
+ * Addresses are never matched across families, although IPv4-mapped IPv6 addresses such as `::ffff:127.0.0.1` are compared as IPv4.
+ *
+ * ### Example
+ * ```php
+ * matches('10.0.1.24', '10.0.0.0/8'); // true
+ * matches('10.0.1.24', '10.0.1.24'); // true
+ * matches('10.0.1.24', '::/0'); // false
+ * ```
+ */
+function matches(string $ip, string $range): bool
```
brendt: With this newly added value object, this function's implementation would be moved to the value object.

The function can be kept, but should be called `ip_matches`. Same for the other functions

packages/support/src/Ip/functions.php
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Determines whether the given IP address belongs to a range that is not routed on the public internet, such as a loopback or private-use address.
+ *
+ * ### Example
+ * ```php
+ * is_private('10.0.1.24'); // true
+ * is_private('203.0.113.9'); // false
+ * ```
+ */
+function is_private(string $ip): bool
```
brendt: Rename `ip_is_private`

packages/support/src/Ip/functions.php
```diff
@@ -0,0 +1,114 @@
...
+{
+    return matches_any($ip, PRIVATE_RANGES);
+}
+
+/**
+ * Converts an IP address to its packed representation, or `null` when it is not an address.
+ *
+ * @internal
+ */
+function to_bytes(string $ip): ?string
```
brendt: Rename `ip_to_bytes`

packages/support/src/Ip/functions.php
```diff
@@ -0,0 +1,114 @@
...
+
+    return ($address[$wholeBytes] & $mask) === ($subnet[$wholeBytes] & $mask);
+}
+
+/**
+ * Determines whether the given IP address falls within any of the given addresses or CIDR ranges.
+ *
+ * @param string[] $ranges
+ */
+function matches_any(string $ip, array $ranges): bool
```
brendt: rename `ip_matches_any`

brendt (changes requested):

packages/support/src/Ip/functions.php
innocenzi: FWIW, I would prefer `matches`, it would be more consistent with our other functions' naming schemes, and we recommend using `Ip\matches` too

brendt: Ah, true. Ok let's keep it as is then

packages/support/src/Ip/functions.php
```diff
@@ -0,0 +1,53 @@
...
+ * Addresses are never matched across families, although IPv4-mapped IPv6 addresses such as `::ffff:127.0.0.1` are compared as IPv4.
+ *
+ * ### Example
+ * ```php
+ * ip_matches('10.0.1.24', '10.0.0.0/8'); // true
+ * ip_matches('10.0.1.24', '10.0.1.24'); // true
+ * ip_matches('10.0.1.24', '::/0'); // false
+ * ```
+ */
+function ip_matches(IpAddress|string $ip, IpAddress|string $range): bool
```
brendt: Let's revert the function names back to without `ip_` prefix as @innocenzi suggested

brendt (changes requested):

osbre:
Accidentally did a force-push by clicking GitHub's "Update branch via Rebase" button... wasn't expecting my commit history to go up the list again 👀 makes it look as if I did new changes ugh

That said, I **renamed those functions back** as requested.

brendt (approved):

brendt:
Great feature and great PR!

PR #2250, opened 2026-08-09
feat(database): add simplePaginate() to SelectQueryBuilder

### LINKED ISSUE #2245: Add simplePaginate() to SelectQueryBuilder
### Description

Tempest currently provides length-aware pagination through `SelectQueryBuilder::paginate()`.

This implementation performs an additional count query to calculate:

* the total number of items;
* the total number of pages;
* the page range.

For large tables or complex queries, the count query may be expensive or unnecessary when the application only needs to know whether another page exists.

It would be useful to provide a simple pagination method, similar to Laravel's `simplePaginate()`, that avoids the count query.

## Proposed API

```php
$pagination = query(Book::class)->select()->simplePaginate(
    itemsPerPage: 20,
    currentPage: 1,
);
```

The implementation could request one additional record:

```php
$itemsPerPage + 1
```

The extra record would be used to determine whether the next page exists and would then be removed from the returned data.

## Proposed result type

Since simple pagination cannot provide `totalItems` or `totalPages`, it may be cleaner to return a separate result type:

```php
/**
 * @template T
 */
final class SimplePaginatedData
{
    /**
     * @param array<T> $data
     */
    public function __construct(
        public array $data,
        public int $currentPage,
        public int $itemsPerPage,
        public int $offset,
        public int $limit,
        public bool $hasNext,
        public bool $hasPrevious,
        public ?int $nextPage,
        public ?int $previousPage,
    ) {}
}
```

This would keep the existing `PaginatedData` contract strict instead of making fields such as `totalItems` and `totalPages` nullable.

## Benefits

* avoids an additional count query;
* improves performance for large or complex datasets;
* works well for APIs and infinite scrolling;
* preserves the existing length-aware pagination behavior;
* keeps simple and length-aware pagination as separate contracts.

I would be interested in implementing this as a first contribution if this direction fits Tempest's API design.

Lanser0614:
## Summary

Adds simple pagination without executing an additional `COUNT(*)` query.

```php
query(Book::class)
    ->select()
    ->simplePaginate(itemsPerPage: 20, currentPage: 2);
```

The query requests one additional item:

```sql
SELECT ...
LIMIT 21 OFFSET 20
```

If the additional item exists, it is removed from the returned data and `hasNext` is set to `true`.

## Changes

- Add `SelectQueryBuilder::simplePaginate()`.
- Add database-independent `SimplePaginator`.
- Add `SimplePaginatedData` without synthetic `totalItems` or `totalPages`.
- Expose navigation metadata such as `hasNext`, `hasPrevious`, `nextPage`, and `previousPage`.
- Support `map()`, array serialization, and JSON serialization.
- Validate `itemsPerPage` and `currentPage`.
- Ensure the public `paginate()` API never returns more than `itemsPerPage` items.
- Document out-of-range page behavior and the trade-offs of offset pagination.

## Tests

Added unit and integration coverage for:

- exactly `itemsPerPage` results;
- an additional result used to detect the next page;
- input larger than `itemsPerPage + 1`;
- empty and out-of-range pages;
- second-page offset calculation;
- invalid arguments;
- `map()`, `toArray()`, and JSON serialization;
- `SelectQueryBuilder::simplePaginate()`;
- execution of a single `SELECT` query without `COUNT(*)`.

Local verification:

- PHPUnit: 51 tests, 159 assertions
- Mago formatter and lint
- PHPStan

## Notes

This is still offset-based pagination. It avoids the potentially expensive count query, but large offsets may remain costly, and concurrent inserts or deletes can shift results between pages. Cursor/keyset pagination remains preferable for very large or frequently changing datasets.

### DIFF
--- a/packages/support/src/Paginator/SimplePaginator.php
+++ b/packages/support/src/Paginator/SimplePaginator.php
@@ -2,7 +2,8 @@
 
 namespace Tempest\Support\Paginator;
 
-use Tempest\Support\Paginator\Exceptions\ArgumentWasInvalid;
+use Tempest\Support\Paginator\Exceptions\CurrentPageWasInvalid;
+use Tempest\Support\Paginator\Exceptions\ItemsPerPageWasInvalid;
 
 final class SimplePaginator
 {
@@ -10,12 +11,18 @@ final class SimplePaginator
         private(set) int $itemsPerPage = 20,
         private(set) int $currentPage = 1,
     ) {
-        if ($this->itemsPerPage <= 0) {
-            throw new ArgumentWasInvalid('Items per page must be positive');
+        if ($this->itemsPerPage <= 0 || $this->itemsPerPage === PHP_INT_MAX) {
+            throw new ItemsPerPageWasInvalid($this->itemsPerPage);
         }
 
-        if ($this->currentPage <= 0) {
-            throw new ArgumentWasInvalid('Current page must be positive');
+        $maximumCurrentPage = min(PHP_INT_MAX - 2, intdiv(PHP_INT_MAX, $this->itemsPerPage)) + 1;
+
+        if ($this->currentPage <= 0 || $this->currentPage > $maximumCurrentPage) {
+            throw new CurrentPageWasInvalid(
+                currentPage: $this->currentPage,
+                itemsPerPage: $this->itemsPerPage,
+                maximumCurrentPage: $maximumCurrentPage,
+            );
         }
     }
 
--- a/packages/support/src/Paginator/SimplePaginator.php
+++ b/packages/support/src/Paginator/SimplePaginator.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Tempest\Support\Paginator;
+
+use Tempest\Support\Paginator\Exceptions\ArgumentWasInvalid;
+
+final class SimplePaginator
+{
+    public function __construct(
+        private(set) int $itemsPerPage = 20,
+        private(set) int $currentPage = 1,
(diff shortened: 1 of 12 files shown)


Lanser0614:
@brendt @innocenzi What is the average review time for a pull request ?

packages/support/src/Paginator/SimplePaginator.php
```diff
@@ -0,0 +1,99 @@
...
+        private(set) int $itemsPerPage = 20,
+        private(set) int $currentPage = 1,
+    ) {
+        if ($this->itemsPerPage <= 0) {
+            throw new ArgumentWasInvalid('Items per page must be positive');
+        }
+
+        if ($this->currentPage <= 0) {
+            throw new ArgumentWasInvalid('Current page must be positive');
+        }
```
xHeaven: Let's set an upper-bound as well. Though edge-case, a big enough argument (e.g. `PHP_INT_MAX`) can cause an overflow, coercing the type to float and causing a `TypeError`.

xHeaven: Let's also create named, past tense exceptions for these, see other exceptions in the codebase. The old paginator follows the currently implemented patterns, but I'd consider it legacy and move towards named exceptions. Only fix these exceptions for now, if you wish, you can open a followup PR for the original paginator exceptions. Though I'd wait for @brendt's approval with this.

xHeaven (changes requested):

xHeaven (approved):

PR #2251, opened 2026-08-10
fix(mapper): preserve native scalar values

### LINKED ISSUE #2246: Mapper serializers cannot preserve native bool and float values
### Tempest version

3.18

### PHP version

8.5

### Operating system

Linux

## Context

While building JSON API responses with Tempest's mapper, I ran into a limitation around scalar serialization.

The `Tempest\Mapper\Serializer` interface currently only allows serializers to return `array|string|int`:

```php
interface Serializer
{
    /**
     * Serializes the given input into a string, array, or integer.
     */
    public function serialize(mixed $input): array|string|int;
}
```

This means a custom serializer cannot legally return native `bool` or `float` values, even though those are valid PHP scalar values and valid JSON values.

## Question

Was this return type chosen intentionally?

I can imagine that some serializer contexts may want string output, for example when serializing values for storage or text-based formats. However, when using the mapper for `toArray()` or `toJson()`, I would expect native scalar values to be preserved:

```php
final class Example
{
    public bool $active = true;
    public float $score = 1.5;
    public int $count = 3;
}
```

Expected array output:

```php
[
    'active' => true,
    'score' => 1.5,
    'count' => 3,
]
```

Expected JSON output:

```json
{
  "active": true,
  "score": 1.5,
  "count": 3
}
```

## Current behavior

The built-in scalar serializers return strings for several scalar types:

```php
// BooleanSerializer
public function serialize(mixed $input): string
{
    return $input ? 'true' : 'false';
}

// FloatSerializer
public function serialize(mixed $input): string
{
    return (string) $input;
}

// IntegerSerializer
public function serialize(mixed $input): string
{
    return (string) $input;
}
```

For JSON API responses this can change the meaning of the output, for example:

```json
{
  "active": "true",
  "score": "1.5",
  "count": "3"
}
```

instead of:

```json
{
  "active": true,
  "score": 1.5,
  "count": 3
}
```

## Custom serializer limitation

I tried solving this with a higher-priority scalar serializer:

```php
use Tempest\Mapper\DynamicSerializer;
use Tempest\Mapper\Exceptions\ValueCouldNotBeSerialized;
use Tempest\Mapper\Serializer;
use Tempest\Reflection\PropertyReflector;
use Tempest\Reflection\TypeReflector;
use Tempest\Support\Priority;

#[Priority(Priority::HIGHEST)]
final readonly class ScalarSerializer implements Serializer, DynamicSerializer
{
    public static function accepts(PropertyReflector|TypeReflector $input): bool
    {
        $type = $input instanceof PropertyReflector
            ? $input->getType()
            : $input;

        return in_array($type->getName(), [
            'bool',
            'boolean',
            'double',
            'float',
            'int',
            'integer',
            'string',
        ], strict: true);
    }

    public function serialize(mixed $input): bool|float|int|string
    {
        if (is_bool($input) || is_float($input) || is_int($input) || is_string($input)) {
            return $input;
        }

        throw new ValueCouldNotBeSerialized('bool|float|int|string');
    }
}
```

However, this is not compatible with the current `Serializer` interface because `bool|float|int|string` is wider than `array|string|int`.

## Possible change

Would it make sense to broaden the serializer return type to include all JSON-compatible scalar values?

```php
interface Serializer
{
    /**
     * Serializes the given input into an array, boolean, float, integer, or string.
     */
    public function serialize(mixed $input): array|bool|float|int|string;
}
```

This should remain compatible with existing implementations that return narrower types such as `string`, `int`, or `array|string`.

## Alternative design question

If the current string-only behavior for `bool` and `float` is intentional, would it make sense to separate serialization contexts?

For example:

- serializers for string/text/database-oriented output,
- serializers for array/JSON-oriented output where native scalar types are preserved.

I am mainly trying to understand whether the current return type is an intentional design decision, or whether supporting `bool` and `float` in mapper serializer output would be considered a valid bug fix / improvement.

radoslav-grencik:
Fixes #2246


xHeaven:
A bit irrelevant to this PR, but it's something we're missing already: `docs/2-features/01-mapper.md:245` - can you add `float` to the supported type list?

Other than that, LGTM.

radoslav-grencik:
> A bit irrelevant to this PR, but it's something we're missing already: `docs/2-features/01-mapper.md:245` - can you add `float` to the supported type list?
> 
> Other than that, LGTM.

Done.

PR #2254, opened 2026-08-17
docs: replace outdated `--allow-dead-links` flag  with `--crawl`

davidtaim:
While reading docs I notices that --allow-dead-links was no available and the new option is --crawl, that is what I think tell if not is correct.


innocenzi:
Thank you!

PR #2255, opened 2026-08-18
fix(http): recursively map nested uploaded files

osbre:
PSR-7 exposes array-style file inputs, such as `files[]` or `files[avatar]`, as nested arrays. The mapper previously assumed a single-level array with a callback typed to `UploadedFileInterface`, so passing a nested array threw `TypeError`.

Uploaded files are now mapped recursively to preserve this nested structure, and the `$files` docblocks on `Request` and `IsRequest` have been widened to `array<array-key, Upload|array>`.

Runtime behaviour for standard file uploads is unchanged, though static analysis tools may report new notices in code that assumes `Upload[]`.

### DIFF
--- a/packages/http/src/IsRequest.php
+++ b/packages/http/src/IsRequest.php
@@ -38,7 +38,7 @@ trait IsRequest
     #[SkipValidation]
     private(set) array $query;
 
-    /** @var \Tempest\Http\Upload[] */
+    /** @var array<array-key, \Tempest\Http\Upload|array> */
     #[SkipValidation]
     private(set) array $files;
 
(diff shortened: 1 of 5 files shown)


xHeaven:
I'd prefer a test that actually exercises the reported behavior, other than that, LGTM.
Something like:
```php
#[Test]
public function nested_uploaded_files_are_mapped_to_uploads(): void
{
    $request = $this->createServerRequest('POST')->withUploadedFiles([
        // ...
    ]);

    $mappedRequest = $this->mapper->map(
        from: $request,
        to: GenericRequest::class,
    );

    // asserts
}
```

osbre:
Thanks @xHeaven - I updated the test

innocenzi (approved):

PR #2256, opened 2026-08-19
feat: allows `get` to resolve services as singletons

aidan-casey:
Based on the conversations with @crell in the Tempest discord, this PR allows services to be shared by default and adds a secondary `make` method that always retrieves a new instance of the class.

Note the definition of the `make` method that PHP-DI has:

> The make() method works like get() except it will resolve the entry every time it is called. Depending on the type of the entry this means:
>
> - if the entry is an object, an new instance will be created every time
> - if the entry is a factory, the factory will be called every time
> - if the entry is an alias, the alias will be resolved every time
>
> Please note that only the entry you ask for will be resolved every time: all the dependencies of the entry will not! That means that if the entry is an alias, the entry the alias points to will be resolved only once.

https://php-di.org/doc/container.html#make

### DIFF
--- a/packages/container/src/Container.php
+++ b/packages/container/src/Container.php
@@ -25,6 +25,13 @@ interface Container extends ContainerInterface
      * @param class-string<TClassName> $className
      * @return TClassName
      */
+    public function make(string $className, mixed ...$params): mixed;
+
+    /**
+     * @template TClassName of object
+     * @param class-string<TClassName> $className
+     * @return TClassName
+     */
     public function get(string $className, string|UnitEnum|null $tag = null, mixed ...$params): mixed;
 
     public function has(string $className, string|UnitEnum|null $tag = null): bool;
--- a/packages/container/src/GenericContainer.php
+++ b/packages/container/src/GenericContainer.php
@@ -54,6 +54,8 @@ final class GenericContainer implements Container
         private(set) ArrayIterator $resettables = new ArrayIterator(),
 
         private(set) ?DependencyChain $chain = null,
+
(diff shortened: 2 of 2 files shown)


innocenzi (commented):
Not a blocker, but I am not a fan of this:
- `get` gets a hidden behavior
- `make` does what `get` used to do
- `get` and `make` aren't clear names, they have minor differences in their behavior, so consumers have to read docs or code dive

I feel like we're unlocking a niche use case while making the container more confusing. I don't have an alternative to suggest, however.

xHeaven:
I've reviewed this PR internally on Discord, but here is the review for clarity:

> A couple things:
> 1. PR says `make()` always returns a new instance. This is true... unless you called `get()` earlier.
> ```php
> $container = new GenericContainer(
>     getClassAsSingletonByDefault: true,
> );
> 
> $get = $container->get(Mailer::class);
> make = $container->make(Mailer::class);
> 
> var_dump($get === $make); // true
> ```
> 
> 2. PR says `make()` always returns a new instance. This _could be_ true... but now you just get a (false) `CircularDependencyEncountered` exception if you call `make()` more than once on the same class because the first call leaves the class constructor in the chain, and that's what the second call sees, so it's not a true circular dependency.
> 3. Though I'm not sure if this was in the scope, but with the current method, constructor dependencies are still not shared. So `$mailer->platform === $platform` is a `false`, even if they are the "same-ish" objects

brendt:
Closing because we'll discuss it here: https://github.com/tempestphp/tempest-framework/issues/2263

PR #2259, opened 2026-08-21
fix(mapper): serialize nested objects

### LINKED ISSUE #2258: ObjectToArrayMapper does not serialize single nested object properties
### Tempest version

3.18

### PHP version

8.5

### Operating system

Linux

### Description

When mapping an object to an array or JSON, `ObjectToArrayMapper` serializes
top-level scalar/date properties correctly, and it also serializes arrays of
objects correctly. However, a single nested object property is left as an object
instead of being recursively mapped to an array.

This becomes visible when the nested object contains a `DateTime` property:
top-level dates are serialized using the configured date serializer, but dates
inside the single nested object are later JSON-encoded as the internal public
structure of the `DateTime` object.

## Current behavior

Given an object like this:

```php
use Tempest\DateTime\DateTime;
use Tempest\DateTime\FormatPattern;
use Tempest\Validation\Rules\HasDateTimeFormat;

final class ChildObject
{
    public function __construct(
        #[HasDateTimeFormat(FormatPattern::ISO8601)]
        public DateTime $createdAt,
    ) {}
}

final class ParentObject
{
    public function __construct(
        #[HasDateTimeFormat(FormatPattern::ISO8601)]
        public DateTime $createdAt,
        public ChildObject $child,
        /** @var ChildObject[] */
        public array $children,
    ) {}
}
```

And this mapper call:

```php
$date = DateTime::parse('2026-08-19T12:34:56+00:00');

$array = map(new ParentObject(
    createdAt: $date,
    child: new ChildObject($date),
    children: [new ChildObject($date)],
))->toArray();
```

The resulting array serializes the top-level date and the array-of-objects date,
but leaves the single nested `child` object unmapped:

```php
[
    'createdAt' => '2026-08-19T12:34:56.000Z',
    'child' => ChildObject { ... },
    'children' => [
        ['createdAt' => '2026-08-19T12:34:56.000Z'],
    ],
]
```

When using `toJson()`, the nested date is then encoded as the public structure of
the `Tempest\DateTime\DateTime` object, instead of as the configured date string:

```json
{
  "createdAt": "2026-08-19T12:34:56.000Z",
  "child": {
    "createdAt": {
      "timezone": "UTC",
      "timestamp": {
        "seconds": 1787142896,
        "nanoseconds": 0
      },
      "year": 2026,
      "month": 8,
      "day": 19,
      "hours": 12,
      "minutes": 34,
      "seconds": 56,
      "nanoseconds": 0
    }
  },
  "children": [
    {
      "createdAt": "2026-08-19T12:34:56.000Z"
    }
  ]
}
```

## Expected behavior

Single nested object properties should be recursively mapped in the same way as
objects inside array properties.

Expected `toArray()` output:

```php
[
    'createdAt' => '2026-08-19T12:34:56.000Z',
    'child' => [
        'createdAt' => '2026-08-19T12:34:56.000Z',
    ],
    'children' => [
        ['createdAt' => '2026-08-19T12:34:56.000Z'],
    ],
]
```

Expected `toJson()` output should contain the nested date as a string as well:

```json
{
  "createdAt": "2026-08-19T12:34:56.000Z",
  "child": {
    "createdAt": "2026-08-19T12:34:56.000Z"
  },
  "children": [
    {
      "createdAt": "2026-08-19T12:34:56.000Z"
    }
  ]
}
```

## Possible cause

`ObjectToArrayMapper::resolvePropertyValue()` currently handles iterable object
properties recursively:

```php
if ($property->getIterableType()?->isClass()) {
    foreach ($propertyValue as $key => $value) {
        if (! is_object($value)) {
            continue;
        }

        $propertyValue[$key] = map($value)->toArray();
    }

    return $propertyValue;
}
```

But single object properties do not seem to go through an equivalent recursive
object-to-array mapping step. If no serializer is selected for that property,
the object is returned as-is.

## Suggested regression tests

This could be covered with tests in `ObjectToArrayMapperTest` and
`ObjectToJsonMapperTest`:

```php
#[Test]
public function object_with_single_nested_object_to_array(): void
{
    $date = DateTime::parse('2026-08-19T12:34:56+00:00');

    $array = map(new ParentObject(
        createdAt: $date,
        child: new ChildObject($date),
        children: [new ChildObject($date)],
    ))->toArray();

    $this->assertSame(
        [
            'createdAt' => '2026-08-19T12:34:56.000Z',
            'child' => [
                'createdAt' => '2026-08-19T12:34:56.000Z',
            ],
            'children' => [
                ['createdAt' => '2026-08-19T12:34:56.000Z'],
            ],
        ],
        $array,
    );
}
```

If this behavior is considered unintended, I can send a focused PR with regression tests for `ObjectToArrayMapper` and `ObjectToJsonMapper`.

radoslav-grencik:
Fixes #2258

### DIFF
--- a/packages/mapper/src/Mappers/ObjectToArrayMapper.php
+++ b/packages/mapper/src/Mappers/ObjectToArrayMapper.php
@@ -14,8 +14,6 @@ use Tempest\Mapper\SerializerFactory;
 use Tempest\Reflection\ClassReflector;
 use Tempest\Reflection\PropertyReflector;
 
-use function Tempest\Mapper\map;
-
 final readonly class ObjectToArrayMapper implements Mapper
 {
     public function __construct(
@@ -30,32 +28,63 @@ final readonly class ObjectToArrayMapper implements Mapper
 
     public function map(mixed $from, mixed $to): mixed
     {
-        if ($from instanceof JsonSerializable) {
-            return $from->jsonSerialize();
+        $visited = [];
+
+        return $this->mapValue($from, $visited);
+    }
+
+    /**
+     * @param array<int, true> $visited
+     */
+    private function mapValue(mixed $value, array &$visited): mixed
+    {
+        if ($value instanceof JsonSerializable) {
+            return $value->jsonSerialize();
         }
 
-        if (is_object($from)) {
-            $class = new ClassReflector($from);
+        if (! is_object($value)) {
+            return $value;
+        }
 
-            $mappedProperties = [];
+        return $this->mapObject($value, $visited);
+    }
 
-            foreach ($class->getPublicProperties() as $property) {
-                if ($property->hasAttribute(Hidden::class)) {
-                    continue;
-                }
+    /**
+     * @param array<int, true> $visited
+     */
+    private function mapObject(object $object, array &$visited): mixed
+    {
+        $objectId = spl_object_id($object);
 
-                $propertyName = $this->resolvePropertyName($property);
-                $propertyValue = $this->resolvePropertyValue($property, $from);
-                $mappedProperties[$propertyName] = $propertyValue;
+        if (isset($visited[$objectId])) {
+            return $object;
+        }
+
+        $visited[$objectId] = true;
+
+        $class = new ClassReflector($object);
+
+        $mappedProperties = [];
+
+        foreach ($class->getPublicProperties() as $property) {
+            if ($property->hasAttribute(Hidden::class)) {
+                continue;
             }
-        } else {
-            $mappedProperties = $from;
+
+            $propertyName = $this->resolvePropertyName($property);
+            $propertyValue = $this->resolvePropertyValue($property, $object, $visited);
+            $mappedProperties[$propertyName] = $propertyValue;
         }
+
+        unset($visited[$objectId]);
 
         return $mappedProperties;
     }
 
-    private function resolvePropertyValue(PropertyReflector $property, object $object): mixed
+    /**
+     * @param array<int, true> $visited
+     */
+    private function resolvePropertyValue(PropertyReflector $property, object $object, array &$visited): mixed
     {
         if (! $property->isInitialized($object)) {
             return null;
@@ -69,9 +98,7 @@ final readonly class ObjectToArrayMapper implements Mapper
                     continue;
                 }
 
-                $propertyValue[$key] = map($value)
-                    ->in($this->context)
-                    ->toArray();
+                $propertyValue[$key] = $this->mapValue($value, $visited);
             }
 
             return $propertyValue;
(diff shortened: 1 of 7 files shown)


xHeaven:
This seems to be a bit problematic right now. You're introducing an infinite recursion (nested objects reference each other) and also a performance regression because you are resolving all configured mappers again for every nested object.

The goal idea is good, but I'm afraid this should be a bit deeper refactor than 2 lines of change. I'll try to think of something, let's also wait for @brendt here.

Edit: added a couple failing tests for clarity.

radoslav-grencik:
> This seems to be a bit problematic right now. You're introducing an infinite recursion (nested objects reference each other) and also a performance regression because you are resolving all configured mappers again for every nested object.
> 
> The goal idea is good, but I'm afraid this should be a bit deeper refactor than 2 lines of change. I'll try to think of something, let's also wait for @brendt here.
> 
> Edit: added a couple failing tests for clarity.

Thanks, that makes sense. I reworked the fix so nested object serialization no longer calls the public mapper pipeline recursively.

The updated approach keeps recursion inside `ObjectToArrayMapper` and tracks visited objects with `spl_object_id()`. This avoids infinite recursion for cyclic graphs; cyclic objects are left for JSON encoding to fail normally instead of hanging.

It also avoids the performance regression from repeatedly resolving configured mappers for every nested object. The new test for unused mapper resolution now passes.

I also updated `ArrayOfObjectsSerializer` to preserve mapping context when it is used directly.

brendt:
I'm can't remember if this was ever intended behavior, but I agree that it should be changed. However, I do wonder if this could be a potential breaking change, because we're suddenly changing how the mapper works… I for example remember manually mapping nested objects in some of my project exactly because of this bug. I don't believe my code would break, but I wonder if others might…

On top of that, I don't see an automated upgrade path for this change; all of that leads me to think we should target 4.x with this one

radoslav-grencik:
> I'm can't remember if this was ever intended behavior, but I agree that it should be changed. However, I do wonder if this could be a potential breaking change, because we're suddenly changing how the mapper works… I for example remember manually mapping nested objects in some of my project exactly because of this bug. I don't believe my code would break, but I wonder if others might…
> 
> On top of that, I don't see an automated upgrade path for this change; all of that leads me to think we should target 4.x with this one

That makes sense, targeting 4.x sounds reasonable.

Is there already a plan or timeline for 4.x? I don’t see a `4.x` branch yet, but I’d be happy to help.

brendt:
> Is there already a plan or timeline for 4.x? I don’t see a 4.x branch yet, but I’d be happy to help.

There is a branch now :) 

We have a milestone for 4.0, but we haven't started with it. I'll mostly be a matter of "tagging when we're done".  (https://github.com/tempestphp/tempest-framework/milestone/24)

Realistically, I think we may need to wait until October, because I have some things planned for September where I suspect more people will try out the new worker mode support. If anything weird comes out of that which requires breaking changes, I'd like to include those in 4.0

PR #2267, opened 2026-08-29
feat(router): allow error responses to carry headers

osbre:
## Problem

Returning response with headers works from a controller:

```php
return new Json(body: ['message' => 'Slow down.'], status: Status::TOO_MANY_REQUESTS)
    ->addHeader('Retry-After', '30');
```

But when using `throw HttpRequestFailed` to return an error response from a different place, there is no way of passing headers, which leads to a few statuses that can only be sent correctly from inside a controller:

- `429` needs `Retry-After`, otherwise the client has no idea when to come back
- `401` needs `WWW-Authenticate` as per RFC 9110; the response is malformed without it
- `405` needs `Allow`
- (same for custom headers such as `X-RateLimit-*`)

## Solution

`HttpRequestFailed` gains a `headers` parameter, and the handler applies it to whatever the renderer returned:

```php
throw new HttpRequestFailed(
    status: Status::TOO_MANY_REQUESTS,
    headers: ['retry-after' => '60'],
);
```

Values may be a string or a list of strings, so multi-value headers work:

```php
throw new HttpRequestFailed(
    status: Status::METHOD_NOT_ALLOWED,
    headers: ['allow' => ['GET', 'HEAD']],
);
```

Declared headers replace whatever the renderer set rather than adding to it: they describe the failure, which outranks anything assembled for the body. The lookup is case-insensitive, so a header the renderer spelled `Retry-After` and a throw site spelling it `retry-after` resolve to one header rather than two.

This mirrors Symfony's `HttpExceptionInterface::getHeaders()`, which Laravel reuses and also exposes as `abort($status, $message, $headers)`.

### DIFF
--- a/packages/http/src/HttpRequestFailed.php
+++ b/packages/http/src/HttpRequestFailed.php
@@ -15,12 +15,14 @@ final class HttpRequestFailed extends Exception implements ProvidesContext
      * @param string|null $message An optional message that will be displayed to the client.
      * @param Response|null $cause The response that caused the failure, if any.
      * @param Request|null $request The request that failed, for debug purposes.
+     * @param array<string, string|string[]> $headers Headers that the response must carry, such as `retry-after` on a `429`.
      */
     public function __construct(
         private(set) readonly Status $status,
         ?string $message = null,
         private(set) readonly ?Response $cause = null,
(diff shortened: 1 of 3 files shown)


brendt:
Great, thanks!

PR #2272, opened 2026-09-04
feat(rate-limit): add rate limiting

osbre:
This PR adds the `tempest/rate-limit` package. It limits how often an action may happen within a window of time, using a fixed-window algorithm. The package is usable both from HTTP routes and from application code that needs to throttle an arbitrary operation.

Counters use the cache by default, so the package works without extra infrastructure. Applications serving concurrent traffic can switch to Redis, which increments counters atomically.

### `#[Throttle]`

Limits how often a route may be requested. The attribute is repeatable and can be applied to a controller method or to the controller itself.

```php
#[Throttle(attempts: 20)]
#[Throttle(attempts: 1000, per: Per::DAY)]
#[Get('/api/posts')]
public function index(): Response { /* … */ }
```

- The window defaults to one minute. `per` and `every` widen it, as in `per: Per::MINUTE, every: 5`.
- Every route gets its own counter, and every client gets its own counter. Naming a `bucket` makes routes sharing that bucket use one allowance per client.
- Limits with different windows use separate counters. When several limits describe the same bucket and window, the smallest allowance applies.
- On a controller, one allowance covers every route it exposes. Method-level limits apply on top rather than replacing it.
- Multiple limits are evaluated from the shortest window to the longest. A rejected request does not consume later quotas.
- Allowed responses carry `X-RateLimit-Limit`, `X-RateLimit-Remaining`, and `X-RateLimit-Reset`. Rejected ones get a `429` with `Retry-After`. Headers can be disabled with `includeHeaders: false`.

The attribute also accepts a `profile` for limits that depend on the request:

```php
#[Throttle(profile: ApiRateLimitProfile::class)]
```

`RateLimitProfile::resolve(Request): RateLimit[]` may return several limits, or an empty array to leave the request unlimited. Unkeyed profile limits use the route or controller scope and are counted per client. Keyed limits created with `withKey()` use the key exactly as provided, so the same counter can be shared or inspected through the generic limiter.

Static limits and profiles are represented by the same `Throttle` attribute. Invalid combinations are rejected when the attribute is constructed, rather than being accepted and failing later in middleware.

### `RateLimitKeyResolver`

Decides what identifies a client. It defaults to `ClientIPKeyResolver`, which normalizes IPv4 and IPv6-mapped addresses. Implement the interface and configure `keyResolverClass` to count by authenticated user, API key, or anything else.

Returning `null` puts the request in a shared unidentified bucket, so an unidentifiable client never escapes its limit. Applications behind a reverse proxy must configure trusted proxies before relying on client IP limits.

### Failure handling

The cache backend can fail to acquire the lock for a counter under contention. In that case the request is rejected with `503 Service Unavailable` rather than guessing whether it should be allowed or throttled. The error is surfaced through the normal HTTP exception rendering path, including the response headers.

## `RateLimiter`

The limiter is not tied to HTTP. Inject `RateLimiter` and count attempts against any key:

```php
$limit = RateLimit::perHour(3)->withKey("verification-email:{$user->id}");

if ($this->limiter->attempt($limit)->exceeded) {
    return;
}
```

- Limits are built with `RateLimit::perSecond()`, `perMinute()`, `perHour()`, or `perDay()`, each taking an optional multiplier.
- `withKey()` scopes a limit, and `scopedTo()` appends to an existing key.
- `attempt()` records an attempt, or several with `by:`, and returns a `RateLimitResult` exposing `allowed`, `exceeded`, `limit`, `hits`, `remaining`, `retryAfter`, and `resetsAt`.
- `peek()` inspects without consuming, and `clear()` discards recorded attempts.
- `throttle()` runs a callback only when the limit allows it, throwing `RateLimitWasExceeded` otherwise.
- A limit must carry a key by the time it reaches the limiter. An unkeyed limit throws `RateLimitKeyWasMissing` rather than sharing a counter with every other unkeyed limit.

`RateLimitWasExceeded` is the caller's to handle. It is not converted into a `429` by a global middleware: code using the generic limiter has opted into handling that operation's rejection itself. Callers can catch the exception or use `attempt()` and branch on the result.

## Configuration and storage

`CacheRateLimitConfig` is the default discovered configuration. `RedisRateLimitConfig` switches the package to `RedisRateLimitStorage`:

```php
// app/rateLimit.config.php
return new RedisRateLimitConfig();
```

`RateLimitConfig` controls the storage backend, key prefix, HTTP headers, and key resolver. Custom backends can implement `RateLimitConfig::createStorage()` and `RateLimitStorage`.

- **`CacheRateLimitStorage`** stores a `RateLimitState` in the configured cache. Read-modify-write is serialized with locks, but the counter is only as durable as the cache.
- **`RedisRateLimitStorage`** performs increment-and-open-window in one Lua script. It counts correctly under concurrency and needs no lock.
- Both backends derive their storage key through the configured prefix and hash arbitrary application keys with `xxh128`.

## Testing

`RateLimitTester` is available on `IntegrationTest` as `$this->rateLimit`. `fake()` swaps in isolated in-memory storage, so tests need neither Redis nor a configured cache, and counters cannot leak between tests.

```php
$this->rateLimit->fake();

$this->rateLimit
    ->hit($limit, times: 2)
    ->assertHits($limit, 2)
    ->assertRemaining($limit, 1)
    ->assertNotThrottled($limit);
```

`exhaust()` and `assertThrottled()` cover rejected attempts. `preventThrottling()` temporarily allows every attempt without recording it, which is useful when testing unrelated behavior. HTTP tests can exercise throttled routes through simulated requests, while keyed profile limits can be asserted directly through `RateLimiter`.

### Fixed windows

A window opens on the first attempt, and its end is fixed at that moment. Later attempts increment the counter but never push the end out, so hammering a limit cannot extend it. Attempts are recorded before evaluating the result, so a rejected attempt cannot nudge the window.

The known trade-off is the boundary burst. Under `perMinute(60)`, 60 requests at `11:59:59` and 60 more at `12:00:00` are both legal. Sliding windows and token buckets are out of scope.

### One storage contract

`RateLimiter` is the API surface to inject and call, while `RateLimitStorage` handles persistence. Keeping them separate means storage operations never leak into the public limiter API, and no storage backend is baked into application code. `RateLimitState` persists hits and the reset timestamp together, so a counter never exists without its window.

### HTTP counter scoping

`Throttle` adds `ThrottleMiddleware` to decorated routes during discovery. The middleware resolves method- and controller-level attributes, derives a counter key from the declaration scope, allowance, window, bucket, and client, collapses duplicate limits, and applies the narrowest limits first. Named buckets are shared per client; keyed profile limits are the explicit escape hatch for counters shared across application code.


packages/rate-limit/src/Config/rateLimit.config.php
innocenzi, line 1: Let's name it `rate-limit.config.php`

packages/rate-limit/src/Http/ThrottleWith.php
```diff
@@ -0,0 +1,40 @@
...
+ * authenticated clients get more of it.
+ *
+ * ```php
+ * #[ThrottleWith(ApiRateLimitProfile::class)]
+ * #[Get('/api/posts')]
+ * public function index(): Response { /* … *\/ }
+ * ```
+ */
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD | Attribute::TARGET_CLASS)]
+final readonly class ThrottleWith implements Throttles
```
innocenzi, line 23: Why not just add a `profile` property to `Throttle`? Would avoid to remember what both attributes do, and would avoid the awkward trait as well.

packages/rate-limit/src/Http/ClientIpKeyResolver.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Tempest\RateLimit\Http;
+
+use Tempest\Http\Request;
+
+/**
+ * Counts requests by client IP. Requires {@see \Tempest\Http\Ip\TrustedProxiesConfig}
+ * for reliable client IPs behind proxies.
+ */
+final readonly class ClientIpKeyResolver implements RateLimitKeyResolver
```
aidan-casey: Can you update the acronym casing to conform to [our guidelines](https://tempestphp.com/3.x/extra-topics/contributing#code-style-and-conventions)?

packages/rate-limit/src/Config/RateLimitConfig.php
```diff
@@ -0,0 +1,58 @@
...
+        public bool $includeHeaders = true,
+
+        /**
+         * Storage for rate limit counters. The default works anywhere a cache is configured, but takes a
+         * lock on every increment. {@see \Tempest\RateLimit\Storage\RedisRateLimitStorage} counts
+         * atomically and is recommended in production.
+         *
+         * @var class-string<RateLimitStorage>
+         */
+        public string $storageClass = CacheRateLimitStorage::class,
```
innocenzi: I'd rather have an interface for the config, and add a dedicated `CacheRateLimitConfig` implementation. We use that pattern in many places.

innocenzi (changes requested):
Looks pretty good overall, I like the API. Not a full review, just a few nitpicks.

packages/rate-limit/src/Http/ThrottleMiddleware.php
```diff
@@ -0,0 +1,139 @@
...
+        private RateLimiter $limiter,
+        private RateLimitKeyResolver $keyResolver,
+        private RateLimitConfig $config,
+        private MatchedRoute $matchedRoute,
+        private Container $container,
+    ) {}
+
+    public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
+    {
+        if (! $this->config->enabled) {
```
aidan-casey: What's the advantage of having this vs. just not adding the middleware?

aidan-casey:
Nice work here! Like @innocenzi, I like the API.

More of a side note, I am expecting we may have to consider potential edge cases with worker mode. There's a chance it may not be an issue given the storage drivers, but just making a mental note.

packages/rate-limit/src/Http/ClientIpKeyResolver.php
osbre: According to this guideline on acronym casing, was https://github.com/tempestphp/tempest-framework/blob/3.x/packages/support/src/Ip/IpAddress.php from https://github.com/tempestphp/tempest-framework/pull/2244 also meant to be named `IPAddress`? 👀 I suppose now we need to rename it as well, and put it out as a breaking change with a Rector rule for easy upgrade, or maybe that's for 4.x?

packages/rate-limit/src/Config/rateLimit.config.php
osbre: I'm gussing the current rule goes that framework-owned configuration files are `camelCase`, whereas user-made ones are `kebab-case`?

- https://github.com/tempestphp/tempest-framework/blob/3.x/packages/command-bus/src/Config/commandBus.config.php
- https://github.com/tempestphp/tempest-framework/blob/3.x/packages/event-bus/src/Config/eventBus.config.php
- https://github.com/tempestphp/tempest-framework/blob/3.x/packages/router/src/Config/staticRoutes.config.php

Or we want to eventually move the framework-owned configs to be `kebab-case` as well?

packages/rate-limit/src/Http/ThrottleWith.php
osbre: That was my initial idea. I split it because merging makes `attempts` nullable and needs a runtime check for the invalid combinations (neither set, both set, or `profile` combined with `per`/`every`/`bucket`) plus an exception type for it - whereas two attributes make those states unrepresentable by splitting responsibilities.

packages/rate-limit/src/Config/RateLimitConfig.php
osbre: Brilliant, thank you! Now we can also move `lockTimeoutInSeconds` into a Cache-specific config 👍

packages/rate-limit/src/Http/ClientIpKeyResolver.php
aidan-casey: Eh... yes. 🫣😛

Let me create an issue for it and we'll track to 4.x, I think.

aidan-casey: For the purpose of this PR, let's rename though.

packages/rate-limit/src/Http/ThrottleWith.php
innocenzi: Makes sense. In my opinion the runtime check is worth the improved DX though. Any opinion here @aidan-casey?

packages/rate-limit/src/Config/rateLimit.config.php
innocenzi: Good catch, we need to actually come to an agreement and document the convention here. Let's forget about that for now, we can always change it later, it's not gonna be a breaking change.

aidan-casey: @osbre you're bringing us to shame with our inconsistencies here. 🙈😅

packages/rate-limit/src/Http/ThrottleMiddleware.php
osbre: It was there to make `preventThrottling()` and `allowThrottling()` work in testing. I just realised we could achieve the same with `UnlimitedRateLimiter` as an alternative to the config parameter: [`a09d24c` (this PR)](https://github.com/tempestphp/tempest-framework/pull/2272/changes/a09d24cc86dd8adcdb835cc5399d89e0e34a4711)

packages/rate-limit/src/RateLimitException.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types=1);
+
+namespace Tempest\RateLimit;
+
+use Exception;
+
+/**
+ * Base class for exceptions thrown by the rate limit component.
+ */
+abstract class RateLimitException extends Exception {}
```
xHeaven, line 12: Could we make this a marker interface, following the exception conventions in the contributing guide?
I'm thinking something like this:
```php
interface RateLimitException {}
```
The existing exceptions would then extend `\Exception` and implement `RateLimitException`, like so:
```php
final class RateLimitWasExceeded extends \Exception implements RateLimitException
```

packages/rate-limit/src/RateLimitHasNoKey.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace Tempest\RateLimit;
+
+final class RateLimitHasNoKey extends RateLimitException
+{
+    public static function forLimit(RateLimit $limit): self
+    {
+        return new self(sprintf(
```
xHeaven, line 11: Could we rename this to `RateLimitKeyWasMissing` and accept the limit directly in the constructor? We are using past tense exception names, and constructors should accept only relevant context as per contribution guidelines.
Preferably the caller would then use `throw new RateLimitKeyWasMissing($limit)`.
Please update the tests and documentation accordingly.

packages/rate-limit/src/Storage/RateLimitStorageFailed.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Tempest\RateLimit\Storage;
+
+use Tempest\RateLimit\RateLimitException;
+
+final class RateLimitStorageFailed extends RateLimitException
+{
+    public static function redisDidNotReportAWindow(string $key): self
+    {
+        return new self(sprintf('Redis did not report a window for `%s` after recording an attempt against it.', $key));
```
xHeaven, line 13: Could we use the same constructor-accepts-context approach that I mentioned for `RateLimitKeyWasMissing`?
Please update tests and docs accordingly.

xHeaven (changes requested):
I've added a couple failing tests as well, please fix them.

packages/rate-limit/src/Http/ThrottleWith.php
osbre: I realised there needs be more runtime checks anyway, so now I dropped `ThrottleWith` and that extra trait.

aidan-casey:
Hey, @osbre! Thanks for putting the work into this.

This PR is too large for us to properly review and merge as-is, and there are also a few things that aren’t working as we’d expect.

I’d suggest breaking this down into smaller, focused contributions. And if you’re using AI to assist with future contributions, please disclose that in the PR.

For now, I’m going to close this one. Thanks again for your contributions thus far!

PR #2273, opened 2026-09-05
perf(container): cap the dependency chain backtrace

osbre:
`resolveChain()` previously captured the entire call stack on every container resolution but only needed the immediate caller's frame.

Added a test covering the reported origin file, which lacked coverage previously.

### Impact

Micro-optimization territory 😄

* **Container lookups:** Changes performance from linear to constant relative to stack depth, dropping from 3.23 µs down to 1.43 µs at depth 60.
* **App performance:** Saves about 35 µs per request on a trivial route. On an application doing real-world work it's a tiny fraction of a percent, but it's a tidy way to remove unnecessary overhead.

### DIFF
--- a/packages/container/src/GenericContainer.php
+++ b/packages/container/src/GenericContainer.php
@@ -653,7 +653,7 @@ final class GenericContainer implements Container
     private function resolveChain(): DependencyChain
     {
         if (! $this->chain instanceof DependencyChain) {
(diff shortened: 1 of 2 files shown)


innocenzi (approved):

brendt:
Thanks!

PR #2274, opened 2026-09-05
fix(database): use static instead of self in IsDatabaseModel

osbre:
## Problem

`IsDatabaseModel` trait methods using `self` caused two issues:

* **Runtime & PHPStan:** `self` inside a trait resolves to the class that *uses* the trait, not the class the method was called on. For example, calling `Admin::create()` (where `Admin extends User` and `User` uses the trait) built queries against `User`, and PHPStan incorrectly typed the return as `User`.
* **Mago:** Mago resolved `self` inside a trait to the trait itself (`IsDatabaseModel`), forcing consumers to manually narrow types with `instanceof`.

## Fix

* Switched return types to `static` and routed calls through `static::queryBuilder()` / `query(static::class)`.
* Inherited models now query themselves, and static analysers correctly infer the concrete model.
* Added 5 `@phpstan-ignore-next-line` annotations (matching the one pre-existing in `resolve()`).

### Why the PHPStan Ignores Are Needed

`QueryBuilder` returns `TModel|object<TModel>`, but `object` is not generic in PHPStan - `object<TModel>` degrades to bare `object`, which doesn't satisfy `static`. That is what forces the suppression today.

The correct annotation (`@template TModel of object` with `@param class-string<TModel>|TModel`) resolves cleanly. But `query()` also accepts raw table names (e.g., `query('users')`), and adding `|string` to the union collapses PHPStan's template inference (upstream issue [phpstan#12985](https://github.com/phpstan/phpstan/issues/12985)). A permanent fix requires binding `TModel` from an instance rather than a class-string (similar to Eloquent), which is out of scope for this PR.

### DIFF
--- a/packages/database/src/IsDatabaseModel.php
+++ b/packages/database/src/IsDatabaseModel.php
@@ -31,17 +31,17 @@ trait IsDatabaseModel
     /**
      * Returns a query builder targeting the specified database connection.
      *
-     * @return QueryBuilder<self>
+     * @return QueryBuilder<static>
      */
     public static function on(string|UnitEnum|null $databaseTag): QueryBuilder
     {
-        return self::queryBuilder()->onDatabase(databaseTag: $databaseTag);
+        return static::queryBuilder()->onDatabase(databaseTag: $databaseTag);
     }
 
     /**
      * Targets a specific database connection for this model instance.
      */
-    public function onDatabase(string|UnitEnum|null $databaseTag): self
+    public function onDatabase(string|UnitEnum|null $databaseTag): static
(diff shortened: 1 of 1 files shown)


aidan-casey:
This honestly should have been static to begin with. This looks good to me, but I'll let @brendt and @innocenzi weigh in.

Thanks!

aidan-casey (approved):

brendt:
It wasn't static because PhpStorm had issues with traits + generics + static in the past. However, I just checked and it seems to work now with static, so let's merge!

PR #2276, opened 2026-09-05
perf(mapper): cache per-class reflection plans during hydration

osbre:
Hydrating rows into models now caches reflection metadata, primary key scans, and resolved mappers to avoid redundant per-row work.

## Improvements

* **Per-Class Reflection Plans:** Caches reflection metadata, property types, and parent-child relations by class rather than recalculating them per row.
* **Cached Primary Key:** Memoizes the primary key property lookup on model inspectors to prevent thousands of redundant scans per nested eager-loading query.
* **Singleton Mapper Cache:** Replaces a factory-level cache that never hit (a fresh factory is built per `map()` call) with a container-scoped `#[Singleton]`, so the cache lives and dies with its container - no cross-container leaks. It implements `Resettable`, so workers calling `resetContainer()` between requests re-resolve mappers instead of holding ones built from unregistered dependencies. `ObjectFactory::__construct()` takes a third argument as a result.

## Performance Impact

<table>
  <thead>
    <tr>
      <th>Benchmark</th>
      <th>Before</th>
      <th>After</th>
      <th>Change</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Mapper stage<br><small>2,000-row SQLite select</small></td>
      <td>73.8ms</td>
      <td>35.3ms</td>
      <td>-52.2%</td>
    </tr>
    <tr>
      <td>End-to-end <code>Model::all()</code></td>
      <td>92.0ms</td>
      <td>53.4ms</td>
      <td>-42.0%</td>
    </tr>
    <tr>
      <td>Eager-loading<br><small>50 authors with nested relations</small></td>
      <td>3,344ms</td>
      <td>1,527ms</td>
      <td>-54.3%</td>
    </tr>
    <tr>
      <td><code>map()</code> in a hydration loop<br><small>Per call, steady state</small></td>
      <td>0.342ms</td>
      <td>0.018ms</td>
      <td>-94.8%</td>
    </tr>
  </tbody>
</table>

### DIFF
--- a/tests/Integration/Mapper/MapperCacheTest.php
+++ b/tests/Integration/Mapper/MapperCacheTest.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Tests\Tempest\Integration\Mapper;
 
 use PHPUnit\Framework\Attributes\Test;
-use ReflectionProperty;
 use Tempest\Container\GenericContainer;
 use Tempest\Mapper\MapperCache;
 use Tempest\Mapper\MapperConfig;
@@ -13,6 +12,9 @@ use Tempest\Mapper\Mappers\ArrayToObjectMapper;
 use Tempest\Mapper\MappingContext;
 use Tempest\Mapper\ObjectFactory;
 use Tests\Tempest\Integration\FrameworkIntegrationTestCase;
+use Tests\Tempest\Integration\Mapper\Fixtures\MapperPrefix;
+use Tests\Tempest\Integration\Mapper\Fixtures\PrefixedString;
+use Tests\Tempest\Integration\Mapper\Fixtures\PrefixedStringMapper;
 
 /**
  * @internal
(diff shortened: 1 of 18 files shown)


xHeaven (changes requested):
I've added 2 failing tests, please fix those regressions.

osbre:
Thanks @xHeaven. Fixed the issue to make the tests pass, and included one more test to ensure it doesn’t accidentally get broken with another change in the future.

tests/Integration/Mapper/MapperCacheTest.php
```diff
@@ -0,0 +1,91 @@
...
+        // re-resolved through the same instance.
+        $this->assertSame($cache, $this->container->get(MapperCache::class));
+
+        $cache->resolve(new MappingContext('default'), $resolve);
+
+        $this->assertSame(2, $resolved);
+    }
+
+    #[Test]
+    public function mappers_are_not_shared_between_containers(): void
```
xHeaven, line 76: In this test we claim that 2 containers get separate mapper caches... which is technically true in the current test implementation, because we actually create 2 separate mapper caches. The currrent test is tautological.
You should resolve the factory from each container instead, and then check the mapping results. You could use a small fixture for the mapper config.

xHeaven (changes requested):
I think this is it, otherwise LGTM.

Edit: ah yes, you mentioned a couple benchmarks in the PR description, could you perhaps add some actual benchmarks into the PR (where the other benchmarks are), so we could verify the performance claims?

osbre:
Thanks! Updated the test and included another benchmark commit to prove the measured numbers. Before that I was using a more DIY-approach at https://github.com/osbre/tempest-hydration-bench - that's where the numbers are from.

xHeaven (approved):

PR #2277, opened 2026-09-06
Fix(database) namespace retrieval in src_namespace function

### LINKED ISSUE #2275: Migration generation fails when placing it outside the default app path
### Tempest version

3.0

### PHP version

8.5

### Operating system

macOS

### Description

Not sure if it's an actual bug or just being a noob who just started with Tempest.

In my composer.json i have the following namespace configurations:
```
"autoload": {
        "psr-4": {
            "App\\": "app/",
            "Migrations\\": "migrations/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Tests\\": "tests/"
        }
    },
```

I'm using the default make migration command and get the following output:
```
php tempest make:migration create_biomes_table                                                                                                                     16:52:22

  │ Choose the migration type
  │ ✓ class

  │ Enter the table name
  │ ✓ biomes

  │ Is this an alteration?
  │ × No

  │ Where do you want to save the file CreateBiomesTable?
  │ ✓ migrations/CreateBiomesTable.php
× // The file could not be written. The path `/Users/{myname}/Projects/{myprojectname}/migrations/CreateBiomesTable.php` could not be mapped to a namespace.
```

yassinebel:
fixes #2275

### DIFF
--- a/packages/core/src/functions.php
+++ b/packages/core/src/functions.php
@@ -59,7 +59,7 @@ function registered_namespace(Stringable|string ...$parts): string
  */
 function src_namespace(Stringable|string ...$parts): string
 {
-    return to_psr4_namespace(Container\get(Composer::class)->mainNamespace, root_path(...$parts), root: root_path());
+    return to_psr4_namespace(Container\get(Composer::class)->namespaces, root_path(...$parts), root: root_path());
 }
 
 /**


brendt:
This is definitely not the fix for the problem.

PR #2278, opened 2026-09-06
fix(router): clear the matched route between requests

osbre:
Implement `RouterReset` to unregister `MatchedRoute` from the container during resets, and clear the binding in `MatchRouteMiddleware` before matching.

## Problem

`MatchRouteMiddleware` binds `MatchedRoute` into the container as a singleton. In long-running workers (FrankenPHP worker mode) or test suites where the container is reused across requests, this state leaks across boundaries: `GenericContainer::reset()` clears resolved singletons and dynamic initializers, but not `singletonDefinitions`, where the binding lives. Anything sharing the container - the next request, a console command, a deferred task - incorrectly inherits the previous request's matched route, causing stale data in URI generation, route model binding, and middleware execution.

## Solution

* Create `RouterReset` implementing `Resettable` to automatically unregister `MatchedRoute::class` during container resets, discovered via `ResettableDiscovery` and following the existing resets from worker mode support (#2172).
* Clear any existing `MatchedRoute` binding inside `MatchRouteMiddleware` before running a new match, ensuring safe reuse even when container resets do not occur, ie in test suite:
```php
$this->http->get('/docs/installation')->assertOk();
$this->http->get('/does-not-exist')->assertNotFound();

is_current_uri([DocsController::class, 'show'], slug: 'installation'); // true, on a 404
```

### DIFF
--- a/packages/router/src/MatchRouteMiddleware.php
+++ b/packages/router/src/MatchRouteMiddleware.php
@@ -21,6 +21,10 @@ final readonly class MatchRouteMiddleware implements HttpMiddleware
 
     public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
     {
+        // In long-running contexts, a previous request may have left a matched route behind.
+        // It must be cleared before matching, since we only rebind it on a successful match.
+        $this->container->unregister(MatchedRoute::class);
+
         $matchedRoute = $this->routeMatcher->match($request);
 
         if (! $matchedRoute instanceof MatchedRoute && $request->method === Method::HEAD && $request instanceof GenericRequest) {
--- a/packages/router/src/RouterReset.php
+++ b/packages/router/src/RouterReset.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Tempest\Router;
+
+use Tempest\Container\Container;
+use Tempest\Container\Resettable;
+
+final readonly class RouterReset implements Resettable
+{
+    public function __construct(
+        private Container $container,
+    ) {}
+
+    public function reset(): void
+    {
+        $this->container->unregister(MatchedRoute::class);
+    }
+}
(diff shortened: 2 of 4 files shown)


aidan-casey:
Nice! Looks good to me.

packages/router/src/MatchRouteMiddleware.php
```diff
@@ -21,6 +21,10 @@ public function __construct(
 
     public function __invoke(Request $request, HttpMiddlewareCallable $next): Response
     {
+        // In long-running contexts, a previous request may have left a matched route behind.
+        // It must be cleared before matching, since we only rebind it on a successful match.
+        $this->container->unregister(MatchedRoute::class);
```
aidan-casey, line 26: What's the thinking here? Use of a different container?

osbre: So far it occurs in the same container. `RouterReset` only fires when something actually calls `Container::reset()`, which today is the worker-mode request boundary in `FrameworkKernel`. Anywhere the container is reused *without* a reset - most obviously a test making two requests - the previous `MatchedRoute` singleton definition survives:

```php
$this->http->get('/docs/installation')->assertOk();
$this->http->get('/does-not-exist')->assertNotFound();

is_current_uri([DocsController::class, 'show'], slug: 'installation'); // true, on a 404
```

Since we only rebind `MatchedRoute` on a successful match, an unmatched request would otherwise keep answering with the prior request's route. Clearing it at the top of the middleware makes "no match" actually mean no matched route, independent of whether a reset happened.

aidan-casey: Makes total sense. We should probably clean this up elsewhere, but I think we merge as is for now.

osbre: Yep. On Discord, `golthem` suggested introducing `scoped registration`. I assume it works similarly to Laravel: https://laravel.com/framework/docs/13.x/container#binding-scoped

Issue #2263, opened 2026-08-28
Add container resolution modes


brendt:
A interesting question was raised in https://github.com/tempestphp/tempest-framework/pull/2256 on how the container should resolve dependencies.

First, we should discuss the current situation: Tempest's container defaults to non-singletons. Whenever you `->get()` a dependency from the container, it will create a new one, _unless_ there's an explicit singleton definition for it (defined either with `$container->singleton`, an initializer with `#[Singleton]`, or a class tagged with `#[Singleton]`).

Now there are three questions to discuss:

- Can we design an API that makes it trivial for users to explicitly retrieve non-singleton objects, even though they have a singleton definition?
- Can we make it so the container's default resolution strategy can be changed?
- What should the default resolution strategy be?

Let's discuss

brendt:
Pitching in myself

## Can we design an API that makes it trivial for users to explicitly retrieve non-singleton objects, even though they have a singleton definition?

Personally I like a two-part answer to this question. First we can build on top of tags, which we already have, and allow "dynamic tags": tags that are "defined" when retrieving a dependency rather than when defining them.

```php
$container->get(BlogRepository::class, tag: 'web');
```

Currently, if there's no explicit definition for `BlogRepository#web`, this call fails. However, we could easily make it so that it is allowed; maybe only when `#[AllowDynamicTag]` is added to `BlogRepository` — to be discussed.

The second part of my answer is to also add a `new` method (instead of `make`, I think that's a much more unclear name):

```php
$container->new(BlogRepository::class);
```

This method will always make a new instance, ignoring all possible singleton definitions.

## Can we make it so the container's default resolution strategy can be changed?

We can, but I think this is an incredible footgun and we shouldn't do it. Let's say the default changes from "non-singleton" to "singleton". Let's then say a third party package relies on the default mode, which is suddenly changed at runtime by the project code.

The only way this can work IMO is if we define a resolution mode on some kind of "context-based" basis. The most logical option would be the namespace level. However, we also have tagged dependencies which you can use for this purpose by defining explicit initializers for specific tags.

## What should the default resolution strategy be?

Assuming we find a good reason to add configurable resolution strategies, then I would argue non-singleton is still a much better default. Shared objects can lead to many unforeseen side-effects, and thus I rather make that an explicit opt-in.

Crell:
As the person who prompted this discussion, I'll offer my thoughts.

[...]

PR #2279, opened 2026-09-06
feat(container): add scoped bindings

osbre:
Adds a new `scoped` lifetime (`Container::scoped()` and `#[Scoped]`) for dependencies that live for a single request or lifecycle, replacing the manual `RouterReset` workaround from #2278.

## Problem

Tempest previously supported only transient and singleton lifetimes. State that is valid for a single request or job - such as `MatchedRoute`, tenant contexts, or authenticated users - had no native lifetime representation.

This resulted in cross-request leaks or clumsy workarounds, such as `RouterReset` and custom `Resettable` implementations, forcing every package holding per-request state to write its own manual cleanup logic.

## Solution

* Introduces a third lifetime, available via container method or attribute:
```php
$container->scoped(MatchedRoute::class, $matchedRoute);
$container->scoped(TenantContext::class, fn () => new TenantContext(...));
```

```php
#[Scoped]
final class TenantContext { /* ... */ }
```

* Scoped bindings act like singletons during a single lifecycle and are cleared when `Container::reset()` is called.
* **Definitions vs. Instances on Reset:** Callable definitions are kept so they can provide fresh instances in the next lifecycle. Concrete instances cannot be rebuilt, so their bindings are removed entirely.
* `#[Scoped]` is supported everywhere `#[Singleton]` is: on autowired classes, on initializer classes, and on `initialize()` methods, including the tagged variant `#[Scoped(tag: 'web')]`.
* Re-registering a scoped dependency with `singleton()` clears its scoped registration, so an explicit singleton always survives a reset.
* `container:show` lists scoped bindings under their own section, instead of mixing them into the singletons.
* `MatchRouteMiddleware` now binds `MatchedRoute` as scoped, and `RouterReset` has been deleted.
* Documented in `docs/1-essentials/05-container.md`.

## Design Choices

* **Defaults remain untouched:** Transient stays the default strategy, and singletons are unmodified.
* **Explicit naming:** Chose dedicated `scoped()` and `#[Scoped]` declarations rather than hiding behavior behind boolean flags on singleton methods.
* **Scope boundaries:** `Container::reset()` is still only triggered via `FrameworkKernel::shutdown()`. This PR introduces the binding mechanism, leaving automatic scope creation per request/job for a future update. This is also why `MatchRouteMiddleware` keeps the `unregister()` call from #2278: two requests within a single test share a container without a reset in between.

## Tests

Added 8 tests in `ContainerTest` covering:

1. Sharing within a lifecycle
2. Discarding an instance on reset
3. Discarding a tagged instance on reset
4. Keeping a callable definition across a reset
5. Leaving plain singletons untouched
6. Re-registering a scoped dependency as a singleton
7. The `#[Scoped]` attribute
8. A tagged `#[Scoped]` initializer

Plus one integration test asserting `container:show` lists scoped bindings separately.


brendt:
I was starting my review but realized a couple of files in that this might not be the right approach. 

What I think we need is singleton lifetime definitions, not another attribute dedicated to scope. My main motivation is that we were already discussing container scopes here: https://github.com/tempestphp/tempest-framework/issues/2263 

The problem with "scope" is that it could mean a "lifetime scope" like with this PR, or a "domain boundary scope", like we have with different highlighters for different contexts (CLI, web, etc.) or tenants.

I think we need two things to happen:

- Allow singletons to be defined with a lifetime (`#[Singleton(lifetime: Lifetime::REQUEST)]`)
- Allow dynamic tags as described in #2263, which will solve the tenant problem

This leads to a clear distinction between lifetime scope and domain boundary scope, which I think is much more clean.

@osbre @tempestphp/core-contributors what are your thoughts about this? @osbre are you willing to make a PR for it?

brendt:
Update: I'll PR it

Issue #1892, opened 2026-01-15
Broken `HeaderSessionIdResolver`


TimWolla:
### Tempest version

git

### PHP version

8.4

### Operating system

Linux

### Description

The `HeaderSessionIdResolver` is broken:

- The `str()` function does not exist (missing `use`).
- The expected header name contains underscores, which cannot appear in `$this->request->headers` (they are normalized to a hyphen).

brendt:
Thanks!

#1893 contains a fix for the broken import, which wasn't detected by CI because our testing environment requires laravel blade for testing the blade compatibility layer. Laravel itself has a global `str` function which is why it all worked in our tests without problems.

Regarding the normalized hyphens: I did add a test for that in the same PR, but as far as I can tell, underscores aren't replaced by hyphens. Maybe you have some more insights in that?

As a sidenote, I'd like to respectfully ask to consider choosing more neutral PR titles and issue descriptions. I'm sure you know the impact an aggressive communication style can have on OSS maintainers. I don't think you had bad intentions with it, but I did want to give you that feedback for the future.

TimWolla:
> Regarding the normalized hyphens: I did add a test for that in the same PR, but as far as I can tell, underscores aren't replaced by hyphens. Maybe you have some more insights in that?

For (Fast)CGI incoming headers are provided as environment variables to PHP. Since they may not contain hyphens, the CGI client (e.g. nginx) replaces them with underscores, resulting in both `tempest-session-id` and `tempest_session_id` request headers being provided to PHP as `$_SERVER['HTTP_TEMPEST_SESSION_ID']` (the same is true for PHP's built-in CLI server).

Since hyphenated headers are much more common than headers with underscores, all “request object” implementations that I'm aware of, including Laminas' ServerRequestFactory as used by Tempest, turn the underscores back into a hyphen:

https://github.com/laminas/laminas-diactoros/blob/ab5853da050bf8aee30f918f7d7e8d60e7b503ce/src/functions/marshal_headers_from_sapi.php#L57

Which makes it impossible for headers containing underscores to appear in practice.

PR #2280, opened 2026-09-06
refactor(http): deprecate HeaderSessionIdResolver

osbre:
Deprecates `HeaderSessionIdResolver`, scheduled for removal in 4.0.

## Why

* **Unused:** Not wired up by any initializer or selectable via configuration.
* **Never a default:** Replaced by `CookieSessionIdResolver` prior to the first tagged release (`v1.0.0-alpha.1`).
* **Undocumented:** Absent from all in-repo and official documentation.
* **Header name never matches:** It looks up an underscore header name (`*_session_id`), but request headers are normalized to hyphens by our own request layer (Laminas `ServerRequestFactory`, after the SAPI collapses both forms to `HTTP_..._..._`), so the lookup can't match a client-supplied identifier. This is the unresolved half of #1892; #1893 fixed the broken import but not the name mismatch.

Given the above, deprecating and later removing it seems cleaner than investing further in fixing it. Happy to adjust if there's an intended use case I've missed.

### DIFF
--- a/packages/http/src/Session/Resolvers/HeaderSessionIdResolver.php
+++ b/packages/http/src/Session/Resolvers/HeaderSessionIdResolver.php
@@ -12,6 +12,11 @@ use Tempest\Http\Session\SessionIdResolver;
 
 use function Tempest\Support\str;
 
+/**
+ * Resolves the session identifier from a request header.
+ *
+ * @deprecated Will be removed in 4.0. Use {@see CookieSessionIdResolver} instead.
+ */
 final readonly class HeaderSessionIdResolver implements SessionIdResolver
 {
     public function __construct(


brendt:
I agree that it might be broken at the moment, but isn't this important for async request? That being said, you could use cookies there as well. I actually don't remember why we ever added support for it. @innocenzi do you perhaps know?

innocenzi:
Hey @brendt, this existed before I touched the session implementation. Honestly unsure about its use case, but you must have had one if you introduced it. Or maybe it was just an easy way of authenticating prior to having cookie-based auth?

I'm fine with removing it unless we actually have a use case that justify its presence

brendt:
If anything, it's a remnant from very early code, I'm fine deprecating

PR #2281, opened 2026-09-06
feat(http): validate session identifiers

osbre:
Adds strict validation to `SessionId` to prevent potential path traversal vulnerabilities when session identifiers are used as filenames.

Validating at the `SessionId` level ensures that all storage drivers and resolvers automatically inherit this security guarantee. This approach matches standard practices used by other frameworks (such as Laravel and Symfony) for file-based session handlers. Existing UUIDs and standard identifiers remain fully compatible.

## Testing

Added unit tests in `packages/http/tests/Session/SessionIdTest.php`.

### DIFF
--- a/packages/http/src/Session/InvalidSessionId.php
+++ b/packages/http/src/Session/InvalidSessionId.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


brendt:
Thanks!

PR #2284, opened 2026-09-06
feat(http)!: regenerate the session identifier on authentication

osbre:
The session identifier used to survive authentication, so an identifier an attacker planted before login stayed valid afterwards and granted them the authenticated session.

`SessionRegenerator` assigns a new identifier to the current session and destroys the session it replaces. `SessionAuthenticator` calls it when authenticating and deauthenticating.

This is what OWASP ASVS 5.0 requires under [7.2.4](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V7-Session-Management.md) at level 1, including the termination of the previous token, and what both Laravel and Symfony do by default.

## Breaking changes

- `SessionManager` gains a `regenerate()` method, which custom resolvers must implement.
- `SessionIdResolver` gains a new `issueNewId()` method, which custom resolvers must implement
- Deauthenticating now discards the session data instead of carrying it over to the new identifier, matching Laravel and Symfony. Authenticating still preserves it, so flash data and the intended URL survive a login.

### DIFF
--- a/docs/2-features/04-authentication.md
+++ b/docs/2-features/04-authentication.md
@@ -90,6 +90,45 @@ final readonly class AuthenticationController
 }
 ```
 
+### Regenerating the session identifier
+
+Tempest automatically regenerates the session identifier when a model is authenticated or deauthenticated. Authentication keeps the existing session data, while deauthentication clears it before creating the new session. In both cases, the previous session is destroyed.
+
+You should also regenerate the session identifier whenever an authenticated session changes privilege level, such as after a password change, enabling two-factor authentication, impersonating another user, or escalating a user's role. Inject {b`Tempest\Http\Session\SessionRegenerator`} for these transitions:
+
+```php app/Authentication/TwoFactorController.php
+use Tempest\Http\Session\Session;
+use Tempest\Http\Session\SessionManager;
+use Tempest\Http\Session\SessionRegenerator;
+
+final readonly class TwoFactorController
+{
+    public function __construct(
+        private Session $session,
+        private SessionManager $sessionManager,
+        private SessionRegenerator $sessionRegenerator,
+    ) {}
+
+    public function enable(): void
+    {
+        // Enable two-factor authentication for the current user...
+
+        $this->sessionRegenerator->regenerate();
+        $this->sessionManager->save($this->session);
+    }
+}
+```
+
+`regenerate()` destroys the old session, assigns a new identifier, and carries the session data over. If the data must not survive the transition, clear the session before regenerating it:
+
+```php
+$this->session->clear();
+$this->sessionRegenerator->regenerate();
+$this->sessionManager->save($this->session);
+```
+
+The new session must be saved after regeneration. The authenticator handles this automatically for normal authentication and deauthentication flows.
+
 ### Accessing the authenticated model
 
 You may access the currently authenticated model by injecting the {b`Tempest\Auth\Authentication\Authenticator`}. The authenticator provides a `current()` method that returns the currently authenticated model, or `null` if no model is authenticated.
--- a/packages/http/src/Session/SessionIdResolver.php
+++ b/packages/http/src/Session/SessionIdResolver.php
@@ -16,5 +16,5 @@ interface SessionIdResolver
      *
      * @see SessionRegenerator
      */
-    public function regenerate(): SessionId;
+    public function issueNewId(): SessionId;
 }
--- a/packages/http/src/Session/SessionIdResolver.php
+++ b/packages/http/src/Session/SessionIdResolver.php
@@ -6,5 +6,15 @@ namespace Tempest\Http\Session;
 
 interface SessionIdResolver
 {
+    /**
+     * Resolves the identifier sent by the client, creating a new one if there is none.
+     */
     public function resolve(): SessionId;
+
+    /**
+     * Creates a new identifier and sends it to the client, replacing the one it was using.
+     *
+     * @see SessionRegenerator
+     */
+    public function regenerate(): SessionId;
 }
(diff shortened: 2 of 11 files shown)


brendt (changes requested):
Why not add the regenerate logic in the manager class?

osbre:
Since `SessionManager` is an interface, adding it there would lead to duplication as all `regenerate` does is:
1. deleting the old session
2. swapping the ID via the resolver
3. saving the new one

It doesn't seem to be storage-specific, so all three managers would end up repeating the same code. It would also mean injecting the `SessionIdResolver` into every implementation.

innocenzi:
I agree with @osbre regarding the responsibility of `SessionManager`.

I cleaned the API a bit by renaming methods to better names, and removing `preserveData` in favor of a dedicated method.

One thing I don't like but don't have a solution for is `replaceId`. It makes no sense for userland to ever reach it, but there isn't anyplace else to put it either. We can't just update `Session` in the container because it could be stale in other dependencies.

If you have any idea, I'll take it; otherwise we can just keep it as `@internal`, but it's not ideal.

brendt:
> It doesn't seem to be storage-specific, so all three managers would end up repeating the same code. It would also mean injecting the SessionIdResolver into every implementation.

I understand, but then I still think it can be better designed. I really don't like replacing `SessionManager` as a dependency. If anything, regeneration is a side effect of the manager doing something, not the entrypoint.

A couple of suggestions:

- Inject the regenerator in the managers. Yes, every manager implementation has to call it then. I don't mind all that much, as long as the regeneration logic itself is contained
- Handle regeneration with events as a side-effect of calling `deauthenticate()` and `authenticate()`
- Directly call regenerate from the authenticator, but also keep the manager calls there. So essentially inject both the manager and regenerator

I think the third option is the best, but am open for suggestions

osbre:
Pushed option 3: the authenticator now takes both, with regeneration as a side effect. Let me know if this matches your vision.

OWASP also recommends regenerating on other privilege changes, like password changes, 2FA, impersonation, and role escalation, so users may need to call `SessionRegenerator` directly. I've updated the auth docs to cover this.

docs/2-features/04-authentication.md
```diff
@@ -90,6 +90,45 @@ final readonly class AuthenticationController
...
+    }
+}
+```
+
+`regenerate()` destroys the old session, assigns a new identifier, and carries the session data over. If the data must not survive the transition, clear the session before regenerating it:
+
+```php
+$this->session->clear();
+$this->sessionRegenerator->regenerate();
+$this->sessionManager->save($this->session);
```
brendt: Sorry for once again circling back to this, but I really find it annoying having two classes that interact with sessions. Yes I understand that it's annoying for each manager implementation to re-implement the regenerate method, but from a user's perspective it's a lot more clean if everything session-related goes through the manager.

Here's another idea: what if we keep the regenerator class, but inject it into each manager; then the `SessionManager::regenerate` method simply uses the underlying regenerator. I understand that this is a little more work for implementations, but it will be so much cleaner for end-users, which is much more important

brendt (changes requested):

docs/2-features/04-authentication.md
osbre: Seems that injecting it would give us circular dependency: `SessionManager -> SessionRegenerator -> Session -> SessionManager`.

Added `regenerate` to `SessionManager` which uses `SessionIdResolver` directly.

packages/http/src/Session/SessionManager.php
```diff
@@ -21,6 +21,18 @@ public function save(Session $session): void;
...
+     * Assigns a new identifier to the session, destroying the session it replaces
+     * and sending the new identifier to the client. Session data is carried over.
+     *
+     * This protects against session fixation, and should be done whenever the session
+     * changes privilege level - such as authentication or a password change.
+     *
+     * @see https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html
+     * @see https://owasp.org/www-community/attacks/Session_fixation
+     */
+    public function regenerate(Session $session): void;
```
brendt, line 34: We need a Rector to automate this change. anything that implements SessionManager but not `regenerate` should add the method and within the body throw an exception saying it's not yet implemented.

packages/http/src/Session/SessionIdResolver.php
```diff
@@ -6,5 +6,15 @@
...
+     * Resolves the identifier sent by the client, creating a new one if there is none.
+     */
     public function resolve(): SessionId;
+
+    /**
+     * Creates a new identifier and sends it to the client, replacing the one it was using.
+     *
+     * @see SessionManager::regenerate()
+     */
+    public function issueNewId(): SessionId;
```
brendt, line 19: We need a Rector to automate this change. anything that implements SessionIdResolver but not issueNewId should add the method and within the body throw an exception saying it's not yet implemented.

brendt (changes requested):

osbre:
Thanks! Assuming it's for 3.20. Followed the same convention as [3.14](https://github.com/tempestphp/tempest-framework/blob/3.x/packages/upgrade/tests/Tempest314/Tempest314RectorTest.php) - included three fixtures: plain implementations, aliased-import implementations, and implementations that already have the methods.

packages/upgrade/src/Tempest320/UpdateSessionImplementationsRector.php
```diff
@@ -0,0 +1,98 @@
...
+                ],
+            ],
+        ],
+        SessionIdResolver::class => [
+            'issueNewId' => [
+                'returnType' => SessionId::class,
+                'parameters' => [],
+            ],
+        ],
+    ];
```
osbre, line 39: Is it best to split those as two separate rules or keep this `foreach` approach?

brendt: I rather have two rules, that's simpler to understand 👍

PR #2285, opened 2026-09-07
feat(database): forward query builder methods to database models

osbre:
This PR introduces static select-side query builder methods to `IsDatabaseModel`, removing the redundant `::select()` boilerplate when starting queries.

### Before

```php
Book::select()->whereField('title', 'Timeline Taxi')->first();
Book::select()->orderBy('published_at')->limit(10)->all();
```

### After

```php
Book::whereField('title', 'Timeline Taxi')->first();
Book::orderBy('published_at')->limit(10)->all();
Book::whereHas('author', fn ($q) => $q->whereField('name', 'Brent'))->all();
```

## Changes

* **Forwarded 39 Select-Side Methods:** Methods like `where`, `whereField`, `whereHas`, `orderBy`, `limit`, `with`, `join`, and pagination helpers can now be called directly statically on models. Each forwarder is a direct delegation to `static::select()->...`.
* **Fixed Return Annotations:** Updated `HasWhereRelationMethods::whereHas()` and `whereDoesntHave()` annotations to ensure proper type inference for the forwarded methods.
* **Excluded Methods:** Chain-continuation methods (`andWhere*`, `orWhere*`) and internal SQL compilation utilities remain builder-only, since neither is meaningful as an opening call.

## Tests

No new tests were added. Every forwarded method is already exercised at the builder level in `tests/Integration/Database/`, and the forwarders add no logic beyond method dispatch. This follows the existing convention of testing behaviour where the builder lives: the `sum()` and `avg()` statics are covered in `AggregateQueryBuilderTest`, not `IsDatabaseModelTest`.

### DIFF
--- a/packages/database/src/Builder/QueryBuilders/HasWhereRelationMethods.php
+++ b/packages/database/src/Builder/QueryBuilders/HasWhereRelationMethods.php
@@ -22,7 +22,7 @@ trait HasWhereRelationMethods
      *
      * @phpstan-param (?Closure(SelectQueryBuilder): void) $callback
      *
-     * @return static<TModel>
+     * @return static
      */
     public function whereHas(
         string $relation,
@@ -45,7 +45,7 @@ trait HasWhereRelationMethods
      *
      * @phpstan-param (?Closure(SelectQueryBuilder): void) $callback
      *
-     * @return static<TModel>
+     * @return static
      */
     public function whereDoesntHave(
         string $relation,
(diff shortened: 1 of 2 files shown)


osbre:
Regarding the benchmark glitch: https://github.com/tempestphp/tempest-framework/pull/2286

innocenzi:
Not something I am in favor of personally. `select` is explicitly creating the request, and it reads well.

brendt:
I agree, I like the explictness of `select`. I'm going to close

PR #2289, opened 2026-09-09
feat(auth): add id_token to GenericOAuthClient

luggesexe:
Hello,
I want to make use of RP Initiated Logout. For that per OIDC Spec I have to pass on the id_token_hint. I preserve this in the GenericOAuthClient and pass it on as a url parameter later on.

Thanks and kind regards
Zoë

### DIFF
--- a/packages/auth/src/OAuth/GenericOAuthClient.php
+++ b/packages/auth/src/OAuth/GenericOAuthClient.php
@@ -153,6 +153,7 @@ final class GenericOAuthClient implements OAuthClient
         $token = $this->requestAccessToken(
             code: $request->get('code'),
         );
+        $this->storeIdToken($token);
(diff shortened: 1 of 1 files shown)


xHeaven:
Could you use the `$token` already passed to the `authenticate()` mapping callback to store `$token->getValues()['id_token']` in your session, then retrieve it when building the logout URL? Is there something in your flow that prevents that?

luggesexe:
That works, thank you!
For reference this was my solution in /oidc/callback

```php
            map: function (OAuthUser $oauthUser, #[\SensitiveParameter] AccessToken $token) {
                $this->session->set(
                    'id_token',
                    $token->getValues()['id_token'] ?? null,
                );
```

PR #2292, opened 2026-09-09
feat(container): add singleton lifetime support

brendt:


brendt:
I'm not sure I'm 100% happy with this approach. It would be more clean if we registered singleton definitions in some kind of config, so that we had a list of attributes available _after_ discovery. 

That's a pretty large refactor however. The container never used this pattern because it was literally the first component ever to be written for Tempest, before we followed the best practice of discovery writing to config. 

It's not the end of the world, and I think I'm fine with the current approach, but I'd love some input @innocenzi @xHeaven

aidan-casey:
I have several thoughts here...

1. I really dislike confusing singletons and scoped objects [which are two separate things.](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection/service-lifetimes) 

2. Remember that the whole idea of discovery is that it is gluing together things that can be manually done under the hood. We have no way of setting the singleton lifecycle when registering here.

innocenzi:
I actually think the idea is good, and does align with .NET. "Transient" is just another lifetime, exactly like .NET does.

@aidan-casey's second point is valid though, we should be able to specify the lifetime when binding services directly through the `Container` instance.

We could have this:

```php
// generic
$container->bind(..., lifetime: Lifetime::REQUEST);

// shorthands
$container->transient(...);
$container->singleton(...);
$container->perRequest(...);
```

PR #2295, opened 2026-09-12
fix(database): don't emit `UNSIGNED` on Postgres

osbre:
Fixes a bug where using `unsigned: true` on an integer column breaks Postgres migrations with a `syntax error at or near "UNSIGNED"`.

Postgres doesn't support unsigned integer types, so it now gets the plain integer type instead. MySQL and SQLite are unchanged. Added missing unit tests for `unsigned` columns to CI so Postgres doesn't break again.


packages/database/src/QueryStatements/IntegerStatement.php
```diff
@@ -29,14 +29,29 @@ public function compile(DatabaseDialect $dialect): string
...
-                is_int($this->size) ? DatabaseIntegerSize::fromBytes($this->size)->toString() : $this->size->toString(),
+                $this->type(),
                 $this->unsigned ? 'UNSIGNED' : '',
                 $this->default !== null ? "DEFAULT {$this->default}" : '',
                 $this->nullable ? '' : 'NOT NULL',
             ),
         };
     }
+
+    private function type(): string
```
brendt: I'd prefer this to be a local variable instead of a method

tests/Integration/Database/QueryStatements/IntegerStatementTest.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        // MySQL supports `UNSIGNED`; dropping it globally would silently halve the requested range.
+        $this->assertStringContainsString(
+            'UNSIGNED',
+            new IntegerStatement('votes', unsigned: true)->compile(DatabaseDialect::MYSQL),
+        );
+
+        $this->assertStringNotContainsString(
+            'UNSIGNED',
+            new IntegerStatement('votes')->compile(DatabaseDialect::MYSQL),
```
brendt: Shouldn't you test that there's no unsigned when the flag is set to true? But then you test for exactly this in the previous test. I don't really understand the use of this assertion?

brendt (changes requested):

brendt (approved):

PR #2296, opened 2026-09-12
Sync latest changes from the original repo

rcalicdan:


aidan-casey:
Hey, @rcalicdan! Not sure what you're attempting to do here, but we will not be merging this PR into the core repository.

I'd suggest creating an issue for discussion before putting any further work in here to make sure it aligns with the direction of the framework.

PR #2299, opened 2026-09-15
fix(framework): read one-line docblock vars in meta:view-component

### LINKED ISSUE #2298: `meta:view-component` misses variables declared in one-line PHPDoc
### Tempest version

3.19.2

### PHP version

8.5

### Operating system

Linux

### Description

While working on a [Carve](https://github.com/markup-carve/carve-tempest) markup integration for Tempest:

`tempest meta:view-component` only reports `@var` declarations when the tag is
on a star-prefixed line inside a multiline PHPDoc block. A valid one-line
PHPDoc declaration is silently omitted.

This affects Tempest's built-in `x-submit` component, whose `$label` declaration
uses the one-line form.

## Reproduction

With Tempest Framework 3.19.2 and PHP 8.5.10:

```sh
composer create-project tempest/app tempest-meta-repro
cd tempest-meta-repro
php tempest meta:view-component x-submit
```

Actual output:

```json
{
    "name": "x-submit",
    "slots": [],
    "variables": []
}
```

The component contains:

```php
/** @var null|string $label The submit button's label */
```

Changing it to the multiline form makes the variable appear:

```php
/**
 * @var null|string $label The submit button's label
 */
```

## Expected behavior

Both conventional PHPDoc forms should expose the same variable metadata. For
`x-submit`, the metadata should include `$label` with the `label` attribute
name.

## Likely cause

`MetaViewComponentCommand::resolveVariables()` matches only lines beginning
with an optional indent followed by `* @var`:

```php
->matchAll('/^\s*\*\s*@var.*$/m')
```

That pattern cannot match `/** @var ... */`.

## Possible fix

Parse both one-line and multiline PHPDoc tags, ideally without making the
metadata extractor depend on a particular comment layout.

dereuromark:
Fixes #2298

`meta:view-component` only picked up var tags on the star-prefixed lines of a multiline docblock. A one-line docblock was skipped, so `x-submit` reported no variables:

```php
/** @var null|string $label The submit button's label */
```

The extractor now reads both forms. `x-submit` lists `$label` with its description.

One-line docblocks also type local variables, as in `x-input`:

```php
/** @var FormSession $formSession */
$formSession = get(FormSession::class);
```

A one-line var tag directly followed by a plain `=` assignment to the same variable is treated as a local and left out, so `x-input` doesn't start listing `$formSession` and `$validator` as attributes. Compound assignments like `??=` still count as attributes, since that's how components set a default.


brendt:
Could you take a look at the failing test? I assume it's something to do with line endings

dereuromark:
"5 workflows awaiting approval"
In settings, you can set to "approval only needed for new github users", this usually helps to speed things up.

PR #2301, opened 2026-09-17
fix(database): resolve uuid primary keys from positional insert bindings

### LINKED ISSUE #2300: UUID primary keys: relation auto-inserts persist the internal last-insert id instead of the generated UUID
### Tempest version

3.19

### PHP version

8.5

### Operating system

Linux

## Summary

When a model uses `#[Uuid]` as its primary key, relation auto-inserts performed by `InsertQueryBuilder`'s `after` callbacks (`BelongsToMany` pivot rows, `HasOne` / `HasMany` child rows) are persisted with the database's internal last-insert id as the parent foreign key instead of the actual generated UUID.

On SQLite this silently corrupts the pivot table:

```
user_roles: user_id = "1", role_id = "01a09f0e-d128-7c50-abf1-3e861d348a9e"
--                    ^^^ the SQLite rowid, not the user's UUID
```

## Environment

- SQLite (reproducible on MySQL as well; PostgreSQL is unaffected because `InsertStatement` compiles `RETURNING *`)

## Reproduction

```php
final class User
{
    use IsDatabaseModel;

    #[Uuid]
    public PrimaryKey $id;

    public string $name;

    /** @var Role[] */
    #[BelongsToMany]
    public array $roles = [];
}

final class Role
{
    use IsDatabaseModel;

    #[Uuid]
    public PrimaryKey $id;

    public string $name;
}
```

```php
$role = query(Role::class)->create(name: 'admin');

$user = query(User::class)->create(
    name: 'Jane',
    roles: [$role],
);

$user->id->value; // "01a09f0e-d13a-7e15-9190-f4a58d51e0aa" (correct UUID)

query('user_roles')->select()->all();
// [ ['user_id' => "1", 'role_id' => "01a09f0e-..."] ]
//                ^^^
// Expected: $user->id. Actual: the internal rowid.
```

The same happens for `HasMany`/`HasOne` children: their owner foreign key column receives the last-insert id instead of the parent's UUID.

## Expected behavior

The pivot row's owner foreign key (and child rows' foreign keys) contain the actual generated UUID primary key of the inserted parent model.

## Actual behavior

The owner foreign key contains `getLastInsertId()` — for SQLite with a `TEXT` primary key this is the internal rowid (`"1"`, `"2"`, ...), for MySQL the value of `last_insert_id()`. Both are meaningless for an explicit string primary key.

## Root cause

`packages/database/src/Query.php`, `Query::execute()`:

```php
return isset($query->bindings[$this->primaryKeyColumn])
    ? new PrimaryKey($query->bindings[$this->primaryKeyColumn])
    : $database->getLastInsertId();
```

Insert bindings are **positional** (`[0 => ..., 1 => ...]`), so `isset($query->bindings[$this->primaryKeyColumn])` (e.g. `$bindings['id']`) is always `false` and the code always falls back to `getLastInsertId()`.

This works for auto-increment primary keys only because the last insert id *is* the primary key there. With `#[Uuid]` (an explicit string primary key generated in PHP before the insert) it is not.

## Known limitation (out of scope for a fix)

Batch inserts (`insert($modelA, $modelB, ...)`) with relation data only attach relations to a single parent — `execute()` returns one `PrimaryKey` and `after` callbacks run once. This is pre-existing behavior for all primary key types and would require a larger redesign (per-row parent ids); it is not part of this issue.

radoslav-grencik:
Fixes #2300


radoslav-grencik:
## What this PR does

Fixes relation auto-inserts for models with `#[Uuid]` primary keys, and adds the missing migration API needed to actually define UUID relationships.

### 1. Bug fix: `Query::execute()` resolved the wrong primary key after insert

`Query::execute()` looked up the primary key in `$bindings[$primaryKeyColumn]`, but insert bindings are positional — the lookup never matched and execution always fell back to `getLastInsertId()`. For `#[Uuid]` primary keys that is the SQLite rowid / MySQL `last_insert_id()`, so `BelongsToMany` pivot rows and `HasMany`/`HasOne` child rows persisted a meaningless integer as the parent foreign key:

```
user_roles: user_id = "1"  -- rowid, not the user's UUID
```

The primary key is now resolved through its column position in the `InsertStatement` before falling back to `getLastInsertId()`. The auto-increment path is unchanged.

Covered by new integration tests (`UuidPrimaryKeyTest`):

- `uuid_primary_key_belongs_to_many_pivot_uses_generated_uuid`
- `uuid_primary_key_has_many_uses_generated_uuid_as_foreign_key`

### 2. Iterable `insert()` now generates `#[Uuid]` primary keys

Previously only the object path (`create()`, `save()`) generated UUID values; the iterable/array path silently omitted the primary key, which surfaced as `Field 'id' doesn't have a default value` on MySQL. `resolveIterableData()` now generates a UUID when the model has a `#[Uuid]` primary key and the key is absent from the given data. Explicitly provided ids are never overwritten.

This also makes UUID primary keys work for relation child rows inserted from arrays (e.g. `chapters: [['title' => ...]]`), which go through the iterable path internally.

Covered by `uuid_primary_key_generated_for_iterable_insert`.

### 3. Feature (bundled): UUID foreign key columns in migrations

While writing the tests it turned out there is no way to define a UUID-typed foreign key column: `belongsTo()` / `foreignId()` always compile `INTEGER`, and `varchar(36)` FK columns fail on PostgreSQL (`character varying` vs `uuid` type mismatch) — this PR's own CI caught it.

This PR therefore also adds:

- `UuidStatement` — a non-PK UUID column (`CHAR(36)` / `UUID` / `TEXT` per dialect)
- `CreateTableStatement::uuidColumn(string $name, bool $nullable = false)`
- `CreateTableStatement::belongsToUuid(string $local, string $foreign, ...)` — UUID counterpart of `belongsTo()`
- `CreateTableStatement::foreignUuid(string $local, string $constrainedOn, ...)` — UUID counterpart of `foreignId()`

The new integration test migrations use `belongsToUuid()` / `foreignUuid()`, so the UUID relation path is exercised end-to-end with real FK constraints on every dialect. Unit tests cover compilation for all three dialects (`UuidStatementTest`, `CreateTableStatementTest::create_a_uuid_foreign_key_constraint`).

## Notes / known limitations (follow-ups, not part of this PR)

- **Pivot tables need no primary key at all.** Pivot rows are always addressed by their FK pair, never by their own id: inserts go through a table-name string (no model metadata), deletes and joins go through the FK columns. The pivot test table therefore declares no PK (matching the Laravel pivot convention). A composite PK was considered but deliberately not added — Tempest's ORM identity is single-column (`PrimaryKey` property, `get($id)`, route binding), so a composite key would require a redesign far beyond this PR's scope.
- **Batch insert + relations**: `execute()` returns a single `PrimaryKey` and `after` callbacks run once, so relations are attached to the first row only. Pre-existing behavior for all PK types, unchanged by this PR.

xHeaven:
What happens when an insert explicitly passes a `0` id?

radoslav-grencik:
> What happens when an insert explicitly passes a `0` id?

Good catch — that was a bug. When an insert explicitly passes `id => 0`, the positional resolution would return `PrimaryKey(0)` instead of falling back to `getLastInsertId()`.

MySQL and PostgreSQL both treat `INSERT ... VALUES (0, ...)` on an auto-increment column as "use the sequence default", so the real id is `1` (or next sequence value). The fix was returning the explicit `0` — a non-existent row.

**Fix:** treat `0` the same as "no id provided" — fall back to `getLastInsertId()`:

```php
$value = $query->bindings[$index] ?? null;

// 0 is treated as "auto-increment" by MySQL/PostgreSQL — not a real id.
if ($value === 0 || $value === null) {
    return null;
}

return $value;
```

This is safe for UUID models too — a UUID entry never contains `0`; the PK is always a generated UUID string.

Added a test (`insert_with_zero_id_falls_back_to_last_insert_id`) that inserts with explicit `id => 0` and verifies the returned id is usable for retrieval.

packages/database/src/QueryStatements/CreateTableStatement.php
```diff
@@ -59,6 +59,16 @@ public function uuid(string $name = 'id'): self
         return $this;
     }
 
+    /**
+     * Adds a UUID column to the table. Uses `CHAR(36)` for MySQL, `UUID` for PostgreSQL, and `TEXT` for SQLite.
+     */
+    public function uuidColumn(string $name, bool $nullable = false): self
```
innocenzi, line 65: This function name is inconsistent with our other names. We should probably repurpose `uuid` for this instead, and use `primary` as the sole way of creating primary UUIDs.

Would be breaking though, so for next major.

innocenzi (changes requested):

packages/database/src/QueryStatements/CreateTableStatement.php
radoslav-grencik: I'm thinking - do we want/need public function for creating uuid column? Now we don't have one and it is actually not needed - for foreign keys i have added `foreignUuid` / `belongsToUuid`. If not I will just delete it for now.

radoslav-grencik: removed for now; naming (uuid vs primary) we can do in next major

PR #2303, opened 2026-09-19
fix(database): escape quote characters in identifiers

osbre:
Escape quote characters in `DatabaseDialect::quoteIdentifier()` to prevent potential SQL injection when identifiers come from untrusted input.

The method wraps an identifier in quote characters, but previously did not escape quote characters already inside it. This meant an identifier containing a quote could end its own quoting, causing the rest of the value to be treated as SQL instead of part of the identifier:

```php
$dialect->quoteIdentifier('ti"tle');

// before: "ti"tle"
// after:  "ti""tle"
```

Doubling the quote character is the escaping format expected by MySQL/SQLite and PostgreSQL. This also matches how other frameworks handle identifier quoting, including Laravel's `Illuminate\Database\Grammar::wrapValue` and Doctrine DBAL's `AbstractPlatform::quoteSingleIdentifier`.

Handling this at the dialect level means all call sites - relations, table aliases, migration statements, etc. - get the same protection without having to sanitise identifiers individually.

## Testing

Added unit test at `packages/database/tests/Config/DatabaseDialectTest.php`.


xHeaven:
Hey, I think this needs a more thorough refactor, as some paths still quote things without calling the fixed method, such as `HasTableAlias.php:45` and `FieldStatement.php:40`.

osbre:
@xHeaven wouldn't that be a separate follow-up PR? to keep the changes small and focused

xHeaven:
> @xHeaven wouldn't that be a separate follow-up PR? to keep the changes small and focused

Good question, I'm not sure what'd be the best here. Those changes are closely related (essentially the same thing) to this one, I would personally keep them in a single PR. @brendt what's your call here?

